Issue #137, opened 2026-06-05
[DESIGN] Contract models: on-disk, wire, config, and runtime schemas


ELares:
IronBus has no shared codebase yet, so the contracts below are the single point of agreement between the storage engine (#4), the record codec (#5), durability (#6), recovery (#7), corruption-skip (#8), consumers (#9), the wire protocol (#11), config (#14), and observability (#16). This issue freezes the concrete, language-neutral data models: their names, their kind (on-disk, wire, config, runtime, or report), their fields with exact byte widths where a format is frozen, and the relationships between them. Field types are language-neutral (u8, u16, u32, u64, bytes, enum, string, varint). Every multi-byte on-disk and wire field is little-endian unless a field is explicitly a varint (LEB128, canonical shortest form). These models are the contract a Rust struct, a fuzz corpus, a CLI decoder, and a future replication peer must all match byte for byte. Nothing here introduces new policy: it transcribes the frozen shapes from the subsystem issues into one normative table.

## On-disk record models (#5)

**RecordHeader** is exactly 36 bytes and starts every record. Fields in order: `magic` u16 (frozen 0x4942), `version` u8, `flags` u8, `seq` u64 (monotonic offset, must fall in the owning segment's `[base_seq, base_seq + record_count)` range), `timestamp` u64 (unix millis), `key_len` u32, `hdr_len` u32, `payload_len` u32, `header_crc` u32. The `header_crc` is CRC32C over bytes `[0..32)`, that is every header field except the crc itself. The header is immediately followed by `key_len` raw key bytes, then `hdr_len` header bytes, then `payload_len` payload bytes.

**RecordTrailer** is exactly 8 bytes and closes every record: `body_crc` u32 (CRC32C over key, header, and payload bytes in wire order), then `total_len` u32. `total_len` equals `36 + key_len + hdr_len + payload_len + 8`. There is no leading copy of `total_len`: the validated header lengths plus the fixed framing sizes are the leading total, and a record is intact only when `header_crc` passes, the trailing `total_len` equals the computed expected total, and `body_crc` passes. Payloads above 64 KiB additionally carry an xxh3-64 in the header region; CRC32C still gates resync.

## On-disk segment models (#4, #5)

[...]

PR #145, opened 2026-06-06
feat(core): frozen v1 on-disk format constants and core value types

ELares:
## What
- `ironbus-core::format`: the frozen v1 on-disk layout as constants and field offsets (record header 36 B, trailer 8 B, segment header 64 B, footer 32 B, record magic, CRC32C algo id, 16 MiB default / 1 GiB ceiling record size), with a compile-time `default <= ceiling` invariant and tests that pin the layout.
- `ironbus-core::types`: `Offset`, `Seq`, and `RecordFlags` value types (documented, tested).
- `#![warn(missing_docs)]` on `ironbus-core` so the public API stays documented.

## Why
First real code. The format constants are the single source of truth the codec and its tests will refer to (#5), and the value types are the contract foundation (#137). Scoped to definitions + tests; encode/decode with CRC32C lands next.

## Test evidence
Local: `cargo fmt --check`, `cargo clippy -- -D warnings` (pedantic + missing_docs), `cargo test` (8 core tests), `cargo deny check`, the SPDX check, and the no-IO lint all pass.

refs #5, #137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,4 +14,5 @@ to follow Semantic Versioning once it reaches a tagged release.
 - Continuous integration: rustfmt, clippy (deny warnings), test matrix on Linux,
   macOS, and Windows, an MSRV 1.78 build, and a lint that keeps `ironbus-core` IO-free.
 - Dual `MIT OR Apache-2.0` license files.
+- `ironbus-core`: frozen v1 on-disk format constants and field offsets (record header, trailer, segment header and footer), and the `Offset`, `Seq`, and `RecordFlags` value types, with unit tests pinning the layout.
 - Supply-chain CI gate (`cargo-deny`) with a permissive-license allowlist, and SPDX license headers on every Rust source enforced by CI.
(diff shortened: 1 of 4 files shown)


ELares:
## Review addressed

- **Saturating `next()` could duplicate a monotonic id (important):** replaced with `checked_next() -> Option` on `Offset` and `Seq`; an exhausted id space now fails loudly rather than silently reusing a value.
- **No known-bits mask (important):** added `RecordFlags::KNOWN` and `unknown_bits()` for forward-compatibility, plus a `Hash` derive.
- **Frozen values not pinned:** added a `frozen_values` test (magic, version, checksum-algo id, xxh3 threshold, segment sizes) and centralized the segment-size and xxh3-threshold constants in `format.rs`.
- **Shift-invariant offset test:** now asserts absolute field offsets, so a consistent-but-wrong interior shift is caught.
- **Magic doc:** clarified the on-disk byte order.

Re-running CI; will merge once green.

Issue #146, opened 2026-06-06
[TASK] Place the xxh3-64 large-payload checksum in the frozen record frame


ELares:
While implementing the record codec (#145 follow-on), a gap surfaced between two frozen decisions:

- #5 froze the record trailer as exactly 8 bytes: `body_crc: u32` then `total_len: u32`. There is no field for a second checksum.
- #8 requires a second, independent `xxh3-64` checksum for payloads at or above 64 KiB (`XXH3_PAYLOAD_THRESHOLD`), stored in the record framing, with CRC32C still gating resync.

The current codec (`ironbus-core::codec`) implements only the mandatory CRC32C path and does not emit or verify the xxh3-64 checksum, because the frozen frame has nowhere to put it.

## Decide
- Where the 8-byte xxh3-64 lives in the frame (most likely: a flag bit on the record plus an extra 8-byte field placed immediately before the trailer, included in `total_len`), so existing 8-byte-trailer parsing still works and `total_len` remains authoritative.
- Whether the threshold is on the uncompressed or stored (compressed) payload size.

## Acceptance criteria
- [ ] The frame layout for the xxh3-64 checksum is specified and added to `ironbus-core::format`.
- [ ] `ironbus-core::codec` encodes and verifies it for payloads >= the threshold, round-trip and corruption tests cover it.
- [ ] CRC32C remains the resync-gating checksum.

Found during the record codec work. Relates to #5, #8.

PR #147, opened 2026-06-06
feat(core): record-frame codec with CRC32C (encode/decode)

ELares:
## What
- `ironbus-core::codec`: `encode`/`decode` for the v1 record frame with CRC32C header and body checksums, typed `EncodeError`/`DecodeError`, a borrowed `RecordView`, and a derived `HAS_KEY` flag.
- Tests: unit cases (truncated, bad magic, header corruption, body corruption, empty-key flag clearing), a proptest round-trip, and a proptest asserting a single bit flip anywhere in the frame never yields a clean decode of the original payload.
- New deps: `crc32c` (runtime, IO-free, license allowlisted) and `proptest` (dev).

## Why
The record codec is the heart of the on-disk format (#5), and the property + corruption-detection tests establish the testing discipline the design requires.

## Discovered gap (filed as #146)
The frozen 8-byte trailer (#5) has no field for the xxh3-64 large-payload checksum that #8 requires. This codec implements the mandatory CRC32C path only; the xxh3 frame layout is tracked in #146.

## Test evidence
Local: `cargo fmt --check`, `cargo clippy -- -D warnings` (pedantic), 18 tests (codec unit + proptests), `cargo deny check`, SPDX, and the no-IO lint all pass.

refs #5, #137, #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,5 +14,6 @@ to follow Semantic Versioning once it reaches a tagged release.
 - Continuous integration: rustfmt, clippy (deny warnings), test matrix on Linux,
   macOS, and Windows, an MSRV 1.78 build, and a lint that keeps `ironbus-core` IO-free.
 - Dual `MIT OR Apache-2.0` license files.
+- `ironbus-core::codec`: record-frame encode and decode with CRC32C header and body checksums, typed `EncodeError`/`DecodeError`, and proptest round-trip plus single-bit-flip corruption-detection tests. The optional xxh3-64 large-payload checksum is deferred (frame layout pending).
 - `ironbus-core`: frozen v1 on-disk format constants and field offsets (record header, trailer, segment header and footer), and the `Offset`, `Seq`, and `RecordFlags` value types, with unit tests pinning the layout.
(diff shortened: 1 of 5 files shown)


ELares:
## Review addressed

- **B1 (BLOCKER, 32-bit usize overflow to panic on crafted input):** `decode` now sums the three `u32` length fields in `u64` and bounds the total against the 1 GiB ceiling before narrowing to `usize`, so a crafted header with `u32::MAX` lengths and a valid CRC returns `TooLarge` instead of panicking. Added `crafted_huge_lengths_are_rejected_not_panicked` and a ceiling-boundary test.
- **N4 (HAS_KEY not validated):** `decode` now rejects a frame whose HAS_KEY flag disagrees with `key_len` (`BadLength`), with a test.
- **I2 (weak proptests):** round-trip now exercises unknown flag bits and asserts preservation; the bit-flip proptest asserts `consumed == len` on the Ok branch.
- **I3 (test gaps):** added multi-frame sequential decode (trailing bytes), all-empty record, and ceiling-boundary tests.
- **N5 / N6 (docs):** added the intentional version exact-match note and the per-record-size caller-responsibility note.
- **32-bit CI coverage** (B1 follow-up) tracked in #148.

23 tests pass locally. Re-running CI; will merge once green.

Issue #21, opened 2026-06-05
[DESIGN] Verification strategy: crash injection, fuzzing, property tests, and simulation


ELares:
IronBus markets itself as hyper resilient to power loss and corrupt files, self-recovering and continuing with only minor, bounded, reported data loss. Those are the hardest claims to make and the easiest to fake, because the failure modes (a torn tail write, a page-cache writeback that silently dropped, a bit-flipped segment on cheap eMMC) almost never appear in a happy-path test. This issue defines how we prove the resilience claims before we trust them: crash and power-cut injection, block-layer and fsync fault injection, a corpus of deliberately corrupted segments, property-based tests for the codec and recovery, coverage-guided fuzzing of the parser, deterministic concurrency testing with loom, and a FoundationDB-style deterministic simulation of the IO and scheduler. The deliverable is an executable specification of the invariants and the harness that checks them, vetted before any production code lands.

## Why this matters for IronBus

This is the Resilient tenet made falsifiable. A resilience claim with no adversarial test is marketing, and the prior art is littered with cases where the test gap was the bug: Postgres shipped fsync error handling for years that silently lost data because nobody injected a writeback EIO (the fsyncgate 2018 episode), and Kafka relied on the high watermark alone for follower truncation until log-divergence bugs forced leader epochs. We will not learn those lessons in production on an unattended edge box. The strategy also serves Edge First: the faults we inject (brownout mid-fsync, SD-card bit-rot, an IPv6 nameserver that wedges IO) are the edge's normal operating conditions, not exotic ones. It serves Simple by keeping the contract small and checkable: a handful of named invariants that hold or do not. And it ties directly to #19, because a durability claim is meaningless without the latency cost attached, so crash-injection runs and benchmark runs share the same workload generator and the same recovery clock.

## Prior art

[...]

Issue #10, opened 2026-06-05
[DESIGN] Backpressure and flow control without head-of-line stalls


ELares:
This issue specifies how IronBus keeps producers fast, consumers isolated, and memory bounded under overload. The modern stance is that producers should rarely block: instead of stalling the writer, the durable log absorbs spikes, slow consumers are quarantined so they cannot stall their peers, and sustained overload is met with explicit, configured shedding rather than silent unbounded growth. The core invariant from every system surveyed is that every buffer must be bounded and the bound must be controllable, because an unbounded queue under sustained overload turns into either an out-of-memory crash or unbounded latency (bufferbloat) long before it ever sheds anything useful. We combine credit-based flow control at consumer boundaries, a bounded in-memory ring backed by spill-to-disk, time-in-queue shedding, per-consumer isolation, and adaptive concurrency limits on egress. This is a design document to be vetted before any code is written.

## Why this matters for IronBus

Backpressure touches four of the five tenets directly. **Resilient**: an overloaded broker that grows buffers without limit dies on a power-constrained edge box; bounded buffers plus spill-to-disk mean a burst becomes queue depth on the durable log rather than a crash. **Edge First**: a single node with a few cores and a tight RAM ceiling cannot afford an unbounded in-flight set, so the in-flight window and the ring size are first-class memory budgets (see #20). **HyperScale**: per-consumer isolation and structural backpressure are what let one fan-out topic feed many consumers of wildly different speeds without the slowest one dictating the rate for all. **Simple**: producers rarely block, and the overflow behavior is a small set of named, per-topic policies with safe defaults rather than emergent, surprising behavior. The natural failure mode we are designing against is a fast producer plus one stuck consumer silently consuming all broker memory.

## Prior art

[...]

PR #149, opened 2026-06-06
feat(core): deterministic clock seam (Clock trait + ManualClock)

ELares:
## What
- `ironbus-core::clock`: the `Clock` trait exposing a wall clock (`now_unix_millis`, for record timestamps, may move backwards) and a monotonic clock (`now_monotonic_nanos`, for lease deadlines and queue sojourn), and a deterministic `ManualClock` that tests and the simulation advance explicitly. Advancing saturates at `u64::MAX`.

## Why
The determinism contract (#21): engine logic must never read `SystemTime::now`/`Instant::now`; it reads time through the `Clock` seam, so the deterministic simulation can replay runs. First piece of the IO/clock/scheduler seam.

## Test evidence
Local: fmt, clippy pedantic `-D warnings`, 29 tests (a self-test caught and fixed a real wrap-instead-of-saturate bug), the no-IO lint (uses only `core::sync::atomic`), cargo-deny, and SPDX all pass.

refs #21, #7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 - Continuous integration: rustfmt, clippy (deny warnings), test matrix on Linux,
   macOS, and Windows, an MSRV 1.78 build, and a lint that keeps `ironbus-core` IO-free.
 - Dual `MIT OR Apache-2.0` license files.
+- `ironbus-core::clock`: the `Clock` trait (wall and monotonic time) and a deterministic `ManualClock`, so engine logic never reads the host clock directly and the simulation controls time.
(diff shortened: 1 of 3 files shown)


ELares:
## Review addressed

- **advance_millis couples wall+monotonic:** documented as the no-skew helper; skew scenarios (NTP step, backwards jump) drive `set_unix_millis` and `advance_monotonic_nanos` independently.
- **Non-atomic snapshot across the two clocks:** documented that the two clocks are independent atomics, so a single advance is not an atomic snapshot; the single-threaded simulation never hits the interleaving.
- **Test gaps:** added `monotonic_never_decreases`, `concurrent_advances_are_all_counted` (exercises the CAS retry loop and the Send/Sync claim), `wall_clock_advance_saturates`, and `set_wall_does_not_touch_monotonic`.
- **Duration ergonomics, nonzero monotonic origin:** deferred to the lease and CoDel consumers (#9, #10), where the duration semantics are defined.
- **SeqCst ordering:** acknowledged, kept as the safe default for a simulation clock.

33 tests pass. Re-running CI; will merge once green.

PR #150, opened 2026-06-06
feat(storage): file IO seam (RandomAccessFile trait + InMemoryFile)

ELares:
## What
- `ironbus-storage::io::RandomAccessFile`: positioned `read_at`/`write_all_at` (pread/pwrite semantics, zero-fill on write past end), `sync_data` (fdatasync), `len`, `set_len`, `is_empty`. All `&self` so a file can be shared with lock-free readers.
- `InMemoryFile`: a `Vec`-backed deterministic implementation that counts `sync_data` calls and can `snapshot()` its bytes, so the simulation can model power loss (only bytes present at the last sync are durable). Poisoned-lock recovery avoids panics.

## Why
The determinism seam (#21): storage must not call the filesystem directly, so the simulation can drive a disk it controls. This is the file half of the IO seam (the clock half landed in #149). The production `std::fs` impl and directory fsync come with the segment store.

## Test evidence
Local: fmt, clippy pedantic `-D warnings`, 7 tests (round-trip, zero-fill gap, partial/past-end reads, truncate/extend, sync counting, cross-thread `Arc<dyn RandomAccessFile>`), cargo-deny, SPDX all pass.

refs #21, #4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 - Continuous integration: rustfmt, clippy (deny warnings), test matrix on Linux,
   macOS, and Windows, an MSRV 1.78 build, and a lint that keeps `ironbus-core` IO-free.
 - Dual `MIT OR Apache-2.0` license files.
+- `ironbus-storage::io`: the `RandomAccessFile` trait (positioned reads/writes, `sync_data`, `len`, `set_len`) and an `InMemoryFile` (sync-counting, snapshot-able) so storage goes through a seam the deterministic simulation can substitute.
 - `ironbus-core::clock`: the `Clock` trait (wall and monotonic time) and a deterministic `ManualClock`, so engine logic never reads the host clock directly and the simulation controls time.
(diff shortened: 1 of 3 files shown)


ELares:
## Review addressed

- **Power-loss model was not implementable (important):** `InMemoryFile` now keeps a durable image captured at each sync. `sync_data`/`sync_all` snapshot live to durable; `durable_snapshot()` exposes it; `simulate_power_loss()` reverts live to the durable image so the simulation can prove only synced data survives. `from_bytes` seeds the durable image. Tests cover the synced/unsynced split and power loss.
- **read_exact_at missing:** added as a defaulted trait method for fixed-size header/footer reads, with an `UnexpectedEof` test.
- **No metadata sync:** added `sync_all` (fsync) distinct from `sync_data` (fdatasync); documented that a `set_len` length is durable only after `sync_all`.
- **Two-mutex lock-order risk avoided:** both images live under one `Mutex<State>`, so there is no lock ordering to get wrong; added a comment on why poison recovery is safe.
- **Torn-write / short-read / injected-IO-error fault injection** (a simulation feature) is deferred to #151.

10 tests pass. Re-running CI; will merge once green.

Issue #17, opened 2026-06-05
[DESIGN] Single static binary, cross-platform build, and distribution


ELares:
IronBus must be trivially installable on an edge device and trivially auditable in a supply chain, and those two goals shape how we build and ship. This issue specifies the build and distribution story: one self-contained, statically linked binary per architecture (musl on Linux), cross-compilation to the targets edge fleets actually run (x86_64, aarch64, armv7, with riscv64 as a watch item), a size-optimized release profile, reproducible builds, supply-chain hygiene (auditable binaries, dependency policy, SBOM, signed releases), and the distribution channels themselves (curl-to-shell installer, OS packages, a minimal container image, and checksummed GitHub releases). The model to copy is Redpanda: a single statically linked binary with no JVM and no ZooKeeper, plus the broader pattern of how Rust CLIs ship. This connects to #20 (what the device can actually hold and tolerate) and #22 (license, versioning, and release policy).

## Why this matters for IronBus

This is the issue where four of the five tenets become concrete artifacts. **Simple**: install is "download one file and run it"; there is no runtime, no interpreter, no sidecar coordination service. Redpanda ships exactly this way (one statically linked binary, no JVM, no ZooKeeper), and we should match that bar. **Cross Platform**: the same source produces binaries for x86_64 dev laptops and aarch64/armv7 gateways, and `crc32fast` already feature-detects at runtime so one binary is fast on both an x86 build host and an aarch64 device. **Edge First**: a static musl binary depends only on the kernel, so it runs across distro and glibc versions on heterogeneous gateway hardware without a dependency hunt, and a small binary costs less flash. **Resilient** extends to the supply chain: an embedded SBOM and signed, checksummed releases mean an operator can verify what is running on a device long after `Cargo.lock` is gone. The fifth tenet, **HyperScale**, is served indirectly: a tiny binary and image start fast and pack densely.

## Prior art

**Redpanda** demonstrates the destination: a Kafka-API-compatible broker compiled to a single statically linked binary with the proxy and schema registry built in, removing the JVM and ZooKeeper operational burden entirely. That is the single-binary ergonomics we want, achieved in a systems language rather than on a managed runtime.

[...]

Issue #151, opened 2026-06-06
[TASK] Fault injection in the simulation file (torn writes, short reads, injected IO errors)


ELares:
The storage IO seam (#150) provides `InMemoryFile` with a durable-image power-loss model (`simulate_power_loss` discards unsynced writes). The deterministic simulation (#21) needs richer fault injection on top of this seam to exercise the storage correctness invariants:

## Scope
- **Torn writes:** a `write_all_at` that, on a simulated crash, persists only a prefix (or a sub-region) of the bytes, modelling a partial-sector write.
- **Short reads:** `read_at` returning fewer bytes than available at chosen points.
- **Injected IO errors:** `sync_data`/`sync_all`/`write_all_at`/`read_at` returning a seeded `io::Error` (including a one-shot fsync `EIO`, per the crash-class matrix in #21).
- All driven by the single seeded PRNG so runs are replayable.

## Acceptance criteria
- [ ] A fault-injecting `RandomAccessFile` wrapper (or an extended `InMemoryFile` mode) can produce torn writes, short reads, and injected errors deterministically from a seed.
- [ ] Recovery/durability tests assert the invariants hold under these faults (no acked-and-synced data lost; recovery stops at a torn tail).

Found during the #150 review. Relates to #21, #7, #8.

PR #152, opened 2026-06-06
feat(storage): production StdFile (cursor-free positioned IO) + sync_dir

ELares:
## What
- `ironbus-storage::io::StdFile`: a production `RandomAccessFile` backed by `std::fs::File`, using cursor-free positioned IO (`FileExt::read_at`/`write_all_at` = `pread`/`pwrite`) so lock-free concurrent readers never contend on a shared file cursor. `open`/`create`/`from_file` constructors; `sync_data`/`sync_all` map to `fdatasync`/`fsync`.
- `sync_dir(path)`: fsyncs a directory so a newly created, renamed, or removed entry is crash-durable (needed before acking a segment roll).

## Scope
Scoped to `cfg(unix)`. The v1 first-class targets are Linux musl (aarch64/x86_64/armv7) and macOS, all Unix. Windows is a committed v1 non-goal, so it compiles and tests the trait + `InMemoryFile` only. Note: macOS `fsync` does not flush to the platter without `F_FULLFSYNC`; that durability nuance belongs to the durability model (#6).

## Test evidence
Local (macOS): fmt, clippy pedantic `-D warnings`, 14 storage tests (4 new `StdFile`: round-trip + reopen persistence, truncate/extend, read-past-end, `sync_dir`), cargo-deny (tempfile licenses allowlisted), SPDX, no-IO lint all pass. `ironbus-core` dep tree stays runtime-free.

refs #4, #6, #17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ to follow Semantic Versioning once it reaches a tagged release.
   macOS, and Windows, an MSRV 1.78 build, and a lint that keeps `ironbus-core` IO-free.
 - Dual `MIT OR Apache-2.0` license files.
 - `ironbus-storage::io`: the `RandomAccessFile` trait (positioned reads/writes, `sync_data`, `len`, `set_len`) and an `InMemoryFile` (sync-counting, snapshot-able) so storage goes through a seam the deterministic simulation can substitute.
+- `ironbus-storage::io::StdFile`: production `RandomAccessFile` over an OS file using cursor-free positioned IO (pread/pwrite), plus `sync_dir` for crash-durable segment creation (Unix targets; Windows is a v1 non-goal).
 - `ironbus-core::clock`: the `Clock` trait (wall and monotonic time) and a deterministic `ManualClock`, so engine logic never reads the host clock directly and the simulation controls time.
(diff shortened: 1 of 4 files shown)


ELares:
## Review addressed

- **macOS durability hole (no F_FULLFSYNC):** documented prominently on `StdFile` (and the sync methods) that plain `fsync`/`fdatasync` on macOS do not flush the device write cache, so acked data is not power-loss-safe there. macOS is a developer and CI target per the vision (#1), not a power-loss-durable production target; production durability targets Linux musl. The macOS `F_FULLFSYNC` path is tracked in **#153**.
- **`create` truncates, no `create_new` (data-clobbering footgun):** added `StdFile::create_new` (`O_EXCL`) as the safe segment-create primitive (errors `AlreadyExists` instead of clobbering); marked `create()` destructive and scratch-only. Test added.
- **`sync_dir` ordering:** documented the contract (fsync the file, then full-fsync the parent directory, then ack).
- **`from_file` doc:** noted intended use (preallocated/handed-off fd).
- **Test gaps:** added `create_new_refuses_existing_file` and `concurrent_positioned_reads_are_race_free` (proving `StdFile` reads are cursor-free and race-free). fsync-error fault injection stays with #151.

16 tests pass. Re-running CI; will merge once green.

PR #154, opened 2026-06-06
feat(core): segment header/footer codec; share raw read helpers

ELares:
## What
- `ironbus-core::segment`: `SegmentHeader` (64 B) and `SegmentFooter` (32 B) `encode`/`decode`, both CRC32C-protected, with a typed `SegmentError` (truncated, bad magic, unsupported version/checksum-algo, bad CRC).
- Freezes the exact header field layout (magic, version, checksum_algo, flags, segment_id, base_seq, base_offset, created_ms, header_crc over `[0,60)`) and footer layout (record_count, last_seq, sealed_ms, footer_crc over `[0,28)`) per #5, with field offsets pinned by a `format` test.
- Factors the little-endian read helpers into `crate::raw`, now shared by the record codec and the segment codec (no behaviour change to the record codec).

## Test evidence
Local: fmt, clippy pedantic `-D warnings`, 44 core tests (segment unit cases for each error + proptest round-trip and single-bit-flip corruption for both header and footer), cargo-deny, SPDX, no-IO lint all pass.

refs #5, #4, #137

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 - Dual `MIT OR Apache-2.0` license files.
 - `ironbus-storage::io`: the `RandomAccessFile` trait (positioned reads/writes, `sync_data`, `len`, `set_len`) and an `InMemoryFile` (sync-counting, snapshot-able) so storage goes through a seam the deterministic simulation can substitute.
 - `ironbus-storage::io::StdFile`: production `RandomAccessFile` over an OS file using cursor-free positioned IO (pread/pwrite), plus `sync_dir` for crash-durable segment creation (Unix targets; Windows is a v1 non-goal).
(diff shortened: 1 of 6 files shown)


ELares:
## Review addressed (freeze-time footer fixes)

- **Footer had no magic/version (important):** the 32-byte footer now carries a 2-byte magic (`0x4653` "SF"), `version`, and `checksum_algo` mirroring the header, so a torn tail cannot masquerade as a valid footer on a CRC collision, and a future format change can gate on the footer version. `decode` checks magic, version, and algo before the CRC.
- **Footer not bound to its segment (important):** added `segment_id` to the footer; the `decode` doc notes the caller must verify it matches the header and `last_seq >= base_seq`. `record_count` is now `u32` and the speculative `sealed_ms` (never in the #5 footer sketch) is dropped, so it all fits the frozen 32 bytes.
- **Footer version/checksum gating:** resolved by the magic/version/algo above.
- **`flags` raw:** documented as preserve-do-not-interpret in v1.
- **Weak bit-flip proptest:** strengthened to assert a single-bit flip is always rejected, for both header and footer.

45 tests pass. Re-running CI; will merge once green.

Issue #80, opened 2026-06-05
[TASK] Segment reaper: time, size, and count bounds over sealed segments


ELares:
Parent: #13

Build the background reaper that evaluates the three retention bounds (`max_age`, `max_bytes`, `max_messages`) over SEALED segments only and selects whole segments for deletion. This is the read-only selection half of retention; the actual delete-and-commit step lives in the crash-safe delete sibling. The reaper never touches the active segment (matching Kafka) and never rewrites individual records.

## Approach

The reaper is event-driven plus timed: it runs whenever a segment seals (new reclaimable bytes appear), on a 60s backstop timer for `max_age` (age can cross a bound with no writes), and out of band when a disk high-water threshold (default 90%) fires. It runs at low priority, yields to the append path, and does bounded work per pass. Selection is greedy: drop every eligible sealed segment in offset order from the oldest. Eligibility per mode: time uses the segment footer's highest BROKER append timestamp clamped `max(footer_time, now)` (never producer time); size compares total on-disk sealed bytes against `max_bytes` and selects oldest segments until under bound; count derives total as (highest sealed `last_offset`) minus (durable log-start offset), with no separate counter. Defaults wired in: `max_bytes = 256MiB`, `max_age = 168h`, `max_messages = 0` (disabled). Final numbers coordinate with #14 and #20. The reaper emits per-run metrics (segments reaped, bytes freed, last-run timestamp) to #16.

## Acceptance criteria

- [ ] Reaper selects only sealed segments and never the active segment, verified against the #4 model.
- [ ] Time, size, and count bounds each independently select the correct oldest segments; combined bounds take the union.
- [ ] Count is derived from offsets with no maintained per-record counter.
- [ ] Age uses broker footer time clamped against backward clock jumps; a backward step never stalls reclamation.
- [ ] Reaper runs on seal, on a 60s timer, and on a 90% high-water trigger; does bounded work per pass and yields to appends.
- [ ] Per-run metrics and a last-run timestamp are exported to #16.

## Failure considerations

A forward NTP step on an RTC-less board could mass-select a topic by age; mitigate by clamping to footer time and deferring age reaping one stabilization interval after a large clock jump while size reaping still protects the disk. A starved reaper that never runs lets the disk fill; the 90% high-water out-of-band trigger guarantees a run before the disk is actually full. A busy-looping reaper could steal append CPU; bounded-work-per-pass plus low priority and append yielding prevent it.

ELares:
[...]

Issue #54, opened 2026-06-05
[TASK] Rebuild and validate offset/time indexes and clamp consumer cursors atomically during recovery


ELares:
Parent: #7

Make the sparse offset index, the time index, and persisted consumer cursors consistent with the recovered prefix, atomically and without trusting any stale on-disk index across a crash. Recovery can move the high-water mark backwards (truncation), so anything that points into the log must be re-derived or clamped, or reads after a self-heal will mis-serve. Coordinates with #4 (index format), #9 (cursor store), and #13 (prefix guarantees).

## Approach
For every segment touched after the checkpoint, unconditionally regenerate the offset and time indexes from the validated record stream into a temp file, fdatasync, atomic rename over the old index, then fsync(parent dir). For sealed segments at or before the checkpoint, accept the existing index only if its trailer CRC and its (first_seq, last_seq) bounds validate against the segment footer; otherwise regenerate the same way. A half-written index from a crash mid-rebuild is detected by a missing or mismatched trailer CRC and triggers regeneration. After truncation, clamp every persisted consumer cursor from #9 to the recovered high-water mark, and record each clamp (old offset, new offset, consumer id) in the RecoveryReport for #16.

## Acceptance criteria
- [ ] Touched-segment indexes regenerated via temp + fdatasync + rename + dir fsync.
- [ ] Sealed-segment indexes validated by trailer CRC and seq bounds before trust; regenerated on mismatch.
- [ ] Half-written index detection (missing trailer CRC) specified.
- [ ] Consumer cursors clamped to the recovered high-water mark, with clamps reported.
- [ ] No index is ever trusted blind across a crash (stated and tested).

## Failure considerations
A crash during index regeneration leaves a temp file, never a corrupt live index, because the rename is atomic and the old index stays valid until then. A cursor left past the truncated end would stall or read past EOF, prevented by clamping. A sealed index silently stale (segment rewritten by retention) is caught by the seq-bounds check against the footer.

ELares:
[...]

Issue #52, opened 2026-06-05
[TASK] Implement the startup recovery state machine: forward validation, torn-tail truncation, idempotent commit


ELares:
Parent: #7

Implement the open-path recovery algorithm that turns whatever is durably on disk into a consistent prefix: read the checkpoint floor, stream-validate records forward, find the clean commit point, truncate a torn tail, and write a fresh checkpoint, all idempotent and bounded. This is the core of #7 and the moment the durability promises of #6 are cashed in against the layouts of #4 and #5.

## Approach
On open: read the CHECKPOINT slot (floor). From durable_offset in last_segment_id, stream the tail through a small fixed buffer, recomputing each record CRC32c per #5 using the hardware path. Stop at the first short, bad-length, or CRC-failing record. Always forward-validate (no clean-shutdown fast path). If the stop is a torn tail (per the classification task), truncate the active segment at the stop offset, then fsync(segment) because a size change is metadata, then fsync(parent dir). Only after both fsyncs return, write a fresh checkpoint slot. The truncation point derives solely from durable bytes, so re-running after a crash mid-recovery yields the identical offset (idempotence). Enforce max_tail_scan_bytes (default 512 MiB) as a warn-and-continue tripwire under the default action from #14. Acquire a directory lock (flock + PID/boot-id) before any of this to prevent double-open.

## Acceptance criteria
- [ ] One unambiguous commit point documented (post-truncation fsync, then fresh checkpoint).
- [ ] Forward validation streams through a fixed buffer; no whole-segment load into RAM.
- [ ] Idempotence proof: re-running after a crash at the truncation or checkpoint step yields identical results.
- [ ] Truncation uses full fsync (size change) plus parent-dir fsync, ordered before the new checkpoint.
- [ ] Directory lock prevents a second concurrent instance.
- [ ] max_tail_scan ceiling and continue/refuse behavior wired to #14.

## Failure considerations
A second power cut during recovery must converge: idempotence guarantees it. A truncation persisted with only fdatasync could lose the size change on some filesystems, so full fsync is required. Double-open would let two writers corrupt the active segment, prevented by the lock. A huge tail triggers warn-and-continue, never a brick.

ELares:
[...]

Issue #47, opened 2026-06-05
[TASK] Implement single-committer group commit on the WAL


ELares:
Parent: #6

Build the group-commit engine that amortises `fdatasync` across many concurrent publishes, which is the only way to reach high write rates on edge flash without one sync per record. This is the load-bearing mechanism behind the `batch` and `sync` levels defined in the levels-spec sub-issue and layered on the WAL from #4.

## Approach
A single dedicated OS thread (not a Tokio task) owns the only writable WAL fd. Producers enqueue `(record bytes, oneshot completion handle)` onto an MPSC channel. The committer loops: drain all currently-queued appends up to `group_commit_max_bytes = 1 MiB`, perform one `write` plus one `fdatasync`, then complete every drained waiter together. Default `group_commit_max_delay_ms = 0`: NO proactive linger (natural coalescing already happens because syncs serialize and appends arriving during sync N ride sync N+1). A non-zero linger is configurable but off by default. Exactly one thread ever calls `write`/`fdatasync` on a given fd, which makes fsync-error observation deterministic. Export commit-queue depth and oldest-unsynced-age as backpressure signals to #10 and #16; bound the queue so overflow triggers spill/shed rather than unbounded RAM growth.

## Acceptance criteria
- [ ] One thread owns the writable WAL fd; verified no other code path syncs it.
- [ ] N concurrent publishes coalesce into one `fdatasync` (metric: syncs-per-record < 1 under load).
- [ ] `group_commit_max_bytes` caps a single batch at 1 MiB.
- [ ] Default linger is 0; a configurable non-zero linger is honoured when set.
- [ ] Commit-queue depth and oldest-unsynced-age are exported and bounded.
- [ ] Queue overflow drives #10 (spill/shed), never unbounded buffering.

## Failure considerations
If sync throughput falls below ingest the queue grows; mitigated by the bound plus backpressure handoff (high-severity resource-exhaustion path). Doing blocking `fdatasync` inside an async task risks cancellation hazards, which is precisely why a dedicated thread is mandated. A second writer on the fd would let a writeback error be stolen, resurrecting fsyncgate; the one-writer invariant prevents it.

ELares:
[...]

PR #155, opened 2026-06-06
feat(storage): segment writer/reader over the file IO seam

ELares:
## What

The first storage-engine PR: a single-segment `SegmentWriter` and `SegmentReader` in `ironbus-storage`, wiring the record and segment codecs (`ironbus-core`) to the `RandomAccessFile` IO seam. This is the heart of the log storage model (log-is-WAL: the active segment is the WAL) and the foundation of recovery.

## Writer

`SegmentWriter<F: RandomAccessFile>` owns one segment file:
- `create` writes the 64-byte CRC32C segment header.
- `append(&RecordView)` frames the record via the codec, writes it at the running `write_pos`, and returns the log `Offset` it was assigned (`base_offset + record_count`).
- `sync` issues `sync_data`; `seal` writes the 32-byte CRC32C footer (bound to this segment id) and `sync_all`s.
- Appends are refused with `SegmentFull` once the record count or byte length would overflow, so a single segment can never silently wrap.

## Reader / recovery primitive

`SegmentReader<F>` validates the header on `open`, then `scan`:
- detects a sealed footer in the trailing 32 bytes, **rejecting a footer whose `segment_id` does not match the header** (`FooterSegmentMismatch`);
- decodes records forward and **stops at the first torn or corrupt frame**, returning the valid prefix and the byte offset where it ends (`valid_end`), with `clean=false` when the tail was truncated.

This gives the core recovery invariant a concrete home: a power-loss tail never reads past the last intact record.

## Tests (deterministic, `InMemoryFile`)

- write then scan round-trip
- seal then scan reads footer and records
- empty sealed segment
- torn tail yields the valid prefix
- power loss keeps only synced records (`simulate_power_loss` drops un-synced writes)
- a footer bound to a different segment is rejected

## Gate

`cargo fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (67 tests), `cargo-deny`, SPDX headers, and the `ironbus-core` IO-free lint all pass locally. CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. Active-segment lifecycle, the offset/time indexes (#54), the startup recovery state machine (#52), the segment reaper (#80/#81), and group commit (#47) are separate follow-up PRs, not absorbed here.

refs #4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Added
+- `ironbus-storage::segment`: `SegmentWriter` (append records, sync, seal with footer) and `SegmentReader` (validate header, scan records, recover the valid prefix on a torn or corrupt tail), wiring the record and segment codecs to the `RandomAccessFile` seam. Tests cover round-trip, seal/scan, empty sealed segments, torn-tail recovery, power-loss (only synced records survive), and rejection of a footer bound to a different segment.
 - Cargo workspace skeleton with six crates: `ironbus-core`, `ironbus-storage`,
   `ironbus-proto`, `ironbus-server`, `ironbus-client`, and `ironbus-cli` (binary
   `ironbus`). `ironbus-core` is IO-free.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,6 +176,7 @@ version = "0.0.0"
 name = "ironbus-storage"
 version = "0.0.0"
 dependencies = [
+ "ironbus-core",
  "tempfile",
 ]
 
--- a/crates/ironbus-storage/Cargo.toml
+++ b/crates/ironbus-storage/Cargo.toml
@@ -8,6 +8,9 @@ license.workspace = true
 repository.workspace = true
 authors.workspace = true
 
+[dependencies]
+ironbus-core = { path = "../ironbus-core", version = "0.0.0" }
+
 [dev-dependencies]
 tempfile = "3"
 
--- a/crates/ironbus-storage/src/io.rs
+++ b/crates/ironbus-storage/src/io.rs
@@ -532,3 +532,48 @@ mod std_file_tests {
         sync_dir(dir.path()).unwrap();
     }
 }
+
+/// Sharing a file behind a reference forwards to the inner implementation, so the
+/// single writer and the lock-free readers can hold the same file.
+impl<F: RandomAccessFile + ?Sized> RandomAccessFile for &F {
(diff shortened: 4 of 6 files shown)


ELares:

An independent review pass over the diff (read against the real `codec`/`segment`/`format`/`types`/`raw` contracts, with throwaway probe tests) found one MAJOR durability defect plus several minor contract/typing gaps. All are addressed in `2ea1ddb`.

**M1 (MAJOR, durability) — footer trusted on content alone.** `detect_footer` accepted the trailing 32 bytes as a sealed footer whenever they decoded (magic/version/algo/CRC valid) and the `segment_id` matched the header, with no cross-check against the records. An unsealed segment whose tail coincidentally, or through attacker-controlled record payload (a ~32-bit CRC grind reachable via `append`), formed a valid footer was misread as sealed, under-reporting the segment length and dropping synced, acknowledged records.
**Fix:** `scan` now trusts a footer only when it describes the body, the record region decodes cleanly up to exactly the footer, and `footer.record_count`/`footer.last_seq` match the recovered records. A footer that disagrees (torn sealed tail, or tail bytes that only look like a footer) is not trusted and the segment is recovered as unsealed. Regression tests `footer_overlapping_record_data_is_not_trusted` and `footer_disagreeing_with_body_is_not_trusted`.

**M2 (MINOR) — footer `record_count`/`last_seq` never cross-checked.** Same root cause; the body match is now part of footer acceptance. Covered by the same regression tests.

**M3 (MINOR) — `wrapping_add` on offsets drifted from the `Offset` no-reuse contract.** `append` now assigns offsets with `checked_add` and returns `SegmentFull` rather than wrapping to a duplicate id; `next_offset` saturates. Regression test `offset_space_exhaustion_is_refused`.

**M4 (MINOR) — sub-header file surfaced a raw IO EOF.** `open` now returns a typed `Segment(Truncated)`. Regression test `short_file_is_typed_truncation_not_io`.

**N1 (NIT) — corrupt-footer-CRC path untested.** Added `corrupt_footer_crc_still_recovers_records`: a sealed segment with a damaged footer is recovered as unsealed and the records survive.

**N3 (NIT) — `valid_end` semantics.** Doc clarified: it is the end of the record region and excludes the trailing footer.

**N2 (NIT, out of scope) — `scan` buffers the whole record region (O(segment) memory).** Correct and bounded for the 64 MiB segment cap; filed as #156 (streaming recovery) rather than absorbed here.

Local gate after the fixes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (72 tests, 27 in storage), `cargo-deny`, SPDX, and the `ironbus-core` IO-free lint all pass.

PR #157, opened 2026-06-06
feat(storage): data-directory Filesystem seam (InMemoryFs + StdFs)

ELares:
## What

A queue's on-disk state is one flat directory of self-describing segment files (#4: "one directory of self-describing files"). This adds the `Filesystem` seam through which the storage engine will create, open, list, remove, and dir-sync those files, rather than calling the OS directly. It is the substrate the multi-segment log and the startup recovery state machine (#52) sit on, and the seam the deterministic simulation substitutes.

## Trait

`Filesystem` (one data directory of flat names): `open`, `create_new` (O_EXCL, never clobbers a segment), `remove`, `list`, `exists`, `sync_dir`. `type File: RandomAccessFile`.

## `InMemoryFs` (sim + tests): directory-entry durability

Models a real filesystem's two-level durability:
- A created or removed directory entry becomes durable only at `sync_dir`.
- `simulate_power_loss` reverts the directory to its last-synced entry set **and** reverts every surviving file to its own last-synced content (composing with `InMemoryFile`'s existing content-durability model).

So the recovery-critical cases are faithfully reproducible:
- a segment created and even **file**-synced, but not **dir**-synced, **vanishes** on power loss;
- an unlink that was not dir-synced is **undone** (the file reappears with its synced bytes);
- a dir-synced create/remove survives.

## `StdFs` (Unix)

Roots at a real directory, hands out positioned-IO `StdFile` handles, and `resolve` rejects any name that is not a single safe path component (`/`, `\`, `.`, `..`, empty, NUL) so a segment name can never escape the data directory. `sync_dir` fsyncs the directory. Windows remains a v1 non-goal (trait + `InMemoryFs` only), consistent with `StdFile`.

## Tests

In-memory: create/open/list/remove/exists, `create_new` refusing an existing file, open-missing → `NotFound`, and all four power-loss directory-durability cases. Unix `StdFs`: create/list/open/remove roundtrip, `create_new` refusal, path-traversal rejection (`../escape`, `a/b`, `.`, `..`, empty, NUL), and that `list` returns files not subdirectories.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (38 in storage), `cargo-deny`, SPDX, and the `ironbus-core` IO-free lint pass locally. CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. Segment file naming/parsing, the multi-segment `Log`, and recovery are separate follow-up PRs.

refs #4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Added
+- `ironbus-storage::fs`: a `Filesystem` seam (open, create-new, list, remove, exists, sync-dir) over one flat data directory of segment files, with an `InMemoryFs` that models directory-entry durability (a created-but-not-dir-synced file vanishes on power loss; a removal that was not dir-synced is undone) and a Unix `StdFs` that rejects names escaping the data directory. This is the substrate for the multi-segment log and recovery.
 - `ironbus-storage::segment`: `SegmentWriter` (append records, sync, seal with footer) and `SegmentReader` (validate header, scan records, recover the valid prefix on a torn or corrupt tail), wiring the record and segment codecs to the `RandomAccessFile` seam. Tests cover round-trip, seal/scan, empty sealed segments, torn-tail recovery, power-loss (only synced records survive), and rejection of a footer bound to a different segment.
 - Cargo workspace skeleton with six crates: `ironbus-core`, `ironbus-storage`,
   `ironbus-proto`, `ironbus-server`, `ironbus-client`, and `ironbus-cli` (binary
--- a/crates/ironbus-storage/src/fs.rs
+++ b/crates/ironbus-storage/src/fs.rs
@@ -46,19 +46,22 @@ pub trait Filesystem: Send + Sync {
     /// Returns [`io::ErrorKind::NotFound`] if the file does not exist, or an IO error.
     fn remove(&self, name: &str) -> io::Result<()>;
 
-    /// Lists the names of the files in the directory, in unspecified order.
+    /// Lists the names of the regular files in the directory, in sorted order.
+    ///
+    /// Only regular files are reported (not subdirectories), and every returned name
+    /// can be passed to [`Filesystem::open`]. The order is deterministic across
+    /// backends so recovery never depends on raw directory order.
     ///
     /// # Errors
     /// Propagates the underlying IO error.
     fn list(&self) -> io::Result<Vec<String>>;
 
-    /// Reports whether a file of this name exists.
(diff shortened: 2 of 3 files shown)


ELares:

An independent review pass attacked the directory-durability model (the whole point of this seam) with probe tests across create/remove/recreate/sync orderings. **The durability model is POSIX-faithful, no blockers.** The key load-bearing property: `sync_dir` clones the directory map's `Arc` pointers, so `live` and `durable` share the same `InMemoryFile` objects (an entry behaves like an inode reference), and a remove-then-recreate-same-name keeps the original inode durable. Four real sim-vs-real consistency issues in the `StdFs` surface were found and fixed in `4ddf096`.

**MAJOR-1 — the Arc-identity invariant was untested.** A future change deep-copying content or reusing the `Arc` on recreate would silently break recovery with every existing test still green. Added `power_loss_after_remove_and_recreate_same_name_restores_original_inode` to lock it in.

**MAJOR-2 — `list`/`open` asymmetry on backslash.** `resolve` rejected any name containing `\`, but `\` is a legal Unix filename byte and `list` did not filter it, so a `a\b` entry was returned by `list` yet refused by `open`/`remove`/`exists` with `InvalidInput`. The recovery walk (list then open each) would choke. `resolve` no longer rejects `\`; the remaining checks (`/`, NUL, `.`/`..`/empty) are what keep a name inside the data directory.

**MINOR (ordering) — `list` order differed between backends.** `InMemoryFs` (BTreeMap) is sorted, `StdFs` (`read_dir`) was arbitrary; code tested only in sim could depend on sorted order and misbehave on ext4/xfs. `StdFs::list` now sorts, and the trait doc promises sorted, openable, regular-file names.

**MINOR (exists) — `StdFs::exists` followed symlinks and returned true for directories** (`try_exists`), while `list` filters to regular files. Now uses `symlink_metadata` and reports true only for a regular file (NotFound → false), matching `list` and the in-memory backend.

**NIT-1 — dead `exists` default removed**; it is now a required trait method (both backends already provide a direct check).

**NIT-2 — a non-UTF-8 entry no longer aborts the whole listing**; foreign non-UTF-8 names (never IronBus segments) are skipped, so everything `list` returns is openable.

**NIT-3 (out of scope) — `InMemoryFile` models `sync_data` as persisting a `set_len` length change**, which a real `fdatasync` need not do. Filed as #158 against the fsync/power-loss contract (#6) rather than absorbed here.

Local gate after the fixes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (40 in storage), `cargo-deny`, SPDX, and the `ironbus-core` IO-free lint all pass.

Issue #37, opened 2026-06-05
[TASK] Specify the advisory manifest plus atomic current-pointer mechanism


ELares:
Parent: #4

Define the mechanism that makes the raw directory listing untrusted while never letting a torn pointer or manifest brick startup. The authoritative source of the live-segment set is the directory scan plus per-segment checksums; the `current` pointer and the `manifest` edit log are accelerators layered on top, and both must survive power loss. This closes a hole in the parent design where the manifest was treated as reliable without specifying its own torn-tail handling.

## Approach

The `current` pointer is a tiny file holding the active segment base offset and the last-flushed offset. It is updated atomically: write a temp file, fdatasync it, rename over `current`, then fdatasync the parent directory so the rename is durable. The `manifest-NNNNNN` file is an append-only, framed, checksummed log of add/seal/delete edits, declared advisory: at startup it is truncated to its longest valid prefix and used only to accelerate the authoritative directory-plus-checksum scan. If `current` fails its checksum or references an absent active segment, recovery falls back entirely to the directory scan and rewrites `current`. All durability primitives go through the cross-platform shim. The replay and fallback rules are handed to #7; the fsync semantics align with #6.

## Acceptance criteria

- [ ] The `current` pointer contents and its atomic write-temp/fdatasync/rename/fdatasync-parent-dir update are specified.
- [ ] The manifest is specified as a framed, checksummed, advisory append-only log truncated to its longest valid prefix.
- [ ] The directory-scan-plus-checksums path is declared authoritative, with the manifest never trusted blindly.
- [ ] The fallback when `current` is corrupt or dangling is defined and handed to #7.
- [ ] Durability calls route through the cross-platform shim and are consistent with #6.

## Failure considerations

A non-atomic pointer update (in-place rewrite) can leave a half-written pointer after power loss; the temp-rename-dirfsync sequence prevents that. A torn manifest tail must be truncated, not replayed, or recovery loads a half-written edit. Making the manifest authoritative would recreate the RocksDB CURRENT single-point-of-failure, so it stays advisory and the directory scan wins.

ELares:
[...]

PR #159, opened 2026-06-06
feat(storage): segment file naming and directory enumeration

ELares:
## What

Recovery discovers segments from the directory of self-describing files (#4), so it needs a canonical, parseable segment file name and a way to enumerate the data directory in segment-id order. This adds `ironbus-storage::naming`.

## Naming

- `segment_file_name(id) -> String`: `seg-<id>.log`, where `<id>` is the 16-digit zero-padded lowercase hex segment id. Sixteen hex digits is exactly a `u64`, so names are **fixed width and their lexicographic order equals numeric segment-id order**, which is the property `StdFs::list` (sorted) relies on so recovery never has to re-sort numerically.
- `parse_segment_file_name(name) -> Option<u64>`: strict and canonical. It accepts only the exact form `segment_file_name` produces (lowercase hex, exact width, no sign), so the two functions are inverses on the set of valid names; any foreign file parses to `None`.
- `segment_ids(fs) -> io::Result<Vec<u64>>`: lists a `Filesystem`, parses the names, skips foreign files, returns ascending ids.

## Why a manifest is not needed for discovery

The directory of self-describing files is the discovery authority. The advisory manifest / atomic current-pointer (#37) is a later optimization, not a correctness dependency, so enumeration is built directly on the filesystem seam.

## Tests

Known names, fixed-width lexicographic ordering, a battery of foreign-name rejections (wrong prefix/suffix, wrong width, non-hex digit, uppercase, leading sign, no prefix), directory enumeration skipping foreign files, and proptest round-trips in **both** directions (`parse∘name = id`, `name∘parse = name`).

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (46 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. Adds `proptest` as a storage dev-dependency (already in the lockfile via `ironbus-core`). CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. The multi-segment `Log` (open data dir, recover the tail, append with rolling) is the next PR and will consume `segment_ids`.

refs #4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Added
+- `ironbus-storage::naming`: segment file naming (`seg-<16-hex-digit-id>.log`, fixed width so lexicographic order equals segment-id order), a strict canonical parser that is the exact inverse of the name function (proptest round-trip both ways), and `segment_ids`, which enumerates a data directory in ascending segment-id order while skipping foreign files. The directory of self-describing files is the discovery authority, no manifest required.
 - `ironbus-storage::fs`: a `Filesystem` seam (open, create-new, list, remove, exists, sync-dir) over one flat data directory of segment files, with an `InMemoryFs` that models directory-entry durability (a created-but-not-dir-synced file vanishes on power loss; a removal that was not dir-synced is undone) and a Unix `StdFs` that rejects names escaping the data directory. This is the substrate for the multi-segment log and recovery.
 - `ironbus-storage::segment`: `SegmentWriter` (append records, sync, seal with footer) and `SegmentReader` (validate header, scan records, recover the valid prefix on a torn or corrupt tail), wiring the record and segment codecs to the `RandomAccessFile` seam. Tests cover round-trip, seal/scan, empty sealed segments, torn-tail recovery, power-loss (only synced records survive), and rejection of a footer bound to a different segment.
 - Cargo workspace skeleton with six crates: `ironbus-core`, `ironbus-storage`,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -177,6 +177,7 @@ name = "ironbus-storage"
 version = "0.0.0"
 dependencies = [
  "ironbus-core",
+ "proptest",
  "tempfile",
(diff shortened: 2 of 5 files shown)


ELares:

An independent review pass empirically verified the naming bijection and the "fixed-width hex ⟹ lexicographic order equals numeric segment-id order" invariant (high-bit set, `u64::MAX`, overflow, Unicode look-alikes, embedded newline, prefix/suffix overlap). **No blockers, no majors.** No `unwrap`/`panic`/indexing in lib paths; `segment_ids` propagates the IO error via `?`. The polish items are addressed in `1cd06b2`.

- **`segment_ids` sort intent.** The re-sort is now documented as deliberate defense in depth: enumeration stays correct even for a hypothetical `Filesystem` backend that lists in arbitrary order, rather than silently depending on the `list()` sorted-order contract. (Both shipped backends already list sorted.)
- **Lowercase-hex check.** Rewrote as `matches!(b, b'0'..=b'9' | b'a'..=b'f')` (clearer; same semantics, uppercase stays rejected so the round trip is canonical, which is why `is_ascii_hexdigit` would be wrong here).
- **Width-drift guard (PR description promised it; now delivered).** `formatter_width_matches_parser_width` ties the `:016x` formatter literal and `ID_HEX_LEN` together so they cannot drift apart unnoticed.
- **Canonical uniqueness.** `near_miss_names_never_alias_a_canonical_id` and a one-digit mutation proptest pin that a near miss parses to a different id or `None`, never a duplicate of the original.
- **Cross-backend equivalence.** A `StdFs`-backed `segment_ids` test asserts the enumeration matches the in-memory backend on a real directory, skipping foreign files and a look-alike subdirectory.

Forward-looking note (no action this PR): `segment_id` is a raw `u64` here, consistent with `SegmentHeader`/`SegmentFooter` in `ironbus-core`; if a `SegmentId` newtype ever lands it is one of several call sites to migrate.

Local gate after the changes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (50 in storage), `cargo-deny`, SPDX, and the `ironbus-core` IO-free lint all pass.

PR #160, opened 2026-06-06
feat(storage): single durable ordered Log (open, recover, append)

ELares:
## What

Wires the segment writer/reader (#155), the filesystem seam (#157), and segment naming (#159) into the first end-to-end durable `Log`: a write path over one active segment that survives a crash with a consistent prefix. This is the first restart-durability property in the storage engine.

## API

- `Log::open(fs, clock) -> Result<Log, StorageError>` — recover or create.
- `Log::append(&Append) -> Result<Offset, StorageError>` — assign the next offset and sequence and write the record.
- `Log::sync()` — fdatasync the active segment (records may be acked after).
- `next_offset()`, `next_seq()`, `active_segment_id()`, `filesystem()`, `into_filesystem()`.
- `Append<'a>` — producer content (timestamp, flags, key, headers, payload); the log owns offset/sequence assignment, the codec derives `HAS_KEY`.

## Recovery (the point of this PR)

`Log::open` enumerates segments via `segment_ids`:
- empty directory → create segment 0, header stamped from the **clock seam** (`now_unix_millis`, no ambient `SystemTime`), `sync` the header, `sync_dir` the entry;
- otherwise → scan the highest segment, **truncate any torn or unsynced tail to the last intact record** (`SegmentScan::valid_end`, the longest-valid-prefix contract), and resume appends after it via the new `SegmentWriter::resume` (reopen at the recovered write head without rewriting the header).

`append` advances the sequence only **after** the write returns `Ok`, so a failed or unsynced append **reuses** its offset and sequence rather than skipping them: no gaps from a dropped tail.

## Tests

- empty open creates segment 0;
- append assigns monotonic offsets/sequences and the records read back;
- reopen recovers the durable records and continues numbering;
- **power loss drops the unsynced tail and resumes**: append+sync A, append B (unsynced), `simulate_power_loss`, reopen → only A survives, `next_offset == 1`, and the next append reuses offset/sequence 1;
- reopening an empty synced log is clean.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (55 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. **Single active segment only** — segment rolling (seal + start the next, #47/#80) and the read path are follow-up PRs. The full startup recovery state machine across many segments is #52; this lays its single-segment core.

refs #4


ELares:

An independent review scrutinized recovery correctness hardest, since this is the first durability/recovery code. **No clean blocker**, but it surfaced two MAJOR defects that the InMemoryFs-only tests structurally could not catch, plus invariant hardening. All addressed in `053af4b`.

**MAJOR-1 — `sync_data` after `set_len` violated the seam's own durability contract.** Recovery truncates the torn tail with `set_len`, whose length change is crash-durable only after `sync_all`; the `RandomAccessFile` contract states this explicitly, but `InMemoryFile` makes `sync_data == sync_all` so no test could expose it. **Switched the post-truncate sync to `sync_all`.**

**MAJOR-2 — a sealed active segment would have its footer destroyed and be overwritten.** For a sealed (footer-bearing) highest segment, `scan.valid_end` excludes the 32 footer bytes, so recovery truncated the footer and `resume`d over it, silently unsealing and corrupting it. **`recover` now returns the typed `StorageError::ActiveSegmentSealed`.** This version never produces a sealed segment, so erroring is strictly safe; it becomes "create the next segment" when rolling lands.

**MINOR-4 — sequence exhaustion was checked after the write.** A successful write at `seq == u64::MAX` then returned `Err`, leaving `next_seq` stale so a retry would reuse the sequence on a durable record. **`append` now reserves the next sequence before the write**, so a write never lands when the space is exhausted.

**MINOR-5 — recovery trusted on-disk sequences blindly.** It now **validates every recovered record carries `base_seq + index`** and derives `next_seq` as `base_seq + record_count` (in lockstep with `next_offset`); a gap is reported as `StorageError::RecoveredSequenceMismatch`.

**MINOR-3 (multiple segments)** — recovery takes the highest segment as the active one, which is the forward-compatible behavior the rolling/read-path PRs build on (predecessor sealed segments are read by the read path, not this write path); no corruption risk, so left as-is rather than erroring.

**Test gaps closed**: sealed-segment open errors, non-zero-`base` recovery, a sequence-gap rejection, a corrupt-mid-record reopen that asserts the file **actually shrank** on truncation, and a **`StdFs`-backed recovery on a real directory** (where `sync_data` and `sync_all` genuinely differ).

Local gate after the fixes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (60 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #161, opened 2026-06-06
feat(storage): segment rolling and multi-segment recovery

ELares:
## What

Lifts the single-active-segment limitation from #160: the `Log` now seals the active segment and rolls to the next when it reaches a configured size, and recovery spans the resulting chain of segments. This is core to the segmented append-only log (#4).

## Rolling

- `LogConfig { max_segment_bytes }` — a **soft** per-segment cap (default 64 MiB, the frozen v1 size).
- `append` seals the active segment and starts the next, **continuing the offset and sequence space**, before the first append that would begin at or beyond the cap. An **empty segment is never rolled**, so a record larger than the cap still gets written (to its own segment) instead of rolling forever. A segment may exceed the cap by at most its last record.

## Crash-safe roll ordering

The old segment is sealed (**durable footer**) BEFORE the new segment becomes discoverable, so a crash mid-roll is recovered by rolling forward. `Log::open` therefore now treats a **sealed highest segment as create-the-next-segment** (the `ActiveSegmentSealed` guard from #160 becomes a handled case), continuing `base_offset`/`base_seq` from the sealed footer. The seal-before-create order is required: creating the new segment first would expose two unsealed segments and lose the old records from the active view.

## Frozen-on-fatal

The active segment is held as an `Option`. A fatal error mid-roll leaves it `None` and **freezes the writer** (`StorageError::WriterFrozen`) rather than resuming over a sealed segment and corrupting it. `next_offset`/`next_seq` are tracked on the `Log`, so the getters work even when frozen.

## Tests

No-roll within a segment; rolling past the cap with **globally monotonic offsets** and sealed predecessors; `base_offset`/`base_seq` continuity across a roll; recovery across multiple segments; **seal-only-crash rolls forward** without overwriting the footer; oversized-record-does-not-roll-forever; power-loss tail recovery; sequence-gap rejection; and a **`StdFs`-backed rolling + recovery** run on a real directory.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (60 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. The read path (reading records across sealed + active segments), the segment reaper/retention (#80/#81), and full predecessor-chain validation (#52/#54) are follow-ups. `Log::open` now takes `clock` by value plus `LogConfig` (an internal API change within this unreleased series).

refs #4


ELares:

An independent review walked every crash point of the roll and the multi-segment recovery. **The roll itself is verified crash-safe** (seal-before-create is necessary and sufficient; every crash point recovers correctly), so no blocker there. The real defects were in recovery, which trusted everything below the highest segment blindly. All addressed in `b31e450`.

**MAJOR-1 — predecessors were never validated.** Recovery scanned only the highest segment, so a corrupt/torn/unsealed predecessor was silently accepted (`Log::open` succeeded on a broken directory; the failure surfaced only when a reader later reached it). **`recover` now scans every segment in ascending order**, so a predecessor that cannot decode fails at open. Non-final segments must be sealed (`UnsealedPredecessor`).

**MAJOR-2 — the highest segment's base was trusted blindly.** A base gap/overlap (hand-built or a recycled/corrupt-but-CRC-valid header) invented a hole in the offset space with no detection. **Recovery now validates the chain**: each segment's `base_offset`/`base_seq` must equal `prev.base + prev.record_count`, else `SegmentChainBroken`. It also checks each segment's stored id matches its file name (`SegmentIdMismatch`).

**MAJOR-3 — `ActiveSegmentSealed` was dead.** The seal-roll path replaced it, leaving it unconstructable with a contradictory doc. **Removed** the variant and its `Display` arm.

**MINOR — `index as u64`** is now `u64::try_from`, matching the module's no-`as` convention.

**MINOR (config floor, m1/n1)** — a `max_segment_bytes` below the header rolls per record; `0` is a legal value that fragments the log. Filed as **#162** (a validating `LogConfig` constructor) rather than changing the config surface here, so small caps still work for tests.

**Test gaps closed**: corrupt-predecessor rejection, base-gap rejection, unsealed-predecessor rejection, resume of an empty active segment after a completed roll (crash point e), **power-loss recovery of the synced prefix across a roll**, a **failed roll freezing the writer** (`WriterFrozen`, getters stay sane, no panic), and a **3+-segment log read back in contiguous global offset order** (the end-to-end continuity check that would catch M2).

Local gate after the fixes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (67 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #156, opened 2026-06-06
[TASK] Streaming segment recovery: bound scan memory instead of buffering the whole record region


ELares:
## Context

`SegmentReader::scan` (added in #155, `crates/ironbus-storage/src/segment.rs`) reads the entire record region into a single `vec![0u8; body_len]` before decoding. For the default 64 MiB segment that is a 64 MiB allocation per scan, with the buffer size driven by `file_len`.

This is correct and bounded for the frozen 64 MiB segment cap, so it is fine for v1. But it forecloses the streaming recovery the module gestures at and scales memory with segment size.

## Task

Recover records by streaming fixed-size windows from the `RandomAccessFile` seam (read a bounded buffer, decode whole frames out of it, slide the window across frame boundaries) so peak recovery memory is O(window + max record), not O(segment). Preserve the exact recovery semantics #155 established:

- stop at the first torn or corrupt frame and return the valid prefix (`valid_end`);
- the body-validated footer decision (footer trusted only if the record region decodes cleanly up to it and `record_count` / `last_seq` match);
- the typed `FooterSegmentMismatch` / `Truncated` errors.

Keep the existing `segment::tests` behavior tests green and add a window-boundary test (a record frame that straddles two read windows).

## Scope

Performance / memory hygiene; no behavior change. Surfaced by the self-review on #155 (item N2). refs #4

ELares:
Resolved by #219 (merged as 4420dbf). Recovery now scans each segment with a streaming reader (`SegmentReader::scan_recovery`) that reads one record at a time and validates it (header + body CRC, sequence continuity, same footer-trust rules) without materializing payloads, bounding peak recovery memory to the largest single record instead of the whole record region. Behavior is pinned equivalent to the buffered scan by a cross-check test and the crash-recovery/determinism sweeps at 20000 cases. The delivery read path (`read_from`) still buffers; streaming it is a separate concern outside this issue's recovery scope and can be a follow-up if needed.

PR #163, opened 2026-06-06
feat(storage): durable read path across segments

ELares:
## What

Adds the read side of the log (#4): consumers can read records back, bounded by what is durable. This is the read half of the at-least-once contract.

## API

- `Log::read_from(start, max) -> Result<Vec<OwnedRecord>, StorageError>` — reads up to `max` records from log offset `start` forward, **crossing segment boundaries** (sealed predecessors into the active segment), in global offset order.
- `Log::flushed_offset() -> Offset` — the durable high-water mark; the first offset not yet on stable storage.

## Durability boundary (the key invariant)

Reads are bounded by `flushed_offset`, so **a reader never observes a record that is not yet durable**. The flush mark advances:
- on `sync` (all appended records become durable and visible), and
- on a **roll** (sealing fsyncs the whole old segment, so its records become durable even before an explicit `sync`).

An unsynced append is therefore invisible to `read_from` until `sync`. `sync` now takes `&mut self` to advance the mark.

## Offset-to-segment mapping

An in-memory segment directory (`id` + `base_offset`, kept sorted) maps an offset to its segment by **binary search**, maintained as segments are created (`start_segment`) and during recovery's chain walk. Reading at or past `flushed_offset` returns empty; reading below the oldest retained offset returns `StorageError::OffsetOutOfRange`.

## Tests

Read within a segment; **flushed-only visibility** (unsynced append invisible until `sync`); read from a middle offset; batch-size limit; empty at/past the flushed offset and `max == 0`; **ordered read spanning multiple segments** with a mid-stream start; read-all after reopen; and the out-of-range guard (a hand-built segment based at offset 5).

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (75 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. CHANGELOG updated. DCO signed.

## Scope

Interim work on the storage engine; `refs #4`. The per-read whole-segment scan is the same buffering tracked for streaming recovery in #156; an offset/time index that avoids re-scanning sealed segments is #54. Lock-free concurrent reads (this is a single-owner `&self` read) and consumer cursors/leases are separate subsystems.

refs #4


ELares:

An independent review traced the durability boundary and the offset-to-segment mapping exhaustively, writing 11 throwaway probe tests. **No blockers, no majors.** The flushed-offset boundary holds: it is impossible for `read_from` to return a record at `offset >= flushed_offset`. `flushed_offset` advances only after `sync` (or a roll's `seal`, which `sync_all`s the whole old segment) returns `Ok`, never past an unsynced record, and never goes backwards on reopen (it is recomputed from the durable on-disk record count). The `segments` vec is always sorted, gap-free, and consistent with disk (no double-push across the recover-walk and seal-roll's `start_segment`), and `segment_index_for` is correct at every boundary. The `sync` → `&mut self` change is sound and `read_from` (`&self`) works against a frozen writer.

Addressed in `668c518`:

- **N1 (ordering)** — `read_from` returns empty for `start >= flushed` before the `start < oldest` out-of-range check. Harmless today (`flushed >= oldest` always holds), now **documented** with a note to reorder once front reaping can advance `oldest` past a small `flushed`.
- **Test gaps** — the review verified six behaviors correct by probe but the PR shipped none. Added: **a roll makes records readable without an explicit sync** (the seal advances the flush mark), read skips an unsynced tail in the active segment, **reads serve the durable prefix from a frozen writer**, a read spanning three segments across two boundaries, a read exactly at `flushed-1`, and a **`StdFs`-backed read across segments**.

**M1 (MINOR, acknowledged)** — `read_from` re-scans the whole segment per call (a 64 MiB read to fetch one record from a full segment). This is the buffering already tracked for streaming in #156 and the offset index in #54; no behavior change here. Noted that the index follow-up is load-bearing for a production read path, not optional.

Local gate after the changes: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (81 in storage), `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #164, opened 2026-06-06
[TASK] Block-layer fault injection: fsync EIO, page-cache reorder, both-slots-torn


ELares:
## Context

The crash-gate harness in `crates/ironbus-storage/tests/crash_recovery.rs` (issues #21, #55) covers power loss before fsync, torn/truncated tails, and byte corruption over the in-memory disk model. The remaining #55 crash classes need a richer fault-injecting file layer than `InMemoryFile::simulate_power_loss`:

- an `fsync`/`sync_data`/`sync_all` that returns `EIO` (the fsyncgate failure mode), asserting the writer freezes read-only and never silently drops the data;
- page-cache reordering: not-yet-fsynced writes persisting out of order, or a strict-prefix subset persisting, rather than the all-or-nothing `simulate_power_loss`;
- a crash at an arbitrary byte boundary of a single write (partial write of N bytes), at every write, rename, and truncation point;
- both-slots-torn and optimistic-checkpoint scenarios once a checkpoint/manifest exists (#37);
- crash-during-recovery idempotence (crash while truncating a torn tail, then re-open).

## Task

Add a `FaultFile`/`FaultFs` fault-injecting layer (a `RandomAccessFile`/`Filesystem` wrapper) that can fail a chosen operation, persist a strict prefix of a write, and reorder/drop unsynced writes deterministically (seeded, no ambient randomness). Extend the crash harness to drive these and assert the same prefix/monotonicity/CRC/bounded-loss invariants, plus: a frozen writer after a fatal fsync, and recovery idempotence under a crash during recovery. Wire the block-layer variants into the nightly CI lane.

## Scope

Extends the per-PR crash gates with the block-layer classes #55 requires. refs #21

ELares:
All block-layer fault classes are now gated and run per-PR (cargo test) and in the nightly deep sweep:
- fatal fsync EIO freezes the writer read-only, no acked record lost: `fatal_fsync_freeze_loses_no_acked_record` and the during-a-roll variant (crash_recovery.rs).
- page-cache reordering / strict-prefix persistence: `power_cut_with_a_holed_unsynced_tail_loses_no_acked_record` (PR #249, a zeroed hole punched into the unsynced tail).
- both-slots-torn checkpoint: `a_both_slots_torn_checkpoint_falls_back_to_redeliver_from_zero` (PR #251, the engine discards a doubly-corrupt checkpoint and redelivers from zero).
[...]

Issue #55, opened 2026-06-05
[TASK] Crash-injection and power-cut harness asserting a consistent recovered prefix at every write boundary


ELares:
Parent: #7

Build the test harness that power-cuts IronBus at every durability-relevant write boundary and asserts that recovery always yields a consistent, monotonic prefix with bounded, correctly reported loss. Recovery correctness is only credible if it is mechanically falsified, so this is the gate that makes the #7 promises real. Lives in #21 and exercises the contracts from #6, #5, and #4.

## Approach
Drive writes through a fault-injecting file layer that can stop the process (simulating SIGKILL and raw power loss) immediately before and after each write, fsync, rename, and truncation, and that can reorder or drop not-yet-fsynced page-cache writes to model power-loss page reordering. After each injected stop, reopen and assert: the recovered log is a prefix of what was acknowledged durable, sequence ids are monotonic with no gaps before the truncation point, every served record passes CRC, and the RecoveryReport's loss_bound matches the actual discarded suffix. Include torn-tail, zero-fill, 0xFF-fill, mid-log rot, both-slots-torn checkpoint, optimistic-checkpoint (must be rejected), and crash-during-recovery (idempotence) scenarios. Run the suite across the platforms in #17 to prove identical behavior.

## Acceptance criteria
- [ ] Injection points cover every write, fsync, rename, and truncation boundary.
- [ ] Page-cache reorder/drop modeling for power loss before fsync.
- [ ] Prefix, monotonicity, and CRC invariants asserted after every injected crash.
- [ ] Idempotence asserted by crashing during recovery and re-running.
- [ ] Loss bound in the RecoveryReport asserted to match the actual discarded suffix.
- [ ] Suite runs in CI across the target platforms of #17.

## Failure considerations
The harness must itself be deterministic and seeded so failures reproduce. A false pass (injection not actually crossing the boundary) is mitigated by asserting the on-disk byte state at each stop. Platform divergence (different fsync semantics) is the key edge risk, surfaced by running the identical suite everywhere rather than on one OS.

ELares:
PR #339 closes the remaining gaps (the suite already covered power-loss-at-every-boundary, torn/truncated tails, mid-log + arbitrary-byte corruption, the fatal-fsync EIO freeze, crash-during-recovery idempotence, both-slots-torn checkpoint, and the sequence-gap reject):

[...]

PR #165, opened 2026-06-06
test(storage): crash-injection gates for the durable log

ELares:
## What

Now that the storage log exists end to end (append, roll, recover, read), this wires in the per-PR **crash gates** the resilience strategy (#21) demands, fulfilling the per-PR slice of #55. `crates/ironbus-storage/tests/crash_recovery.rs` power-cuts the log at durability boundaries and corrupts segments, then asserts the recovery invariants.

## Invariants asserted after every injected crash

- the recovered log is a **consistent, monotonic prefix** (offsets and sequences contiguous from zero, correct payloads);
- **no acknowledged record is lost** (`recovered >= durable`, and exact for the in-memory model);
- **bounded, valid loss** (`recovered <= appended`; recovery never reads past the damage);
- recovery **never panics or hangs**, and CRC-validates every served record (a corrupt frame stops the scan).

## Crash classes gated (in-memory disk model)

- **Power loss before fsync** — at every operation boundary of a fixed workload, **with and without rolling**, plus a **randomized proptest** over both configs. Recovery recovers exactly the durable prefix.
- **Torn tail** — the file truncated mid-record; the partial record is dropped, the prefix kept.
- **Last-record and mid-log corruption** — recovery stops at the first bad record and serves the valid prefix before it (equal-size records make the mid-log slot exact).
- **Arbitrary single-byte corruption** (proptest) — recovery either fails cleanly on a damaged header or returns a valid monotonic prefix; never a panic.

The flushed-offset high-water mark is used as the durability oracle, so the gates hold across rolling (a roll's seal advances durability) without hand-tracking sync points.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (7 crash-gate tests incl. 2 proptests at 256 cases each), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. CHANGELOG updated. DCO signed.

## Scope

`refs #21`, `refs #55`. The block-layer classes that need a fault-injecting file layer (fsync `EIO` → frozen writer, page-cache reorder, partial single write, both-slots-torn, crash-during-recovery idempotence) are filed as **#164** for the nightly lane.

refs #21
refs #55


ELares:
## Self-review (independent mutation-testing pass) and resolution

The review injected 8 plausible recovery bugs into production code and re-ran the harness. **The power-loss / no-acknowledged-loss core is genuinely tight**: a `flushed`-advances-on-append bug, a dropped-durable-record bug, an inflated-head bug, and a zeroed-payload bug each fail 3-6 gates, and the durability oracle was confirmed **non-circular** (pre-crash value from `sync`, post-crash from `recover`'s independent accumulation). But it found two properties the harness advertised yet could **not falsify** (false confidence). Both fixed in `c63c504`.

**MAJOR-2 — body checksum never exercised.** `mid_log_corruption` always flipped a byte at `frame/2 = 26`, inside the 36-byte header's CRC range, so disabling the record **body** CRC left every gate green. Fixed: split into a **header-byte** and a **body-byte** case (the body byte sits past `RECORD_HEADER_LEN`), and added a **payload assertion to both proptests** (a checksum that wrongly accepts a corrupt record now surfaces as a mismatched payload). **Verified by mutation**: with the body CRC disabled, `mid_log_body_corruption_stops_at_the_first_bad_record` and `arbitrary_byte_corruption_...` now both FAIL (they passed before); reverted, tree clean.

**MAJOR-1 — sequence-gap guard not falsifiable by the harness.** No injected corruption ever produced a non-contiguous sequence, so a removed contiguity guard passed all gates (the property was covered only by a separate `log.rs` unit test). Fixed: added `recovery_rejects_a_synthesized_sequence_gap`, which hand-builds a segment with a gap and asserts `RecoveredSequenceMismatch`, so the harness itself owns the monotonic-no-gap invariant it claims.

**MINOR-1/2** — both proptests now assert payload content (was offset/seq only), closing the content-corruption blind spot; the arbitrary-corruption proptest comment notes survived records must have intact bytes.

**NIT-1 (roll density)** — left as-is; the review confirmed `power_loss_at_every_boundary_with_rolling` genuinely crosses rolls (the 16-op workload rolls at appends 5/9/13 under the 256-byte cap), and the proptest covers both configs.

Harness now 9 tests (7 deterministic + 2 proptests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #63, opened 2026-06-05
[TASK] Ack vocabulary, MaxDeliver, BackOff, and crash-atomic DLQ move


ELares:
Parent: #9

Implement the per-message ack vocabulary and the poison-message quarantine path. Verbs: `ack` (done), `nack` (retry now or after a delay, faster than waiting out the timeout), `progress` (extend lease, owned by the lease task), `term` (stop redelivery without DLQ). MaxDeliver caps attempts; on exhaustion the message moves to the DLQ exactly once and emits a MAX_DELIVERIES advisory (#16). The wire verbs are carried by #11; this task owns their server-side semantics and the durable DLQ transition.

## Approach
Default `max_deliver` = 5; unlimited (0) is allowed only behind an explicit `allow_unlimited_deliver` flag plus a startup WARN, else rejected at config load. `nack` optionally takes a delay; a per-consumer `backoff_ms` list supplies escalating delays indexed by attempt. Cumulative ack is offered for broadcast consumers only and hard-rejected for work-groups (the JetStream AckAll trap). The DLQ move is ONE durable consumer-state event committed under a single group-commit fsync (#6): it appends the payload reference to the DLQ topic and tombstones the source in-flight entry atomically, reconciled idempotently on recovery by `(group, offset, attempt)`. DLQ preserves the original enqueue timestamp; config validation requires DLQ retention >= source. DLQ on key_shared groups is opt-in, default OFF; when off, exhaustion parks the message and fires an advisory instead of reordering the stream.

## Acceptance criteria
- [x] A message failing `max_deliver` times moves to the DLQ exactly once, emits an advisory, and no longer redelivers.
- [x] `nack` triggers retry faster than the visibility timeout; `term` stops redelivery without DLQ.
- [x] A crash mid-DLQ-move recovers with the message in exactly one place (idempotent reconciliation).
- [ ] Cumulative ack is rejected on work-groups and accepted on broadcast consumers.
- [ ] DLQ retention < source is rejected at config load; original timestamp is preserved.

## Failure considerations
A crash between appending to the DLQ and tombstoning the source would double-DLQ or drop the message; folding both into one fsynced transition with an idempotency key prevents this. A misconfigured short DLQ retention silently expires poison messages; the load-time check refuses to start. Unlimited max_deliver is reachable but loud, so a single bad payload cannot loop forever by accident.

ELares:
[...]

Issue #62, opened 2026-06-05
[TASK] Lease lifecycle: visibility timeout, progress extend, janitor reclaim, generation fencing


ELares:
Parent: #9

Implement the SQS-style lease as the delivery primitive: a fetched message becomes in-flight for a visibility timeout, only an explicit ack removes it, and on expiry it is reclaimed and redelivered. This task owns the visibility timeout, the lease-extension (progress) call with a hard cap from first receipt, the janitor sweep that reclaims expired leases, and the fencing token that makes cross-member reclaim race-free (the mechanism the parent's no-double-ack criterion depends on).

## Approach
Defaults: visibility 30000 ms, hard lease cap 300000 ms (5 min) from first receipt. Live deadlines are derived from monotonic time within a process run; across a restart all leases are treated as expired. Each lease grant stamps a monotonically increasing per-`(group,offset)` generation token; ack, nack, and progress carry the token they were issued under, and the broker rejects any stale-token operation as a no-op. A `progress` call extends the deadline by one visibility window but never past the hard cap. The janitor runs every 1000 ms (configurable 100 ms to 60000 ms): a single bounded pass over the in-flight set ordered by deadline, early-exiting once deadlines are in the future; expired leases have their generation bumped and are requeued for redelivery to a live member.

## Acceptance criteria
- [ ] A message unacked past its visibility timeout is redelivered; an extended lease (progress) defers redelivery without resetting the hard cap.
- [ ] A dead member's leases are reclaimed by another member after expiry with no double-ack race; a late ack from the original holder carries a stale generation and is rejected.
- [ ] The hard cap bounds total lease from first receipt across any number of progress calls.
- [ ] Restart expires all prior-process leases; the janitor redelivers them on the next sweep.
- [ ] The janitor sweep is O(expired) per pass via deadline ordering, not O(total in-flight).

## Failure considerations
A clock step cannot extend or prematurely expire a lease because deadlines use monotonic time; wall-clock is read only for the persisted first-receipt timestamp, bounded by the cap. A janitor that falls behind under load redelivers slightly late, never incorrectly. Generation-token comparison prevents the zombie-ack corruption entirely; without it a reclaimed-then-reacked offset would corrupt the in-flight set.

ELares:
[...]

Issue #61, opened 2026-06-05
[TASK] Work-group cursor advancement under out-of-order acks (sparse acked-ahead set)


ELares:
Parent: #9

Implement the single hardest mechanism in the consumer model: how a work-group's ONE shared committed cursor advances when acks arrive out of order while older messages are still in-flight. Naively advancing to the highest acked offset silently skips unacked gaps (data loss); refusing to advance until strictly in-order ack collapses the group into an exclusive consumer (no parallelism). This task delivers the correct middle, preserving the at-least-once guarantee from #3 while allowing parallel draining for HyperScale.

## Approach
Maintain `committed_offset` plus an explicit acked-ahead structure: a run-length-encoded set of offsets acked above `committed_offset`. An out-of-order ack lands in the sparse set; whenever the offset immediately after `committed_offset` becomes acked, advance `committed_offset` over the now-contiguous prefix and collapse those entries out of the sparse set. The sparse set is bounded by `max_ack_pending` (it can only hold acked-but-not-yet-collapsed entries, which are a subset of outstanding work), tying it to the #10 lever and guaranteeing it cannot grow without limit. The committed offset and a compact RLE snapshot of the sparse set persist via the consumer-state store.

## Acceptance criteria
- [ ] Acking every offset in a window except the oldest leaves `committed_offset` unmoved and all acked offsets recorded; acking the oldest then jumps the cursor over the whole contiguous run in one step.
- [ ] The acked-ahead structure's memory is bounded by `max_ack_pending` under any ack ordering.
- [ ] After restart, `committed_offset` and the sparse set are restored exactly; redelivery targets only genuinely unacked offsets.
- [ ] Property test over random ack permutations asserts: no offset below `committed_offset` is ever unacked (no silent skip).

## Failure considerations
A crash with a partially-filled sparse set is safe: unacked offsets in the gap are simply redelivered (at-least-once). An adversarial all-but-oldest ack pattern is the worst memory case and is exactly what the `max_ack_pending` bound caps. RLE degenerates to per-offset entries only under maximally fragmented acks, still bounded by the in-flight cap.

ELares:
[...]

PR #166, opened 2026-06-06
feat(core): work-group ack cursor with a sparse acked-ahead set

ELares:
## What

Implements the single hardest mechanism in the consumer model (#61): how a work-group's **one shared committed cursor** advances when acks arrive **out of order** while older messages are still in flight.

- Advancing to the highest acked offset would silently skip the unacked gaps below it: at-least-once violated, because recovery resumes past records that were never processed.
- Refusing to advance until the next offset is acked in strict order collapses the group into an exclusive consumer: no parallelism (no HyperScale).

`AckCursor` is the correct middle.

## Design

- `committed` is the watermark: every offset below it is acked, and it is where a fresh or recovered consumer resumes.
- A compact **run-length-encoded** set holds offsets acked *ahead* of the watermark (the out-of-order acks). `ack` merges adjacent runs so the set stays sorted, disjoint, and non-adjacent.
- Whenever the offset *at* the watermark becomes acked, the watermark **jumps over the now-contiguous run** (and only that one run, since ranges are non-adjacent).
- **Duplicate acks** (below the watermark or already in the set) return `false` and are ignored, which at-least-once permits.
- `resume(committed)` rebuilds the cursor from a recovered offset; the ahead set is **not durable**, so in-flight messages redeliver after a restart (deliberate, matches the lease model where a restart expires all leases).

Pure and **IO-free** — it lives in `ironbus-core`, so the engine logic stays deterministic. Durability of `committed` and bounding the ahead set (via the consumer's max-in-flight) are the caller's responsibility, documented on the type.

## Tests

Unit: in-order advance, out-of-order wait-then-collapse, run bridging, duplicate handling, resume. Proptests:
- acking **any permutation** of `[0, n)` leaves the cursor fully committed at `n` with an empty ahead set;
- after any ack sequence the ahead ranges are **sorted, disjoint, non-adjacent, and above committed**, `committed` is **monotonic**, and `ahead_len` equals the distinct out-of-order offsets;
- `is_acked` agrees with the full ack history.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, and the `ironbus-core` **IO-free lint** all pass locally. CHANGELOG updated. DCO signed.

## Scope

First PR of the consumer subsystem; `refs #61`. The lease lifecycle (visibility timeout, janitor reclaim, fencing — #62), the ack vocabulary and DLQ (#63), and durable persistence of `committed` are follow-ups that build on this cursor.

refs #61


ELares:

An independent review verified the cursor-advancement and RLE-merge logic correct across **5M+ brute-force trials** against an independent oracle (no gap-jump, no mis-merge, no swallowed unacked offset; `advance`'s "at most one contiguous range" assumption is sound given non-adjacency). It found **one MAJOR boundary defect**, fixed in `4ac3255`.

**MAJOR — `offset + 1` overflowed at `u64::MAX`.** `insert` computed the half-open range end with raw `u64` arithmetic. At `offset == u64::MAX`:
- debug builds **panicked** (`attempt to add with overflow`) — a caller-reachable panic in a lib path, which the goal forbids;
- release builds (no `overflow-checks`) **wrapped to 0**, storing an inverted range `(u64::MAX, 0)` so `advance` collapsed `committed` from `u64::MAX` to **0** — a silent at-least-once / data-integrity failure, the exact thing this cursor exists to prevent.

It was also inconsistent with the codebase's own `Offset::checked_next` contract, which deliberately makes `u64::MAX` a loud exhaustion boundary.

**Fix:** `ack` now computes the range end with `checked_add` and **refuses the ack at the boundary** (returning `false`), matching the offset-space-exhausted contract; a real deployment never approaches `u64::MAX`. `committed` can no longer wrap or go backwards.

**MINOR (coverage gap that let it through CI)** — all proptests drew from dense small ranges, so the boundary was structurally unreachable. Added: `acking_the_max_offset_never_overflows_or_collapses_committed` (fresh and resumed), `far_apart_offsets_do_not_merge`, a proptest over the **full `u64` domain including `u64::MAX`**, and a **resume-from-non-zero-base** proptest.

The review also positively cleared (verified, not just asserted): the dup check runs before mutation, the `merge_left && merge_right` index math, `prop_shuffle` determinism, `resume` dropping the ahead set (matches the lease model), and the O(runs) linear scans.

Cursor now 13 tests (5 proptests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #167, opened 2026-06-06
feat(core): lease lifecycle with generation fencing

ELares:
## What

Implements the SQS-style lease as the delivery primitive (#62) — the mechanism the at-least-once **no-double-ack** guarantee depends on. A claimed message is in-flight for a visibility timeout; an explicit ack removes it; on expiry the next claim redelivers it.

## API (`LeaseTable`)

- `claim(offset, now) -> Claim`: grants a free or expired offset (`Granted { token, deliveries }`) or returns `InFlight` for one still actively leased. A redelivery starts a **new attempt**, bumps the generation, and increments `deliveries` (which the caller reads for the max-deliver / DLQ decision in #63).
- `ack(token) -> Acked | Fenced`: removes the lease only if the token's generation matches the current lease.
- `extend(token, now) -> Extended(deadline) | CapReached | Fenced`: defers redelivery by one visibility window, clamped to the hard cap.
- `expired(now) -> Vec<Offset>`: the reclaimable offsets, for the janitor.
- `in_flight()`: count of tracked leases.

## Generation fencing (the crux)

Each grant stamps a strictly increasing generation. `ack`/`extend` carry the token they were issued under; an operation whose generation no longer matches the current lease is a **no-op**. So once a message is redelivered (generation bumped), the original holder's late ack is rejected — **no double-ack race** across members. Acceptance criteria from #62 all covered, including "a late ack from the original holder carries a stale generation and is rejected."

## Hard cap (design note)

`progress`/`extend` never pushes the deadline past `attempt_start + hard_cap`. The cap is measured **per delivery attempt** (reset on redelivery), so each attempt gets its own extension budget — I read "hard cap from first receipt" as the first receipt *of that lease/attempt* (a redelivery is a new lease under a new generation). The total in-flight time across attempts is then bounded by `max_deliver × hard_cap`. If the intended semantic is a single cap from the very first delivery, that's a one-line change — flagging the interpretation.

## Purity

IO-free: the caller supplies monotonic `now` (nanoseconds, from the clock seam) on each call, so the table is fully deterministic and testable with integer time. Across a restart it is empty, so every previously in-flight message redelivers (matches the model). Size is **caller-bounded** via max-in-flight; documented.

## Defaults

Visibility 30 s, hard cap 5 min (`LeaseConfig::from_millis`, frozen #62 defaults).

## Tests

Grant/in-flight, redelivery with a new generation and delivery count, ack-then-fenced, late-ack-after-redelivery fencing, extend with the hard-cap clamp and `CapReached`, stale-token fencing, `expired` ordering, delivery-count increment, and a **proptest** that only the latest token ever acks at any redelivery depth.

## Gate

## Scope

`refs #62`. The janitor's deadline-ordered early-exit sweep (this PR's `expired` is a correct linear scan; a deadline index is the optimization), the ack vocabulary / `nack` / `term` / `MaxDeliver` / DLQ (#63), and durable persistence of the committed cursor are follow-ups built on this and the `AckCursor` (#166).

refs #62


ELares:

An independent review hunted double-ack and message-loss holes and **reproduced a BLOCKER and a MAJOR with failing tests**. Both fixed in `3b9f53a`. It also verified correct (not just assumed): cross-offset ack safety, the `now == deadline` boundary, `expired` idempotency, and panic-freedom.

**BLOCKER — generation saturation → same-offset double-ack.** `next_generation` advanced with `saturating_add`, so at `u64::MAX` it pinned and every later grant, including a redelivery of the **same offset**, reused `u64::MAX`. The redelivery then did **not** fence the prior holder, whose stale token still acked — a silent double-ack, the exact thing this primitive prevents (reproduced: `ack(old_token)` returned `Acked`). **Fix:** `claim` advances the generation with `checked_add` and returns a new `Claim::Exhausted` rather than reusing one, mirroring `Offset::checked_next`'s loud-failure contract. Unreachable in practice, but a correctness primitive must not have a silent-double-ack terminal state. Regression test added.

**MAJOR — initial grant ignored the hard cap.** Only `extend` clamped to `attempt_start + hard_cap`; `claim` set `deadline = now + visibility` unclamped. With `visibility > hard_cap` the first delivery was held past the cap, breaking the documented "can't hold forever" bound (reproduced). **Fix:** `claim` now clamps the initial deadline to the cap, so the cap is a true invariant regardless of config. Regression test added.

**Documented the assumptions the review surfaced** (not fixable inside `LeaseTable`):
- **Cross-restart fencing is not self-sufficient** here — the table resets its generation counter on restart, so the caller must invalidate pre-restart tokens via an outer epoch (server/connection generation). Now stated in the module doc.
- **Orphaned leases** are evicted by the caller driving `expired → claim → max-deliver`; the table bounds count via the caller's max-in-flight. Documented.
- **Degenerate configs** (`visibility`/`hard_cap` of 0, or `visibility > hard_cap`) behave as the clamp dictates. Documented on `LeaseConfig`.

**Test gaps closed:** generation exhaustion, initial-grant hard cap, per-attempt cap reset, multi-offset independence, and the at-deadline boundary.

Lease module now 14 tests (1 proptest). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #168, opened 2026-06-06
feat(core): delivery policy (max-deliver, backoff, ack vocabulary)

ELares:
## What

The pure decision layer of the consumer model (#63): the ack vocabulary, max-deliver poison detection, and the nack backoff schedule. It drives the durable side but performs no IO itself.

## API

- `AckVerb`: `Ack` (done), `Nack { delay_nanos }` (retry, optionally sooner/later than the visibility timeout), `Progress` (extend the lease), `Term` (stop redelivery without DLQ).
- `DeliveryConfig::new(max_deliver, allow_unlimited, backoff_nanos) -> Result`: a **validating constructor** — `max_deliver = 0` (unlimited) is rejected without `allow_unlimited`, since unlimited delivery lets a poison message loop forever.
- `disposition(deliveries) -> Deliver | DeadLetter`: a message that exceeds `max_deliver` (default 5) is poison and routed to the DLQ; an unlimited cap never dead-letters.
- `nack_backoff(attempt) -> u64`: the escalating per-attempt retry delay, **clamped to the schedule's last entry** (empty schedule → no delay).

## How it composes

The lease grant (`LeaseTable::claim`, #167) returns the running `deliveries` count; `disposition(deliveries)` turns that into deliver-or-dead-letter, and `nack_backoff(attempt)` sets the retry delay. The committed cursor (`AckCursor`, #166) advances on `Ack`/`Term`.

## Tests

Cap-then-dead-letter threshold, unlimited opt-in (rejected without the flag, never dead-letters with it), backoff escalation + clamping, empty schedule, verb distinctness. Proptests: **dead-letter iff a finite cap is exceeded**, and backoff **clamps and stays monotonic** on a sorted schedule.

## Gate

## Scope

`refs #63`. The **durable DLQ move** (append to the DLQ topic + tombstone the source in-flight entry under one fsync, reconciled idempotently by `(group, offset, attempt)` on recovery), **cumulative-ack rejection for work-groups**, **DLQ retention validation**, and **original-timestamp preservation** are the server's job and build on this policy with the lease (#167) and cursor (#166).

refs #63


ELares:

An independent review verified the max-deliver boundary, unlimited semantics, and `nack_backoff` indexing correct (no panic / no out-of-bounds across `attempt = 0, 1, len, >len, u32::MAX`, empty/single schedules). No blockers. Two MAJOR items and the minors are fixed in `eece85c`.

**MAJOR — `max_deliver = u32::MAX` was a silent poison loop.** The lease delivery counter saturates at `u32::MAX` (`saturating_add(1)`), so `deliveries > u32::MAX` can never fire: a poison message with that cap would **redeliver forever, never dead-lettered** — a second, undocumented way to get unlimited delivery with no opt-in. **Fix:** `new` now treats both `0` and `u32::MAX` as unlimited and rejects them without `allow_unlimited`. Regression tests added (rejected without opt-in; saturated count still dead-letters under a finite cap).

**MAJOR — the backoff proptest oracle was tautological.** Its expected value recomputed the implementation's own `saturating_sub(1).min(last)` index formula, so a bug edited into both (the natural failure mode) would pass vacuously. **Fix:** the oracle now states the contract **directly** (attempts `1..=len` map to `schedule[0..len]`, beyond `len` clamps to the last, attempt `0` is the first), derived independently of the impl's arithmetic.

**MINOR — Nack-delay vs schedule precedence was undefined.** Added `effective_nack_delay(attempt, explicit)`: an explicit per-nack delay (from `AckVerb::Nack`) **overrides** the schedule; otherwise the schedule applies. The decision is now owned and tested here, not reinvented per server call site.

**MINOR (docs)** — reworded `DeadLetter` ("delivered **more than** `max_deliver` times", matching the `> max_deliver` code) and the module doc (the **server applies** each verb's effect; this layer decides).

**NIT (coverage)** — added `max_deliver = 1` (deliver-once-then-DLQ), `deliveries = u32::MAX`, `max_deliver = u32::MAX`, single-element schedule with `attempt = u32::MAX`, and the precedence cases.

Delivery module now 12 tests (2 proptests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #11, opened 2026-06-05
[DESIGN] Wire protocol and client API


ELares:
This issue specifies the client-facing wire protocol and client API for IronBus: how producers and consumers talk to the broker over a TCP (optionally TLS) connection. The protocol must be trivial to implement and debug like the NATS text protocol, yet carry the length-prefixed binary framing, capability negotiation, and explicit credit frames that an edge message queue needs to stay cheap on the wire and forward-compatible over time. It defines the verb set (PUB, SUB, ACK/NACK, INFO, CONNECT, PING/PONG, FLOW), the connection lifecycle and auth handshake hook, and how frames map onto persisted records. This is a design document to be vetted before any code is written.

## Why this matters for IronBus

The protocol is where four of the five tenets meet on the wire. **Simple**: a client should be implementable in an afternoon and debuggable with a hex dump, the way NATS Core is debuggable with `netcat` because PUB carries an explicit byte count (`PUB <subject> <#bytes>\r\n<payload>\r\n`). **Edge First**: links are metered cellular or lossy Wi-Fi, so per-message overhead, round-trips, and reconnect cost dominate; MQTT's fixed header is as small as 2 bytes precisely for this reason, and one long-lived connection per client avoids per-message setup. **HyperScale**: the broker fans one logical topic out to many consumers, so framing must support zero-copy slicing into the storage path rather than re-encoding. **Cross Platform**: a binary, self-describing frame with explicit lengths parses identically on a 32-bit ARM edge box and an x86 dev laptop with no endianness or alignment surprises. The protocol is also the surface where #10 (credit frames), #18 (the auth handshake), and #9 (ack vocabulary) become concrete.

## Prior art

**NATS Core** uses a tiny line-oriented text protocol over TCP terminated by CRLF: `INFO`/`CONNECT` handshake, `PUB`/`HPUB`, `SUB`/`UNSUB`, `MSG`/`HMSG`, `PING`/`PONG`, `+OK`/`-ERR`. Every payload-bearing verb carries an explicit byte length so the parser never scans for a delimiter inside the body. The server advertises `max_payload`, `headers`, and `jetstream` flags in `INFO`; the client replies in `CONNECT` with `verbose`, `pedantic`, `headers`, `lang`, `version`, and auth fields. `+OK` is sent only in verbose mode to keep the wire quiet. Header frames (`HPUB`/`HMSG`) add a `NATS/1.0` block of `name: value` lines with separate header-bytes and total-bytes counts.

[...]

PR #169, opened 2026-06-06
feat(proto): wire frame envelope (length-prefixed, type-tagged)

ELares:
## What

Starts the wire protocol (#11) with the **frame envelope** every protocol message rides in. Layout: `[ len: u32 LE ][ type: u8 ][ body: len-1 bytes ]`, where `len` counts the type byte plus the body.

## Why this shape

- **Length-prefixed** so framing is independent of the per-type body codecs (later work), and a reader knows a frame's full size up front (zero-copy slicing into the storage path).
- **Cap-before-allocate**: the length is validated against `MAX_FRAME_LEN` (16 MiB + 64 KiB, sized for a max record plus protocol overhead) on the 4-byte prefix alone, so a hostile/corrupt prefix can't force a large read or reservation (the same discipline as the record codec).
- **Forward-compatible**: an unknown type tag still decodes at the envelope level (the length lets a reader skip a frame it doesn't understand), so new verbs don't break old readers.

## API

- `FrameType` — the verb set (`Connect`, `Info`, `Ping`/`Pong`, `Pub`, `Sub`/`Unsub`, `Ack`/`Nack`, `Flow`, `Ok`/`Err`) with a stable one-byte tag (`as_u8`/`from_u8`); `#[non_exhaustive]` for future verbs.
- `encode_frame(type, body, out)` — appends a frame; errors on an oversized body, writing nothing.
- `decode_frame(input) -> FrameDecode` — a **streaming** parser returning `Frame { type_tag, body, consumed }` or `Incomplete { needed }` (the minimum input length to make progress), so a connection frames a byte stream without over-reading. Errors only on a zero or over-cap length prefix.

## Tests

Tag bijection, body and empty-body round-trips, partial-stream `Incomplete` with the exact needed length, consecutive frames from one buffer, zero/oversize rejection (and that nothing is written on a rejected encode). Proptests: **any frame round-trips**, **any strict prefix is `Incomplete`** (never a wrong frame or an error), and an **unknown type tag still frames** its body and length.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. Adds `proptest` as a proto dev-dependency (already in the lockfile). CHANGELOG updated. DCO signed.

## Scope

First PR of the wire protocol; `refs #11`. The per-type body codecs (the PUB subject + byte count, SUB group, CONNECT capability negotiation, FLOW credit, etc.), the connection lifecycle, and the auth handshake hook are follow-ups that fill in the bodies this envelope carries.

refs #11


ELares:

An independent review hammered the streaming parser against hostile network input. **No blockers, no majors.** It verified (not assumed): no reachable panic / out-of-bounds (the two slice ops are reached only after `input.len() >= needed`, with `frame_len >= 1` so the body range `[5..5]` is empty-valid), no DoS allocation (`decode_frame` borrows; the cap check precedes all size-proportional work; `needed` can't overflow `usize` on 32-bit), no desync (`consumed == LEN_PREFIX + frame_len` exactly; advancing lands on the next frame), `Incomplete` always has `needed > input.len()` (no caller infinite-loop), little-endian portability, and forward-compatible unknown tags. Three cheap hardening items addressed in `b60f19c`:

- **M1 — wire tags now pinned exactly.** The bijection test only proved uniqueness + round-trip, so a future reorder could silently shift `Pub` from 5 to 6 and still pass. Added an explicit value assertion per variant, so a reorder breaks a test, not a deployed protocol (the tags are a frozen wire contract).
- **N1 — `decode_frame_with_cap(input, max_len)`** added: the receiving side can reject a frame above a **per-connection negotiated maximum** (the `Connect`/`Info` capability handshake will set it). The effective cap is `min(max_len, MAX_FRAME_LEN)`, so a caller can only tighten the absolute cap, never raise it; `decode_frame` delegates with the absolute cap.
- **N2 — success-boundary regression tests**: a frame at exactly `MAX_FRAME_LEN` decodes, trailing bytes after a frame are not consumed (`consumed` is exactly one frame), and a tighter negotiated cap rejects a frame the absolute decoder accepts.

**M2 (note for the connection PR):** a `decode_frame` `Err` (zero or over-cap length) is unrecoverable for a length-prefixed stream, so the connection layer must treat it as "close the connection," not "skip and retry" — there is no resync point. Carried to the server/client PR.

Frame module now 15 tests (3 proptests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #170, opened 2026-06-06
feat(proto): PUB and ACK frame body codecs

ELares:
## What

Fills in the two message bodies the at-least-once produce/consume path rides in (#11), inside the frame envelope (#169, which owns the length prefix and type tag).

- **PUB** (producer → server): record `flags`, a producer `timestamp_ms`, a **u16-length** `key` and `headers`, then `payload` as the remainder of the body.
- **ACK** (consumer → server): a fixed **25-byte** layout — the `op` (ack/nack/term/progress), the message `offset`, the lease `generation` (the fencing token from #167), and a nack `delay_ms`.

## Safety

Decoding goes through a **bounds-checked, panic-free `Reader`**: a short or inconsistent body is a typed `BodyError` (`Truncated`, `BadLength`, `FieldTooLarge`, `BadAckOp`, `TrailingBytes`), never a slice out of range. Fields are little-endian fixed width with explicit `u16` lengths for the variable parts, so a body parses identically on every target. ACK rejects trailing bytes (fixed layout); PUB treats the payload as the recoverable remainder.

These are **wire types** — the server maps them to the storage (`RecordView`) and consumer (`AckVerb`/`Offset`/lease token) domain types, keeping the wire format decoupled from the on-disk and in-memory representations.

## Tests

PUB and ACK round-trips (empty fields, every op), oversized-key rejection, a key length past the body, **truncation safety** (the framed header errors on a short cut while the payload is the recoverable remainder), unknown-op and short/overlong ACK rejection. Proptests: both round-trips, and a **decode-arbitrary-bytes-never-panics** property over random input.

## Gate

## Scope

`refs #11`. The remaining verb bodies (CONNECT capability negotiation, INFO, SUB group/options, FLOW credit, OK/ERR) and the connection lifecycle are follow-ups; the server wires these bodies onto the storage and consumer layers.

refs #11


ELares:

An independent review traced every slice/index in the decoders and fuzzed short-length hostile inputs (multiple fill patterns, the `0xFFFF`-length-at-offset-9 case, the u16 boundary). **No blockers, no exploitable panics.** The `pos <= buf.len()` invariant holds (`take` advances `pos` only after `get(pos..end)` succeeds), so `rest()` can never panic; `take`'s `checked_add` guards the overflow; no `unwrap`/`panic`/lossy-`as` in lib paths. Three hardening items addressed in `8c81920`:

- **M1 — `AckOp` wire tags now pinned exactly.** `ack_round_trips_every_op` only proved encode/decode move together, so a reorder would silently shift the on-the-wire op and still pass. Added `ackop_tags_have_their_exact_frozen_wire_values` (Ack=0, Nack=1, Term=2, Progress=3), matching the frame PR's bar.
- **M2 — u16 boundary tested.** A `key` and `headers` each at exactly `u16::MAX` round-trip, exercising `push_var`/`var` at the largest field a length prefix can name (the round-trip proptest had capped at 300).
- **M3 — `decode_pub` contract documented.** Payload is the remainder, so the caller must pass exactly one frame's body or trailing bytes fold into the payload (the frame layer hands out an exact body slice; `decode_ack`, fixed-layout, rejects trailing bytes instead).

**Confirmed intentional (no change):** `delay_ms` is transported verbatim for non-Nack ops — a wire codec faithfully carries the bytes; zeroing it would break `decode(encode(x)) == x`, and the server ignores it for non-Nack. `BadLength` is an unreachable-in-practice defensive guard (kept). Raw `u64`/`u8` instead of `Offset`/`RecordFlags` is the correct wire/domain decoupling.

Message module now 13 tests (3 proptests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #133, opened 2026-06-05
[DESIGN] End-to-end golden-path acceptance scenario from install to power-loss recovery


ELares:
Every issue has local acceptance criteria, but nothing exercises the whole promised product story end to end: install the single binary, produce, fan out to many consumers, survive a power cut, recover with bounded reported loss, and read the loss report. The tagline (Simple, Resilient, HyperScale, Edge First, Cross Platform) and #1's success criteria (install-to-first-message under one minute, a stated loss bound, tens of thousands of msgs/sec/core) are currently unbacked by any single executable scenario. This issue owns one scripted golden-path acceptance run that ties the subsystems together and is a release gate.

### The scenario (one script, runs on the aarch64 reference edge device)
1. **Install** the signed static musl binary via the fail-closed curl-to-shell installer (#17); assert fail-closed verification rejects a tampered artifact; assert install-to-first-message < 60 s (#1).
2. **Boot** zero-config on the balanced profile (#14); assert /healthz and /readyz come up, bound to loopback only with no auth required on loopback (#16, #18).
3. **Produce** N records of mixed sizes with default durability; assert every ack carries a durable offset and ack-implies-durable I2 holds (#3, #6).
4. **Fan out** to a broadcast consumer and a competing group with a keyed subset; assert single total durable order, per-group at-least-once, per-key head-of-line in the group (#3, #9).
5. **Overload** producers past the ring; assert spill-to-disk engages, then drop-new with a reported counter, never a silent drop and never an indefinite hang (#10, #13).
6. **Power-cut** mid-batch via the #21 simulated power cut (unsynced page-cache drop + reorder) and via a real dm-flakey run on the reference device.
7. **Recover**: assert a consistent durable prefix (I1), torn-tail truncation, and a structured loss report in three units plus offset range and reason (#7, #8); assert the Prometheus loss counter and the on-disk report file agree (#16).
8. **Resume**: consumers reconnect via stored cursor; a consumer whose offset fell below earliest_retained gets exactly one truncation event and resets (#11, #13).
9. **Inspect offline** with the broker stopped: #15 peek/dump reads only up to the durable high-water mark and reports the same loss as recovery did; exit codes follow the fixed scheme.
10. **Upgrade** in place (atomic swap, ironbus.prev retained) and assert the data dir opens cleanly with no format migration needed within the major version (#17, compat policy).

### Acceptance criteria
[...]

Issue #98, opened 2026-06-05
[TASK] Resilience-counter durability: checkpoint plus replay reconciliation with a monotonicity contract


ELares:
Parent: #16

Make sure a crash never zeroes the numbers that prove how much data was lost. Resilience counters (skipped records/bytes, last-skip-offset, append totals) must survive power loss and remain monotonic non-decreasing across restarts, reconciled against the durable log so no skip is ever forgotten or double-counted. This is co-owned with #7, which is the source of truth.

## Approach

Persist a small counters checkpoint (skipped_records, skipped_bytes, last_skip_offset, append_records, append_bytes) into the #7 manifest alongside the durable log head, fsynced on the same cadence as segment metadata so it is consistent with the log. On restart, seed the in-memory counters from the checkpoint, then let the #7 replay scan ADD any skips discovered in the unscanned tail beyond the checkpoint head. The log is the source of truth; the checkpoint is purely an optimization that avoids a full O(disk) rescan every boot. Counters are monotonic non-decreasing by contract, so a stale or lost checkpoint can only under-count transiently until replay reconciles, never reset to a lower value (which Prometheus would misread as a counter rollover). Any checkpoint-versus-replay discrepancy increments `ironbus_counter_checkpoint_repair_total` and logs once.

## Acceptance criteria

- [ ] After a kill -9 between a skip and the next checkpoint fsync, restart reconciles the skip from replay and the counter is not lower than before the crash.
- [ ] Counters are monotonic non-decreasing across an arbitrary sequence of restarts (property test).
- [ ] `ironbus_last_skip_offset` after restart reflects the true last skip, taking the max of checkpoint and replay.
- [ ] A corrupt or absent checkpoint falls back to a full replay-derived count without panicking.
- [ ] `ironbus_counter_checkpoint_repair_total` increments when replay finds skips the checkpoint missed.

## Failure considerations

The headline risk is the draft's own worst case: a crash zeroing the loss numbers. The seed-then-add design plus monotonicity contract closes it. A torn checkpoint write must be detected and ignored in favor of replay, so the checkpoint carries its own integrity check. Resetting a counter downward would corrupt rate() math fleet-wide, so downward movement is forbidden by contract and tested.

ELares:
[...]

PR #171, opened 2026-06-06
feat(server): single-topic queue engine integrating storage and consumer

ELares:
## What

The **integration milestone**: one synchronous engine that wires the durable log to the consumer primitives, proving the whole design end-to-end and forming the heart the async network server will wrap. It owns a `Log` plus, for one work-group, an `AckCursor` (committed offset, #166), a `LeaseTable` (in-flight leases, #167), and a `DeliveryConfig` (max-deliver/backoff, #168).

## API

- `produce(msg) -> Offset` — append + **fsync before returning**, so a producer's ack is post-durability.
- `poll(now) -> Poll` — claim and return the next deliverable `Message` under a fencing token, or `Parked` (poison), or `Idle`. It scans a **sliding `max_in_flight` window** above the committed cursor (the max-ack-pending bound, so in-flight work never grows unbounded), skips acked and actively-leased offsets, redelivers an expired lease (new generation, higher delivery count), and **parks** a message over max-deliver (commits past it, emits the dead-letter advisory) rather than looping forever.
- `ack(token) -> Acked | Fenced` — removes the lease (fenced if stale) and advances the cursor over the newly contiguous prefix, so **out-of-order acks** work.

## Determinism

Synchronous; the caller supplies monotonic `now` on each call, so the **produce → consume → recover loop is fully testable without a runtime** and the single-writer rule holds (one owner). The async server wraps this; one append actor owns the engine.

## Tests (deterministic, `InMemoryFs` + `ManualClock`)

produce/poll/ack advances the cursor; an in-flight message isn't redelivered until it expires, then redelivers with the **old token fenced**; out-of-order acks advance over the contiguous prefix; the **in-flight window bounds delivery** and slides on ack; a message over max-deliver is **parked**; reopen recovers the durable log and redelivers uncommitted messages.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. Adds `ironbus-core`/`ironbus-storage` as server deps. CHANGELOG updated. DCO signed.

## Scope

`refs #1`, `refs #133` (the golden-path scenario, now reachable in-process). Follow-ups: the **durable consumer cursor** (#98, so a restart resumes instead of redelivering from zero), explicit **nack/term** verbs, the **DLQ topic write** (this parks; #63), **group commit** for produce throughput, multiple work-groups, and the **async TCP server** that frames the wire protocol (#169/#170) onto this engine.

refs #1
refs #133


ELares:

An independent review hammered the integration empirically for at-least-once violations. **No data-loss / double-delivery / unbounded-growth blocker.** It verified: the cursor never commits past a live unacked message (only `ack`-on-Acked and the poison-park advance it); the **max-deliver-to-park boundary is exact** (default 5 → delivered exactly 5 times, parked on the 6th claim, matching the README); no active lease is ever re-delivered (InFlight skip); liveness on expiry holds; `in_flight() <= max_in_flight` under out-of-order churn; and restart is at-least-once-safe (no acked-to-producer message lost). One MAJOR and the silent-boundary minors fixed in `40204e8`.

**MAJOR — `max_in_flight = 0` silently wedged the queue.** A zero window made the delivery loop never run, so every produced message was durably stored but **never delivered and never parked** — a silently dead queue. **Fix:** `open` now rejects it with `EngineError::ZeroMaxInFlight`, the same defensiveness `DeliveryConfig` applies to `max_deliver = 0`.

**MINOR — silent boundaries made loud.** Introduced a typed `EngineError` (`Storage`, `ZeroMaxInFlight`, `GenerationExhausted`, `MissingRecord`). `poll` now surfaces a lease-generation exhaustion and the unreachable empty-read invariant break as **loud errors** rather than an `Idle` indistinguishable from "caught up," so an operator can tell a wedged queue from an idle one (the storage layer is loud at its analogous boundary).

**Confirmed correct, no change:** `produce`'s per-message fsync (durability is correct; a failed sync leaves the record invisible to `poll`, no delivered-offset gap; throughput batching is a network-layer follow-up), and `ack` advancing the cursor only on a genuine ack.

**Test gaps closed:** zero-window rejection; the **default max-deliver=5** boundary end-to-end; a full **deliver → expire → redeliver → ack** retry cycle with delivery-count progression and fenced earlier tokens; **`in_flight` never exceeds the window** across 40 rounds of out-of-order acks under drifting time; and a parked message holds no lease.

Engine now 11 tests. Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #172, opened 2026-06-06
feat(storage): crash-safe double-slot checkpoint for small durable state

ELares:
## What

The durable small-state primitive the consumer cursor (#98), the resilience counters, and a manifest current-pointer (#37) all need: a `Checkpoint` of **two fixed-size, CRC32C-protected slots**.

## Crash-safety mechanism

- Each `write` goes to the slot its **sequence number selects, alternating**, and is **fsynced**; the previous value stays intact in the other slot until the new one is durable.
- On recovery the **higher-sequence slot whose CRC validates wins**, so a slot torn by a crash mid-write fails its CRC and is ignored, and the prior slot survives.
- The checkpoint may **regress** to an earlier durable value after a crash, never to a torn or invented one. For an at-least-once cursor that just means some already-processed messages redeliver (safe); it never advances past a value that was not fully, durably written.
- Sequence `0` means "never written," so a fresh or zeroed file recovers nothing. A payload over the 64-byte slot cap is rejected.

This is the classic two-slot durable checkpoint (the "both-slots-torn" scenario from the crash strategy #21/#55), the substrate for #98 and #37.

## Tests

Fresh recovers nothing, write/reopen round-trip, latest-of-several wins, empty payload, over-cap rejection, and the crash cases: **a torn newest slot falls back to the previous value**, **both slots torn recovers nothing**, and **a power loss before the write syncs keeps the prior value**.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. Adds `crc32c` as a storage dependency (already in the lockfile via `ironbus-core`). CHANGELOG updated. DCO signed.

## Scope

`refs #98`, `refs #37`. The follow-up wires this into the engine so the committed cursor is checkpointed and recovered on restart (resume instead of redeliver-from-zero), and later extends the payload with the resilience counters (#98).

refs #98
refs #37


ELares:

An independent review traced every crash point and fuzzed 5000 random slots. **The crash-safety primitive is verified sound, no blockers**: recovery can never return a torn, partial, or invented value, never advances past a non-durable write, and never panics. The CRC covers seq+len+payload (a tampered length or sequence fails it), the seq-0 sentinel is correct, sequence wrap is unreachable (`checked_add`), and `decode_slot`'s slices are all in-bounds. One **MAJOR** test defect (false confidence in the exact mechanism) and the missing crash cases are fixed in `0d07018`.

**MAJOR — the power-loss test passed for the wrong reason.** It wrote the simulated in-flight seq-2 record to **slot 1** (the just-synced slot) instead of seq-2's alternating **slot 0**, so the double-slot **fallback was never exercised** — the test would have passed even if the fallback logic were broken. **Fix:** two real synced writes to alternating slots (`v1`→slot 1, `v2`→slot 0), then an **unsynced** seq-3 write to slot 1, then power loss; it asserts recovery returns `v2` (the higher durable sequence in slot 0), genuinely taking the fall-back-to-the-other-durable-slot path.

**Crash cases the suite was missing (added):**
- consecutive writes land in **different slots** (a single torn write never destroys both),
- an **exactly-`MAX_PAYLOAD`** (64-byte) round-trip (the cap boundary),
- a **corrupted length field** is rejected (the CRC covers the original length),
- a **corrupted sequence** is rejected (the CRC covers the sequence).

**NITs:** documented the caller's `create_new` + parent-dir-fsync contract on `open` (so the file survives a power loss right after creation), and added a `debug_assert` in `encode_slot` so a removed guard fails loudly rather than silently encoding a wrong length.

Checkpoint module now 12 tests. Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #173, opened 2026-06-06
feat(server): durable consumer cursor via the checkpoint

ELares:
## What

Wires the crash-safe `Checkpoint` (#172) into the engine so **consumer progress survives a restart**. The engine's previous limitation was that a restart redelivered from zero; now it resumes from the last committed offset.

## How

- The committed offset lives in a crash-safe `cursor.ckpt` in the data directory. `checkpoint_cursor()` records it; `Engine::open` recovers it and **resumes the `AckCursor`** there, so a restart redelivers only the **uncommitted tail**.
- The checkpoint is an **optimization, not the source of truth**: it may lag the true committed cursor (a crash then redelivers a few already-processed messages, which at-least-once permits) but **never records an offset that was not committed**, and the recovered offset is **clamped to the durable log head** so it can never resume past a record that did not survive.
- The lease table still starts empty, so anything in flight at the crash redelivers.
- `open` creates the checkpoint file (`create_new` + parent-dir fsync) on first run and opens it thereafter, through the log's filesystem seam.

## Tests

- **checkpoint then reopen resumes** from the committed offset (only the uncommitted tail redelivers, not the acked prefix);
- a fully consumed checkpointed queue is **idle after reopen**;
- a **stale checkpoint** (lagging the true cursor) redelivers only the uncheckpointed tail — **duplicates, never loss**;
- a **`StdFs`-backed** run resumes on a real directory.

## Gate

## Scope

`refs #98`. The checkpoint cadence is caller-driven (`checkpoint_cursor`); a periodic/automatic checkpoint policy, and extending the payload with the resilience counters (skipped records/bytes, #98), are follow-ups.

refs #98


ELares:

An independent review traced the resume logic and verified the critical no-loss property empirically. **No blockers.** Confirmed: `committed` is checkpointed only at the genuinely-acked watermark (it advances only via `AckCursor::ack` over the contiguous prefix, or the deliberate poison-park, never silently); `offset == committed` redelivers after resume (no off-by-one); `produce` syncs before returning, so `flushed_recovered >= committed_checkpointed` always and no durable-but-unacked message can be skipped; and the clamp is down-only (duplicates, never loss). Two hardening items addressed in `38ecf43`.

**MINOR — forward-compatible cursor decode.** The decode required an exact 8-byte payload (`<[u8;8]>::try_from(p.as_slice())`). Once the checkpoint payload is extended to also carry the resilience counters (#98, `MAX_PAYLOAD=64`), an exact-length decode would fail and silently reset the cursor to zero, **redelivering the whole log** (mass duplicates). **Fix:** decode the **leading 8 bytes** (`p.get(..8)`), so recovery survives the format extension.

**NIT — loud clamp.** The clamp (`recovered.min(flushed)`) can only fire if the log recovered below a valid checkpoint (corruption/truncation). Added a `debug_assert` so that invariant break is visible rather than silently swallowed, matching the codebase's surface-invariants-loudly style.

**Test gaps closed:** the resume test now **drains the whole window and asserts the exact set** that redelivers (only the uncommitted tail; the acked prefix never reappears; nothing skipped), and a new test **produces more messages after the checkpoint** and confirms the whole post-checkpoint durable tail survives and delivers after reopen with **no loss**.

Engine now 15 tests (incl. a `StdFs` real-directory resume). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

PR #174, opened 2026-06-06
feat(server): connection session framing the wire protocol onto the engine

ELares:
## What

The connection **session** that turns a byte stream into engine calls and responses, the synchronous glue the async TCP server will drive. It wires the wire protocol (#169/#170) onto the engine (#171/#173).

## API

`Session::process(engine, input, out, now) -> Result<consumed, SessionError>`: consumes the **complete frames at the front of `input`** (leaving a partial trailing frame for the next read, via the frame parser's `Incomplete` signal) and appends response frames to `out`. Dispatches the request/response verbs:

- `Connect` → `Info` (marks the session connected);
- `Ping` → `Pong`;
- `Pub` → decode the message, append it via the engine, reply `Ok` carrying the assigned 8-byte offset;
- `Ack` → decode, commit the delivered message by its fencing token, reply `Ok` (a fenced/stale ack is an idempotent no-op, still `Ok`).

## Error handling

`Pub`/`Ack` before `Connect`, a malformed body, and a recognized-but-unwired or unknown verb reply **`Err` without closing**. A **malformed frame envelope ends the session** (`SessionError::BadFrame`), since a length-prefixed stream cannot resync.

Synchronous and **buffer-driven** — it owns no IO, so the async server feeds it from a socket and it stays fully deterministic/testable.

## Tests

ping/pong; connect/info; **pub-after-connect appends and replies `Ok` with the offset** (and the message is then deliverable from the engine); pub-before-connect rejected; **ack commits a delivered message** (committed advances); a **partial trailing frame is not consumed**; a malformed frame ends the session; an unsupported verb replies `Err` without closing.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (23 in server), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. Adds `ironbus-proto` as a server dep. CHANGELOG updated. DCO signed.

## Scope

`refs #11`. The **streaming consumer-fetch path** (`Sub`/`Flow` credit, the push/pull delivery of messages to a consumer), `nack`/`term`, capability negotiation in `Connect`/`Info`, and the **async TCP accept loop** (one task per connection, a shared engine behind the single-writer append actor) are follow-ups built on this session.

refs #11


ELares:

An independent review fuzzed the frame-dispatch loop against hostile input. **No blockers**: no infinite loop (a frame is always ≥ 5 bytes so `consumed` strictly advances), no out-of-bounds on `&input[consumed..]`, no desync on a bad body (the envelope already delimits it), the connect gate holds, and responses are bounded 1:1. Two MAJORs and the minors fixed in `2be6285`.

**MAJOR (M2) — a fatal engine error is no longer masked as a transient retry.** Every `produce` error was collapsed to a generic `Err("produce failed")` with the session kept open, so a **permanently frozen writer** (no thaw path) would accept Pub after Pub and fail every one forever, the real cause dropped. `EngineError::is_fatal` now marks the unrecoverable variants (`WriterFrozen`, the poll invariant breaks); `handle_pub` queues an `Err` response then returns `SessionError::EngineFatal(e)`, so the caller flushes and **closes the connection**, and the diagnostic is carried for logging.

**MAJOR (M1) — ack ownership documented + filed.** Acks are not connection-scoped: the generation token is the sole authority, so any session with a valid `(offset, generation)` commits (confirmed by a test where a second connection commits the first's message). In the single-work-group model this isn't trivially exploitable, but it's a real authorization gap once consumers share the engine. Documented on `handle_ack` and filed as **#175** for the consumer path.

**MINOR (m3) — ack status reported.** The ack `Ok` now carries a **status byte** (1 = committed, 0 = fenced). A fenced/stale ack no longer silently tells the client the commit took effect, so it knows not to drop state for a message that will redeliver.

**MINOR/NIT (m4, n5):** documented that a repeated `Connect` is idempotent for now (decide when it carries capabilities), and added a `debug_assert` pinning the reply-encode-can't-fail invariant.

**Verified safe (no action):** client `flags` pass-through (the codec normalizes `HAS_KEY` and never acts on unknown bits), and borrowed body slices are copied by `produce` before return.

**Tests added:** a malformed body does not desync (the trailing valid frame still dispatches), a fenced ack replies status 0, empty input consumes nothing, ack-before-connect rejected, and the cross-connection ack (documenting #175). Session now 13 tests. Local gate green (28 in server).

Issue #175, opened 2026-06-06
[TASK] Per-connection ack ownership: scope acks to the delivering session


ELares:
## Context

`Session::handle_ack` (in `crates/ironbus-server/src/session.rs`) builds a `LeaseToken` straight from the wire `(offset, generation)` and calls `Engine::ack`, which checks only the token against the shared `LeaseTable` (the generation fences a stale ack) but never which **connection** holds the lease. So any connected session can ack (commit) any in-flight message if it presents a matching `(offset, generation)`.

In the current single-work-group model the generation token is the sole authority and is hard to guess, so this is not trivially exploitable. But once multiple consumers share the engine (the streaming `Sub`/`Flow` delivery path, a follow-up), a misbehaving or compromised client could commit and suppress redelivery of messages destined for a different consumer, which the legitimate consumer never receives, silent loss from the victim's perspective. Surfaced by the self-review on #174.

## Task

Decide and implement the ownership model when the consumer-delivery path lands:
- track which session (connection) a lease was delivered to (the lease grant returns to a specific consumer), and reject (fence) an ack whose token did not originate from that session;
- or document explicitly that acks are intentionally not connection-scoped (the generation token is the whole authority) and the security model relies on the token being unguessable plus transport auth (#18).
- Add a regression test: a second connection cannot ack a message delivered to the first.

Until then, the behavior is documented on `Session::handle_ack` / `Engine::ack` as a known gap.

## Scope

Consumer authorization boundary; pairs with the streaming consumer path. refs #11 refs #9

PR #176, opened 2026-06-06
feat(server): blocking thread-per-connection TCP server + SystemClock

ELares:
## What

Makes the durable queue **reachable over the network** — the first time the whole stack runs as a real broker over TCP.

- **`SystemClock`** — the production `Clock` seam over the OS clock (wall from `SystemTime`, monotonic from `Instant`), the real implementation `ironbus-core` cannot hold while staying IO-free. The engine takes any `Clock`; tests use `ManualClock`, production wires this.
- **`server::serve`** — a blocking, **thread-per-connection** TCP server. For the bounded connection count an edge box carries, this is the minimal model: **no async runtime**, small binary, simple. It accepts until a shutdown flag is set, **refuses connections past a configurable cap** (so a flood cannot spawn unbounded threads — honoring no-unbounded-queues), and spawns a thread per connection that reads bytes, runs the synchronous `Session` against the engine shared behind a `Mutex`, and writes responses.

## Concurrency

The `Mutex` serializes all engine access into the **single logical writer** the storage layer requires. Group-commit batching behind a dedicated append actor (the goal's append-actor + bounded-channel model, for fsync amortization) is a **throughput follow-up**; correctness (single-writer, fsync-before-ack) holds with the `Mutex`. A malformed frame or fatal engine error flushes any queued response then closes the connection.

## Tests (real localhost sockets)

- **CONNECT + PUB round-trip over TCP** appends to the engine and returns the assigned offset (`Ok` body = offset 0), and the message is then deliverable from the engine;
- a **malformed frame prefix closes the connection** (the client read returns EOF);
- `SystemClock`: monotonic time never goes backwards, wall time is plausibly current.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (32 in server), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally. No new external dependencies (std only). CHANGELOG updated. DCO signed.

## Scope

`refs #11`, `refs #1`. Follow-ups: the **streaming consumer-fetch path** (Sub/Flow delivery to consumers) so a client can receive and ack over TCP; the **append-actor + group-commit** for produce throughput; a graceful connection **drain** on shutdown; per-connection ack ownership (#175); and the **`ironbus` CLI binary** + **client** that connect to this server (toward the golden path #133).

refs #11
refs #1


ELares:

An independent review verified the concurrency core sound: **no deadlock**, **no cap over-admission** (the single accept thread serializes load+add), the lock scope is exactly `session.process` (the socket read and `write_all` are outside the lock), poison recovery is reasonable, and `SystemClock` never goes backwards or panics. It found one BLOCKER and availability hazards, fixed in `7b9e466`.

**BLOCKER — a panic in a connection thread leaked a connection-cap slot.** The decrement ran after `let _ = handle_connection(...)`, which catches a returned `Err` but **not a panic unwind**, so a panicking handler left the slot incremented forever; after `max_connections` panics the server refuses every connection (reproduced by the reviewer). **Fix:** a `ConnectionSlot` drop-guard decrements on both a normal return and an unwind, with a test proving release-on-panic.

**MAJOR — no socket timeout (slowloris).** A client that connects and never sends, or never reads its responses (blocking `write_all`), parked its handler forever holding a cap slot, eventually refusing legitimate clients. **Fix:** a 30s read/write timeout on each accepted stream closes a stalled connection (a ping keeps it alive).

**MINOR — a transient accept error tore down the whole listener.** `EMFILE`/aborted/interrupted accepts now back off and keep serving (the listener always recovers) rather than killing the server.

**MAJOR (tracked, not blocking) — the engine `Mutex` is held across the produce fsync**, so one stalled disk head-of-line-blocks every connection. This is the documented group-commit follow-up; filed as **#177** (append actor + bounded channel + group commit, which also folds in the O(n²) re-decode mitigation from the review). Documented the caveat on the module.

Also fixed the **Windows CI failure**: the server tests now use the cross-platform `InMemoryFs` (`StdFs` is a Unix-only v1 target), so they compile and run on all three platforms.

Server now 33 tests (incl. the panic-slot guard and two real-socket integration tests). Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint all pass.

Issue #177, opened 2026-06-06
[TASK] Append actor + group commit: stop holding the engine lock across fsync


ELares:
## Context

The TCP server (#176) shares the engine behind a `Mutex` and holds it across `Session::process`, which for a `Pub` runs the produce **fsync** while the lock is held. Consequences (surfaced by the self-review on #176):

- A single slow or hung fsync (a stalled/degraded SD card or full flash, the edge deployment target) head-of-line-blocks **every** connection, including pings and acks, for the whole stall, because they all wait on the same `Mutex`.
- Per-`Pub` fsync means no batching: each producer serializes behind every other producer's flush.

## Task

Implement the goal's concurrency model so producers do not block on each other's fsync:

- A single **append actor** thread owns the `Engine` (the offsets and the active segment). Producers fan in via a **bounded channel** (`std::sync::mpsc::sync_channel` is fine, no custom lock-free structure, no loom needed), each carrying a reply channel.
- A **group commit**: the actor drains a batch of pending appends, writes them, issues ONE `fdatasync` per group, then acks the whole batch (the storage layer's one-write-plus-one-sync-per-group). This both removes the head-of-line block (connection threads never hold a lock across fsync) and amortizes the fsync.
- The connection handler sends a command and awaits its reply on the bounded channel instead of locking the engine.

Also fold in the O(n^2) re-decode mitigation from the #176 review: the handler currently re-runs `Session::process` over the whole input buffer on every read; use the frame parser's `Incomplete { needed }` hint to avoid re-decoding until enough bytes have arrived (matters for a client trickling a near-cap frame).

## Acceptance

- No code path holds a lock across an fsync; a stalled fsync on one producer does not block other connections' pings/acks (testable with a fault-injecting file that blocks `sync_data`).
- A batch of concurrent produces issues one fsync, not N.

## Scope

Server concurrency + throughput; refs #176 refs #1

ELares:
[...]

PR #178, opened 2026-06-06
feat: consumer fetch path — the produce/consume loop closes over TCP

ELares:
## What

A client can now **receive and acknowledge** messages, not just produce, so the full at-least-once loop works over the wire. This is the **golden-path round-trip** (#133): produce → fetch → ack over TCP.

## Pieces

- **`ironbus-proto`** — a `Deliver` frame type (tag 13, the frozen wire tag pinned) and a `DeliverBody` codec carrying the message plus the `offset` that names it and the lease `generation` to ack it with. Round-trip + never-panic proptests.
- **`ironbus-storage`** — `Log::now_monotonic` exposes the log's clock so the engine can poll without an externally supplied time.
- **`ironbus-server`** — `Engine::poll_now` polls using the engine's own clock; the **session handles a `Flow` (fetch)** request: decode a `u32` credit, poll up to that many messages, stream each as a `Deliver` frame, **skip a parked (poison) message**, and terminate the batch with an `Ok` carrying the delivered count (so the client knows the batch is complete). A fatal engine error ends the session.

Because the TCP server already drives the session, the **consumer path works over sockets with no server change**.

## Pull-based fetch protocol

`Flow(credit)` → `[Deliver]* Ok(count)`. The client reads `Deliver` frames until the `Ok` terminator, then acks each by its fencing token. This is the SQS-style pull model (no server push), which fits the synchronous session and the edge.

## Tests

The session fetches messages as `Deliver` frames then `Ok`; replies `Ok(0)` when nothing is available; rejects a fetch before connect; a full **produce → fetch → ack round-trip** both in the session and **over a real TCP socket** (the headline test). Proto: `Deliver` round-trip and arbitrary-bytes-never-panics.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (38 in server), `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally on all three platforms. CHANGELOG updated. DCO signed.

## Scope

`refs #11`, `refs #133`. Follow-ups: `nack`/`term`/`progress` over the wire (the verbs reply `Err` for now), the lease's nack-with-delay, capability negotiation, the append-actor/group-commit (#177), and the **`ironbus` CLI** + client library that speak this protocol.

refs #11
refs #133


ELares:

An independent review verified the consumer path safe empirically. **No blockers.** Confirmed: `handle_flow` only **leases** (never acks/commits), so a fetched-but-unsent message (encode error, write failure, client gone) just **redelivers** after the timeout — no loss; the response `out` is bounded by **`max_in_flight`** (16 Deliver frames for `Flow(1_000_000)`), not by client credit, so **no memory-amplification DoS**; the engine commits past a parked message before returning `Parked`, so **no infinite loop** on poison; and `decode_deliver` is panic-free (the bounds-checked Reader). One MAJOR and the doc/test items fixed in `ed5589f`.

**MAJOR — one terminator per fetch.** A non-fatal fetch error sent **both** an `Err` and the `Ok` terminator for one `Flow`, desyncing the length-prefixed stream (the client expects exactly `[Deliver]* <one terminator>`). **Fix:** the non-fatal error path now returns after the `Err` (the `Err` is the terminator).

**MINOR — parked breadcrumb.** Documented that a parked (poison) message is committed past and skipped, and that the dead-letter advisory + DLQ write (#63) is not yet wired, so the consumer isn't told.

**MINOR — overloaded `Ok`.** `Ok` now carries three body shapes (Pub→offset, Ack→status, Flow→count). Documented the conventions on the session module and **filed #179** to split `Ok` into per-verb response frames.

**NIT (test seam) closed.** Added a blanket `Clock for Arc<C>` in `ironbus-core` so a test can share a controllable clock with the engine. New tests now exercise the at-least-once path **through the fetch**: an unacked fetched message **redelivers** after the visibility timeout with a new (fencing) generation, and a **poison message is parked past and skipped** in the fetch. Plus a core test that the `Arc`-shared clock forwards and shares state.

Server now 40 tests (incl. the redelivery/parked fetch tests and the real-TCP round-trip); core 84. Local gate: `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass on all three platforms.

PR #180, opened 2026-06-06
feat(client): synchronous client library (connect, produce, fetch, ack)

ELares:
## What

The **client side** of the wire protocol, so external programs (and the upcoming `ironbus` CLI) can talk to a broker. `ironbus-client` owns one TCP connection.

## API

- `Client::connect(addr)` — connect + handshake (`Connect` → reads `Info`).
- `produce(&PubBody) -> u64` — `Pub`, returns the assigned offset.
- `fetch(max) -> Vec<Message>` — `Flow(max)`, reads the `[Deliver]* Ok` batch, returns each message with its `offset` and fencing `generation`.
- `ack(offset, generation) -> bool` — `Ack`, returns whether it **committed** (`true`) or was **fenced** (`false`, the message will redeliver).
- `ping()` — keepalive.

Responses are framed with a **persistent buffer** so a single read carrying several frames is never lost (the bug the server test helper hit). A server `Err` frame maps to a typed `ClientError::Server(message)`. Blocking and minimal, matching the edge-first server.

## Tests

A full **produce → fetch → ack round-trip against a real in-process server** over a localhost socket: connect, ping, produce, fetch the message, ack it, confirm nothing remains; plus fetching an empty queue returns no messages. (The server is `ironbus-server` driving the real `serve` loop, so this exercises the whole stack client↔TCP↔session↔engine↔storage.)

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace`, `cargo-deny`, SPDX, `ironbus-core` IO-free lint pass locally on all three platforms. Adds `ironbus-proto` as a client dependency. CHANGELOG updated. DCO signed.

## Scope

`refs #11`, `refs #133`. The **`ironbus` CLI binary** (produce/consume/serve from the command line) builds on this client and is the next PR, reaching the golden path. `nack`/`term`, batching, and reconnect are follow-ups.

refs #11
refs #133


ELares:

An independent adversarial review traced the client against the `ironbus-proto` codecs and the `ironbus-server` session it talks to, and empirically confirmed each hypothesis with throwaway tests against a hostile raw-TCP server (all reverted; tree clean).

**Verdict: no BLOCKER.** The proto layer's `MAX_FRAME_LEN` cap already prevents any parse panic, frame desync, and unbounded buffer growth from the frame layer (all verified: a coalesced `Info`+`Ok` is read in order; an over-cap length prefix is rejected before any body is buffered; `decode_deliver`/`from_le_bytes`/`try_from`/`first()` are all bounds-safe; no `unwrap`/`expect`/`panic` in any non-test path). The findings were robustness gaps against a slow, silent, or hostile broker. All addressed in `dd1e7ac`:

| M1 | MAJOR | No connect/read/write timeout: a silent or slow broker hangs the calling thread forever (the server self-defends with a 30s timeout; the client had none). | `ClientConfig` with finite defaults (connect 10s, read 30s, write 30s) applied after connect; `dial` honors a connect timeout via `TcpStream::connect_timeout`. `connect_with` exposes overrides. |
| M2 | MAJOR | `fetch` reads `Deliver` until `Ok` with no count check, so a server can stream deliveries past `max` into an unbounded `Vec` (OOM). | Reject the `(max+1)`-th `Deliver` as `ClientError::BadResponse`. |
| m1 | MINOR (data-loss stakes) | `ack` read `body.first() == Some(&1)`, so an eight-byte pub-shaped `Ok` with low byte 1 was misread as committed, silently dropping the message (violates at-least-once). | Require exactly a one-byte status (`[status]`); any other length is `BadResponse`. |
| m2 | MINOR | `Unexpected(FrameType::Err)` was a misleading stand-in for an unknown tag / wrong-shape `Ok`. | New `ClientError::UnknownFrameType(u8)` (names the raw tag) and `ClientError::BadResponse(&str)`. |
| m3 | MINOR | Tests were happy-path only; server-`Err` mapping, fenced ack, and multi-message fetch were untested. | Added multi-message fetch (real server) and scripted raw-socket cases: server `Err`, fenced ack -> `false`, the eight-byte-`Ok` ack confusion, the over-credit fetch guard, and an unknown tag. |
| n1 | NIT | `ping` ignored a server `Err`. | Handle `Err` consistently with the other four verbs. |

**Gate (local):** `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings -W clippy::pedantic`, `RUSTDOCFLAGS=-D warnings cargo doc`, and `cargo test -p ironbus-client` (8 tests) all green. Awaiting CI on all three platforms.

Issue #136, opened 2026-06-05
[DESIGN] CLI command map: the full command surface, online and offline


ELares:
IronBus ships its broker and its operator/developer tooling in one static binary, so `ironbus` is both the server and the CLI (per #1 and the single-binary build story in #17). This issue freezes the full command map: the verb/noun tree, global flags, output discipline, exit codes, and (the differentiator) which commands run **offline** against a data directory with no broker running versus **online** against a listening broker. The headline capability no competitor ships cleanly is offline forensics: `peek`, `dump`, `segments`, `scrub`, `repair`, and `recovery report` decode the on-disk record frames defined in #5, bounded by the durable high-water mark from #7, and they show exactly what recovery (#7) and corruption-skip (#8) decided to drop, with offsets attached. The design goal is to be strictly better than the `nats` CLI: deeper diagnostics, a frame-accurate view into the stored bytes, and a versioned `--json` contract that is safe to script against.

## Command tree

Top-level verbs and noun groups (every leaf is enumerated in the structured command list):

- `serve` start the broker (online-producing). The default zero-config path from #14.
- `pub` / `sub` quick produce and consume over the wire protocol (#11).
- `tap` live-mirror frames from a running broker; `wire` decode a captured frame dump offline.
- `bench` open-loop load generator (#19) with edge flash-endurance guards.
- `info` summary: against a broker (live offsets, consumers, durability mode) or against a data-dir (segment count, log span, last recovery).
- `consumer` `ls`, `info`, `lag`, `reset`, `purge`, `pause`, `resume`. The mutating leaves (`reset`, `purge`, `pause`, `resume`) are online and require `--confirm`.
- `peek` paged record listing (offset, seq, ts, key, size, codec); `dump` full decode of a single record or range including headers and payload, with `--raw` for the on-disk compressed frame bytes.
- `segments` `ls` (per-segment id, byte span, offset span, seal state) and `verify` (re-CRC every record, report skips).
- `scrub` read-only corruption survey of sealed segments; `repair` produces a quarantine/truncate plan, mutating only under `--apply` (#7, #8).
- `dlq` `ls`, `redrive`, `purge` over the dead-letter queue from #9.
- `retention` `info` (current bounds and what is reclaimable) and `reap` (force an out-of-band reaper pass, online, #13).
- `top` the live TUI (#16 metrics), strictly read-only, degrades to plain text on a dumb terminal.
[...]

Issue #91, opened 2026-06-05
[TASK] CLI surface and contracts: subcommands, flags, exit codes, versioned --json


ELares:
Parent: #15

Specify and implement the `clap`-parsed command surface and the cross-cutting output contracts that every subcommand obeys. This is the spec the acceptance criteria demand: documented synopsis, flags, exit codes, and a committed versioned `--json` schema for `pub`, `sub`, `bench`, `info`, `consumer ls`, `consumer lag`, `peek`, `dump`, `repair`, `scrub`, and `top`.

## Approach

Define global flags once: `--data-dir`, `--json`, `--no-color` (plus honor `NO_COLOR`), `--limit`, and `--confirm` for mutating verbs. Bind defaults (refresh rate, output mode, data-dir) through #14 rather than hardcoding. Adopt a fixed exit-code scheme: 0 clean, 1 usage error, 2 not-found, 3 corruption-handled or threshold-exceeded, 4 corruption-blocked, 5 broker-unreachable, 70+ internal; `--json` still emits a result object on non-zero exit. Make `--json` a committed contract: every object and NDJSON header carries `"schema":"ironbus.cli.<command>.vN"`; additive fields keep N, removals/renames bump N under #22 semver; human tables carry no stability guarantee. Color and any pager are TTY-gated and a pager is never auto-spawned. Online subcommands reuse #18 server credentials resolved via #14; there is no separate CLI auth. `consumer lag` emits lag-records, lag-bytes, and lag-age (age from the stored next-unread record timestamp, never a two-host clock diff) by default.

## Acceptance criteria

- [ ] Every subcommand has a documented synopsis, flags, exit codes, and example output (human and `--json`).
- [ ] The exit-code table is implemented and unit-tested per command.
- [ ] Every `--json` payload carries a `schema` field; an additive-field change is shown not to bump N and a rename is shown to bump it.
- [ ] `NO_COLOR` and non-TTY produce plain ASCII; no pager is auto-spawned.
- [ ] `consumer lag` outputs all three dimensions by default; age is clamped non-negative with a skew note.
- [ ] Online auth resolves through #14/#18 with no CLI-specific credential store.

## Failure considerations

[...]

Issue #60, opened 2026-06-05
[TASK] Durable consumer-state store: cursors, PEL, attempt counters, snapshot and replay


ELares:
Parent: #9

Build the per-consumer durable state layer that survives an unclean restart: committed cursors for broadcast consumers, and for work-groups the shared committed offset plus the per-group Pending Entries List (in-flight set) with a per-message persisted attempt counter. State lives in a separate append-only log under `consumers/<name>/`, NOT inline with the data segments, so the immutable log from #4 is never rewritten to record a per-consumer ack and #8 / #7 invariants hold.

## Approach
Each consumer/group gets a directory holding a compact append-only state-event log: lease-grant, ack, nack, attempt-increment, dlq-move, cursor-commit, each keyed by message offset. Writes fold into the existing WAL group-commit (#6) rather than fsync-per-ack, so ack throughput does not amplify flash wear. Periodically write a snapshot (committed_offset + serialized in-flight set + attempt counters) and truncate the tail behind it. On startup, load the latest valid snapshot and replay the tail to reconstruct exact state; treat all prior-process leases as expired (monotonic basis gone). Per-record CRC (#5) gates every replayed event; a failing record invokes corruption-skip and rebuilds forward from the last snapshot.

## Acceptance criteria
- [ ] A broadcast cursor and a work-group's in-flight set + attempt counter are reconstructed exactly after `kill -9` plus restart, verified by crash injection (#21).
- [ ] Snapshot + tail replay bounds startup cost; replay time does not grow with total log length, only with events since the last snapshot.
- [ ] A corrupt state record is skipped, state rebuilds from the last snapshot, and the rewind is reported as bounded loss.
- [ ] Acks batch into group-commit; no per-ack standalone fsync.
- [ ] State is fully per-consumer: one consumer's corruption does not affect another or the data log.

## Failure considerations
Power loss between an ack and its group-commit fsync loses that ack; recovery redelivers (at-least-once, documented). A torn snapshot is detected by CRC and the prior snapshot plus replay is used. A consumer-state log that grows faster than snapshots compact (pathological nack storm) is bounded by `max_ack_pending` on the in-flight set, capping the live event count.

ELares:
[...]

PR #181, opened 2026-06-06
feat(cli): the ironbus binary, first verbs (serve, pub, sub)

ELares:
## What

The first three verbs of the frozen command map (#136) in the single `ironbus` binary, so the broker and a producer/consumer can be driven from the command line. This is the first end-to-end command-line story toward the golden path (#133).

## Commands

- `ironbus serve --data-dir <dir> [--addr <host:port>] [--max-connections <n>]` run an on-disk broker. Creates the data dir, opens the durable engine on `StdFs`, binds loopback by default, and serves until the process is signalled. **Unix only in v1** (on-disk storage is Unix-only); a clear error on other platforms so the binary still compiles and runs there.
- `ironbus pub [--addr <host:port>] [--key <key>] [<payload>]` connect, append one message (payload from the argument or stdin), print the assigned durable offset.
- `ironbus sub [--addr <host:port>] [--max <n>] [--ack]` fetch up to a credit, print one line per message, and with `--ack` acknowledge each (committed or fenced).

## Conformance to frozen contracts

- **Verb names** `pub` / `sub` / `serve` from the #136 command map (not `produce`/`consume`).
- **Exit-code scheme** from #91: `0` clean, `1` usage, `5` broker unreachable, `70` internal. (`2` not-found and `3`/`4` corruption belong to the offline verbs not yet implemented.)

## Scope and deferrals (explicit, not silent)

- Argument parsing is **hand-rolled** to keep this small surface dependency-free. `clap` is the parser the #91/#136 *approach* suggests; I deferred it together with the larger command surface so the dependency lands under its own review against the **MSRV 1.78** and **cargo-deny** gates (the `deny.toml` allowlist is strict), rather than pulling clap + serde_json + base64 into a 3-command slice. If the maintainer considers clap frozen rather than approach-level, say so and I will adopt it in the next CLI PR.
- The **versioned `--json` contract** and the rest of the tree (`bench`, `info`, `consumer`, `peek`, `dump`, `segments`, `scrub`, `repair`, `top`, offline forensics) are follow-ups under #91 and #136.

## Tests

- `pub` then `sub --ack` round-trip over an **in-process broker** (in-memory fs, cross-platform), and a second `sub` confirms the acked message is gone.
- The same round-trip over a **real on-disk broker** (`StdFs`, temp dir, Unix), exercising the `serve` disk wiring (create dir, `StdFs`, `Engine::open`).
- Usage-error paths (unknown subcommand, no subcommand, missing flag value, bad number, `serve` without `--data-dir`) all map to exit `1`.
- An **unreachable broker** maps to exit `5`.
- Smoke-tested as the built binary end to end (serve, pub from arg and from stdin, sub --ack, exit codes), confirming the durable segment and cursor checkpoint on disk.

## Gate

`fmt --check`, `clippy --all-targets -D warnings` (pedantic), `cargo test --workspace` (274 tests), all green locally. CHANGELOG updated. DCO signed.

refs #91
refs #136
refs #133


ELares:

An independent adversarial review traced the CLI against the `ironbus-client` API, the `serve` loop, and the frozen contracts in #91 (exit codes) and #136 (command map). It verified the high-risk axes are correct and found no blocker.

**Verified correct (so no one re-investigates):** the cross-platform cfg matrix compiles clean on Windows (no unused/missing imports, no dead-code under `-D warnings`); exit-code conformance is faithful (0/1/5/70); a connection-refused correctly maps to Unreachable (5); no argument input panics; usage-before-dial ordering holds (a parse error never dials the broker); engine-open precedes bind (a failed engine never prints "listening"); the connection cap bounds threads and the `ConnectionSlot` drop-guard releases slots on panic, so `serve` leaks nothing; the never-flipped shutdown flag is acceptable for v1 because acks are fsynced before they return.

**Findings addressed in `983654a`:**

| 1 | MAJOR | Mid-session broker death (Io/Closed from produce/fetch/ack) was hard-coded Internal (70), inconsistent with `connect()` which classifies the same variants as Unreachable (5). | Shared `classify()` used at every client call site: `Io`/`Closed` -> Unreachable (5), any answered-but-wrong-shape error -> Internal (70). |
| 2 | MINOR | `--max-connections 0` binds, prints "listening", then refuses every connection (`0 >= 0` in the accept loop). | Reject `--max-connections 0` as a usage error (exit 1). |
| 3 | MINOR | No `--` end-of-options, so a payload beginning with `--` was unpublishable from the argument form. | `pub` honors a bare `--`: every remaining token is the payload. |
| 4 | MINOR | The on-disk test proved a round-trip, not durability. | Reworked to restart: reopen the same data dir and assert the fsynced record survives (a fresh publish lands at offset 1) and redelivers. |
| 7 | NIT | An empty stdin silently publishes an empty message. | Documented in the usage text. |

**Out of scope, filed:** the restart test surfaced that the server never calls `checkpoint_cursor` on the online ack path, so the durable cursor lags at 0 and a restart redelivers all acked messages (safe at-least-once, never loss). The engine already has the checkpoint; only the online wiring is missing. Filed as #182 (relates to #60, #9), not fixed here.

NITs 5 (`--addr=value` equals form) and 6 (a single-dash token treated as payload) are deferred with the `clap` adoption tracked under #91/#136, as called out in the PR description.

**Gate (local):** `fmt --check`, `clippy --all-targets -D warnings` (pedantic), `cargo test --workspace` (274 tests) all green. Smoke-tested as the built binary (the `--max-connections 0` rejection, the `--` payload escape, exit codes). Awaiting CI on all three platforms.

Issue #103, opened 2026-06-05
[TASK] Signed checksummed releases and fail-closed curl-to-shell installer


ELares:
Parent: #17

Produce signed, checksummed per-arch release assets and a curl-to-shell installer that verifies them fail-closed, including on an air-gapped or intermittently-connected edge device. This makes the signed-release work actually meaningful rather than theater under MITM, and it is the channel most operators will touch first.

## Approach

Sign with cosign keyless (Fulcio/Rekor, OIDC identity = the release workflow), and pin the expected signer identity (workflow ref + repo) so verifiers check identity, not merely a valid signature. Because an edge device may not reach Rekor, also publish a minisign/age detached signature using a long-lived offline public key checked into the repo and docs. Publish `SHA256SUMS` for all per-arch binaries and attach the SBOM. The installer detects OS/arch, downloads the matching asset, and verifies both the checksum and a signature against a trust root embedded in the installer script (not fetched over the same channel). It is fail-closed: it refuses to place the binary on any mismatch, with a single, loudly-warned INSECURE override documented separately for air-gapped manual installs. Coordinate the canonical trust-root statement with #22.

## Acceptance criteria

- [ ] Every release asset has a published SHA256 and a cosign signature, plus an offline minisign signature.
- [ ] cosign verification pins the expected signer identity, not just signature validity.
- [ ] The installer verifies checksum and signature against an embedded trust root before install and refuses on mismatch.
- [ ] An offline verification path works without Rekor reachability.
- [ ] A single documented INSECURE override exists for air-gapped use and prints a loud warning.
- [ ] Versioning and signing policy are consistent with #22.

## Failure considerations

A leaked or misconfigured signing identity signing a bad artifact is the critical risk; identity-pinned keyless verification plus a separately-held offline minisign key (both must agree) bounds it, and Rekor makes an unexpected signing event auditable. A warn-only verifier training operators to ignore warnings is avoided by fail-closed behavior. Keyless verification failing offline (pushing operators to bypass) is mitigated by the embedded offline signature path. Trust-root bootstrap over the fetch channel is closed by embedding the key in the script.

ELares:
[...]

PR #183, opened 2026-06-06
ci: static ironbus musl binary for the three edge triples

ELares:
## What

Add a `cross-build` CI job that cross-compiles the static `ironbus` binary for the three edge triples with [`cross`](https://github.com/cross-rs/cross), on every PR:

- `x86_64-unknown-linux-musl`
- `aarch64-unknown-linux-musl`
- `armv7-unknown-linux-musleabihf`

Each target builds `-p ironbus-cli --release --locked`, asserts the result is statically linked (fails if `file` reports a dynamic interpreter), and uploads the binary as an artifact.

## Why now

The single static musl binary for these three triples is a Done(v1) criterion and the goal sequences the musl builds as part of the foundational CI pipeline. Adding it now (rather than at release time) guards every future PR against breaking the musl edge targets: the CLI merged in #181 uses `#[cfg(unix)]` storage paths, and nothing yet proved it cross-compiles to aarch64/armv7 musl. This is the build half of #17; signing, SBOM, reproducibility, and the curl-to-shell installer are follow-ups under #17 and #103.

## Notes

- `cross` is docker-based and covers all three musl targets uniformly (native cross-linking for aarch64/armv7 musl is not apt-installable). The job uses the same `taiki-e/install-action` already used for `cargo-deny`.
- The static-link assertion checks for the *absence* of a dynamic interpreter (`grep -q "dynamically linked"` then fail), which is robust to static-pie vs plain-static `file` wording.
- This job is not yet in branch protection's required set; once it is green here I will add the three matrix checks to required status checks so it gates future PRs.

## Gate

YAML validated. The job validates itself by running on this PR (it cannot be exercised on the macOS dev host, which lacks a linux-musl linker). CHANGELOG updated. DCO signed.

refs #17


ELares:

An independent adversarial review traced the new `cross-build` job for vacuous-pass and false-failure modes. Verdict: 1 BLOCKER, 1 MAJOR, 2 MINOR, 2 NIT. Addressed in `136deb6`:

| 1 | BLOCKER | The `file(1)`-prose static-link check (`grep "dynamically linked"`) is version-dependent and semantically wrong for static-PIE musl binaries. It passed here only because `ubuntu-latest` ships `file 1:5.45` ("static-pie linked"); on older `file` it would **false-negative a correct static binary**, and the substring cannot tell static-PIE from genuinely dynamic. | Replaced with a `readelf` structural check: fail if there is a `PT_INTERP` segment or any `NEEDED` shared library. Robust across `file` versions and target arches. |
| 2 | MAJOR | `rust-cache` + cargo fingerprinting could let the assertion inspect a stale cached binary (`test -x` only proves a file exists, not that this run produced it), masking a future link-level break. | `rm -f` the target binary before the build, so the assertion can only ever see a freshly produced artifact. |
| 3 | MINOR | No `timeout-minutes`; the docker-based cross builds are the long pole and a hung image pull would ride the 6h default. | Added `timeout-minutes: 25`. |
| 4 | MINOR | The global `RUSTFLAGS: -D warnings` is forwarded into the cross container, an undocumented coupling (this job also gates the edge targets for warnings). | Documented with a comment. |

**Verified correct by the review (no change needed):** a failed `cross build` does fail the job (no `continue-on-error`/`|| true`); `crc32c` (the only third-party runtime dep) builds on all three targets via its pure-Rust software fallback with runtime feature detection, no C in `build.rs`; `upload-artifact@v4` works under `permissions: contents: read`; `-p ironbus-cli --release --locked` is the correct invocation and the output path matches the `ironbus` bin name.

**Deferred (NIT):** pinning the GitHub Actions to commit SHAs (and the `cross` docker images via `Cross.toml`) is a real supply-chain hardening but is consistent with the existing jobs' floating-tag style; it belongs in a separate hygiene pass across the whole workflow, not this PR. The `pull_request`-from-fork blast radius is low (read-only token, no secrets).

The previous CI run already proved the binaries build static for all three triples (12/12 green); this revision makes the assertion that guards that claim trustworthy. Awaiting the re-run.

Issue #135, opened 2026-06-05
[DESIGN] WAL under high write rates and on-disk file cleanup (segment janitor and checkpointing)


ELares:
IronBus is one durable, totally ordered, append-only log per instance where the active segment **is** the write-ahead log: a publish is a single framed, CRC32C-checksummed, record-aligned append to that segment, group-commit `fdatasync`'d before ack. That frozen model (see #4, #6) closes one gap and opens another. We specified how a record becomes durable, but never how the log absorbs a sustained high write rate without the on-disk segment set growing without bound, and we never named **who** retires segment files so a 4 GB eMMC does not fill. This issue specifies the seal-then-retire pipeline, the janitor and checkpointer that own it, the exact gates that make a file deletable, and what happens when cleanup cannot keep up. It is a planning issue: the design is the text below.

The central reframing: in PostgreSQL, RocksDB, and SQLite the WAL is a *separate* structure that is recycled or truncated once its contents land in the main store. IronBus has no separate WAL, so there is nothing to truncate ahead of the main log. "WAL cleanup" here means **sealing the active segment and retiring sealed segments**. The active segment is never deleted, exactly as Kafka never deletes its active segment and only retention-deletes whole *closed* segments ([Strimzi](https://strimzi.io/blog/2021/12/17/kafka-segment-retention/)). Retirement is gated by three independent watermarks, and a file is eligible only when it is below the minimum of all three.

## Absorbing a high write rate

Sustained high write throughput is bounded on the ingress side, not by spilling into a second log. The single append actor (#4) serializes appends to one preallocated active-segment fd; the dedicated commit thread (#6) drains queued appends up to the 1 MiB group-commit cap and issues one `write` + one `fdatasync`, with no proactive linger by default so batch size grows naturally under load. fsync stall is a first-class backpressure source: the writer publishes admission credits derived from group-commit headroom (default 4 batches in flight), and exhaustion fires the topic overflow policy in #10 rather than surfacing as an invisible latency spike. This is the lesson PostgreSQL teaches in reverse: there, too-small `max_wal_size` causes "checkpoints are occurring too frequently" and an I/O storm ([PostgreSQL docs](https://www.postgresql.org/docs/current/wal-configuration.html)); IronBus turns the same pressure into explicit producer-visible backpressure instead of a hidden checkpoint stampede.

[...]

PR #184, opened 2026-06-06
feat(server): persist the consumer cursor on the online path (closes #182)

### LINKED ISSUE #182: Server does not checkpoint the consumer cursor on the online ack path: restart redelivers all acked messages
## Problem

The engine has all the durable-cursor machinery (a crash-safe `Checkpoint`, `Engine::open` recovers the committed offset, `Engine::checkpoint_cursor` writes it), but the **server never calls `checkpoint_cursor`** on the online path. `handle_ack` (`crates/ironbus-server/src/session.rs`) commits the in-memory cursor via `engine.ack(...)` and replies, but nothing persists the advance. The only callers of `checkpoint_cursor` are engine unit tests.

Consequence: the on-disk cursor checkpoint stays at its initial value (0). After a broker restart, `Engine::open` recovers committed offset 0 and redelivers the entire uncommitted range, i.e. **every previously acked message redelivers**. This is safe at-least-once (a duplicate, never a loss), and the durable log itself is fine (a fresh publish continues at the correct next offset), but the durable-cursor optimization is effectively inert online: a restart replays everything already processed.

Verified by the `ironbus-cli` on-disk restart test (PR #181, `serve_on_disk_round_trips_and_survives_a_restart`): after ack + reopen, the acked record redelivers and a new publish lands at offset 1.

## What is needed

Wire the existing checkpoint into the online path. Decide and implement the trigger policy, for example:
- periodic (every N acks or every T milliseconds), and/or
- opportunistic (on a clean client disconnect, on idle, on shutdown), and
- bound the redelivery-on-restart window by that policy.

The checkpoint is explicitly an optimization that may lag (a stale checkpoint costs duplicates, never loss), so the policy only trades restart-replay volume against checkpoint write amplification (relevant to edge flash endurance, see #135). No new durability invariant is required; this is an integration + policy task, not a format change.

## Acceptance criteria

- [ ] The server persists the committed cursor on a defined policy, documented with its worst-case restart-replay bound.
- [ ] A restart after steady-state acking redelivers at most the bounded tail, not the whole log (an integration test that reopens a real on-disk broker and asserts the bound).
- [ ] The checkpoint write cadence is bounded (no checkpoint-per-ack fsync storm) and the flash-endurance tradeoff is noted against #135.

Relates to #60 (durable consumer-state store) and #9 (consumer model). Surfaced while implementing the CLI in #181.

ELares:
## What

Close the durable-cursor gap filed as #182: the engine already had the crash-safe checkpoint machinery (`Checkpoint`, `Engine::open` recovers the committed offset, `checkpoint_cursor`), but the **server never invoked it on the online path**, so a restart redelivered every acked message (safe at-least-once, but the durable cursor was inert online).

## Change

- **Engine:** add `EngineConfig.checkpoint_interval` and `Engine::maybe_checkpoint`, which writes the committed cursor once it has advanced at least the interval since the last checkpoint and reports whether it wrote. This bounds crash redelivery to roughly `checkpoint_interval` while keeping the checkpoint write rate far below one per ack (edge flash endurance, see #135). `checkpoint_cursor` now skips a redundant write when nothing advanced. `ack` stays pure.
- **Server:** drive `maybe_checkpoint` best-effort after each successful dispatch, and force a checkpoint on a clean disconnect so a reconnect resumes past acked messages. The checkpoint is a documented lagging optimization: a write failure only costs redelivery on restart, never correctness, so it never fails an ack or tears down a connection.
- **CLI:** `ironbus serve` uses a default interval of 1024; `open_disk_engine` takes the interval so the restart test can pin it to 1.

## Why this is safe

The committed cursor only ever reflects acked messages, so checkpointing it is always a valid lower bound. A checkpoint that lags (or fails) costs duplicates on restart, never loss. This is exactly the contract the checkpoint was already documented to provide.

## Tests

- **Engine bounded-replay** (`maybe_checkpoint_bounds_replay_and_reopen_resumes`): with interval 2, a single ack does not persist (`maybe_checkpoint` returns false), the second ack does, and reopen resumes past both, only the uncommitted tail redelivers.
- **Real on-disk restart** (`serve_on_disk_round_trips_and_survives_a_restart`, updated): with interval 1 the server persists synchronously on ack, so reopening the same data dir resumes (a re-`sub` returns nothing) and the log continues at offset 1.
- **Smoke-tested** across two `ironbus serve` processes on one data dir at the default interval (1024), exercising the force-on-close path: `pub` then `sub --ack`, restart, `sub` returns `fetched 0 message(s)`, `pub` returns offset 1.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (275 tests) all green locally. CHANGELOG updated. DCO signed.

Closes #182


ELares:

An independent adversarial review focused on the central durability question. Verdict: **0 BLOCKER, 0 MAJOR, 2 MINOR, 3 NIT**.

**The key question, answered with a proof: a swallowed checkpoint error CANNOT cause acknowledged data loss.** Message-byte durability and cursor-checkpoint durability are decoupled: `Engine::produce` fdatasyncs the log and advances `flushed_offset` before returning the offset, so the producer's `Ok` is post-fsync; an ack only mutates the in-memory cursor; the checkpoint writes `cursor.committed()` (a watermark, everything below is acked) and `Engine::open` clamps `recovered_offset.min(flushed)`, with the checkpoint's two-slot torn-write protection guaranteeing recovery regresses to an earlier durable value, never a torn/invented one. So `let _ = maybe_checkpoint()` at worst lets the checkpoint lag, costing redelivery of up to `interval-1` already-processed messages on a crash. That is the defined at-least-once behavior, not loss.

Also verified by the review (no change needed): the interval bound is exactly `interval-1` (brute-forced including run-collapse jumps), `last_checkpointed` is updated only after a successful write (so it never leads the persisted checkpoint and the next attempt retries on error), `last_checkpointed` is initialized to the recovered committed offset in `open`, the skip-when-not-advanced guard makes a close-path checkpoint a no-op for a ping-only/produce-only connection, and concurrency is clean (all checkpoint methods run under the engine `Mutex`).

**Addressed in `8cba08c`:**

| 1 | MINOR | No deterministic test for the force-on-clean-close path (the CLI restart test uses interval 1, so its resume is driven by `maybe_checkpoint`, not the close path). | Added `a_clean_disconnect_checkpoints_the_cursor`: drives one connection through `handle_connection` directly (joinable, so the EOF-triggered checkpoint is deterministically complete before reopen), acks one message at the default interval 1024 so `maybe_checkpoint` never fires, then reopens the same filesystem and asserts resume at 1. It fails if the close-path block is removed. |
| 2 | MINOR | The checkpoint fsync runs under the engine `Mutex` (head-of-line blocking), same hazard already documented for produce. | Extended the module CAVEAT to name the cursor checkpoint alongside produce; both move off the lock with the append-actor + group-commit follow-up. |

NIT 1 (`maybe_checkpoint` runs after every dispatch incl. pings) is a harmless cheap threshold check; NIT 2 (the `bool` return is a test-only seam) and NIT 3 (the clamp `debug_assert!` wants a release-mode warn once logging exists) are noted for the observability follow-up, none are correctness issues.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (276 tests) all green. Awaiting CI.

PR #185, opened 2026-06-06
feat: nack over the wire (request redelivery, optionally after a delay)

ELares:
## What

Wire the `Nack` op of the ack vocabulary (#63) end to end. A consumer can negatively acknowledge a delivered message to ask the broker to redeliver it, immediately or after a delay. Until now the session recognized `Nack`/`Term`/`Progress` but rejected everything except `Ack`; a consumer that failed to process a message could only wait out the 30s visibility timeout.

## Change

- **`ironbus-core`** `LeaseTable::nack(token, now, delay_nanos)`: requeues the lease at `now + delay` and **fences the nacking token with a fresh generation**, so a later ack of the same token cannot commit an unprocessed message (a real at-least-once protection against a buggy nack-then-ack). The delivery count is kept, so the next claim escalates it for the `MaxDeliver` decision. A stale token is a no-op.
- **`ironbus-server`** `Engine::nack(token, delay_ms)`: reads the engine's own clock, converts the wire delay (milliseconds) to monotonic nanoseconds, and requeues. The `MaxDeliver` / dead-letter decision stays in the existing `poll` path, so a message nacked past its cap is **parked, not looped**. Returns `EngineError::GenerationExhausted` only on the (astronomically unlikely) generation-space exhaustion.
- **session**: wires the `Nack` op to a one-byte status reply (`1` requeued, `0` fenced), the same convention as ack. `Term`/`Progress` now reply a clear "not yet supported" error instead of a generic one.
- **`ironbus-client`** `Client::nack(offset, generation, delay_ms) -> bool`: `true` if requeued, `false` if fenced.

## Why it is safe

A nack and an ack are both terminal for a delivery attempt. By fencing the token on nack, the broker guarantees that once you nack, your token is spent: a stray ack with it is rejected and the message redelivers, so a nack can never be silently overridden into committing unprocessed data. Redelivery itself goes through the unchanged claim path (generation bump, delivery-count increment, `MaxDeliver` park), so all the existing invariants hold.

## Tests

- **Lease**: requeue + fence-the-holder + reclaimable-at-deadline with escalated delivery count; stale-nack no-op; immediate (delay 0) nack reclaimable at once.
- **Engine**: produce, deliver, nack, the old token is fenced, redeliver with delivery count 2 and a fresh generation, the fresh token commits; stale-nack no-op.
- **Client**: a full produce, fetch, nack, redeliver (fresh generation), stale-token-ack-rejected, fresh-token-ack-commits round trip over a real socket. The default 30s visibility means the redelivery is caused by the nack, not a timeout.

## Scope

`Term` (stop without dead-lettering), `Progress` (extend the lease), the server-side backoff *schedule* (this slice uses the client-supplied `delay_ms`), and a CLI nack surface are follow-ups under #63 / #62.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (282 tests) all green locally. CHANGELOG updated. DCO signed.

refs #63


ELares:

An independent adversarial review traced and empirically probed every loss vector. Verdict: **0 BLOCKER, 0 MAJOR, 2 MINOR, 2 NIT**.

**Headline, confirmed with probes: a nack cannot cause acknowledged data loss or commit-without-processing.** The nack-then-ack fence is airtight (`LeaseTable::nack` verifies the generation matches, then bumps it via `checked_add` before requeueing, so the original token's later ack returns `Fenced` and the cursor never advances for the unprocessed message; the bump and deadline write happen in one `&mut self` call under the single-writer rule, so no claim can interleave). Verified also: a zero-delay nack is reclaimable on the next poll (equality redelivers, monotonic time is non-decreasing); a poison nack-loop parks at `MaxDeliver` (each redeliver `claim` increments `deliveries`, `poll` parks at the cap, probed with `max_deliver=2`); a `u64::MAX` delay saturates to a far-future deadline, never wrapping to immediate; and `Engine::nack` never touches the cursor.

**Addressed in `c631fea`:**

| 1 | MINOR | `Engine::nack`'s `GenerationExhausted` (which `is_fatal`) was reported as a flat string with the connection kept open, unlike produce/flow which end the session on a fatal error, leaving a client hammering a wedged engine. | `handle_ack` now returns `Result` and ends the session with `SessionError::EngineFatal` on a fatal nack error, matching the produce path. |
| 2 | MINOR | `nack` does not clamp the delay to the hard cap (unlike `claim`/`extend`), and the behavior was unstated, inconsistent with the `hard_cap_nanos` doc. | Documented that a nack delay is a retry backoff intentionally exempt from the per-attempt visibility cap (a nack ends the attempt and schedules the next). |
| 3 | NIT | Stale dispatch comment implied the `Nack` frame type and the `Nack` ack-op share a path. | Clarified: a nack arrives as an `Ack` frame with the `Nack` op; the catch-all covers `Sub`/`Unsub` and response-only verbs. |
| 4 | NIT | Client `nack` doc said a `false` return means "already redelivered or acked" only. | Added "or you nacked it before." |

The fatal-nack path is unreachable without `u64::MAX` lease grants, so (consistent with how the produce fatal path is handled) it is verified structurally rather than by a dedicated test.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (282 tests) all green. Awaiting CI.

PR #186, opened 2026-06-06
feat: term and progress over the wire (complete the ack vocabulary)

ELares:
## What

Wire the last two ops of the consumer ack vocabulary (#63). With `nack` already merged (#185), `ack` / `nack` / `term` / `progress` are now all supported; the session previously rejected `Term` and `Progress`.

## Change

- **Term** (an intentional drop): `Engine::term` commits past the message, the same mechanism as `ack`, so it never redelivers and is **not** dead-lettered. It is distinct from `ack` only in the caller's intent (a future metrics or dead-letter-policy split can diverge them); sharing the commit path keeps the cursor and lease invariants identical to a normal ack.
- **Progress** (extend the lease): `Engine::progress` (new `ProgressResult`) reads the engine clock and maps the lease `extend` to `Extended` / `CapReached` / `Fenced`, so a consumer still working on a message is not redelivered, bounded by the existing hard cap.
- **session**: both ops reply a one-byte status (`term`: 1 dropped, 0 fenced; `progress`: 1 extended, 2 cap reached, 0 fenced), the same framing as ack/nack.
- **`ironbus-client`**: `Client::term(offset, generation) -> bool` (dropped vs fenced) and `Client::progress(offset, generation) -> ProgressOutcome`.

## Why it is safe

`Term` reuses the ack commit path, so all the ack invariants (cursor advances only over the contiguous acked prefix, fenced tokens are no-ops) hold unchanged; a term of an unprocessed message is an explicit consumer choice to drop, not a loss. `Progress` only extends a deadline and is clamped to the hard cap, so it cannot keep a message in flight forever.

## Tests

- **Engine**: `term` drops the message and it never redelivers (cursor advances, a stale term is fenced); `progress` extends the lease (the message stays in flight after the original deadline) then returns `CapReached` at `attempt_start + hard_cap`, under a controlled `Arc<ManualClock>`; a stale progress is fenced.
- **Client**: a round trip producing two messages, `progress` on one (Extended), `term` on the other (dropped), then ack the first and confirm nothing remains, plus stale-token `progress`/`term` both fenced.

## Scope

The crash-atomic dead-letter-queue move (a message past `MaxDeliver` is currently parked, committed past, not written to a DLQ) and the server-side backoff *schedule* (nack uses the client-supplied delay) remain follow-ups under #63. A CLI surface for these verbs is a follow-up.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (285 tests) all green locally. CHANGELOG updated. DCO signed.

refs #63


ELares:

An independent adversarial review traced the term/progress paths and confirmed the two definitive questions with five throwaway probe tests (reverted; tree clean). Verdict: **0 BLOCKER, 0 MAJOR, 0 MINOR, 2 NIT**.

**(1) Can a term cause silent loss beyond the consumer's explicit drop? NO.** `Engine::term` delegates to `ack` (`leases.ack` then `cursor.ack` only on `Acked`). A term of a never-delivered offset, a stale token, or a forged `(u64::MAX, u64::MAX)` is fenced and never advances the cursor. The out-of-order case is clean: term offset 5 while 0..5 are unacked lands 5 in the cursor's acked-ahead set without moving `committed`, and acking 0..4 then collapses with no skip and no double-count (probed). A term is mechanically identical to an ack (commit, no DLQ), the consumer's explicit drop, which at-least-once permits; single work-group means no other consumer is owed it.

**(2) Can a progress loop pin a message past the hard cap? NO.** `Engine::progress` → `leases.extend`, which computes `cap = attempt_start + hard_cap_nanos`, returns `CapReached` once `now >= cap`, and clamps any successful extension to the cap. The boundary is exact: at `now == attempt_start + hard_cap` it is `CapReached`, not one more extension (probed by hammering progress every nanosecond: last `Extended` at t=99, `CapReached` at t=100, then the message redelivers with a fresh generation). No off-by-one.

Other checks all passed: no status-byte collision (length-prefixed body, one-byte status, unknown byte is `BadResponse` not a panic), progress is correctly non-terminal (does not bump the generation, so the same token still acks afterward), term is terminal (a later op of the same token fences), no panic on a hostile token, and not-connected / malformed-body are guarded before the op match.

**Addressed in `4da0c0a`:**

| 1 | NIT | The `handle_ack` doc still said "acks or nacks" and the status comment described only the 1/0 convention, now that term/progress are handled and progress can reply `2`. | Updated the doc and the per-op status comment. |
| 2 | NIT | No direct session-layer test for the term/progress wire dispatch (only engine + client round-trip). | Added `progress_then_cap_then_term_over_the_wire` (asserts the raw status bytes 1 then 2 then 1 then 0, and the cursor advance) and `term_before_connect_is_rejected`. |

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (287 tests) all green. Awaiting CI.

PR #187, opened 2026-06-06
feat(cli): golden-path acceptance slice (#133) + serve --checkpoint-interval

ELares:
## What

The first **golden-path acceptance slice (#133)**: an integration test that drives the real, built `ironbus` binary end to end through the part of the promised product story that is reachable today, plus the `serve` knob that makes it deterministic.

## The scenario (`crates/ironbus-cli/tests/golden_path.rs`)

1. **Boot** the broker (`ironbus serve` on an ephemeral loopback port, the test parses the bound port from the broker's first stdout line).
2. **Produce** three messages; assert each `pub` returns the next durable offset (0, 1, 2).
3. **Consume** the batch with `sub --max 10 --ack`; assert all three are delivered and each is `ack committed`.
4. **Restart** the broker on the same data directory.
5. **Resume**: assert the acked messages do **not** redeliver (`fetched 0 message(s)`, the cursor was durable) and the durable log continues (`pub` returns offset 3).

A `ChildGuard` kills and reaps the broker on drop, so a failing assertion never leaks a serve process. The test is gated to Unix because `serve` is Unix only in v1.

## Also: `ironbus serve --checkpoint-interval <n>`

Tune how many messages may be redelivered after an abrupt crash (the cursor-checkpoint interval), default 1024. A low value persists the consumer cursor more eagerly. The golden-path test uses `1`, so the cursor is checkpointed synchronously on each ack and the restart-resume is **race-free** (no dependence on the asynchronous close-path checkpoint timing).

## Scope

This is a faithful slice, not all of #133. The remaining steps (broadcast plus competing consumer groups, overload spill-to-disk, a simulated power cut, the structured loss report, the fail-closed installer, in-place upgrade) need subsystems that are not built yet, so #133 stays open and this is `refs #133`.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (288 tests, including the new end-to-end binary test) all green locally. CHANGELOG updated. DCO signed.

refs #133


ELares:

An independent adversarial review verified the test is a genuine end-to-end proof and exercised it for flakiness. Verdict: **0 BLOCKER, 0 MAJOR, 2 MINOR, 3 NIT**.

**Proven genuine (not a false pass):** the reviewer neutered `Engine::maybe_checkpoint` and `Engine::checkpoint_cursor` to no-ops and re-ran, the test FAILED exactly at the restart-resume assertion (all three messages redelivered), then reverted. This confirms the second broker is a fresh process reading the same on-disk dir, `fetched 0 message(s)` is load-bearing on the durable cursor, and `pub` returning offset 3 is a real log-durability check (produce fsyncs before returning the offset). **Resume is race-free:** under the engine lock, the ack advances the cursor and `maybe_checkpoint` fsyncs it BEFORE the lock drops and BEFORE the ack reply is written, so the cursor is durable before the `sub` client sees the reply and exits, well before the test kills the broker. No port race (the listening line prints after bind), no stdout-pipe deadlock, ephemeral ports only, PID-suffixed dirs. Ran 5x stable.

**Addressed in `961ce14`:**

| 1 | MINOR | The `ChildGuard` was constructed AFTER reading stdout, so a panic in `start_broker` before that (read error or unparseable listening line) dropped a bare `Child`, which does not kill on drop, orphaning the broker. | Construct the guard immediately after `spawn`; every early exit now kills and reaps it. |
| 2 | MINOR | No timeout on the listening-line read: a broker that wedged without printing or exiting would hang the test until the CI job timeout. | Read the line on a worker thread bounded by a 10s `recv_timeout`. |

NITs (all confirmations, no action needed): the `DEFAULT_CHECKPOINT_INTERVAL` cfg change is correct on both platforms (the uncfg'd `run_serve` uses it, the non-unix `cmd_serve` consumes it via `let _`); the test cannot false-pass on a bad addr (every pub/sub asserts exit 0); the CHANGELOG honestly scopes the slice and does not over-claim true multi-consumer fan-out.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (288 tests, golden-path ran 5x stable) all green. Awaiting CI.

PR #188, opened 2026-06-06
feat(cli): sub --nack and --term dispositions

ELares:
## What

Expose the full consumer ack vocabulary through the `ironbus` binary. The library and wire protocol gained `nack`/`term`/`progress` in #185 and #186, but the CLI `sub` could only `--ack`. Now `sub` takes at most one disposition for the fetched batch:

- `--ack` commit each message (existing)
- `--nack [--delay-ms <n>]` requeue each for redelivery after the delay (immediately if omitted)
- `--term` drop each without dead-lettering
- (none) **peek**: print only, the messages stay in flight and redeliver after the visibility timeout

A second disposition is a usage error, and `--delay-ms` requires `--nack`. Each message prints its per-disposition result (`committed` / `requeued` / `dropped` / `fenced`).

(`progress` is for long-running single-message work, not a batch disposition, so it stays a library-only verb for now.)

## Tests

- Usage guards: `sub --ack --nack` and `sub --ack --delay-ms 5` both exit 1 (usage).
- Round trip over an in-process broker: `sub --nack` requeues a message and it redelivers on the next fetch (the default 30s visibility means the nack, not a timeout, brings it back), then `--ack` commits it; `sub --term` drops a message and a re-fetch is empty.
- Smoke-tested as the built binary: `sub --nack` then `sub --term` then a peek returning `fetched 0 message(s)`, and `--ack --nack` rejected.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (291 tests) all green locally. CHANGELOG updated. DCO signed.

refs #63
refs #136


ELares:

An independent adversarial review verified the parsing, dispatch, and tests. Verdict: **0 BLOCKER, 0 MAJOR, 1 MINOR, 3 NIT**, no correctness defect.

**Verified correct:** mutual exclusion is airtight (every disposition routes through `set_dispose`, a second is rejected, all orderings); `--delay-ms` validation is order-independent (checked after the parse loop); no parse panics; mid-batch disposition failure stops cleanly via `?` with `classify` mapping Io/Closed to exit 5, already-disposed messages stay committed, un-reached ones redeliver, and a re-fetch issues a fresh generation so there is no double-act; `fenced` reports without erroring; the nack round-trip redelivery is the nack's doing (the in-process server's 30s default visibility means it cannot be a coincidental timeout in a sub-second test).

**Addressed in `3a0c52f`:**

| 1 | MINOR | The disposition slot was stringly-typed (`Option<&'static str>` matched against `Some("nack")`), so a typo in a verb literal would silently fall through to peek with no compile error. | Replaced with a typed `DispositionKind` enum; the build-the-`Disposition` match is now exhaustive and compiler-checked (`None => Peek` is explicit). |
| 2 | NIT | No test covered `--delay-ms` before `--nack` (order independence). | Added `nack_delay_is_order_independent`: with no broker it parses then fails to connect (exit 5), not a usage error. |

NIT 1 (confirming `--delay-ms 0` is not special-cased) and NIT 3 (the help line could mention the delay default) need no code change; the CHANGELOG covers the defaults.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (292 tests) all green. Awaiting CI.

PR #189, opened 2026-06-06
feat: server-side escalating nack backoff schedule (#63)

ELares:
## What

Apply the per-attempt **nack backoff schedule** (#63). A nack with no explicit client delay now defers redelivery by the configured escalating schedule instead of redelivering immediately, so a flapping consumer backs off rather than hot-looping a retry. Until now `nack` honored only the client-supplied `delay_ms`; the `nack_backoff` schedule existed in `delivery.rs` but was never applied.

## Change

- **`ironbus-core`** `LeaseTable::deliveries(token) -> Option<u32>`: the live attempt count (1 for the first delivery) the schedule is indexed by, or `None` for a stale token.
- **`ironbus-server`** `Engine::nack`: when `delay_ms == 0`, the delay is `nack_backoff(attempt)` (the configured schedule, indexed by the delivery count, clamped to the last entry); an explicit `delay_ms > 0` still overrides. An empty schedule yields `0` (immediate), so existing tests and the in-process server are unchanged.
- **`ironbus-cli`** `ironbus serve` ships a default schedule of **100 ms, 500 ms, 2 s, 10 s, 30 s**.

## Semantics

The schedule is the default retry policy; a client overrides per-nack with `--delay-ms`. `delay_ms == 0` means "no explicit delay, use the schedule" (an empty schedule then means immediate, the prior behavior). Escalation is by attempt: the first nack uses `schedule[0]`, the second `schedule[1]`, clamped to the last entry, since each redelivery increments the lease's delivery count.

## Tests

- **Lease**: `deliveries` returns the attempt for a live token, `None` for a stale token (after redelivery) and after ack.
- **Engine**: with a `[50 ns]` schedule, a `nack(0)` defers redelivery to `now + 50` (a poll before that is `Idle`, after it redelivers with the attempt escalated); with the same schedule a `nack(1 ms)` is governed by the client's 1 ms, not the 50 ns backoff (proving the override).
- **Smoke-tested** as the binary: `sub --nack` (no delay) requeues, and an immediate re-`sub` returns `fetched 0 message(s)` because the 100 ms backoff has not elapsed.

## Scope

This is the BackOff part of #63. The crash-atomic dead-letter-queue move (a message past `MaxDeliver` is still parked, not written to a DLQ) is the remaining piece and stays open. A per-consumer (vs per-broker) backoff configuration is a follow-up; this wires the server-wide default.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (295 tests) all green locally. CHANGELOG updated. DCO signed.

refs #63


ELares:

An independent adversarial review confirmed the core question with a multi-attempt probe. Verdict: **0 BLOCKER, 0 MAJOR, 2 MINOR, 2 NIT**.

**Verified correct:** the attempt-to-schedule indexing is correct and escalating (attempt 1 -> `schedule[0]`, attempt 2 -> `schedule[1]`, then clamp; `attempt 0` from a missing token maps to `idx 0` via `saturating_sub`, but `leases.nack` fences it so the delay is moot); the `deliveries` read happens before `leases.nack` and `nack` does not bump the count, so the index reflects the attempt being nacked; the empty-schedule path (all in-process servers, the engine `config()` helper) yields `0`, i.e. unchanged immediate redelivery, so no existing nack timing changed; the `cfg(unix)` gating of the default schedule is clean on Windows.

**Addressed in `c8f0819`:**

| 1 | MINOR | The engine flattened the delay to a bare `u64` with `0 == use-schedule`, so a client could not request truly immediate redelivery, and this contradicted the core's own tested-but-dead `effective_nack_delay(attempt, Option<u64>)`. | Carry the Option end to end: `u64::MAX` is the wire sentinel for "no explicit delay" (apply the schedule), any other value is explicit ms (`0` = immediate). `Engine::nack` now uses the live `effective_nack_delay`; `Client::nack` takes `Option<u64>`; the CLI sends `None` when `--delay-ms` is omitted; the `AckBody.delay_ms` doc states the convention. |
| 2 | MINOR | The override test only proved "still deferred", never "redelivers at the client's delay". | Advance to the 1 ms mark and assert the message redelivers there with the attempt escalated. |
| 3 | NIT | Multi-attempt escalation was not covered by an engine test. | Added `the_backoff_schedule_escalates_across_attempts` with a `[50, 200]` schedule. |

NIT 2 (backoff still applied on the `max_deliver`-th attempt before the inevitable dead-letter) is harmless added latency before a message that was never going to redeliver to a consumer; noted, not changed.

**Verified end to end:** a binary smoke test shows `sub --nack` (None) deferred by the 100 ms schedule while `sub --nack --delay-ms 0` (Some(0)) redelivers immediately, the exact distinction the fix restores.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (296 tests) all green. Awaiting CI.

Issue #16, opened 2026-06-05
[DESIGN] Observability: metrics, tracing, health, and introspection


ELares:
IronBus must make its internal state legible to an operator who is standing in front of a degraded edge box (or, more often, looking at a dashboard a continent away) without paying for that visibility in RAM, flash wear, or CPU it does not have. This issue specifies what an operator can see: a Prometheus-style metrics surface, OpenTelemetry tracing hooks, health and readiness endpoints, structured JSON logs, and an introspection/admin endpoint that the #15 consumes for deep diagnostics. The central requirement, and the thing that distinguishes IronBus from a generic exporter bolt-on, is that the resilience story must be observable: corruption, poison-record skips, backpressure overflow, and bounded data loss can never be silent. A metric or event must fire for every byte IronBus drops, skips, or fails to fsync, with the offset attached.

## Why this matters for IronBus

Observability is where four tenets meet. Resilience is only real if it is reported: #8 promises self-recovery with bounded, reported loss, and "reported" is an observability contract, not a log line nobody reads. Edge First means the metrics path must be cheap enough to leave on permanently on an ARM box with a few hundred MB of RAM, because you cannot SSH into a brownout. Simple means one HTTP port, OpenMetrics text, JSON logs, no sidecar, no agent. HyperScale means per-consumer lag and fsync-latency histograms that let #19 prove the SLOs are met on real hardware. The anti-goal is the Kafka/JMX experience: hundreds of mbeans, a JVM, and a separate exporter process just to learn that a partition is under-replicated.

## Prior art

[...]

PR #190, opened 2026-06-06
feat: health and readiness endpoints (observability slice, #16)

ELares:
## What

The first slice of the observability surface (#16): minimal HTTP **liveness** and **readiness** endpoints on a loopback port, so an orchestrator probe or an operator can tell a live broker from a degraded one.

## Endpoints (`ironbus-server::health`)

- `GET /healthz` — **liveness**: `200 ok` whenever the health loop is running (the process is up).
- `GET /readyz` — **readiness**: `200 ready` if the durable log writer is not frozen, `503 writer frozen` if a fatal fsync has frozen it (the exact "resilience must be observable" signal #16 asks for).
- A non-`GET` is `405`, an unknown path `404`, an over-long request line `414`.

The parser reads **only the bounded request line** (chunked, capped at 8 KiB), sets read and write timeouts, and closes after one response (`Connection: close`), so a slow or hostile client cannot wedge the loop. A query string is stripped; a malformed bare-newline request yields `405`, never a panic.

## Wiring

- `Engine::is_healthy()` -> `Log::is_writable()` reports the writer-frozen state without attempting a write (a frozen writer keeps serving reads but refuses writes).
- `ironbus serve --health-addr <host:port>` binds a second loopback listener and runs `serve_health` in a thread alongside the wire server, sharing the engine `Arc`. Without the flag, nothing changes.

## Tests

- Module tests: `/healthz` and `/readyz` return 200 on a live broker, a query string is stripped, an unknown path is 404, a non-GET is 405, and a malformed bare-newline request does not panic. Stable across repeated runs (socket-based).
- Smoke-tested with `curl` against the built binary: `/healthz` -> `200 ok`, `/readyz` -> `200 ready`, `/nope` -> `404`, `POST` -> `405`, and the wire port still serves `pub` alongside.

## Scope

This is liveness + readiness only. `OpenMetrics` `/metrics` (per-consumer lag, fsync-latency histograms, the drop/skip/loss counters), OpenTelemetry tracing, and the structured introspection endpoint are follow-ups under #16. The `/readyz` writer-frozen path is structurally wired but not unit-tested (freezing the writer needs the fault-injecting fs from #164); noted.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (299 tests) all green locally. CHANGELOG updated. DCO signed.

refs #16


ELares:

An independent adversarial review probed the parser against giant lines, non-UTF8, bare LF, empty method/path, EOF-mid-line, and slow clients. Verdict: **1 BLOCKER, 2 MAJOR, 2 MINOR, 1 NIT**. The parser cannot panic, but it had real wedge and contract defects, all addressed in `f72c807`:

| 1 | BLOCKER | The accepted socket inherited the listener's non-blocking flag, so the read/write timeouts were a no-op AND a request split across TCP segments was dropped with no response. | `stream.set_nonblocking(false)` after accept, mirroring the wire handler. A new test sends the request line in two TCP segments to pin it. |
| 2 | MAJOR | The read loop had only a per-read timeout, so a client dribbling one byte per window could hold the inline loop (starving every other probe) for hours. | A total deadline: once `REQUEST_TIMEOUT` elapses the request is answered `408` and closed, so a connection is bounded to ~5s regardless of dribble. |
| 3 | MAJOR | `/readyz` was documented to surface a frozen-on-fatal-fsync writer, but `Log::is_writable` only goes false on a failed segment ROLL: `Log::sync` takes `&self` and does not freeze on an fsync error, so an fsync failure left `/readyz` reporting ready. | Corrected the health/engine/log docs and the CHANGELOG to state what `/readyz` actually catches; filed the real storage gap (sync should freeze the writer on a fatal fsync, per the frozen design) as **#191**. |
| 4 | MINOR | `/readyz` takes the engine lock, so a slow produce fsync delays the probe. | Documented (readiness latency tracking write latency is the intent). |

The reviewer confirmed the rest is sound: the parser is panic-free on hostile input, the HTTP response framing is correct (Content-Length = byte length, CRLFs, blank line, `Connection: close`), the `PoisonError::into_inner` keeps readiness answering after a panicked wire handler, and the module compiles on Windows. MINOR-2 (cfg-gate the module) and the NIT (the rarely-reached shutdown/join cleanup is correct) need no change; the module is genuinely cross-platform and its tests run on all platforms.

**Verified:** the module tests (including the split-segment regression), and a binary slowloris smoke test, a partial request does not wedge the server and a normal probe right after still answers.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (300 tests) all green. Awaiting CI.

PR #192, opened 2026-06-06
feat: Prometheus /metrics endpoint with the core engine gauges (#16)

ELares:
## What

Extend the health HTTP port (added in #190) with `GET /metrics` in Prometheus/OpenMetrics text format (#16), the first metrics slice. Operators get the headline signals (consumer lag, durable head, in-flight depth, writer health) on the same loopback port as `/healthz` and `/readyz`, no sidecar.

## Gauges

```
ironbus_committed_offset   the committed consumer cursor (every offset below it is acked)
ironbus_flushed_offset     the durable log head (offset of the next record to be written)
ironbus_consumer_lag       flushed minus committed (durable records produced but not committed)
ironbus_in_flight          messages delivered (leased) but not yet acked
ironbus_writer_healthy     1 if the durable log writer is live, 0 if frozen
```

Each is emitted with its `# HELP` and `# TYPE gauge` header. All are read once off the engine under a single lock (`Engine::flushed_offset` is added; the rest already existed). Consumer lag is the one most operators watch and is exactly `flushed - committed`.

## Tests

- Module test: after producing two records, `/metrics` returns 200 with `ironbus_flushed_offset 2`, `ironbus_consumer_lag 2`, `ironbus_committed_offset 0`, `ironbus_in_flight 0`, `ironbus_writer_healthy 1`, and the `# TYPE` headers.
- Smoke-tested with `curl` across a full cycle: after producing 2, `lag 2`; after `sub --ack`, `committed 2` and `lag 0`.

## Scope

Gauges only. The resilience counters #16 cares most about (bytes dropped, records skipped on corruption, bounded loss, with the offset attached) require engine instrumentation and are a follow-up, as are the fsync-latency histograms and per-consumer-group lag (there is one work-group today). The content type is `text/plain`, which Prometheus scrapes; an explicit OpenMetrics version header is a follow-up.

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (301 tests) all green locally. CHANGELOG updated. DCO signed.

refs #16


ELares:

An independent adversarial review verified the gauge semantics and the wire format. Verdict: **0 BLOCKER, 0 MAJOR, 1 MINOR, 2 NIT**.

**Confirmed correct:** `ironbus_consumer_lag = flushed - committed` is exactly "durable but not committed" (`flushed_offset` is the exclusive durable head, `committed` is the acked watermark); `committed > flushed` cannot legitimately occur (clamped at open via `recovered_offset.min(flushed)`, and `poll` only advances `committed` within `[committed, flushed)`), so the `saturating_sub` is defensive, not masking an inversion. `in_flight` is a subset of the lag window (complementary, not double-counted). The emitted text is valid Prometheus exposition format, verified on the wire: every line ends with exactly one `\n`, the Rust `\` line-continuations eat the source indentation so no leading whitespace leaks, values are bare integers, names match the metric grammar, the body ends with a trailing newline, and `Content-Length` matches.

**Addressed in `aa6e40f`:**

| 1 | MINOR | The CHANGELOG said "OpenMetrics text format", but the endpoint emits Prometheus text exposition format (no `# EOF`, `text/plain`), contradicting the title and module doc. | Corrected the CHANGELOG to "Prometheus text exposition format". |
| 2 | NIT | The test used unanchored substring matches (`ironbus_flushed_offset 2` also matches `20`). | Anchored every assertion to a full line (`\n...\n`). |
| 3 | NIT | The test never exercised a non-zero committed / in-flight / lag-shrink path. | Extended it: lease one message (assert `in_flight 1`), ack it (assert `committed 1`, `consumer_lag 1`, `in_flight 0`), exercising the real `flushed - committed` arithmetic. |

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (301 tests) all green. Awaiting CI.

PR #193, opened 2026-06-06
feat: operational metrics counters (the dead-letter drop signal, #16)

ELares:
## What

Extend `/metrics` (#16) with monotonic operational **counters**, the resilience-observability core: #16's central requirement is that a dropped, skipped, or redelivered message is a *countable event*, not a buried log line.

## Counters

```
ironbus_produced_total       messages appended by produce
ironbus_delivered_total      message deliveries handed out (a redelivery counts again)
ironbus_redelivered_total    deliveries that were a redelivery
ironbus_dead_lettered_total  messages parked past MaxDeliver  <-- the drop signal
ironbus_acks_total           commits via ack (a term commits through the same path)
```

`ironbus_dead_lettered_total` is the one that matters most: a poison message dropped past `MaxDeliver` is now a counter, so loss is never silent.

## Implementation

- `Engine` gains a `Counters` struct (5 `u64`s) and increments them on the produce / poll / ack paths: `produced` on a successful `produce`; `delivered` (and `redelivered` when `deliveries > 1`) when `poll` hands out a `Message`; `dead_lettered` when `poll` parks a message; `acks` on a successful `ack`. They are in-memory statistics, not durable state, so a restart resets them to zero (standard for process counters).
- `Engine::counters()` returns a `Copy` snapshot, read under the same single lock as the gauges (a consistent snapshot, no tearing).
- The `/metrics` body building moved into a `metrics_body` helper so the request handler stays small.

## Tests

- The metrics test now asserts `ironbus_produced_total 2` after producing two, `ironbus_delivered_total 1` after leasing one, and `ironbus_acks_total 1` after acking, with `dead_lettered`/`redelivered` at 0 (anchored to full lines).
- Smoke-tested with `curl`: `pub` x2 then `sub --ack` yields `produced_total 2`, `delivered_total 2`, `acks_total 2`.

## Scope

These are the core lifecycle counters. The fsync-latency histograms, per-byte loss accounting with the offset attached, and per-consumer-group counters are follow-ups under #16 (there is one work-group today).

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (301 tests) all green locally. CHANGELOG updated. DCO signed.

refs #16


ELares:

An independent adversarial review verified the counting semantics statically, on a live broker over the wire, and with a unit probe. Verdict: **0 BLOCKER, 0 MAJOR, 2 MINOR, 2 NIT**.

**Confirmed correct:** every counter increments exactly once per real event, in the right place. `produced` is after `append` + `sync` (a failed produce does not count); `delivered`/`redelivered` are in `poll`'s `Deliver` arm (returns immediately), `redelivered` only when `deliveries > 1`, and an idle poll / already-leased / acked offset never increments; `dead_lettered` is in the mutually-exclusive `DeadLetter` arm (the parked attempt is not a delivery and the re-poll is skipped by `is_acked`, so never re-counted); `acks` is in the `Acked` arm only (a fenced ack does not count). The snapshot is read under one lock with the gauges (tear-free, `Counters: Copy`), names are `_total`-suffixed `counter` type, and the body stays valid Prometheus text. `term` counting as an ack is the documented design.

**Addressed in `a4dd4f2`:**

| 1 | MINOR | `redelivered` and `dead_lettered` were only ever asserted as 0, so a broken increment would not fail CI. | Added `counters_track_delivery_and_redelivery` (a lease that expires and redelivers → `delivered 2`, `redelivered 1`) and `counters_track_a_dead_letter` (`max_deliver 1`, the second attempt is parked → `dead_lettered 1`, `delivered 1`, `redelivered 0`). |
| 2 | MINOR | The `delivered 1` assertion could not distinguish once-per-delivery from once-per-poll. | The redelivery test asserts `delivered 2` after two deliveries of the same message, pinning once-per-delivery. |

NIT 1 (the final poison attempt is counted only in `dead_lettered`, not `delivered`/`redelivered`) is now both asserted by the dead-letter test and noted in the CHANGELOG. NIT 2 (`term` reuses the `acks` counter) is the documented design; a separate `ironbus_terms_total` is a cheap future seam (the `term`/`ack` engine split already exists), no change now.

**Gate (local):** `fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (303 tests) all green. Awaiting CI.

PR #194, opened 2026-06-06
docs: a usage / getting-started guide for the implemented product

ELares:
## What

`docs/USAGE.md`, a getting-started guide for the `ironbus` binary as it exists today. It documents the implemented subset (the README holds the broader vision):

- **Build**: `cargo build --release -p ironbus-cli`, the static musl artifacts from CI, and the Unix-only `serve` caveat.
- **Run a broker**: `serve` with a flag table covering every flag and its default (`--data-dir`, `--addr` 127.0.0.1:7777, `--max-connections` 256, `--checkpoint-interval` 1024, `--health-addr`).
- **Produce / Consume**: `pub` (arg or stdin), `sub` with the per-message line format and the `--ack` / `--nack [--delay-ms]` / `--term` / peek dispositions, plus the at-least-once and `MaxDeliver` notes.
- **Restart and resume**, the health/metrics endpoints with the exact metric names, a complete end-to-end example, and the fixed exit-code scheme (0/1/5/70).

## Verification

Every factual claim was checked against the code and the running binary, not just written from memory:

- All documented constants match `main.rs`: `DEFAULT_ADDR` = 127.0.0.1:7777, `DEFAULT_FETCH` = 10, `DEFAULT_MAX_CONNECTIONS` = 256, `DEFAULT_CHECKPOINT_INTERVAL` = 1024, `DEFAULT_NACK_BACKOFF_NANOS` = [100ms, 500ms, 2s, 10s, 30s], `MaxDeliver` = 5, exit codes 1/5/70, bin name `ironbus`.
- All 10 metric names match `health.rs::metrics_body`; the `sub` line format matches `"#{} gen={} key={} payload={}"`.
- A full binary run confirmed the runtime behavior: a usage error exits 1, an unreachable broker exits 5, `pub` returns the offset, `sub --ack` prints the documented lines, `/healthz` -> `ok`, `/readyz` -> `ready`, `/metrics` -> the gauges, and the restart-resume claim is the exact scenario `golden_path.rs` already exercises.

## Scope

Docs only, no code change. Linking `docs/USAGE.md` from the canonical README is left as a follow-up so the front-door doc is changed deliberately, not in passing.

## Gate

No code change, so the test/clippy/fmt/musl gates are unaffected and green. CHANGELOG updated. DCO signed.

refs #133


ELares:

An independent adversarial reviewer cross-checked every claim against the code and the running binary (including a live SIGKILL-with-socket-open test). Verdict: **1 BLOCKER, 1 MAJOR, 2 MINOR**.

The reviewer confirmed all the mechanical facts are accurate (every flag and default, the `sub` line format, the four disposition result lines, the backoff schedule applied to the CLI broker, all ten metric names, the `/healthz`/`/readyz` bodies, the Unix-only caveat, the three musl targets, the exit codes verified live). The defects were concentrated entirely in the durability story, exactly where a wrong doc is harmful.

**Addressed in `41a4bf7`:**

| 1 | BLOCKER | "every ack is fsynced before it returns" is false: only produce fsyncs; an ack advances an in-memory cursor, and cursor durability is the lagging checkpoint. | Removed it; the doc now states produce is synchronously durable and the cursor is a separate lagging checkpoint. |
| 2 | MAJOR | "Restart and resume" implied durability the default config does not provide; there is no graceful-shutdown drain, so signalling the broker is an abrupt stop, and the everyday flow only resumes because short-lived `sub` connections flush on close. | Qualified the resume claim: durable on clean disconnect or every `--checkpoint-interval` commits; an abrupt stop (including a signal) can redeliver up to `--checkpoint-interval` recently-acked messages (safe at-least-once); the durable-log-survives half stays. |
| 3 | MINOR | Peek is not side-effect-free: a bare `sub` claims a lease and counts as a delivery attempt. | Documented, with the 30s visibility timeout value (not flag-configurable). |
| 4 | MINOR | The 30s visibility timeout value was never stated. | Stated. Also noted `progress`/`ping` exist in the protocol but are not yet CLI flags. |

The corrected durability claims are verified against the code I authored (#184): `Engine::produce` appends then `log.sync()` before returning the offset; the cursor checkpoint fires only via `maybe_checkpoint` (interval) and `checkpoint_cursor` on the EOF close path; no signal handler exists.

**Gate:** docs-only change, so the test/clippy/fmt/musl gates are unaffected and green.

PR #196, opened 2026-06-06
feat(storage): freeze the writer on a fatal fsync; add fault injection (closes #191)

### LINKED ISSUE #191: Log::sync does not freeze the writer on a fatal fsync (contradicts the frozen-on-fatal-fsync design)
## Problem

The frozen design (and `Log::sync`'s own doc, "A fatal sync must freeze the writer read-only") says a fatal `fdatasync` freezes the writer read-only and is never retried. But `Log::sync` does **not** actually freeze the writer on an fsync error:

- `Log::sync` (`crates/ironbus-storage/src/log.rs`) calls `SegmentWriter::sync(&self)` (which takes `&self` and cannot mutate the writer's state) and on failure simply propagates `StorageError::Io` via `?`. The `active` segment stays `Some`.
- `active` is set to `None` (the "frozen" state, surfaced by `Log::is_writable`) **only in `roll()`**, i.e. a failed segment seal/create when a segment fills, not on a failed sync.

Consequence: after a genuine `fdatasync` failure, `Engine::produce` returns an error to the producer, but the writer is NOT frozen, so:
- A later `produce` may retry the write/sync against a disk that is failing, contradicting "never retried".
- `Engine::is_healthy()` / the new `/readyz` endpoint (#190) still report healthy/ready, even though the host's fsync is failing and the broker is not genuinely write-ready.

## What is needed

Make `Log::sync` (and any other path that performs a fatal fsync) freeze the writer on a fatal sync error: set the active segment to `None` (or an explicit frozen flag) and return `StorageError::WriterFrozen` on subsequent writes, matching the documented contract. Decide which sync errors are "fatal" (likely all `fdatasync` failures, since a failed durability barrier cannot be safely retried) and whether to distinguish transient vs fatal.

## Acceptance criteria

- [ ] A fatal fsync error freezes the writer: subsequent `append`/`sync` return `WriterFrozen`, and `is_writable()` / `is_healthy()` / `/readyz` report the degraded state.
- [ ] The writer is never retried after a fatal fsync (no silent re-sync against a failing disk).
- [ ] A crash-recovery test or fault-injection test (the #164 fs) exercises the fsync-failure-then-frozen path.

Surfaced by the #190 health-endpoints review: `/readyz` was documented to surface a frozen-on-fsync writer, but `Log::sync` never enters that state, so #190 corrects its docs to "frozen on a failed roll" and this issue owns making the stronger contract true. Relates to #116 (edge durability defaults: fatal-fsync).

ELares:
## What

Two coupled durability changes: close #191 (a fatal fsync must freeze the writer) and lay the first slice of #164 (block-layer fault injection), with the fault injection justified by immediately using it to test the freeze.

## Change

- **`Log::sync` freezes the writer on a fatal fdatasync.** It now drops the active segment on a sync error:
  ```rust
  if let Err(e) = self.active()?.sync() { self.active = None; return Err(e); }
  ```
  So every later `append`/`sync` returns `WriterFrozen`, a failed durability barrier is never silently retried (the design's "a fatal fsync freezes the writer read-only, never retried"), and `Engine::is_healthy` / `/readyz` report the degraded state. Previously `sync` propagated the IO error but left `active` `Some`, so the writer stayed live and `/readyz` reported ready after an fsync failure (the gap the #190 review found).
- **`ironbus-storage::fault`**: a deterministic fault-injecting layer. `FaultFs<F>` / `FaultFile<F>` wrap any `Filesystem` / `RandomAccessFile` and fail their syncs while a shared `FaultControl` is armed. Directory ops, reads, writes, length, and truncation delegate unchanged. This is the first slice of #164 (the fsyncgate/EIO mode); injected write failures, partial writes, and page-cache reordering are follow-ups there.

## Why it is safe

The freeze only changes behavior when a sync actually fails (a healthy fs never triggers it, so existing tests and the crash gates are unaffected). On a fatal sync, `flushed_offset` is left unchanged (the unsynced record was never made durable), so no acknowledged write is lost and reads keep serving exactly the durable prefix.

## Tests

- `a_fatal_fsync_freezes_the_writer_and_preserves_the_durable_prefix`: produce + sync one record (durable), append another, arm a fatal fsync, and assert: the sync surfaces the IO error; the writer is frozen (`is_writable()` false, `append`/`sync` return `WriterFrozen`); `flushed_offset` is unchanged; and a read still returns exactly the durable prefix.
- The pre-existing roll-failure freeze test still passes (both freeze paths covered).

## Docs

Updated `Log::is_writable`, `Engine::is_healthy`, and the health module doc now that `/readyz` catches an fsync-failure freeze (they previously noted it as tracked-in-#191).

## Gate

`fmt --check`, `clippy --workspace --all-targets -D warnings` (pedantic), `cargo test --workspace` (304 tests) all green locally. CHANGELOG updated. DCO signed.

Closes #191
refs #164


ELares:
## Adversarial self-review (multi-lens) + resolution

Ran a 3-lens review over the freeze/fault-injection change, then had each raw finding independently verified by a skeptical verifier (refute-by-default, code re-read plus a real probe). Result: **8 raw findings, 4 confirmed (1 MAJOR, 2 MINOR, 1 NIT), 4 dismissed.** All 4 confirmed are now fixed in `b80daa8`.

### Confirmed and fixed

**MAJOR. The produce that triggered the freeze was classified non-fatal, so the session did not end on the first failure.**
`Log::sync` dropped `self.active` but returned the original `StorageError::Io`. `EngineError::is_fatal()` matches only `WriterFrozen`, so `handle_pub` took the `Err(_) => reply_err("produce failed"); Ok(())` arm and kept the connection open. Only the *next* produce (which hits `active()? -> WriterFrozen` in `append`) was fatal. Net effect: a client is told the exact write that froze the broker is retryable, and can retry once into a permanently read-only writer before being closed.
Fix: `Log::sync` now returns `WriterFrozen` after `self.active = None`, so the freezing produce is fatal on the same call and `handle_pub` ends the session immediately. The freezing **roll** path (`seal`/`start_segment` after the active writer is taken) is mapped to `WriterFrozen` the same way, since it had the identical shape. Added an engine-level test (`a_freezing_produce_is_fatal_and_marks_the_engine_unhealthy`) that drives a `FaultFs`-backed `Engine` and asserts the freezing produce is `is_fatal()` and flips `is_healthy()` to false. Updated the `Log::sync` doc (it previously said "Propagates the IO error").

**MINOR. "Preserves the durable prefix" overclaimed a crash-recovery guarantee the in-memory test cannot exercise.**
A real `StdFs` reopen after the freeze recovers *more* than the acked prefix (the unsynced bytes were physically written before the fsync fault fired; a reopen rescans and is not gated by the old in-process `flushed_offset`). The in-memory test only returned the acked prefix because `read_from` is bounded by `flushed_offset`, not because the bytes were absent.
Fix: renamed the test to `a_fatal_fsync_freezes_the_writer_and_reads_keep_serving_the_flushed_prefix` and tightened its comment and the CHANGELOG to the *proven* live-handle property (the in-process flush mark is unchanged and an in-process reader keeps serving exactly the acked prefix). Durability across a restart is the recovery tests' job (`power_loss_drops_the_unsynced_tail_and_resumes`), and the comment now says so.

**NIT. The `sync_all` (roll/seal) freeze path was real but untested; only the `sync_data` path had a fault test.**
Fix: added `a_fatal_fsync_during_a_roll_freezes_the_writer`, which arms the fault then appends (never syncing) until the active segment fills and the next append rolls, so the freeze happens inside `roll()`'s `seal()` -> `sync_all`. It asserts `is_writable()==false`, `WriterFrozen` on further append/sync, and an unchanged flush mark.

**MINOR. The new `fault.rs` shipped with zero tests of its own.**
Fix: added `#[cfg(test)] mod tests` to `fault.rs`: unarmed read/write round-trip and sync passthrough, armed `sync_data` + `sync_all` faults (with IO still flowing), disarm-restores-passthrough, control shared across files and clones, and directory-op + `set_len`/`len` delegation while armed.

### Dismissed (verified false positives)

1. **`read_exact_at`/`is_empty` not overridden on `FaultFile`** — the trait defaults compose correctly through the overridden `read_at`/`len` to the inner file; no concrete impl gives these non-default semantics. Behaviorally correct today; purely hypothetical future-fidelity nit.
2. **No getter on `FaultControl` to observe armed state** — the only consumer arms inline then asserts (the idiomatic deterministic pattern); the getter would be unused public surface. The finding's motivation rested on self-removed throwaway probes.
3. **Sync-only injection scope wants a doc note** — the live/durable divergence is faithful fsync-EIO behavior, already named in the module doc; the finding conceded "no code change needed."
4. **InMemoryFs reopen "recovers" the unsynced record** — accurate mechanism, but the shipped test asserts no reopen behavior; this is forward-looking guidance for the #164 reopen-after-fault follow-ups (drive durability via `simulate_power_loss()`), correctly requiring no change here.

Local gate green: `cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, and `cargo test --workspace` (0 failed; storage and server test counts up by the new cases). Pushing for CI.

PR #197, opened 2026-06-06
test(storage): wire fsync-EIO block-layer faults into the crash gates (#164)

ELares:
## What

Wire the merged `FaultFs` fsync-EIO fault (from #191 / #196) into the per-PR crash-recovery gates, advancing the block-layer fault-injection scope of #164 now that the fault layer exists.

- **`FaultFs::inner`**: borrow the wrapped filesystem so a crash gate can reach the underlying disk (drive `simulate_power_loss`, inspect the durable image) after arming a fault. `InMemoryFs` is not shareable, so the harness needs this to model the crash that accompanies a failed fsync.
- **Two crash gates** (`crates/ironbus-storage/tests/crash_recovery.rs`):
  - `fatal_fsync_freeze_loses_no_acked_record`: a fatal `fdatasync` on an explicit `Log::sync` freezes the writer, and a reopen after the crash recovers exactly the acked prefix (monotone offsets, intact payloads). Swept over a few durable counts.
  - `fatal_fsync_freeze_during_a_roll_loses_no_acked_record`: the same invariant when the faulting fsync is the roll's `seal` (`sync_all`) rather than an explicit sync.
- Generalize the harness `append_at` / `assert_prefix` helpers over `Filesystem` so both the in-memory and fault-layer gates share them.

The invariant asserted: **no acknowledged record is lost and none is invented** across an fsync-EIO crash, complementing the existing in-memory power-loss and corruption gates.

## Out of scope (follow-ups under #164)

Page-cache reordering and partial (torn) writes, both-slots-torn, and crash-during-recovery idempotence (which needs a disk handle that survives a failed `Log::open`, e.g. a cheaply-cloneable `InMemoryFs`). These will be filed/tracked under #164.

## Testing

`cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, and `cargo test --workspace` all green locally (crash_recovery: 9 -> 11 gates; storage lib: +1 for `inner`).

refs #164
refs #21


ELares:

Ran a 3-lens review (durability-correctness, test-rigor, API/drift) over the diff, each finding then independently verified by a skeptical refute-by-default verifier. Result: **1 raw finding, 1 confirmed (MINOR), 0 dismissed.** Fixed in `ad75e54`.

**MINOR (test rigor). The roll-freeze gate's assertion was looser than the non-roll gate's.**
`fatal_fsync_freeze_during_a_roll_loses_no_acked_record` asserted `recovered >= acked`, while the non-roll gate asserts `assert_eq!(recovered, durable)`. The verifier confirmed (by tracing the code) that when the roll's `seal()` fsync faults, `start_segment()` is never reached so segment 1 is never created; on power loss segment 0 reverts to its durable image (the acked prefix only), so recovery yields *exactly* `acked`. The `>= acked` form would silently accept an over-recovery from a partially-surviving uncommitted segment if a future change introduced one.
Fix: tightened to `assert_eq!(recovered, acked)` with a comment explaining why equality holds (seal faults before segment 1 exists). The secondary `assert_prefix` already validated record count/content; this makes the count invariant fail-fast and self-documenting.

The durability-correctness and API/drift lenses found nothing: the gates correctly drop the unsynced tail via `simulate_power_loss()` before reopen (so `recovered == durable` is a real durability claim, not an artifact of the in-memory model), and `FaultFs::inner()` plus the `<F: Filesystem>` helper generalization are sound and accurately described.

Local gate green: `cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, `cargo test --workspace` (crash_recovery: 11 gates). CI was 12/12 green on the prior commit; re-running on `ad75e54`.

PR #198, opened 2026-06-06
feat(storage): inject write faults (clean EIO + torn write) and gate them (#164)

ELares:
## What

Extend the `FaultFs` block-layer fault layer past fsync faults (the next slice of #164) and gate the durability invariant under a faulting write.

- **`FaultControl::set_fail_write`**: a clean write EIO (every `write_all_at` errors, no bytes persisted).
- **`FaultControl::arm_torn_write(prefix_len)`**: a one-shot torn write that persists the first `prefix_len` bytes of the next `write_all_at`, then errors, modelling a write interrupted mid-record by a crash. A separate armed-flag keeps `Default` disarmed.
- `FaultFile::write_all_at` honors both; reads, length, and truncation still delegate unchanged.
- **Two crash gates** (`tests/crash_recovery.rs`):
  - `a_write_eio_during_append_recovers_the_acked_prefix`: a clean write failure leaves the record unacked, the writer stays live (a write fault does not freeze, only a fatal fsync does), and recovery yields exactly the acked prefix.
  - `a_torn_write_during_append_is_truncated_by_recovery`: a mid-record torn write leaves partial bytes in the live image; recovery truncates the torn tail and recovers exactly the acked prefix, never reading past it.
- Unit tests for both new fault modes (clean failure persists nothing; torn write persists exactly the prefix and fires once).

## Out of scope (follow-ups under #164)

Page-cache reordering (a strict-prefix subset of unsynced writes persisting out of order) and both-slots-torn (needs a checkpoint/manifest, #37), plus crash-during-recovery idempotence (needs a disk handle that survives a failed `Log::open`).

## Testing

`cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, `cargo test --workspace` all green (crash_recovery 11 -> 13 gates; storage lib +2 for the fault modes).

refs #164
refs #21


ELares:

Ran a 3-lens review (durability-correctness, fault-injection semantics, API/lib-safety/drift), each finding to be independently verified by a skeptical refute-by-default verifier. Result: **0 findings.** No defects to address.

What the lenses confirmed against the real code:
- **Durability**: both gates prove a real invariant, not an artifact. The torn-write gate's `flushed_offset == durable` is meaningful because the 4 torn bytes genuinely persist to the live image (`FaultFile::write_all_at` writes `buf[..n]` to the inner file before erroring) and recovery truncates that torn tail; this is independently pinned by the unit test `a_torn_write_persists_a_prefix_then_fails_once` ("only the torn prefix persisted"). The write-EIO gate correctly asserts the record stays unacked and the writer stays live (a write fault is non-fatal; only a fatal fsync freezes).
- **Fault semantics**: the one-shot torn write is race-free (an `AtomicBool` armed-flag consumed via `swap(false)`, prefix loaded only when armed), `Default` is disarmed for all three new fields, the `buf.len().min(...)` clamp is correct for `prefix == 0` and `prefix >= len`, and torn is checked before the clean-fail path.
- **API/drift**: the new public methods are documented and named consistently, no `unwrap`/`expect`/`panic` on the lib path, `usize::try_from(prefix).unwrap_or(usize::MAX)` is 32-bit-safe and clippy-pedantic clean, and the module + `FaultFile` docs and the CHANGELOG accurately scope this (write faults and torn writes covered; page-cache reorder and both-slots-torn explicitly left as #164 follow-ups).

Local gate green: `cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, `cargo test --workspace` (crash_recovery 13 gates, storage lib +2). CI 12/12 green on `a8614b7`.

PR #199, opened 2026-06-06
test(storage): property-based fsync-fault crash gate (seeded sim lane, #21)

ELares:
## What

Generalize the fixed-point fsync-EIO crash gates into a **property sweep**, the per-PR seeded sim lane #21 calls for.

A proptest (`a_sync_fault_at_any_point_loses_no_acked_record`) injects a fatal `fdatasync` at an arbitrary point `k` of an arbitrary append/sync workload (rolling and non-rolling configs), then a power loss, and asserts recovery yields **exactly** the prefix that was durable at the fault:
- `recovered >= durable_at_fault` (no acknowledged record lost),
- `recovered <= appended` (none invented),
- `recovered == durable_at_fault` (exactly the durable prefix), and
- the recovered records are a valid monotone run (offsets, seq, payloads).

It uses only the existing `set_fail_sync` + `FaultFs::inner` (no new primitive). The key invariant it leans on: once the fault is armed (from op `k` onward) no `sync` or roll's `seal` can succeed, so the durable mark cannot advance past `durable_at_fault`, and the modeled crash drops every unsynced byte.

Verified locally across **2048** randomized cases (`PROPTEST_CASES=2048`) in addition to the default 256.

## Testing

`cargo fmt --check`, `clippy -p ironbus-storage --tests -D warnings -W clippy::pedantic`, `cargo test --workspace` all green (crash_recovery: 13 -> 14 gates).

refs #21
refs #164


ELares:

Ran a 2-lens review (invariant-soundness, test-meaningfulness), each finding independently verified by a skeptical refute-by-default verifier. Result: **3 confirmed (1 MAJOR + 2 MINOR), 0 dismissed**, all one root issue from the test-meaningfulness lens. The invariant-soundness lens confirmed `recovered == durable_at_fault` is a genuine invariant (no counterexample). Fixed in `fd4af76`.

**MAJOR + 2 MINOR (one root cause). The proptest could pass vacuously.**
The fault was armed in phase 2 but only fired if `ops[k..]` happened to contain a `sync` (or a roll). A phase-2 of pure appends, a large `fault_after`, or a shrunk case (e.g. `ops=[]`) left the fault armed-but-unfired, so `recovered == durable_at_fault` held trivially without ever exercising the freeze. The two MINORs (shrinking to a fault-free case; coverage of the fault-firing path) were the same root cause.

Fix: instead of hoping the workload contains a sync, the test now **forces** the fault to fire with an explicit `log.sync()` right after arming and **asserts** the writer freezes every case (`matches!(.., Err(WriterFrozen))` + `!is_writable()`). The remaining `ops[k..]` are then applied on the frozen writer and asserted to make no durable progress. This makes every generated case non-vacuous and shrink-proof: a shrunk failure still exercises the freeze. (Note: the reviewer's suggested `!is_writable() || k == ops.len()` would have false-failed on an all-appends phase 2 where the writer legitimately stays live; forcing the sync is the correct, stronger fix.)

Verified across **4096** cases (`PROPTEST_CASES=4096`). The `recovered == durable_at_fault` invariant itself was unchanged and confirmed sound.

Local gate green: `cargo fmt --check`, `clippy -p ironbus-storage --tests -D warnings -W clippy::pedantic`, `cargo test --workspace`. CI was 12/12 green on the prior commit; re-running on `fd4af76`.

Issue #120, opened 2026-06-05
[TASK] Invariant checkers I1 to I4 and the versioned structured loss-report schema


ELares:
Parent: #21

Turn the four resilience invariants into executable checks that every verification layer can assert, and define the structured loss report as a stable, versioned artifact rather than an ad-hoc log line. This sub-issue is the contract the simulation, corpus, fuzz, and property layers all check against, so it must land early and be the single source of truth for what 'recovered correctly' means.

## Approach

Encode I1 (no acknowledged write lost below its durability level), I2 (recovery = longest valid prefix, kPointInTimeRecovery style), I3 (skip loss is bounded and reported), and I4 (recovery is a pure function of durable bytes) as reusable checker functions over a recovered-log state plus the ack history. Define a serde-serializable, versioned LossReport (schema_version, plus a list of events each with segment_id, byte_offset_start, byte_offset_end, bytes_skipped, records_lost_estimate, reason_code) emitted by recovery and consumable by #16 and #15. Enforce the I3 caps decided here: per-event cap of one segment or 64 MiB (whichever is smaller) and a 1% global cap of durable bytes, with freeze-read-only and non-zero exit on exceed. Ties to #7 for torn-tail handling and #8 for resync-at-segment-boundary.

## Acceptance criteria

- [ ] I1 to I4 exist as reusable checker functions invoked by the sim, corpus, and property layers.
- [ ] A versioned serde LossReport struct is defined with the fields above and a schema_version.
- [ ] I3 enforces the per-event (1 segment / 64 MiB) and 1% global caps; exceeding either freezes the log read-only and exits non-zero.
- [ ] The loss report is emitted by recovery and asserted by corpus fixtures against expected values.
- [ ] I4 is verified by recovering twice from identical durable bytes and asserting identical output.
- [ ] Defaults for both caps are documented as safe and are configurable per #14.

## Failure considerations

The key failure is I3 silently exceeding its bound (a cascade of skipped corrupt segments past the cap), which turns bounded reported loss into unbounded silent loss; freeze-on-exceed plus an over-cap report mitigate it. A second risk is schema drift: if each layer invents its own report shape, fixtures cannot assert exact values, so the versioned struct must be the only emitter. A buggy checker that always passes is guarded by negative fixtures (known-bad inputs that must fail each invariant).

ELares:
## Status (loss-report stack landed across 4 PRs)

The core of this issue is now implemented and CI-gated:

[...]

Issue #96, opened 2026-06-05
[TASK] Resilience-observability contract: count every skip, freeze, drop, and overflow with its offset


ELares:
Parent: #16

Guarantee the central promise of this issue: nothing the resilience machinery drops, skips, or freezes on is ever silent. Every corruption-detection, poison-skip, backpressure spill/drop/shed, and integrity-freeze path must increment a counter and/or set a gauge with the relevant offset captured. This is the contract #8, #10, and #7 are measured against.

## Approach

Define and wire the must-have metric set: `ironbus_corruption_skipped_records_total`, `ironbus_corruption_skipped_bytes_total`, `ironbus_last_skip_offset` (gauge), `ironbus_readonly_frozen` (0/1 gauge), `ironbus_backpressure_events_total{policy}` over the fixed policy enum (spill, drop_oldest, drop_new, shed), plus `ironbus_append_records_total`/`_bytes_total`. Each #8 and #10 code path calls the increment at the moment the event is decided, with the offset passed in (uint64, base 10). Alert-worthy state transitions (first skip after clean, freeze, backpressure policy escalation) additionally emit exactly one edge-triggered structured event at WARN/ERROR carrying offset and reason, which also becomes an OTel event when export is on. Events are edge-triggered on transition, never per scrape, so a wedged node does not storm.

## Acceptance criteria

- [ ] Every skip/freeze/drop/spill/shed path is shown (by test) to increment its counter or set its gauge with the correct offset.
- [ ] `ironbus_last_skip_offset` equals the offset of the most recent skipped record or segment.
- [ ] `ironbus_readonly_frozen` is 1 exactly while the node is integrity-frozen and 0 otherwise.
- [ ] Policy labels are confined to the fixed enum; no offset, message-id, or subject ever appears as a label.
- [ ] State transitions emit one edge-triggered WARN/ERROR event with offset+reason, not one per scrape.

## Failure considerations

The risk is a skip path that returns without counting, producing silent loss, so the test matrix must enumerate every path and assert the increment. Using an offset as a label would explode cardinality; offsets go in values and event attributes only. Level-triggered events on a stuck node would storm the log, so transitions are edge-triggered while the counter carries the steady state.

PR #200, opened 2026-06-06
feat(server): report recovery loss via ironbus_recovery_truncated_bytes (#16, #96)

ELares:
## What

The durability contract is "skip loss bounded and **reported**", but recovery truncated a torn or unsynced active-segment tail **silently**. This surfaces that loss, the first slice of the resilience-observability contract (#16, #96).

- **`Log::recovered_truncated_bytes`**: `Log` records the bytes it truncates from a torn/unsynced active-segment tail at recovery (zero for a fresh log or a clean recovery). `Engine::recovered_truncated_bytes` delegates it.
- **`ironbus_recovery_truncated_bytes` gauge** on `/metrics`: an operator can now see how much unsynced data the last restart discarded.
- Tests: storage asserts the reported count equals (pre-recovery length minus post-recovery length) on a torn tail and zero on a clean reopen; the metrics test asserts the gauge is exposed. `docs/USAGE.md` metric list updated.

## Out of scope (follow-up)

The structured, versioned loss-report schema (per-cause, per-offset accounting of every skip/freeze/drop/overflow) is #120; this PR is the single raw bytes-truncated signal that lands it on the existing metrics surface.

## Testing

`cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, `cargo test --workspace` all green (storage lib +1, server health test extended).

refs #16
refs #96
refs #120


ELares:

Ran a 2-lens review (correctness, wiring/drift), each finding independently verified by a skeptical refute-by-default verifier. Result: **1 confirmed (MINOR, doc grammar), 0 dismissed**. Fixed in the latest commit.

- **Correctness lens: nothing.** Verified `len - scan.valid_end` is the right quantity captured at the right moment (before `set_len`); the unsealed-active torn-tail branch is the *only* silent-truncation path (sealed roll-forward and empty-active-resume drop nothing; a corrupt sealed predecessor is a reported hard `Err`, not silent loss); the field is zero-initialized in both constructors; and the test's `reported == torn_len - post_recovery_len` assertion genuinely pins the capture (not vacuous).
- **MINOR (fixed):** the field/getter doc phrase "the silent loss recovery truncates" was reworded to "the silent loss that recovery truncates" on both comments.

Local gate green (`fmt --check`, `clippy`, `cargo test`). CI was 12/12 green on `9f429ee`; re-running on the doc commit.

PR #201, opened 2026-06-06
test(storage): gate crash-during-recovery idempotence (#164)

ELares:
## What

Close the last reachable block-layer crash class from #164: a crash **while recovery is truncating a torn tail** must be retryable and never leave the log unopenable.

- **`InMemoryFs` is now cheaply cloneable**: its directory state is shared behind an `Arc`, so a clone aliases the same disk. `Log::open` takes the fs by value (a failed open drops it), so the gate needs a probe handle that survives the failure. This is generally useful test infra (a probe can drive `simulate_power_loss` / inspect the durable image after the fs is moved into an engine or fault layer).
- **`recovery_is_idempotent_after_a_fault_during_recovery`**: tear a tail, arm a sync fault so recovery's truncation `sync_all` faults mid-recovery and the first `Log::open` fails cleanly (no panic, no hang), then assert a second reopen recovers the same valid prefix.
- **`a_clone_aliases_the_same_disk`** unit test: a power loss driven through one handle reverts the other handle's view; synced bytes survive, an unsynced overwrite does not.

This completes the crash-gate foundation alongside the merged power-loss, torn-tail, corruption, fsync-EIO, write-EIO, torn-write, and property-sweep gates. The remaining #164 items (page-cache reorder, both-slots-torn) are either subsumed by the truncation+corruption gates for the sequential append-only log, or blocked on the checkpoint manifest (#37).

## Testing

`cargo fmt --check`, `clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic`, `cargo test --workspace` all green (storage lib 105 -> 106, crash_recovery 14 -> 15).

refs #164
refs #21


ELares:

Ran a 2-lens review (clone-soundness, gate-validity), each finding independently verified by a skeptical refute-by-default verifier. Result: **3 confirmed (2 MAJOR + 1 MINOR), 0 dismissed** -- all one root issue. The clone-soundness lens found nothing (the `Arc`-shared state is Send/Sync, `Default` still yields an empty disk, no existing usage assumed distinct state). Fixed in the latest commit.

**MAJOR x2 + MINOR (one root cause). The gate passed for a weaker reason than it claimed.**
The test omitted `simulate_power_loss()` between the failed and retried opens. Because the disk is shared, the first (failed) recovery's `set_len(valid_end)` had already truncated the **live** image, so the second open saw an already-clean file, found `valid_end == len`, and **skipped recovery's truncation branch**. It verified retry-ability but not that recovery's truncation is genuinely idempotent under a crash.
Fix: model the crash that accompanies the failed `sync_all` by reverting live to durable (`disk.simulate_power_loss()`); the durable image still holds the torn tail (the truncation never synced), so the retry **re-runs** truncation. I also added `assert!(log.recovered_truncated_bytes() > 0)` to prove the truncation branch actually re-executed, not just that the final content was correct.

This is exactly the "passes for the wrong reason" class the adversarial pass is meant to catch.

Local gate green (`fmt`, `clippy`, `cargo test --workspace`). CI was 12/12 on `8cc1c1e`; re-running on the fix.

PR #202, opened 2026-06-06
feat(server): scope acks to the delivering session (#175)

### LINKED ISSUE #175: [TASK] Per-connection ack ownership: scope acks to the delivering session

ELares:
## What

`Engine::ack` fences only by **generation**, never by **connection**, so any session presenting a matching `(offset, generation)` could commit or requeue an in-flight message, silently suppressing redelivery to the consumer it was destined for (a consumer-authorization gap surfaced by the #174 self-review). Issue #175 asked to either scope acks to the delivering session or document the gap; this scopes them.

- `Session` tracks the leases it handed out via `Flow` in a `HashMap<offset, generation>`.
- `handle_ack` fences (status 0, **without** touching the engine) any ack/nack/term/progress whose `(offset, generation)` was not delivered to **this** session. The generation token still fences a stale op on an own-but-already-redelivered lease.
- The map is bounded to the in-flight window: keyed by offset so a redelivery overwrites the stale generation, terminal ops remove the entry, and **each `Flow` batch prunes offsets below the committed cursor** so a dead-lettered (committed-past) lease cannot accumulate. (Addresses "no unbounded queues".)

## Tests

- **Regression (#175):** `a_second_connection_cannot_ack_a_message_delivered_to_another` — B's ack of A's lease is fenced (status 0, nothing committed); A, the owner, commits it. (Inverted from the prior test that documented the gap.)
- `ack_commits_a_delivered_message` now delivers **through the session** (`Flow`) rather than out-of-band, matching the real consumer path.
- The existing progress/cap/term, redelivery, and fenced-ack tests pass unchanged (a double-term and a never-delivered ack now fence via ownership, same status 0).

Closes #175


ELares:

Ran a 3-lens review (security-correctness, bounding-semantics, test-drift), each finding independently verified by a skeptical refute-by-default verifier. Result: **4 raw, 2 confirmed (both MINOR test-coverage), 2 dismissed.**

The **security-correctness and bounding-semantics lenses found nothing**: no path reaches the engine without the ownership guard passing (the guard runs before the op match); the map is bounded (per-offset overwrite on redelivery + the committed-offset `retain` per batch never drops a live lease, since an unacked offset is always `>= committed`); and the per-op removals are correct (keep on Progress Extended/CapReached, remove on terminal/Fenced).

Two confirmed MINOR gaps, both fixed in `c2d508f`:
- **Stale-generation at the session layer was untested.** The redelivery test now asserts that after a redelivery grants a new generation, acking the **old** generation is fenced by the session guard (status 0, nothing committed) and the current one commits.
- **The regression test only covered `Ack`.** The cross-session test now also fences B's foreign **`Nack`, `Term`, and `Progress`** (all status 0, nothing committed), backing the CHANGELOG's "all four ops" claim.

Dismissed: a duplicate of the all-ops gap (the unified guard makes one op logically sufficient, but the explicit coverage is still worth having), and a note that the Progress-only-removes-on-Fenced logic is "subtle" (verified correct as designed).

Local gate green (`fmt`, `clippy --workspace --all-targets -D warnings -W clippy::pedantic`, `cargo test --workspace`). CI was 12/12 on `68ded2b`; re-running on the test commit.

PR #203, opened 2026-06-06
feat(server): report the most recent dead-lettered offset on /metrics (#16, #96)

ELares:
## What

`ironbus_dead_lettered_total` says **how many** messages were dropped past `MaxDeliver` but not **which**, so an operator alerting on drops can't find the offending message. This reports the offset of the most recent dead-letter.

- `Engine` tracks `last_dead_lettered: Option<Offset>` (set on the park path) and exposes `Engine::last_dead_lettered_offset`.
- `/metrics` adds the `ironbus_last_dead_lettered_offset` gauge, with **-1** as the unambiguous "none yet" sentinel (offsets are never negative).

## Testing

- Engine: `a_message_over_max_deliver_is_parked` now asserts `last_dead_lettered_offset()` is `None` before and `Some(0)` after a dead-letter (alongside `dead_lettered == 1`).
- Health: the metrics test asserts `ironbus_last_dead_lettered_offset -1` on a clean broker.
- `docs/USAGE.md` metric list updated.

A step of the resilience-observability contract (#16, #96: count every drop with its offset); the full per-offset structured loss report stays #120.

refs #16
refs #96


ELares:

Ran a 2-lens review (correctness, drift); each finding independently verified. Result: **1 confirmed (MINOR doc consistency), 0 dismissed.** The correctness lens found nothing: `last_dead_lettered` is set only on the dead-letter path and only there, the -1 sentinel is unambiguous (offsets are never negative), the value is read under the same lock as the other gauges, and the engine test proves None -> Some(0).

**MINOR (fixed):** the private field comment was less precise than the public getter doc; both now mention 'parked past `MaxDeliver`'.

Local gate green; CI was 12/12 on `95e5126`, re-running on the doc commit.

Issue #111, opened 2026-06-05
[TASK] Macro-bench harness: open-loop generator, HdrHistogram, coordinated-omission proof


ELares:
Parent: #19

Build the custom macro-bench harness that is the real SLO instrument: it drives the shipping IronBus binary over real sockets and measures end-to-end throughput (msg/s and MB/s), p50/p99/p99.9 latency, steady-state RAM, and write amplification. The single most important property is that it does not commit coordinated omission, because the parent's entire p99/p99.9 SLO depends on the tail being measured honestly under overload.

## Approach
Use an open-loop load generator with a constant arrival rate and Poisson jitter, with a separate sender and a separate latency-recording receiver, never a closed-loop request/response that waits for the prior message. Record latency from the intended send time (wrk2 style), not from when the generator actually got around to sending, so a stalled broker shows up in the tail. Drive the broker through the real client from #11, not a privileged test path. Record latencies into an HdrHistogram (1 us to 60 s, 3 significant figures) and archive the raw histogram per run so percentiles are recomputable and mergeable across windows. Measure end-to-end latency against a single CLOCK_MONOTONIC_RAW by embedding a send-time token in each payload, defaulting to co-located sender and receiver on one device. Emit the versioned results JSON (git SHA, build flags, device, config, raw histogram, reproduce command). Tie the overload workload to #10 so the harness measures shed-not-OOM behavior.

## Acceptance criteria
- [x] Open-loop generator with intended-send-time latency accounting.
- [ ] An injected-stall self-test (SIGSTOP the broker ~200 ms) proves the stall appears in recorded p99/p99.9; CI fails if it does not.
- [x] Latencies recorded in HdrHistogram and archived raw per run.
- [x] Harness drives the real #11 client against the shipping binary.
- [x] Reports msg/s, MB/s, p50/p99/p99.9, steady-state RAM, write amplification.
- [x] Emits the versioned provenance JSON for every run.

## Failure considerations
Closed-loop generation would silently erase tail latency under overload, so the injected-stall self-test gates every harness change to keep the instrument honest. A sorted-array of all samples would blow memory on multi-hour runs and an approximate sketch would lose tail fidelity; HdrHistogram avoids both. Cross-host clock skew larger than the sub-5 ms target is avoided by the default co-located single-clock design.

ELares:
Harness delivered in #279 (merged, squash `55223970`).

[...]

Issue #110, opened 2026-06-05
[TASK] Ratify and version the SLO target table from measured floors


ELares:
Parent: #19

Turn the illustrative SLO sketch in the parent into a committed, versioned table of falsifiable target numbers for the three reference devices (edge-min Pi4, edge-mid RK3399, x86-ref N100). The deliverable is a checked-in document plus a machine-readable schema, not a wiki page, so CI and the macro-bench can read targets and compare against them. Every cell is a tuple of {target, measured-baseline, margin} annotated with message size, fan-out, and durability mode, and the headline marquee number is locked to group-commit-fsync.

## Approach
Commit each SLO cell as a measured floor: after the first real macro-bench run on each device, set the committed gate to the measured p50/p99 minus a 20 percent margin, never an aspirational figure. A cell with no measured baseline ships as informational, not a gate. Define one marquee figure: 256 B, fan-out 1, group-commit-fsync on edge-min, provisionally >= 60k msg/s sustained and p99 < 6 ms until measured. Always publish async page-cache and sync-per-message rows alongside, with the page-cache row carrying a literal `not power-loss safe` label. Store the table as versioned JSON with a schema version so #14 and #16 can consume it. Cross-check every row against #6, #4, #10, and #21 so targets never contradict those subsystems' guarantees.

## Acceptance criteria
- [ ] Versioned JSON SLO table exists with one row per {device, message size, fan-out, durability mode}.
- [ ] Each row carries {target, measured-baseline, margin} and a durability label.
- [ ] Marquee figure is defined and tagged durability=group-commit-fsync.
- [ ] page-cache rows are labeled `not power-loss safe`; sync-per-message rows present.
- [ ] Unmeasured cells are marked informational and cannot gate CI.
- [ ] Schema is reviewed against #6, #4, #10, #21.

## Failure considerations
The main risk is committing aspirational numbers that no build can hit, which manufactures false confidence; the measured-floor-minus-margin rule prevents this. A second risk is durability-mode mislabeling (the Kafka flush trap): a schema lint enforces that the marquee field is group-commit-fsync and that page-cache rows carry the safety label, so an operator cannot quietly promote a page-cache number to the headline.

ELares:
[...]

Issue #97, opened 2026-06-05
[TASK] Metric registry: fixed histogram buckets, consumer-lag cardinality cap, allocation-free hot path


ELares:
Parent: #16

Build the bounded, allocation-free metric registry that makes leaving metrics on permanently affordable on a few-hundred-MB ARM box. The hot path must not allocate per message, the registry must have a hard memory ceiling, and per-consumer lag must be O(1) to update and to scrape, with #19 signing off the budget.

## Approach

Counters are relaxed-atomic increments; gauges are plain atomics read on scrape; histograms use a fixed compile-time bucket set in seconds {0.0005, 0.001, 0.002, 0.005, 0.01, 0.02, 0.05, 0.1, 0.2, 0.5, 1, 2, 5} (plus +Inf), reused for `ironbus_fsync_duration_seconds` and append latency. Per-consumer lag (`ironbus_consumer_lag_records{consumer}`) is maintained incrementally on append and commit, never scanned on scrape. Enforce a hard cap of 1024 distinct consumer series; on overflow, refuse new series, fold excess into `{consumer="__overflow__"}`, and increment `ironbus_consumer_labels_dropped_total`. Publish the registry memory ceiling: a fixed sub-100 series core plus 1024 consumer series times the per-series cost, asserted by a test. Add self-monitoring series `ironbus_build_info`, `ironbus_start_time_seconds`, `ironbus_uptime_seconds` (monotonic-derived).

## Acceptance criteria

- [ ] The append/scrape hot path performs zero heap allocation (verified by an allocation-counting test).
- [ ] Histogram buckets match the fixed set exactly and are not runtime-configurable.
- [ ] The 1024th-plus consumer triggers `__overflow__` folding and increments the dropped counter; total lag stays visible.
- [ ] A scrape is O(number of series), independent of record count or disk size.
- [ ] The documented registry memory ceiling is asserted in a test and signed off by #19 against the edge RAM ceiling.

## Failure considerations

Unbounded consumer cardinality would OOM the very node metrics protect, so the cap and overflow fold are mandatory. A scan-on-scrape lag would turn a scrape into a disk walk and stall the thread-per-core core, so lag is incremental. Runtime-tunable buckets would unbound per-series memory, so buckets are a compile-time constant.

ELares:
PR #320 implements this in `crates/ironbus-server/src/registry.rs`, wired into the engine produce/commit hot paths and rendered on `/metrics`. All five acceptance boxes are met:

[...]

PR #204, opened 2026-06-06
feat(server): expose the produce fsync latency as a Prometheus histogram (#16, #97)

ELares:
## What

fsync (the durability barrier) latency is the headline performance signal for a log-is-WAL queue, but nothing measured it. This exposes it as a Prometheus histogram (#16, #97).

- **`ironbus-server::metrics::LatencyHistogram`**: a fixed-bucket histogram (frozen nanosecond bounds 100us..1s + `+Inf`), allocation-free on `observe`. Per-bucket counts are stored non-cumulatively and made cumulative only at exposition.
- **`Engine::produce`** times the fsync through the clock seam (`now_monotonic` around `log.sync`) and observes it, so the deterministic sim stays reproducible (logical time does not advance during an in-memory sync). `Engine::fsync_histogram()` snapshots it.
- **`/metrics`** emits `ironbus_fsync_seconds` as a Prometheus histogram: cumulative `le`-in-seconds buckets, `+Inf`, `_sum` (nanosecond-precise, formatted without floating point), and `_count`. The growing metric snapshot is consolidated into a `Copy` `MetricsSnapshot` read under one lock.

## Testing

- Histogram type: empty-is-zero; bucketing + cumulative counts + sum on mixed observations (including a `> 1s` value landing only in `+Inf`); the `le` labels equal `bound / 1e9`.
- Engine: `produce` records exactly one observation per call.
- Health: `/metrics` emits the `histogram` TYPE line, `ironbus_fsync_seconds_count 2`, and the `+Inf` bucket `2` for the two produces (bucket distribution is timing-dependent under the system clock, so it is not pinned).

The durability-performance signal the SLO work (#110, #111) will gate on.

refs #16
refs #97


ELares:

Ran a 3-lens review (histogram-correctness, exposition-conformance, timing-drift), each finding independently verified. Result: **7 raw, 1 confirmed (MINOR), 6 dismissed.**

The histogram-correctness and exposition-conformance lenses thoroughly probed and **confirmed the implementation is correct**: the 6 dismissed were all verified-correct false positives (trailing newlines via `writeln!`, the float-free `_sum` formatting `{}.{:09}` valid for all nanos, consistent `_bucket`/`_sum`/`_count` naming, `le` labels == bound/1e9, ascending cumulative buckets, and `+Inf` == `_count`).

**MINOR (fixed):** the health module doc still listed fsync histograms as a follow-up; it now names `ironbus_fsync_seconds` and drops it from the follow-ups.

Local gate green; CI was 12/12 on `47d9e58`, re-running on the doc commit.

PR #205, opened 2026-06-06
feat(cli): add serve --max-deliver to tune the dead-letter threshold (#63, #91)

ELares:
## What

The max delivery attempts before a poison message is dead-lettered was hardcoded to 5 in `serve`. Expose it as `--max-deliver <n>`.

- Threads through `run_serve` -> `cmd_serve` -> `open_disk_engine` into `DeliveryConfig` (default 5).
- **Validated**: must be at least 1; `0` (which `DeliveryConfig` treats as *unlimited* delivery, looping a poison message forever) is rejected as a usage error rather than silently enabled from the CLI.

## Testing

- `serve_rejects_a_non_numeric_max_deliver` and `serve_rejects_a_zero_max_deliver` assert the usage-error paths (exit code 1).
- The on-disk round-trip test passes through the updated `open_disk_engine` signature unchanged.
- Engine behavior for varying `max_deliver` is already covered by the poison/park tests in `ironbus-server`.

refs #63
refs #91


ELares:

Ran a 2-lens review (correctness, consistency/drift); each finding independently verified. Result: **1 confirmed (MAJOR), 0 dismissed.**

**MAJOR (fixed):** `--max-deliver u32::MAX` (4294967295) is also "unlimited" per `DeliveryConfig` (the lease counter saturates there), but the CLI only validated `== 0`, so it slipped past parse and surfaced as an Internal error (exit 70) instead of a Usage error (exit 1), inconsistent with the 0 case. Fix: reject both `0` and `u32::MAX` at parse with a usage error, and added `serve_rejects_an_unlimited_max_deliver` to pin it (exit 1, message mentions 'unlimited'). CHANGELOG/message updated to name both forms.

The consistency lens otherwise found the wiring correct (threaded into both `cmd_serve` cfg variants and `open_disk_engine`, usage text + USAGE.md table consistent).

Local gate green; CI was 12/12 on `5ee55a0`, re-running on the fix.

PR #206, opened 2026-06-06
feat(cli): add serve --max-in-flight to tune the delivery window (#91)

ELares:
## What

The max-ack-pending window (at most this many messages leased above the committed cursor at once) was hardcoded to 1024 in `serve`. Expose it as `--max-in-flight <n>`.

- Threads into `open_disk_engine` -> `EngineConfig` (default 1024). Must be at least 1; `0` (which delivers nothing and `Engine::open` rejects) is a usage error.
- The four serve tuning knobs (`max-connections`, `checkpoint-interval`, `max-deliver`, `max-in-flight`) are grouped into a `Copy` `ServeConfig` struct so `cmd_serve` stays within the `too_many_arguments` lint (the non-unix variant reads each field in its discard to avoid a dead-code warning on Windows).

## Testing

- `serve_rejects_a_non_numeric_max_in_flight` and `serve_rejects_a_zero_max_in_flight` assert the usage-error paths.
- Engine behavior for the window size is already covered in `ironbus-server` (the in-flight window / `ZeroMaxInFlight` tests).

refs #91


ELares:

Two independent skeptical reviewers (correctness, consistency/drift), refute-by-default. Both found **no logic/consistency defects**: the correctness reviewer traced the full flow and confirmed no argument transposition (`open_disk_engine(data_dir, config.max_in_flight, config.checkpoint_interval, config.max_deliver)` matches the signature, and `max_in_flight` reaches the right `EngineConfig` field); the consistency reviewer confirmed the `ServeConfig` field types, usage text, USAGE.md table, CHANGELOG, and tests are all aligned, and the non-unix discard reads every field.

**CI caught a real cross-platform bug (fixed):** `test (windows-latest)` failed with E0425 because `DEFAULT_MAX_IN_FLIGHT` was `#[cfg(unix)]` but my `run_serve` change now uses it in non-gated code (as the `--max-in-flight` default). Local macOS (unix) testing could not see this; the Windows lane did. Fixed by un-gating the const (it is now referenced on every platform). CI is now 12/12 green on `6c466ef`.

This is exactly why the Windows lane and the cfg-gated-const class of bug are worth gating on.

Issue #162, opened 2026-06-06
[TASK] Validate LogConfig: reject a max_segment_bytes below a sane floor


ELares:
## Context

`LogConfig.max_segment_bytes` (added in #161) is a soft cap on total segment size, compared against `SegmentWriter::write_pos()`, which starts at the 64-byte header. The empty-segment guard prevents an infinite roll on an oversized record, but a tiny cap is still a footgun: any `max_segment_bytes <= SEGMENT_HEADER_LEN` rolls after every single record (one record per segment), and `max_segment_bytes: 0` (a legal value, the field is `pub`) fragments the log into one-record segments with no diagnostic.

Surfaced by the self-review on #161 (items m1, n1).

## Task

- Add a validating constructor `LogConfig::new(max_segment_bytes) -> Result<LogConfig, _>` (or a builder) that rejects a cap below a practical floor, and have `Log::open` reject (or the constructor enforce) it. The floor should comfortably exceed the header plus footer plus a small minimum record so a segment can hold more than one record. Tests that deliberately use small caps to exercise rolling should use a documented test-only path or the smallest valid floor.
- Decide whether the cap counts total segment bytes (current: includes the 64-byte header) or record bytes only (`write_pos - SEGMENT_HEADER_LEN`); document the chosen meaning and the overshoot bound precisely.
- Keep the field accessible or make it private behind the constructor; if kept `pub`, document that `0` and sub-header values fragment the log.

## Scope

Config ergonomics and a degenerate-value guard; no durability change. refs #4

PR #207, opened 2026-06-06
feat(cli): add serve --max-segment-bytes and reject a tiny cap (#91, #162)

ELares:
## What

The soft per-segment size cap used the hardcoded storage default (64 MiB) in `serve`. Expose it as `--max-segment-bytes <n>`, and guard the #162 footgun at the operator boundary.

- Threads through `ServeConfig` -> `open_disk_engine` into `LogConfig` (default 64 MiB, via a **CLI-local non-gated const** so the non-unix build does not regress, the lesson from #206).
- **Rejected below 4096 bytes** as a usage error: smaller caps make segments proliferate one record at a time (slow recovery, inode pressure on edge devices). This is the operator-facing half of #162; the library-level `LogConfig` floor stays open there, since the storage tests use deliberately tiny caps to force rolling.

## Testing

- `serve_rejects_a_non_numeric_max_segment_bytes` and `serve_rejects_a_tiny_max_segment_bytes` assert the usage-error paths.
- The on-disk round-trip test threads through the new `open_disk_engine` signature.

refs #91
refs #162


ELares:

Two independent skeptical reviewers (correctness/cross-platform-cfg, consistency/floor/drift), refute-by-default. **No defects found.**

The first reviewer specifically verified the cross-platform cfg-gating that broke #206's Windows build: both new consts (`DEFAULT_MAX_SEGMENT_BYTES`, `MIN_MAX_SEGMENT_BYTES`) are non-gated, `run_serve` references only non-gated items (no `LogConfig` leak), both consts are used on every platform (no dead-code on Windows), and the value lands in `LogConfig { max_segment_bytes }` with no argument transposition. CI confirms: `test (windows-latest)` is green this time.

The second reviewer confirmed the default (67108864 = 64 MiB) matches the storage `LogConfig` default, the `< 4096` floor is sensible (4096 allowed; the test uses 100, unambiguously below), and the usage text, USAGE.md table, CHANGELOG (`refs #162`, not `Closes`, since the library floor stays open), and tests are all aligned.

CI 12/12 green on `13a260f`.

Issue #119, opened 2026-06-05
[TASK] Deterministic simulation core: IO/Clock/Scheduler seam, seeded disk model, determinism meta-test


ELares:
Parent: #21

Build the in-house deterministic simulation that is the centerpiece of the resilience proof. Production code is written against IO, Clock, and Scheduler traits with two impls: a real impl (io_uring / memmap2 with flush at commit points) and a simulated impl that models a sector-atomic, reordering, droppable disk. A single seeded ChaCha8Rng drives every fault decision, and a printed seed must replay the exact interleaving and fault schedule. This sub-issue delivers the seam, the simulated disk, and the guard that makes the whole replay claim trustworthy.

## Approach

Define the three traits as the single abstraction point (per the decided one-seam policy) and route all time, ordering, and IO through them. The simulated disk models the #4 hardware assumptions: linear-not-atomic sector writes, reorder of unsynced writes, garbage file growth, simulated power cut that drops unsynced page-cache pages, and fsync EIO exactly once (the fsyncgate behavior). Thread the ChaCha8Rng explicitly; forbid ambient nondeterminism in sim mode (no SystemTime/Instant::now, no real thread::spawn or tokio, no unordered-map iteration in trace paths) via a cfg-guarded lint deny-list. Add the same-seed-twice meta-test that runs the harness twice and compares the event trace and recovered-log hash byte-for-byte. Crash classes kill -9, simulated power cut, and fsync EIO-once run here on every PR.

## Acceptance criteria

- [ ] IO, Clock, and Scheduler traits exist with real (io_uring/memmap2) and simulated impls; production code depends only on the traits.
- [ ] The simulated disk models sector-atomic-but-linear writes, unsynced reorder, garbage growth, power-cut page-cache drop, and fsync EIO-once.
- [ ] A single seeded ChaCha8Rng drives all faults; every failure prints a replayable seed.
- [ ] Same-seed-twice meta-test compares trace + recovered-log hash byte-for-byte and is a hard CI gate.
- [ ] Sim-mode lint deny-list rejects now(), real thread spawn, and unordered-map iteration in trace paths.
- [ ] Per-PR fixed 256-seed sweep runs under a 5-minute cap; kill -9, power-cut, and EIO-once classes are green.

## Failure considerations

[...]

PR #208, opened 2026-06-06
test(storage): determinism meta-test for the deterministic-sim foundation (#119)

ELares:
## What

The deterministic simulation (#119) and crash-class reproduction both rely on the storage layer being a **pure function of its inputs**: a fixed workload (or seed) must always produce the same disk image. This pins that with a meta-test.

`the_same_workload_produces_a_byte_identical_disk_image` runs the same produce/sync workload twice under a fresh `ManualClock` and asserts the **full disk image** (every segment file's bytes) is byte-identical across runs. It fails the moment a `SystemTime`, `Instant::now`, or ambient RNG leaks into a header, record, or footer, which the design forbids (IO and the clock go through trait seams). The workload uses a small segment cap so it rolls across several segments, exercising the per-segment header (`created_unix_ms`, sourced from the clock seam) and the sealed footers.

The first concrete slice of #119; the seeded disk model and the scheduler seam are follow-ups there.

## Testing

`cargo fmt --check`, `clippy -p ironbus-storage --tests -D warnings -W clippy::pedantic`, `cargo test --workspace` all green (the new `tests/determinism.rs` passes).

refs #119


ELares:

Two independent skeptical reviewers (meaningfulness/would-it-catch-the-bug, drift/style). Result: **1 MINOR (doc), 0 substantive defects.**

The meaningfulness reviewer confirmed the test is **non-vacuous and would actually fail on a regression**: it compares the complete byte image of every segment file (headers + records + footers, CRC-protected), the segment header's `created_unix_ms` is the only ambient-time vector and comes from the clock seam (so swapping it for `SystemTime::now` would make the two runs differ and fail the assert), the >= 2 segments assertion holds (24 records / 256-byte cap -> ~5 segments), and there is no ordering/flakiness (`fs.list()` is a sorted BTreeMap, `snapshot()` is deterministic).

**MINOR (fixed):** the CHANGELOG used `(#119)`; changed to `(refs #119)` to match the first-slice framing (consistent with the commit/PR).

CI 12/12 green on `42053ad`; re-running on the doc commit.

PR #209, opened 2026-06-06
feat(cli): add serve --visibility-timeout-ms (#91, refs #133)

ELares:
## What

How long a delivered message stays in flight before it may redeliver used the hardcoded lease default (30 s) in `serve`. Expose `--visibility-timeout-ms <n>`, which also unblocks driving the redelivery and dead-letter paths **fast** in an end-to-end test (#133, where the 30 s default is impractical).

- Threads through `ServeConfig` -> `open_disk_engine` into `LeaseConfig::from_millis` (default 30000 ms). The lease hard cap is derived as `max(5 minutes, visibility)`, so it is never below one redelivery window.
- Rejected at `0` (a hot redelivery loop) as a usage error.
- The serve-flag validation is extracted into a `validate_serve_config` helper so `run_serve` stays within the `too_many_lines` lint. Unix-only consts (`DEFAULT_HARD_CAP_MS`) are `#[cfg(unix)]`-gated to keep the non-Unix build dead-const-free (the #206 lesson).

## Testing

- `serve_rejects_a_non_numeric_visibility_timeout` and `serve_rejects_a_zero_visibility_timeout`.
- The on-disk round-trip test threads through the new `open_disk_engine` signature.

refs #91
refs #133

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Added
+- `ironbus serve --visibility-timeout-ms <n>` (#91): tune how long a delivered message stays in flight before it may redeliver (default 30000 ms), instead of the hardcoded lease default. Must be at least 1 (a zero window is a hot redelivery loop). The lease hard cap is derived as the larger of 5 minutes and the visibility window, so it is never below one redelivery window.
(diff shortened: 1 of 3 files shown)


ELares:

Independent skeptical reviewer focused on the cross-platform cfg-gating (the #206 lesson), the `validate_serve_config` refactor, and the lease flow. **No defects found.**

Verified: `DEFAULT_VISIBILITY_MS` is non-gated (used in non-gated `run_serve`), `DEFAULT_HARD_CAP_MS` is correctly `#[cfg(unix)]`-gated (referenced only from unix `open_disk_engine`, never from `run_serve`), so the Windows build has no E0425 and no dead-const, confirmed by CI (`test (windows-latest)` green). The `validate_serve_config` extraction preserves all five checks in order before the broker opens; `visibility_ms` threads into `LeaseConfig::from_millis` with no transposition and `hard_cap = max(300000, visibility) >= visibility` always holds; `open_disk_engine` is 6 args (under the limit); no panics on the parse path.

CI 12/12 green on `ab6fbc2`.

Issue #123, opened 2026-06-05
[TASK] Corruption corpus plus block-layer and ALICE conformance gate on the reference edge device


ELares:
Parent: #21

Assemble the versioned corruption corpus that exercises the #8 path and asserts I3, and stand up the block-layer and ALICE-style conformance gate that keeps the deterministic simulation honest against real hardware. This is the oracle: without it the sim is self-certifying and can be comprehensively wrong, which is the exact trap this whole issue exists to prevent.

## Approach

Build a three-tier corpus: Tier A hand-written canonical poison fixtures, one per taxonomy entry (bad CRC, truncated length, mid-record torn tail, zeroed block, header bit-flip, swapped/duplicated segment, garbage-grown file), run every PR; Tier B fuzzer-minimized crashers auto-promoted from nightly; Tier C real bit-flipped segments captured from a wear-stressed eMMC/SD device, refreshed per hardware generation. A versioned manifest records sha256, provenance, and the expected LossReport per fixture. Wire the conformance gate: identical scenarios run on the sim and on the real impl on a reference aarch64 device (or container) under dm-flakey (dropped/intermittent writes) and dm-dust (per-block read errors), plus an ALICE/CrashMonkey-style crash-prefix run that catches a missing parent-directory fsync the seam abstracts. Recovered state and loss reports must match; divergence is release-blocking. These device classes run nightly and pre-release and are hard release gates.

## Acceptance criteria

- [x] Tier A poison fixtures cover the full taxonomy and assert I3 (bounded skip + structured loss report) every PR. (#298)
- [ ] Tier B promotes fuzzer-minimized crashers; Tier C captures real bit-flipped segments per hardware generation.
- [ ] A versioned manifest records sha256, provenance, and expected LossReport per fixture.
- [ ] dm-flakey and dm-dust runs execute on a reference aarch64 device/container nightly and pre-release.
- [ ] An ALICE/CrashMonkey-style crash-prefix run is wired into the same gate.
- [ ] Differential conformance asserts sim-vs-device recovered state and loss reports match; divergence is release-blocking.

## Failure considerations

[...]

PR #211, opened 2026-06-06
ci: add a nightly deep proptest sweep lane (#21, #123)

ELares:
## What

Done(v1) wants the crash-class gates to run per-PR (sim) **and nightly**. The per-PR pipeline runs the crash gates and codec proptests at proptest's default 256 cases; this adds a scheduled workflow that re-runs the workspace tests with `PROPTEST_CASES=50000` (~195x), so a rare non-determinism or a crash-class regression that only a specific seed exposes is caught within a day.

- `.github/workflows/nightly.yml`: `schedule` (04:00 UTC daily) + `workflow_dispatch` (manual, also used to validate the lane). Same toolchain/cache actions as `ci.yml`; ubuntu-only (the sweep is OS-independent); `permissions: contents: read`.
- The deep-sweep command (`cargo test --workspace --all-features --locked` with the env) was validated locally: the whole workspace is clean at 50000+ cases after the #210 fragile-proptest fix (which this lane's investigation surfaced).

## Validation

The per-PR pipeline does not run `nightly.yml` (it triggers only on schedule/dispatch), so this PR's checks gate the change itself; I will trigger the lane manually (`workflow_dispatch`) on main after merge to confirm it runs green.

Block-layer crash gates on a reference device, fuzz soak, Miri, and coverage are follow-ups under #21 / #123.

refs #21
refs #123


ELares:

Independent reviewer focused on workflow validity, the deep-sweep command, ci.yml consistency, and security. **No issues found.**

Verified: valid YAML and cron (`0 4 * * *`), `workflow_dispatch` present for manual triggering, `PROPTEST_CASES=50000` is the correct proptest env var (~195x the default) placed on the run step, `--locked` works (committed `Cargo.lock`), action versions (`actions/checkout@v4`, `dtolnay/rust-toolchain@stable`, `Swatinem/rust-cache@v2`) and the env block match `ci.yml` exactly, `permissions: contents: read` is least-privilege with no secret/injection exposure, and the comments/CHANGELOG accurately scope this as the proptest deep sweep (block-layer-on-device, fuzz, Miri, coverage correctly deferred).

Per-PR CI is 12/12 green on `0ce9411`. The lane itself triggers only on schedule/dispatch, so I will trigger it via `workflow_dispatch` on main after merge to confirm it runs green end to end.

PR #212, opened 2026-06-06
test(cli): golden-path poison-message dead-letter gate (#133, refs #63)

ELares:
## What

A step of the golden-path acceptance scenario (#133): exercise the poison-message resilience story **end to end over the wire**.

An in-memory broker with a short visibility timeout and `max_deliver = 1`: pub a message, `Peek` it (delivery 1, leased, not acked), wait past the visibility timeout, then fetch again. Delivery 2 exceeds the cap, so the engine **dead-letters** it (parked, not redelivered) and the re-fetch is empty. The test then asserts on the engine handle that it recorded exactly one dead-letter and its offset (`counters().dead_lettered == 1`, `last_dead_lettered_offset() == Some(0)`).

## Non-flaky

Platform-agnostic (in-memory fs, so it runs on every OS in the matrix) but uses the real clock for the visibility timeout, with a **10x margin** (50 ms visibility, 500 ms sleep). Verified stable over repeated local runs (5/5).

refs #133
refs #63


ELares:

Independent reviewer focused on correctness and flakiness. It confirmed the dead-letter mechanic and the timing margin are correct, and raised one MEDIUM concern, a potential serve-thread startup race. After verification, **that finding is a false positive** and the test is unchanged.

**Why it is not a race:** `TcpListener::bind()` puts the socket in the LISTEN state (with a backlog) BEFORE the thread is spawned. A client `connect()` to a listening socket completes the TCP handshake into the kernel accept-backlog regardless of whether the server has called `accept()` yet, so `connect()` never fails; `serve()`'s non-blocking accept loop picks up the queued connection microseconds later, well inside the client's multi-second read timeout. Evidence: every existing CLI broker test (`pub_then_sub_with_ack_round_trip`, `sub_nack_requeues_and_sub_term_drops`, the on-disk round-trip) uses the identical bind-before-spawn pattern and passes reliably across macOS/Ubuntu/Windows; a real race would have flaked one of them over many runs. This test passed 5/5 locally and 3/3 OS jobs in CI. The reviewer's suggested barrier-before-`serve()` would not even guarantee `accept()` was reached, so it would be a placebo that also diverges from the established race-free pattern.

Confirmed-correct items: the timing margin (50 ms visibility, 500 ms sleep, monotonic clock, strict deadline) is robust; `Peek` leaves the message leased so it expires and re-polls as delivery 2 > max_deliver 1 -> dead-lettered (parked, empty re-fetch); `fetched 0 message(s)` is the exact empty-batch output; `counters().dead_lettered == 1` and `last_dead_lettered_offset()` are recorded; the thread is joined.

CI 12/12 green on `b6a032f` (all three OSes).

PR #213, opened 2026-06-06
test(server): engine-level determinism meta-test, cursor checkpoint (#119)

ELares:
## What

Extend the deterministic-sim foundation (#119) from storage to the engine. The storage determinism gate (#208) covers the segment bytes; the full produce / poll / ack lifecycle also writes the durable consumer-cursor checkpoint (`cursor.ckpt`), and that must be deterministic too.

`the_same_engine_lifecycle_produces_a_byte_identical_disk_image` runs the same produce/poll/ack workload twice under a fresh `ManualClock` and asserts the **full disk image (segments AND `cursor.ckpt`)** is byte-identical. `checkpoint_interval = 1` ensures the checkpoint is written every commit and is part of the compared image (the test also asserts a `ckpt` file is present). Ambient time or randomness leaking into the checkpoint or the commit path would fail here.

## Testing

New `crates/ironbus-server/tests/determinism.rs` passes. `cargo fmt --check`, `clippy -p ironbus-server --tests -D warnings -W clippy::pedantic`, `cargo test --workspace` all green.

refs #119


ELares:

Independent reviewer probed whether the test is non-vacuous and would catch a regression. It found a **real HIGH defect**: the test was **vacuous for its stated goal**. `engine.ack()` only advances the *in-memory* cursor; the checkpoint is written by `maybe_checkpoint()` (called by the server session loop, not by the test). So `cursor.ckpt` was created-but-empty, and the test compared an empty file, never exercising the checkpoint write path it claimed to.

Fix (latest commit): explicitly call `engine.maybe_checkpoint()` after the acks (as the session loop does) and assert it returned `true` (the checkpoint was written), and assert `cursor.ckpt` is present **and non-empty**, so the byte comparison genuinely covers the checkpoint write determinism. The misleading 'advancing and checkpointing' doc comment is corrected.

This is exactly the 'passes for the wrong reason' class the adversarial pass is meant to catch. The other checks (segment determinism via the ManualClock-sourced header, fs.list ordering handled by the sort, no ambient RNG/SystemTime, the 3-message workload correctness) were confirmed sound.

Re-running CI on the fix.

Issue #101, opened 2026-06-05
[TASK] Size-optimized, reproducible release profile with byte-identical gate


ELares:
Parent: #17

Define the release build profile that minimizes binary size for scarce edge flash and makes builds bit-for-bit reproducible, then enforce reproducibility as a CI release gate. Reproducibility is the load-bearing claim behind 'an operator can confirm exactly what is on a device,' so it must be mechanized, not asserted.

## Approach

Profile: `opt-level = "s"` (default), `lto = "fat"`, `codegen-units = 1`, `panic = "abort"`, `strip = true`. Record the before/after size delta. Gate any move to `opt-level = "z"` on a #19 throughput measurement: ship `"z"` only if it costs under 5 percent on steady-state throughput. For determinism, set `SOURCE_DATE_EPOCH` from the release tag's commit date, pass `--remap-path-prefix` to strip absolute build paths and `$HOME`, set `CARGO_INCREMENTAL=0`, keep `codegen-units = 1` (parallel codegen is nondeterministic), and run the `cargo-auditable` embed before `strip`. Build the same tag twice on two different runners and assert byte-identical SHA256; a mismatch fails the release, with a documented diffoscope triage step. The `panic = "abort"` choice is binding: handlers must be panic-free by construction (see #7/#8), so this task records the constraint and links the recovery design rather than re-deciding it.

## Acceptance criteria

- [ ] The size-optimized profile is applied and the before/after size delta is recorded.
- [ ] A throughput measurement confirms no unacceptable regression from the chosen `opt-level`.
- [ ] Two builds of the same commit on two runners produce byte-identical binaries, enforced in CI.
- [ ] All determinism inputs (`SOURCE_DATE_EPOCH`, `--remap-path-prefix`, `CARGO_INCREMENTAL=0`, `codegen-units=1`, embed-then-strip order) are set and documented.
- [ ] The `panic = "abort"` constraint is cross-referenced with #7.

## Failure considerations

The main failure is a determinism input regressing (an embedded path or timestamp, or parallel codegen creeping in), which either blocks releases via the diff gate (good) or ships a nondeterministic binary that breaks an operator's independent rebuild (bad); the two-runner gate plus enumerated inputs prevents the latter. `opt-level = "z"` regressing hot loops is caught by the #19 gate. `panic = "abort"` removing the `catch_unwind` safety net is handled by requiring explicit Result-based skip logic in #7.

ELares:
[...]

PR #214, opened 2026-06-06
build: size-optimized release profile (#101, refs #17)

ELares:
## What

Apply the #101 frozen `[profile.release]` to shrink the static binary for scarce edge flash:
```toml
opt-level = "s"   # not "z" (gated on a #19 throughput measurement)
lto = "fat"
codegen-units = 1
panic = "abort"  # binding; handlers are panic-free by construction (#7)
strip = true
```

**Size delta (recorded per #101's acceptance criterion):** the native release `ironbus` binary drops **703 KiB -> 410 KiB (41% smaller, 292 KiB)**; the static musl binaries (also stripped) shrink similarly.

## Safety of `panic = "abort"`

Only the musl `cross build --release` jobs use `[profile.release]`. The per-PR and nightly test jobs run `cargo test` on the dev/test profile (cargo forces unwind for the test harness regardless), so `panic = abort` does not conflict. Verified: no CI job runs `cargo test --release`. The dev gate (fmt, clippy, full test suite) is green and unaffected.

## Out of scope (follow-up under #101)

The byte-identical two-runner reproducibility gate (`SOURCE_DATE_EPOCH` from the tag commit, `--remap-path-prefix`, `CARGO_INCREMENTAL=0`, cargo-auditable embed-then-strip, two-runner SHA256 compare) is the release-workflow slice and stays open.

The three musl build jobs in this PR's CI validate the profile (fat LTO + panic=abort + musl cross-compile) end to end.

refs #101
refs #17


ELares:

Independent reviewer checked the profile against #101's spec, the panic=abort safety, the musl/LTO risk, and drift. **No code issues found.**

Verified: the profile matches #101 exactly (opt-level "s" not "z"; fat LTO; codegen-units 1; panic abort; strip); panic=abort is safe because no CI job runs `cargo test --release` (the test/nightly jobs use the dev profile, where cargo forces unwind for libtest) and there is **no `catch_unwind` anywhere in the codebase** (handlers are panic-free, #7); `Cargo.lock` and the non-release jobs are unaffected; and the CHANGELOG accurately records the 703 KiB -> 410 KiB (41%) delta with the reproducibility gate deferred.

The reviewer's one flagged item was a 'watch CI' note on the fat-LTO + panic=abort + musl cross-compile interaction (the local validation gap). **CI closed it: all three musl builds (aarch64, armv7, x86_64) are green**, so the profile cross-compiles and statically links on every edge triple. CI 12/12 green on `987d143`.

PR #215, opened 2026-06-06
test(cli): golden-path durability across a restart with a partial ack (#133)

ELares:
## What

A step of the golden-path acceptance scenario (#133) that exercises the **core durability invariant** end to end: produce three messages, ack only the first (a `sub --ack` with credit 1 leases and acks exactly offset 0), restart the broker on the same data directory, then confirm the acked message stays committed (never redelivers) while the **uncommitted tail (offsets 1 and 2) redelivers**, via the durable cursor checkpoint.

This validates 'no acked write lost below its durability level' and 'the uncommitted tail redelivers after a restart' through the real CLI/wire/disk stack, not just at the engine layer.

## Deterministic

No timing dependence: the credit-of-1 ack commits exactly offset 0, and the restart (reopen the same dir) is synchronous. Mirrors the existing on-disk round-trip test's restart pattern (`#[cfg(unix)]`, `checkpoint_interval = 1`). Verified stable over repeated local runs (3/3).

refs #133


ELares:

Independent reviewer traced the durability claim and checked flakiness. **No issues found.**

Verified: a credit-of-1 `sub --ack` delivers and acks exactly offset 0 (committed -> 1); the cursor is durably checkpointed before restart via BOTH `maybe_checkpoint` (checkpoint_interval=1) and `checkpoint_cursor` on the client's clean disconnect, which happens before the broker shuts down; the reopen recovers committed=1 and redelivers only offsets 1 and 2. The assertions are tight enough to catch a cursor-not-durable regression: if the checkpoint were not written, the reopen would redeliver all three and both `fetched 2 message(s)` and the `NOT payload=m0` checks would fail. No timing dependence (30s visibility >> runtime, leases never expire mid-test), no port/dir collision (random ports, distinct per-pid dir name), race-free bind-before-spawn, clean shutdown + dir removal.

CI 12/12 green on `1700a3c` (all three OSes).

PR #216, opened 2026-06-06
test(cli): golden-path backpressure gate, the in-flight window bounds delivery (#133)

ELares:
## What

Adds an end-to-end golden-path test (`the_in_flight_window_bounds_each_delivery_batch`) for the **overload / backpressure** scenario in #133. It proves the **in-flight window**, not the consumer credit, is what caps each delivery batch.

## Why

The golden path (#133) calls for an explicit demonstration of the bounded-in-flight / no-unbounded-queues invariant: a consumer that asks for more than the window allows must still only be handed a window's worth at a time. Without a test, a regression that let credit override the window (re-introducing unbounded in-flight) would pass CI silently.

## How it reads

- Open an in-memory broker with `max_in_flight = 2`, `max_deliver = 5`, a 30s visibility lease.
- Produce four messages `a, b, c, d` (assert offsets `0..3`).
- `cmd_sub(credit = 10, Ack)` → delivers exactly **2** (`a`, `b`) despite the credit of 10 and four available. The acks advance the committed cursor.
- A second `cmd_sub(credit = 10, Ack)` → the freed window delivers the next **2** (`c`, `d`).
- A final `cmd_sub(credit = 10, Peek)` → `fetched 0 message(s)`, the stream is drained.

Deterministic: each `cmd_sub` acks synchronously before returning, so the committed cursor (and therefore the next window) is fully settled at each step. Passed 3/3 locally plus the full `cargo test --workspace --all-features --locked` suite.

refs #133


ELares:
## Independent adversarial review: SOLID (no defects)

An independent reviewer (not the author) was tasked to **refute** the test, defaulting to skepticism. Verdict on all five probes:

1. **Window vs credit is genuinely distinguishable.** `Engine::poll` bounds the batch at `window_end = committed + max_in_flight` (clamped to flushed), independent of consumer credit. With 4 produced, credit 10, `max_in_flight = 2`: if the window were not enforced, the first Flow (credit 10) would deliver all 4 at once and `fetched 2 message(s)` would fail. So the test does distinguish window=2 from credit=10. **SOLID**
2. **Acks advance `committed`, freeing the window.** `Engine::ack` -> `AckCursor::ack` -> `advance()` jumps the watermark; the first batch's two acks move committed 0->2, so the second fetch's window [2,4) yields c,d. A broken ack path would redeliver a,b and fail the `payload=c && payload=d` assertion. **SOLID**
3. **Deterministic, non-flaky.** `cmd_sub` is fully synchronous (fetch, then ack loop, then close); the engine is Mutex-serialized; connections run to completion before the next starts; no sleeps or time-dependent assertions. **SOLID**
4. **The `fetched N message(s)` count line is the authoritative guard.** A broken window delivering all 4 in batch 1 prints `fetched 4 message(s)` and fails the count assertion. Single-byte payloads make the `payload=` substring checks collision-free and supplementary. **SOLID**
5. **Conventions clean.** Test-local `unwrap` matches every other test in the file; no clippy/style/edition concern. **SOLID**

**Overall:** a genuine guard of the bounded-in-flight invariant; would fail decisively if window enforcement or committed advancement regressed. No vacuity, no flakiness, no misleading assertions. CI is 12/12 green on `af5bde3` (3-OS test matrix + clippy + msrv 1.78 + 3 musl triples + no-IO lint + cargo-deny + SPDX). Merging.

PR #217, opened 2026-06-06
docs(storage): pin the macOS F_FULLFSYNC durability barrier, resolve #153

### LINKED ISSUE #153: [TASK] macOS durable fsync needs F_FULLFSYNC (plain fsync does not flush the device cache)
On macOS (and iOS), `fsync(2)`/`fdatasync` push data to the drive but do NOT issue a write-cache flush barrier; the bytes can sit in the drive's volatile cache and be lost on power loss. Only `fcntl(fd, F_FULLFSYNC)` forces a platter/NAND flush. Rust's `File::sync_data()`/`sync_all()` map to plain `fsync`/`fdatasync` on Darwin, so `StdFile` (#152) acking after `sync_data` is not power-loss-safe on macOS.

macOS is a developer and CI target (per the vision in #1), not a power-loss-durable production target (production durability targets Linux musl, where `fdatasync` is a true barrier on ext4/f2fs). So this is documented as a caveat for now rather than blocking. If macOS ever needs production durability, the durability contract (#6) requires the real barrier.

## Scope
- On `target_os = "macos"`/`"ios"`, make `StdFile::sync_data`/`sync_all` use `F_FULLFSYNC` (via a safe wrapper, for example `rustix::fs::fcntl_fullfsync`), with a documented fallback to plain `fsync` on filesystems that return `ENOTSUP`.

## Acceptance criteria
- [ ] On macOS, the durable-sync path issues `F_FULLFSYNC`.
- [ ] A test (or documented manual procedure) verifies the barrier path is taken.

Found during the #152 review. Relates to #6.

ELares:
## What

Resolves #153. The issue worried that `StdFile::sync_data`/`sync_all` map to a plain `fsync` on Darwin, which does not flush the drive's volatile write cache and would break ack-implies-durable (I2) on macOS.

## Finding (verified, not assumed)

Checked the actual `std` source (`std::sys::fs::unix`, tag 1.92.0, and unchanged across our MSRV 1.78+ range): on Apple targets `std` maps **both** `datasync` (`sync_data`) and `fsync` (`sync_all`) to `fcntl(fd, F_FULLFSYNC)`, the real device-cache flush, with **no** `ENOTSUP` fallback.

```rust
#[cfg(target_vendor = "apple")]
unsafe fn os_fsync(fd: c_int) -> c_int { libc::fcntl(fd, libc::F_FULLFSYNC) }
#[cfg(target_vendor = "apple")]
unsafe fn os_datasync(fd: c_int) -> c_int { libc::fcntl(fd, libc::F_FULLFSYNC) }
```

`StdFile` delegates straight to these, so **the durable-sync path already issues the `F_FULLFSYNC` barrier on macOS.** The issue's premise ("maps to plain fsync") is false for current `std`. Adding our own `libc`/`rustix` `fcntl` wrapper would only duplicate the barrier, so this PR deliberately adds **no** redundant sync code.

## What this PR does instead

- **A load-bearing comment** on the `StdFile` sync path: why we delegate to `std`, and why we keep `std`'s no-`ENOTSUP`-fallback behaviour on purpose. On a filesystem that cannot honour the barrier, the sync errors, the writer freezes (`/readyz` -> 503), and the broker stops acking. Failing closed beats silently downgrading the barrier and letting an ack outrun a real flush.
- **A platform durability table** in `docs/USAGE.md`: Linux musl is the production target (`fdatasync`/`fsync` true barriers); Apple uses `F_FULLFSYNC`; Windows is not a v1 broker target.
- **An Apple-gated test** (`apple_durable_sync_takes_the_full_fsync_barrier_path`) that exercises the `F_FULLFSYNC` path on the macOS CI runner and asserts the round-trip. It honestly does **not** claim to prove the physical platter flush in CI (that needs a real power cut, owned by #133 and the device gates); it pins that the barrier syscall path is taken and does not error, so a regression that stopped issuing it would be caught.

This satisfies #153's acceptance criteria: the durable-sync path issues `F_FULLFSYNC`, and a test plus a documented procedure verify the path is taken. Relates to #6.

## Verification

- Full `cargo test --workspace --all-features --locked` green; the new Apple test runs and passes on a darwin host.
- No behaviour change: docs, one comment, one test only.

Closes #153


ELares:
## Independent adversarial review: SOLID (central claim corroborated)

An independent reviewer was tasked to **refute** the premise (that closing #153 without code is a cop-out resting on a false claim). It instead corroborated it with additional evidence and returned SOLID on all five probes.

1. **The load-bearing claim is TRUE.** Verified against the `std` source and its history: Apple targets map BOTH `File::sync_all` (`fsync`) and `File::sync_data` (`datasync`) to `fcntl(fd, F_FULLFSYNC)`, no ENOTSUP fallback. Corroborated by rust-lang/rust PR #60121 (2019, `sync_all` -> F_FULLFSYNC) and PR #77900 (2020, confirms `sync_data` -> F_FULLFSYNC on Apple, even debating whether it is "too much" and keeping it). Both predate our MSRV 1.78 by years, so the guarantee holds across the supported range. **SOLID**
2. **`StdFile` delegates straight to `std`** (`self.file.sync_data()`/`sync_all()`, no wrapper), and the hot produce path syncs before ack via `sync_data`, so the barrier is on the path that matters. **SOLID**
3. **The Apple-gated test is honest and non-vacuous:** it calls both sync entry points on a real file and asserts a round-trip, the comment honestly states it does not prove the physical platter flush in CI, and the `target_vendor = "apple"` gate keeps non-Apple builds warning-clean. **SOLID**
4. **Fail-closed reasoning is correct:** a sync error sets the writer frozen (`WriterFrozen`) and propagates out of `produce`, failing the ack rather than being swallowed, so the no-ENOTSUP-fallback behaviour is genuinely fail-closed. **SOLID**
5. **No em/en dashes; CHANGELOG `### Documentation` matches the existing heading style; USAGE table renders.** **SOLID**

**Overall:** closing #153 with documentation plus a guard test, and deliberately NOT adding a redundant `fcntl` wrapper, is the correct engineering call; the premise is verified, not assumed. CI is 12/12 green on `0109f95` including `test (macos-latest)` (which ran the new barrier test). Merging.

PR #218, opened 2026-06-06
fix(storage): model fdatasync not persisting a set_len truncation (#158)

### LINKED ISSUE #158: [TASK] InMemoryFile: model fdatasync not persisting set_len metadata (sync_data vs sync_all)
## Context

`InMemoryFile` (`crates/ironbus-storage/src/io.rs`) models `sync_data` (fdatasync) and `sync_all` (fsync) identically: both capture the full live image, including the file length, as the durable image. On a real disk, `fdatasync` flushes data but is not required to persist metadata such as a length change from `set_len` (truncation/extension). So a `set_len` followed by only `sync_data` is modeled as durable but may not be on real hardware.

Surfaced by the self-review on #157 (item N3). It is a sim-vs-real divergence that the recovery code consuming the IO/fs seam could rely on: a recovery step that truncates a torn tail with `set_len` and then only `sync_data`s would pass in the simulation but might not persist the truncation on a power cut.

## Task

Decide and encode the durability contract for length metadata, then make the in-memory model match:

- Either treat `set_len` as requiring `sync_all` (fsync) to become durable, and have `simulate_power_loss` restore the length from the last `sync_all`, not the last `sync_data`; or
- Document that IronBus always pairs a recovery truncation with `sync_all` (which is the conservative real-disk requirement) and add a debug assertion / test that the truncation path uses `sync_all`.

Add a test that a `set_len` + `sync_data` (only) does NOT survive a simulated power loss under the chosen contract, and that `set_len` + `sync_all` does.

## Scope

Durability-model fidelity for the fsync policy and power-loss contract; refs #6.

ELares:
## What

Resolves #158. `InMemoryFile` modeled `sync_data` (fdatasync) and `sync_all` (fsync) identically: both snapshotted the full live image, including the file length, as the durable image. A real `fdatasync` flushes data but is **not** required to persist metadata such as a length shrink from `set_len`, so a `set_len` truncation that was only `sync_data`'d was modeled as durable when on real hardware it may not be.

The `RandomAccessFile` trait **already documents** the contract ("The new length is metadata, so it is only crash-durable after `sync_all`"). This PR makes the in-memory implementation actually enforce it.

## The contract (now encoded)

- `sync_data` (fdatasync): persists file data and any growth that data caused, but does **not** persist a `set_len` length shrink. The durable image keeps its old length and the bytes beyond the new live length.
- `sync_all` (fsync): a full barrier, persisting data **and** the length, including a truncation.
- `simulate_power_loss` therefore reverts a truncation that was only `sync_data`'d, restoring the pre-truncation length and the tail bytes; an in-place edit made under an unsynced truncation still survives via fdatasync.

Truncation is the **only** length change IronBus makes through `set_len` (recovery drops a torn tail, then `sync_all`s, in `Log::recover` at `log.rs`), and this model now enforces that pairing: a regression that truncated with only `sync_data` would fail the crash-recovery sweep.

## Why the one test change

`recovery_is_idempotent_after_a_fault_during_recovery` built its durable torn tail with `set_len(len - 3)` + `sync_data`. Under the faithful contract that truncation is **not** durable, so the subsequent `simulate_power_loss` would restore the full (un-torn) image and recovery's truncation branch would never re-fire, making the test vacuous. It now uses `sync_all` to make the torn-tail image genuinely durable, which is exactly what the test intends ("the torn-tail image is the durable on-disk reality"). This is the model correctly surfacing an unfaithful shortcut.

## Tests added

- `a_truncation_is_durable_only_after_sync_all`: a `set_len` shrink + `sync_data` is reverted by a power loss (old length and tail return); a `set_len` shrink + `sync_all` survives.
- `fdatasync_persists_an_in_place_edit_under_an_unsynced_truncation`: a data edit within the surviving range is durable via fdatasync even while the truncation is not, so the power-loss image keeps the edit and the un-truncated tail (per-byte faithfulness).

## Verification

- `crash_recovery` + `determinism` sweeps pass at `PROPTEST_CASES=5000`; all 109 storage unit tests green.

Relates to #6 (durability contract) and #21 (crash-class gates).

Closes #158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Fixed
+- `InMemoryFile` modeled `fdatasync` (`sync_data`) and `fsync` (`sync_all`) identically, treating a `set_len` truncation as durable after either (#158). A real `fdatasync` flushes data but need not persist a length-shrink (metadata), so the simulation now reverts a truncation that was only `sync_data`'d on a power loss and persists it only after `sync_all`, while in-place edits and data growth stay durable under `fdatasync`. This pins the conservative real-disk contract that `Log::recover` already follows (truncate a torn tail, then `sync_all`); a regression that truncated with only `sync_data` would now fail the crash-recovery sweep. Includes two model tests and updates the recovery-idempotence gate's torn-tail setup to make the truncation durable via `sync_all` (refs #6, #21).
 - The `dead_letter_iff_a_finite_cap_is_exceeded` delivery proptest no longer relies on a `prop_assume!` that rejected ~2.5% of inputs: under a deep (nightly) sweep that exhausted proptest's global-reject budget and aborted the test. The strategy now generates only valid configs (max == 0 forces the unlimited opt-in), so it rejects nothing and is robust at any case count (refs #21).
 
 ### Added
--- a/crates/ironbus-storage/src/io.rs
+++ b/crates/ironbus-storage/src/io.rs
@@ -108,7 +108,10 @@ fn invalid_input(msg: &'static str) -> io::Error {
 struct State {
     /// The current (possibly-unsynced) bytes.
     live: Vec<u8>,
-    /// The bytes as of the last sync: the only bytes that survive a power loss.
+    /// The bytes that survive a power loss. Its length is the durable file length, which
+    /// `sync_data` (fdatasync) advances for data but never shrinks for a `set_len`
+    /// truncation (length is metadata): only `sync_all` (fsync) makes a truncation
+    /// durable. See the sync methods below and [`InMemoryFile`] for the contract.
     durable: Vec<u8>,
 }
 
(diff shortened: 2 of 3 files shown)


ELares:
## Independent adversarial review: SOLID (model faithful, test change load-bearing not masking a regression)

An independent reviewer was tasked to refute the model and to prove the one test change does NOT hide a regression. It hand-traced every case and returned SOLID on all six probes.

1. **New `sync_data` is internally correct and POSIX-faithful.** Cases traced by hand: write-growth stays durable; `set_len(3)` + `sync_data` keeps durable at the old length (truncation reverted on power loss); an edit under an unsynced truncation yields `XBCDEF` (edit durable, tail returns); a `sync_all`'d truncation persists. The `durable[..live.len()]` slice is guarded by the `live.len() < durable.len()` else-branch, so no panic; the `>=` boundary (equal lengths) takes the full-clone path. **SOLID**
2. **Matches POSIX (`fdatasync` persists data + metadata needed to read it; a shrink is not).** Independently grepped non-test `set_len`: only `log.rs` recovery (shrink) and `checkpoint.rs` `set_len(0)`-before-write (an extension, durable under any model). No `set_len` extension elsewhere, so the documented scoping is accurate and the model is sound for actual usage. **SOLID**
3. **The `sync_data` -> `sync_all` test change is load-bearing, not a cover-up.** The reviewer traced that under the new model with the OLD test code, `set_len(len-3)` + `sync_data` leaves the durable image at full length, so the power loss restores the intact image, recovery finds nothing to truncate, and `recovered_truncated_bytes() > 0` FAILS. Switching to `sync_all` restores the test's original intent (a genuinely durable torn tail) and still exercises (i) the first recovery's faulting `sync_all`, (ii) a power loss reverting that truncation, (iii) a second recovery that re-truncates. Non-vacuous. **SOLID**
4. **Blast radius independently re-derived: exactly one test needed updating.** Every other `set_len`+`sync_data` site (`crash_recovery.rs` 150/172, `log.rs` 728/1159) either does not simulate a power loss (recovery reads the truncated live image) or pairs `set_len(0)` with a re-extending write. No missed regression. **SOLID**
5. **Both new tests are non-vacuous:** each uses `simulate_power_loss` and would fail under the old (identical-sync) model or a naive tail/edit-dropping model. **SOLID**
6. **No em/en dashes.** **SOLID**

**Overall:** the model faithfully enforces the fdatasync/fsync distinction that `Log::recover` already assumes, and the crash-recovery sweep will now fail if a regression reintroduces sync_data-persists-truncation. CI is 12/12 green on `4c06d2e` (3-OS tests + clippy + msrv 1.78 + 3 musl triples + no-IO lint + cargo-deny + SPDX); sweeps verified at PROPTEST_CASES=5000 locally. Merging.

Issue #115, opened 2026-06-05
[TASK] Specify and enforce the tiny profile RAM budget


ELares:
Parent: #20

Turn the asserted 64 MiB RAM ceiling on the `tiny` profile into an itemized, provably-bounded budget and a boot-time guard that refuses to start when the configured caps cannot fit. The current draft lists a ceiling, an 8 MiB segment, and a 32 MiB mmap cap but never shows the per-buffer split or what happens on overflow, so the headline guarantee (every buffer bounded, no unbounded queue, all summing under the cap) is unverifiable. This sub-issue makes the ceiling a derived, enforced invariant and feeds the bounded-buffer machinery owned by #10.

## Approach

Publish an explicit per-buffer budget that sums to <= `ram_ceiling_bytes` (64 MiB on tiny): producer accumulation 8 MiB, active-segment write buffer 8 MiB, per-topic spill index 8 MiB, in-flight/unacked set 8 MiB, read/replay cache 16 MiB, codec plus checksum scratch 4 MiB, and a 12 MiB hard reserve. Set `mmap_max_bytes=0` on tiny so no uncounted mapped pages exist. Make `ram_ceiling_bytes` a global cap divided across active topics by equal-share with a documented per-topic floor; reject topic creation when the floor cannot be granted from the remaining budget. At startup, measure RSS and refuse to boot if RSS plus configured caps exceed the ceiling. Wire each cap to a config key under #14 so the same arithmetic holds for `small` and `server` profiles.

## Acceptance criteria

- [ ] An itemized RAM budget is documented and provably sums under `ram_ceiling_bytes` for `tiny`.
- [ ] `mmap_max_bytes=0` on tiny; no buffer is unbounded.
- [ ] The binary refuses to boot when measured RSS plus configured caps exceed the ceiling, with a clear error.
- [ ] Topic creation is rejected when the per-topic floor cannot be granted from the remaining global budget.
- [ ] A RAM-burst test asserts RSS stays under the ceiling under sustained write pressure.

## Failure considerations

The dominant risk is OOM-kill of the single binary under burst because uncounted buffers (mmap pages, per-topic spill across N topics, codec scratch) push RSS past the cap; the itemized budget, `mmap_max_bytes=0`, per-topic floors, and refuse-to-boot guard close this. A secondary risk is the budget drifting out of sync with code as buffers are added, so a test must assert the sum of configured caps never exceeds the ceiling, failing CI if a new buffer is introduced uncounted.

[...]

PR #219, opened 2026-06-06
perf(storage): stream segment recovery to bound scan memory (#156)

ELares:
## What

Resolves the memory hazard in #156. Recovery scanned each segment by reading the **whole record region** into one buffer (`vec![0u8; body_len]`, up to ~64 MiB) and then materializing **every** record (key, headers, payload) into a `Vec<OwnedRecord>`, so peak was roughly **2x the segment** on a device whose RAM budget is deliberately scarce (#115). Yet recovery never uses the payloads: it needs only each record's sequence, the valid record count, and the torn-tail boundary (`valid_end`).

## How

- New `codec::decoded_len(header)` (ironbus-core): reads a frame's total on-disk length from its 36-byte header alone (validating magic, version, header CRC), so a streaming reader can read exactly one record at a time. A codec test pins that it returns `decode`'s consumed length and rejects the same bad headers.
- New `SegmentReader::scan_recovery` (ironbus-storage): a streaming scan that, per record, reads the header to learn the length, reads the rest of the frame, validates it **exactly as the buffered scan does** (header + body CRC, sequence continuity, identical footer-trust rules), then **drops the bytes**. It returns only the metadata `recover()` consumes (`record_count`, `last_seq`, `valid_end`, `clean`, `footer`). Peak memory is the largest single record instead of the whole region (a reused scratch buffer).
- `Log::recover` switches to `scan_recovery`; the per-record sequence-continuity check moves into the streaming walk (same `RecoveredSequenceMismatch` hard error for a CRC-valid frame with an out-of-order seq).
- The buffered `SegmentReader::scan` (with `OwnedRecord` payloads) is **retained** for the delivery read path (`read_from`), which genuinely needs payloads. The delivery-path streaming is out of scope here (a possible follow-up).

## Equivalence (no behavior change)

- `scan_recovery_agrees_with_scan_across_shapes`: asserts `scan_recovery` matches the buffered `scan` on count, last seq, `valid_end`, `clean`, and footer across clean, sealed, empty, header-only, torn-tail, and corrupt-body segments.
- `scan_recovery_reports_a_recycled_frame_with_a_bad_seq`: a CRC-valid out-of-order frame is a hard error, not a silent truncation.
- The whole `crash_recovery` + `determinism` sweeps pass at `PROPTEST_CASES=20000`; `decoded_len` agreement pinned in core.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` clean; `cargo build -p ironbus-core -p ironbus-storage` clean under `-D warnings`.
- `crash_recovery`, `determinism`, codec, and storage-lib suites green at `PROPTEST_CASES=20000`.

refs #156, #115, #21


ELares:
## Independent adversarial review: SOLID (behavior-preserving, memory-bounded)

An independent reviewer audited this line-by-line against the old buffered scan, defaulting to deep skepticism since it touches the longest-valid-prefix guarantee. SOLID on all seven probes:

1. **`decoded_len` is exactly `decode`'s header-validation prefix** (same checks, same order: short -> Truncated, magic, version, header CRC, overflow-safe length sum -> TooLarge). The `HAS_KEY`/`BadLength` body check correctly lives only in `decode`, which the streaming scan still calls on the full frame, so no frame slips through. **SOLID**
2. **`scan_body_streaming` matches `scan_body`** across clean / torn-tail (short header AND body-cut-off) / corrupt-body / corrupt-header: same `cursor`, same `clean`, and `pos` only advances on `decode` success (so a corrupt frame leaves the cursor at its start, identical to the buffered break). **SOLID**
3. **The seq-continuity check moved into the walk is equivalent:** `base_seq + count` == `base_seq + index`, same `RecoveredSequenceMismatch{index,expected,found}`. The only change is which hard error surfaces if a frame has BOTH a wrong seq AND a footer mismatch (a near-impossible case, since seq is header-CRC-covered); the accepted/rejected prefix is identical, so it is benign. Per-segment `base_seq` is each segment's own header. **SOLID**
4. **Footer trust rules identical:** `ends_at_footer`, `body_matches`, `FooterSegmentMismatch`, the empty-body `last_seq == base_seq` default, and `valid_end` (sealed `body_end` / unsealed `header_end + cursor`) all match `scan`. **SOLID**
5. **`recover` edits correct:** `Option<RecoveryScan>`, `record_count`/`last_seq` substitutions, `SegmentScan` import removed (no longer named) while `OwnedRecord` stays for `read_from`, which still uses buffered `scan` and is unchanged. **SOLID**
6. **Memory is bounded, no hostile allocation:** the order is `decoded_len` (caps at the 1 GiB ceiling) -> check `total > remaining` -> THEN `scratch.resize(total)`, so a header claiming 1 GiB with few bytes left breaks before allocating. Peak is one record, not the region. **SOLID**

**Overall:** a faithful, behavior-preserving, memory-bounded replacement; no silent loss/gain, no off-by-one, no unbounded allocation. CI is 12/12 green on `23e0c7d` (3-OS tests + clippy + msrv 1.78 + 3 musl triples + no-IO lint + cargo-deny + SPDX); crash-recovery, determinism, codec, and storage-lib suites verified at PROPTEST_CASES=20000. This completes #156's scope (recovery). Merging.

PR #220, opened 2026-06-06
feat(proto): distinct response frames per verb instead of overloading Ok (#179)

### LINKED ISSUE #179: [TASK] Distinct response frame types per verb instead of overloading Ok
## Context

The session (#174, #178) overloads the single `Ok` frame with three incompatible body shapes: a `Pub` `Ok` carries a u64 offset (8 bytes), an `Ack` `Ok` a u8 status (1 byte), and a `Flow` `Ok` a u32 delivered-count (4 bytes). Disambiguation relies entirely on the client remembering its last request. This works for the current strict synchronous request/response per connection, but it is undocumented at the frame level, will break a generic `Ok` handler, and forecloses request pipelining. Surfaced by the self-review on #178.

## Task

Replace the overloaded `Ok` with explicit, self-describing response frames (or a correlation id on each response):
- `PubAck { offset }`, `Ack { status }` (committed/fenced), `FlowEnd { delivered }`, or a generic response carrying a request-correlation id plus a typed payload.
- Keep `Err { message }` distinct.
- Update the wire-tag table and the session/client to match, and pin the new tags.

## Scope

Wire protocol clarity / forward-compat; refs #11 refs #178

ELares:
## What

Resolves #179. The session overloaded the single `Ok` response frame (tag 11) with **three incompatible body shapes**: a `Pub` `Ok` carried an 8-byte offset, an `Ack` `Ok` a 1-byte status, a `Flow` `Ok` a 4-byte delivered count. Disambiguation relied entirely on the client remembering its last request, which was undocumented at the frame level, would break a generic `Ok` handler, and foreclosed request pipelining.

## How

Three self-describing response frames, each reply routed to its own (tags pinned by the frozen-tag test):

| Frame | Tag | Body | Sent for |
|-------|-----|------|----------|
| `PubAck` | 14 | offset, 8-byte LE u64 | `Pub` |
| `AckStatus` | 15 | status, 1 byte (0 fenced / 1 committed-requeued-extended / 2 progress cap) | `Ack`, `Nack`, `Term`, `Progress` |
| `FlowEnd` | 16 | delivered count, 4-byte LE u32 | `Flow` batch terminator |

`Err` stays distinct. `Ok` is **retained** as a reserved, body-less generic success, documented not to be overloaded again (kept rather than removed to avoid churning the frozen tag table; tags are never reused).

## The ambiguity is now structural

The client's former defense, a length check that rejected an "eight-byte `Ok` whose low byte is 1" so a pub offset could not be misread as a committed ack, is no longer needed: a pub offset arrives as a `PubAck`, a different frame type, handled as a clean `Unexpected`. The test that pinned that hazard is rewritten to assert exactly this (`a_pub_ack_in_response_to_an_ack_is_unexpected_not_misread`).

## Scope

- `ironbus-proto`: three new `FrameType` variants with body-layout docs, tags 14/15/16, `from_u8`, the all-types array, and frozen-tag assertions; the unknown-tag forward-compat sweep now starts at 17.
- `ironbus-server`: the 8 overloaded `reply(..., Ok, ...)` sites routed to `PubAck` / `AckStatus` / `FlowEnd`; session + integration tests updated.
- `ironbus-client`: `produce` decodes `PubAck`, `fetch` terminates on `FlowEnd`, the four ack-path methods decode `AckStatus`; tests updated.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` clean; `cargo build --workspace` clean under `-D warnings`.
- Full `cargo test --workspace --all-features --locked` green (proto frozen-tag + bijection, session, client, and server integration tests all pass).

Closes #179


ELares:
## Independent adversarial review: SOLID (server/client agree per verb); review comments addressed

An independent reviewer audited the protocol change for any verb where the server and client now disagree, and returned SOLID on all seven probes:

1. **Per-verb agreement is exact.** Pub -> `PubAck` (both sides), Ack/Nack/Term/Progress -> `AckStatus` (all six server reply paths in `handle_ack` and all four client decode methods), Flow terminator -> `FlowEnd` (both sides); Connect->Info, Ping->Pong, Err unchanged. Grep confirmed zero `reply(out, FrameType::Ok` in non-test server code and zero `(FrameType::Ok` decode arms in the client. **SOLID**
2. **Body layouts unchanged** (only the tag changed): PubAck = 8-byte LE u64 offset, AckStatus = the same 1-byte status values (0/1/2 with identical per-path meaning), FlowEnd = 4-byte LE u32 count; client decoders read the same widths. **SOLID**
3. **Wire-tag table intact:** 14/15/16 added to both `as_u8` and `from_u8` (bijection holds), `ALL_TYPES` is length 16, frozen-tag assertions cover the three new tags, `Ok` stays tag 11, no existing tag (1-13) changed. **SOLID**
4. **Forward-compat sweep** correctly excludes 14/15/16 (now known) by starting the unknown-tag proptest at 17. **SOLID**
5. **The rewritten ambiguity test is genuine:** a `PubAck` sent in response to an ack is `Unexpected(PubAck)`, proving the type (not a body-length guess) now disambiguates. **SOLID**
6. **No verb left behind;** the CLI goes through the client API and is unaffected. **SOLID**

The reviewer also flagged (non-blocking) several **stale doc comments** in session.rs still describing the old overloaded `Ok` (the module response-conventions doc, `handle_flow`'s terminator docstring, the `ack_reply` helper doc, the Flow-terminator and empty-batch comments, and a test name). I addressed all of them in a follow-up comments-only commit (`docs(session): refresh stale Ok-frame comments...`), renaming the test to `..._then_flow_end`. Code and wire protocol are unchanged. Re-running CI on the doc commit before merge.

PR #221, opened 2026-06-06
feat(storage): validate LogConfig against a minimum segment-size floor (#162)

### LINKED ISSUE #162: [TASK] Validate LogConfig: reject a max_segment_bytes below a sane floor

ELares:
## What

Resolves #162. `LogConfig.max_segment_bytes` is a `pub` field with no lower bound. A cap of `0` (a legal value) or any sub-header value silently fragments the log into **one-record segments** with no diagnostic: the cap is compared against the write position (which starts at the 64-byte header), so a tiny cap rolls after every record.

## How

- `LogConfig::MIN_MAX_SEGMENT_BYTES`: a floor derived from the frozen format constants, `SEGMENT_HEADER_LEN + SEGMENT_FOOTER_LEN + 2 * (RECORD_HEADER_LEN + RECORD_TRAILER_LEN)` = **184 bytes**, so a segment always has room for the header, the footer, and at least two minimum records (more than one record).
- `LogConfig::new(max_segment_bytes) -> Result<LogConfig, LogConfigError>`: the validating constructor, returning a typed `LogConfigError::MaxSegmentBytesTooSmall { value, floor }` below the floor.
- The CLI `serve` path now builds its `LogConfig` through `new` (its existing 4 KiB user-facing `--max-segment-bytes` minimum already clears the floor, so this is defense in depth and gives `new` a real production consumer).

## Decisions (per the issue's acceptance criteria)

- **Constructor-enforce, not `Log::open`-reject.** The issue allows either; I chose the constructor so the field can stay public for tests that deliberately use a tiny cap to force rolling (they construct via the struct literal, the documented test path). The floor does not affect recovery or rolling, only the cap value, so no existing test needed migration.
- **The cap counts total segment bytes** (the 64-byte header included), documented on the field; a record larger than the cap still gets its own segment (the empty-segment guard is unchanged).
- **The field stays `pub`** and its doc now records that setting it below the floor directly (including `0`) is a footgun; prefer `new`.
- `LogConfig` now derives `PartialEq`/`Eq`.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` clean; `cargo build --workspace` clean under `-D warnings`.
- Full `cargo test --workspace --all-features --locked` green; new tests cover `new` rejecting `0`/`1`/`64`/`floor-1` and accepting the floor and the default (and opening + rolling a `new`-built config).

Closes #162


ELares:
## Independent adversarial review: SOLID (no bugs)

An independent reviewer probed for an off-by-one in the floor and a CLI error-message regression, and returned OVERALL SOLID:

1. **Floor is correct and meaningful.** `MIN_MAX_SEGMENT_BYTES = 64 + 32 + 2*(36+8) = 184`. Traced against the roll condition (`write_pos >= cap && record_count > 0`, write_pos starting at the 64-byte header): cap 184 holds ~3 minimum (44-byte) records before rolling (record 1 -> 108, 2 -> 152, 3 -> 196 rolls), so "more than one record" holds with margin. usize arithmetic before the u64 cast cannot overflow. **SOLID**
2. **Validation boundary exact:** rejects 0/1/64/183, accepts 184 and the 64 MiB default, with correct `value`/`floor`. **SOLID**
3. **Constructor-enforce, field stays pub:** the reviewer flagged the residual struct-literal footgun as WEAK-but-acceptable, since #162 explicitly allows "constructor enforce" and "if kept pub, document that 0 and sub-header values fragment the log", which the field doc now does. The tiny-cap roll tests pass unchanged via the struct literal. Deliberate, documented, issue-sanctioned. **Acceptable**
4. **CLI routing:** `open_disk_engine` uses `LogConfig::new(...).map_err(...)?`; the existing 4 KiB arg-level minimum runs first and always clears the 184 floor, so this is harmless defense in depth (and gives `new` a real production consumer). **SOLID**
5. **`PartialEq`/`Eq` derive** is needed (the tests assert_eq! a `Result<LogConfig, _>`) and safe (a single u64, not a map key). **SOLID**
6. **Error type** matches the manual `StorageError` style, Display is correct with a valid line-continuation, no dashes. **SOLID**
7. **Tests non-vacuous** (reject below floor / accept floor+ / open+roll a `new`-built config); honest `small_config` comment. **SOLID**

**Overall:** floor correct, validation sound, design coherent with #162, nothing broken. CI is 12/12 green on `a9e471f`. Merging.

PR #222, opened 2026-06-06
ci: run the byte-parser tests on a 32-bit target (#148)

### LINKED ISSUE #148: [TASK] Run the parser tests on a 32-bit target in CI
The record-codec review (#147) found a 32-bit `usize` overflow in `decode` (now fixed by doing the length arithmetic in `u64`). The fix is correct on all targets, but CI currently only builds/tests on 64-bit, so a future `usize`-width regression in the parsers would not be caught.

## Scope
Run the test suite on at least one 32-bit target (i686-unknown-linux-gnu or armv7 via `cross`/qemu), at least for the crates that parse untrusted bytes (`ironbus-core`). This complements the musl cross-builds in #128, which are build-only.

## Acceptance criteria
- [ ] CI runs `cargo test -p ironbus-core` on a 32-bit target.
- [ ] The `crafted_huge_lengths_are_rejected_not_panicked` regression test passes there.

Found during the #147 review. Relates to #128, #21.

ELares:
## What

Resolves #148. The record-codec review (#147) found a 32-bit `usize` overflow in `decode`, fixed by doing the length arithmetic in `u64`. That fix is correct on all targets, but CI only **tested** on 64-bit (the musl cross-builds are build-only), so a future `usize`-width regression in a parser would ship uncaught.

## How

A new `parse-32bit` job runs `cargo test -p ironbus-core` on `i686-unknown-linux-gnu` via `cross`:
- `ironbus-core` is the crate that parses untrusted bytes (the record and segment codecs), so its tests are exactly the ones that must hold on a 32-bit target.
- `i686-unknown-linux-gnu` has 32-bit `usize`, and its binaries run **natively** on the x86_64 runner (no qemu), so the job is fast and not flaky.
- The global `RUSTFLAGS: -D warnings` is forwarded into the cross container, so this also gates a warning-clean 32-bit compile.

## Acceptance criteria (#148)

- [x] CI runs `cargo test -p ironbus-core` on a 32-bit target.
- [x] `crafted_huge_lengths_are_rejected_not_panicked` (ironbus-core, codec.rs) runs there.

## Verification

- YAML validated (parses; the new `parse-32bit` job is registered alongside the existing jobs).
- `ironbus-core`'s 89 tests pass on 64-bit locally; the codec is 32-bit-safe by construction (u64 length math), so the new 32-bit job is expected green. The job run on this PR is the real gate. CI-config only; no Rust source changed.

Closes #148


ELares:
## Independent adversarial review: PASS (high confidence), plus verified non-vacuous execution

A reviewer probed specifically for a **false pass** (a green job that does not actually run 32-bit code) and returned SOLID on all six probes:
1. **Genuinely 32-bit:** `i686-unknown-linux-gnu` is a tier-1 target with 32-bit `usize`; `cross test` builds for i686 and runs the test binary natively on the x86_64 runner (no qemu), with no fallback to the host target. **SOLID**
2. **Cross invocation correct:** `cross` installed via `taiki-e/install-action`, `cross test --locked -p ironbus-core --target i686-unknown-linux-gnu` forwards valid cargo args, i686 has a prebuilt cross image, and `-D warnings` is forwarded into the container (same pattern as the proven `cross-build` job). **SOLID**
3. **Low flake risk:** no qemu; `ironbus-core`'s only deps (`crc32c` runtime, `proptest` dev) are pure Rust and i686-cross-compilable, no C toolchain needed. **SOLID**
4. **Right tests present:** `crafted_huge_lengths_are_rejected_not_panicked` (codec.rs) and the full codec/segment suite run under `-p ironbus-core`; scope (ironbus-core is the untrusted-byte parser) is honest. **SOLID**
5. **No silent skip:** required job, no `if:`, single-entry matrix valid, command executes. **SOLID**
6. **No em/en dashes.** **SOLID**

**Verified from the job log (not just a green check):** the run genuinely executed on i686 with `test result: ok. 89 passed` for the ironbus-core suite, and `codec::tests::crafted_huge_lengths_are_rejected_not_panicked ... ok` (the #148 acceptance-criterion regression). So the gate is functional, not cosmetic. CI is 13/13 green on `ceada06`. Merging.

Issue #15, opened 2026-06-05
[DESIGN] CLI and diagnostics: better than the NATS CLI, with a real view into the data


ELares:
IronBus ships its operator and developer tooling inside the same single static binary as the broker. One file gives you a server (`ironbus serve`), a full pub/sub client, a load generator, deep introspection, a live TUI dashboard, and (the differentiator) the ability to decode and display the on-disk records of a data directory with no server running at all. The goal is to be strictly better than the NATS CLI for everyday work and to expose the things operators of Kafka, Pulsar, and Redis Streams have always wished their tools showed: fsync latency, backpressure state, lag broken down by cursor, and a frame-accurate view of the WAL and segments including the records the recovery path decided to skip. Every command emits human-readable output by default and structured `--json` for scripting.

## Why this matters for IronBus

A great CLI is the most visible expression of the **Simple** tenet: a newcomer should publish, subscribe, benchmark, inspect lag, and read raw records within five minutes of `curl | sh`, with no client library, no extra package, and no companion daemon. It is also a **Resilient** tenet feature. When a power-cut leaves an edge node with a torn WAL tail or a poison segment, the operator needs to see exactly what #7 truncated and what #8 quarantined, offline, from the data directory alone, without risking a second crash by starting the broker. **Edge First** shapes the constraints: the tooling must run over a flaky link (long-poll style waits, not busy spins), draw a TUI over a 9600-baud-feeling SSH session, and add zero bytes and zero dynamic dependencies to the shipped artifact (**Cross Platform**, one musl binary per arch). Diagnostics are where the storage engine (#4, #5) and the metrics from #16 become something a human can act on.

## Prior art

The **NATS CLI** is the bar to clear for ergonomics: subcommands like `nats pub`, `nats sub`, `nats bench`, and `nats stream info` map cleanly onto a queue, and `nats server report` gives a live operational view. Its limit is that it speaks only the wire protocol; it cannot read a JetStream file store directly, so it is blind when the server will not start. JetStream does expose rich state (per-consumer `num_pending`, `num_ack_pending`, `redelivered`, AckFloor, plus `$JS.EVENT.ADVISORY.*` events), which is exactly the data a good `info`/`consumer` view should surface.

[...]

PR #223, opened 2026-06-06
feat(storage): add the versioned structured loss-report schema (#120)

ELares:
## What

The first piece of #120: the versioned, structured **loss report** that recovery will emit and that the metrics endpoint (#16) and the offline inspector (#15) will read. #120 calls for this schema to "land early and be the single source of truth for what 'recovered correctly' means," so corpus fixtures can assert exact values instead of each layer inventing its own shape.

New module `ironbus_storage::loss`:
- `LossReport { schema_version, events }`, stamped to `SCHEMA_VERSION = 1` by the constructors, with saturating `total_bytes_skipped` / `total_records_lost_estimate`.
- `LossEvent { segment_id, byte_offset_start, byte_offset_end, bytes_skipped, records_lost_estimate, reason_code }`, with a `span()` constructor that derives `bytes_skipped` from the offsets (saturating, so a reversed range is an empty span, never an underflow).
- `ReasonCode` (`TornTail`, `CorruptRecordHeader`, `CorruptRecordBody`, `CorruptSegmentHeader`, `SequenceGap`) with **frozen numeric `code()`s** (1..5) for the metrics side; `serde` serializes the variant name for a human-readable report.
- The I3 bounded-loss caps as documented constants: `PER_EVENT_BYTE_CAP` (64 MiB) and the 1% global fraction (`global_loss_cap_bytes`).

## Serialization (per the chosen approach)

Following the agreed minimal-footprint decision: `serde` (derive only) is a normal dependency so the report is a versioned, fixture-assertable artifact; **`serde_json` is a dev-dependency**, so the static edge binary does not pay for a JSON format crate until an emitter needs one. The derives are the contract; the dev test exercises a `serde_json` round-trip.

## Scope and follow-ups

This PR is the schema (acceptance criterion: "a versioned serde `LossReport` struct ... with a schema_version"). Linked follow-ups under #120: the I1 to I4 invariant checkers, the cap enforcement (freeze read-only and exit non-zero on exceed), and recovery emitting the report (wiring it to the existing `recovered_truncated_bytes`, with the `records_lost_estimate` semantics).

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-storage --all-targets --all-features` clean; `cargo build -p ironbus-storage` clean under `-D warnings`.
- Full `cargo test --workspace --all-features --locked` green; six new tests cover the schema_version stamp, `span` byte math + reversed-range saturation, saturating totals, frozen reason codes, the 1% cap math, and a `serde_json` round-trip that pins the serialized shape (`"schema_version":1`, `"reason_code":"CorruptRecordBody"`).
- `Cargo.lock` updated; `serde`/`serde_json` and their transitive deps carry MIT/Apache-compatible licenses (the cargo-deny job gates this).

refs #120, #16, #15, #21


ELares:
## Independent adversarial review: SOLID (9/9)

A reviewer audited the schema against the #120 spec and probed the dependency posture, returning SOLID on every point:

1. **Schema faithful to #120:** all six `LossEvent` fields plus a stamped `schema_version` present and correctly typed. **SOLID**
2. **Numerically safe:** `LossEvent::span` uses saturating subtraction (reversed range -> empty), the totals saturate (verified with a `u64::MAX + u64::MAX` case), and `global_loss_cap_bytes` divides before multiplying so it cannot overflow. **SOLID**
3. **Frozen reason codes** (1..5) pinned and asserted distinct; serde serializes the variant name (human-readable) while `code()` gives the numeric for metrics, no inconsistency. **SOLID**
4. **Dependency posture correct:** `serde` is a normal derive-only dep; `serde_json` is dev-only. The reviewer confirmed via `cargo tree -e normal` that serde_json is NOT in the production tree, and the derives are exercised by the round-trip test (not dead code). **SOLID**
5. **Licenses pass:** ran `cargo deny check licenses` -> "licenses ok"; every new crate (serde/serde_json/serde_derive/itoa/ryu/memchr) has an allowed option in its OR expression (e.g. ryu via Apache-2.0). Core untouched, so the IO-free check is unaffected. **SOLID**
6. **MSRV 1.78 safe:** serde MSRV 1.56 (the binding normal dep); serde_json is dev-only so the `cargo build` msrv job never compiles it. **SOLID**
7. **API quality:** `Default == new` with the version stamped, `#[must_use]` on constructors/queries, clippy-pedantic clean, module doc honest about deferred work. **SOLID**
9. **Tests non-vacuous** (each would break under the corresponding regression). **SOLID**

CI confirms all of the above: 13/13 green on `7ee9528`, including cargo-deny, msrv (1.78), ironbus-core is IO-free, and the 32-bit parser job. This lands the schema (single source of truth) per #120; the I1 to I4 checkers, cap enforcement, and recovery emission are linked follow-ups, so #120 stays open. Merging.

PR #224, opened 2026-06-06
feat(storage): recovery emits the structured loss report (#120)

ELares:
## What

Makes the #120 loss-report schema **live**: recovery now builds a `LossReport` while it truncates a torn or corrupt active-segment tail, and `Log::loss_report()` exposes it for the metrics endpoint (#16) and the offline inspector (#15). This turns the raw `recovered_truncated_bytes` counter into a structured, per-segment, reason-carrying artifact, advancing #133 step 7 ("recover with bounded reported loss").

## How

The streaming recovery scan (`scan_recovery` / `scan_body_streaming`, from #156) now records **why** the valid prefix ended, mapped to a `ReasonCode`:
- `TornTail`: fewer bytes than a record header remain, OR an intact header names a frame that runs past the region (the body was never fully written).
- `CorruptRecordHeader`: a bad magic, version, or header CRC.
- `CorruptRecordBody`: a header-valid frame whose body or trailer fails its CRC.

`recover()` turns the truncation into one `LossEvent::span(segment, valid_end, len, reason)` with a `records_lost_estimate` that is a documented **lower bound** (a torn or corrupt span is, by definition, not fully parseable, but at least the frame at `valid_end` is gone).

## Faithful, non-breaking

- The report **agrees** with the existing `recovered_truncated_bytes` counter: `loss_report().total_bytes_skipped() == recovered_truncated_bytes()`, and the event's `[byte_offset_start, byte_offset_end)` is exactly the dropped region.
- It is **empty** for a fresh log or a clean recovery, so nothing else changes; `recovered_truncated_bytes` is retained (the metrics gauge and existing tests still use it).
- The `tail_reason` only changes the recovery scan's *reporting*, not which prefix it accepts (a sequence mismatch remains a hard error, unchanged).

## Tests

- `recovery_reports_the_truncated_tail_bytes` extended: a torn tail yields exactly one `TornTail` event in segment 0 whose span is the dropped region and whose bytes equal the counter; a clean reopen reports an empty report.
- `recovery_reports_a_corrupt_body_in_the_loss_report` (new): a flipped body byte (length unchanged) yields a `CorruptRecordBody` event, not a torn tail.

## Verification

- Full `cargo test --workspace --all-features --locked` green; `crash_recovery` + `determinism` sweeps pass at `PROPTEST_CASES=20000`.

## Follow-ups (still #120)

The I1 to I4 invariant checkers and the bounded-loss cap enforcement (freeze read-only and exit non-zero on exceed) remain linked follow-ups.

refs #120, #16, #15, #21


ELares:
## Independent adversarial review: SOLID (7/7)

A reviewer probed for a `ReasonCode` misclassification, a span/counter disagreement, and any change to which prefix recovery accepts, and returned SOLID on every point:

1. **`clean` is computed identically:** all four old `clean = false; break;` are now `tail_reason = Some(...); break;`, the normal loop exit leaves `tail_reason = None`, and `clean: tail_reason.is_none()` is returned, so the footer-trust logic is unchanged. **SOLID**
2. **Classification is faithful:** `remaining < header` and `total > remaining` -> `TornTail`; `decoded_len` error -> `CorruptRecordHeader` (verified `decoded_len` cannot return `Truncated` on a full RECORD_HEADER_LEN buffer, so this is never a misclassified torn tail); `decode` error after a valid header -> `CorruptRecordBody` (only body/trailer failures reach here). **SOLID**
3. **Span agrees with the counter:** `byte_offset_start = valid_end`, `byte_offset_end = len`, `bytes_skipped = len - valid_end` via saturating sub, all equal `recovered_truncated_bytes`; the `valid_end < len` guard keeps it > 0; `last_id` is the truncated active segment; exactly one event per truncating recovery. **SOLID**
4. **Prefix-acceptance provably unchanged:** seq-mismatch stays a hard error, `valid_end`/`record_count`/`last_seq`/footer trust and cursor advancement are untouched, and the `scan_recovery_agrees_with_scan_across_shapes` equivalence test (clean/sealed/empty/torn/corrupt) would fail if the accepted set changed. **SOLID**
5. **`records_lost_estimate = 1` is an honest lower bound:** a loss event is only created when the scan broke on an unparseable frame at `valid_end`, so at least one record-worth is gone; the schema documents it as a best-effort lower bound. **SOLID**
6. **Tests pin the behavior** (torn tail -> one `TornTail` event with the exact span and counter-equal bytes; flipped body byte -> `CorruptRecordBody`, not torn; clean reopen -> empty report); they would fail under a swapped classification or off-by-one span. **SOLID**
7. **API safe:** `loss_report` initialized empty in both constructors, `loss_report()` is `#[must_use] -> &LossReport`; no dashes; clippy-clean. **SOLID**

CI is 13/13 green on `0078f63` (incl cargo-deny, msrv 1.78, IO-free, 32-bit). This makes the #120 schema live (recovery emits it); the I1 to I4 checkers and cap enforcement remain follow-ups, so #120 stays open. Merging.

PR #225, opened 2026-06-06
feat(storage): enforce the bounded-loss caps at recovery (#120, I3)

ELares:
## What

Implements the **I3 invariant** (skip loss is bounded and reported) from #120: recovery previously accepted any amount of torn or corrupt tail loss (truncate, report, never refuse). I3 requires that a cascade of skipped spans must not silently become unbounded loss. Recovery now **fails closed** with `StorageError::ExcessiveRecoveryLoss(CapViolation)` when the loss report exceeds the caps.

## How

- `LossReport::check_caps(per_event_cap, global_cap)` (pure, reusable): no single event may exceed the **per-event cap** (one segment or 64 MiB, whichever is smaller), and the total may not exceed the **global cap**. Returns the first `CapViolation` (PerEvent / Global). Being pure, the recovery path and future sim/corpus/property fixtures share the exact same check.
- `recover()` computes the caps from the config and the durable byte count (accumulated across the recovered segments) and calls `check_caps` before returning the `Log`.

## Small-log-safe rule (the approved deviation from a literal 1%)

The issue specifies a global cap of "1% of durable bytes." Taken literally, that **freezes normal recovery of a small log**: a single torn record can exceed 1% of a few dozen durable bytes. So the global cap here is `max(1% of durable bytes, per_event_cap)`: a single in-cap event (the normal torn tail) is always within bounds, while a genuine multi-segment cascade past both the floor and 1% fails closed. This is the small-log-safe behavior I flagged and you approved.

## Fail-closed, not freeze-read-only

`recover()` returns an `Err`, so the broker refuses to start after excessive recovery loss (the CLI exits non-zero). The "open read-only for inspection" refinement (so #15's offline inspector can still read the durable prefix via a force/inspect path) is a follow-up; failing to open is the conservative choice that guarantees the broker never acks writes after excessive loss.

## Tests

- `check_caps`: accepts in-bounds loss, rejects a single oversized event (PerEvent), rejects a cascade over the global cap (Global) using a crafted multi-event report.
- `recovery_fails_closed_when_loss_exceeds_the_per_event_cap`: a 5000-byte corrupt tail with a 4096 segment cap -> `Err(ExcessiveRecoveryLoss(PerEvent { cap: 4096, .. }))`.
- `a_normal_small_log_torn_tail_recovers_despite_exceeding_one_percent`: proves the floor (a tiny log's torn tail still recovers).

## Verification

- Full `cargo test --workspace --all-features --locked` green; `crash_recovery` + `determinism` sweeps pass at `PROPTEST_CASES=20000` (legitimate torn tails stay within the per-event cap, so the enforcement is non-breaking).

## Follow-ups (still #120)

The I1 / I2 / I4 invariant checkers as reusable functions, their negative fixtures, and the offline-inspector force-open path remain linked follow-ups.

refs #120, #21


ELares:
## Independent adversarial review: SOLID (8/8); cannot fail a normal recovery

A reviewer was tasked above all to determine whether this could make a NORMAL crash recovery fail (a denial-of-recovery regression). Conclusion: **NO, with HIGH confidence.**

1. **`check_caps` correct:** per-event checked before global, strict `>` so a loss exactly at the cap is allowed, empty report -> Ok; unit tests (in-bounds / single oversized -> PerEvent / cascade -> Global) are non-vacuous. **SOLID**
2. **Small-log-safe floor prevents a denial-of-recovery regression:** `global_cap = max(1% durable, per_event_cap) >= per_event_cap`, so a single event within the per-event cap always passes BOTH caps; a normal torn tail is at most one active segment, bounded by `per_event_cap = min(max_segment_bytes, 64 MiB)`, so it can never trip the caps. Only genuinely excess loss (e.g. 10 GiB unsynced) fails closed. **SOLID**
3. **`durable_bytes` sound:** accumulates `valid_end - header` per segment (the kept record-region bytes) before `scan` is moved, saturating, no double-count, no underflow. **SOLID**
4. **Prefix-acceptance unchanged:** the cap check is purely additive at the end; truncation, valid_end, emission, fresh-log, clean-recovery, and sealed roll-forward paths are untouched and pass trivially (empty report). **SOLID**
5. **Failure test strong:** 5000-byte 0xff tail vs a 4096 segment cap -> `CapViolation::PerEvent { cap: 4096 }` regardless of whether the bytes read as a corrupt header or TooLarge. **SOLID**
6. **Error plumbing complete:** the variant, Display, and `source()` (CapViolation impls Error), `.map_err(ExcessiveRecoveryLoss)` correct, clippy-clean. **SOLID**
7. **Fail-closed interpretation sound:** Err from `Log::open` means the broker never acks writes after excessive loss and the CLI exits non-zero; the read-only-inspection path is a reasonable follow-up. **SOLID**

CI is 13/13 green on `0290fcc`; crash-recovery + determinism sweeps pass at PROPTEST_CASES=20000 (legitimate torn tails stay within the cap, confirming non-breaking). This lands I3 enforcement; the I1/I2/I4 checkers remain follow-ups, so #120 stays open. Merging.

Issue #14, opened 2026-06-05
[DESIGN] Hyper-configurable configuration system with safe defaults


ELares:
IronBus must be trivial to start with zero configuration and yet expose every durability, retention, compression, and backpressure knob an operator could reasonably want to tune for a specific edge box. This issue specifies the configuration system: a layered precedence model (built-in defaults < config file < environment variables < command-line flags < runtime admin API), a TOML config format, strict validation with actionable errors, hot reload for the subset of keys that are safe to change while running, and named profiles (`edge-tiny`, `balanced`, `throughput`) that set coherent groups of knobs in one move. The single static binary embeds sane defaults so `ironbus` with no arguments starts a durable, bounded, self-recovering queue. This realizes two tenets from #1 directly: Simple (zero-config startup, one obvious way to configure) and the implicit "hyper-configurable with safe defaults" promise that lets the same binary run on a microcontroller-class gateway and a multi-core hub.

## Why this matters for IronBus

Configuration is where the five tenets meet a real operator. Simple means a new user runs one command and gets a working, durable queue without reading a manual; the defaults must be the safe ones, not the fast-and-lossy ones. Edge First means a fleet operator can stamp `profile = "edge-tiny"` on a constrained device and trust that RAM ceilings, fsync cadence, and segment sizes move together coherently rather than having to hand-tune a dozen interacting knobs. Resilient means the config system itself never wedges the broker: a malformed file is rejected at load with a precise error, and a bad hot-reload is rejected without tearing down the running queue. HyperScale and Cross Platform mean the same key surface and the same precedence rules behave identically whether IronBus runs as a foreground binary, a systemd unit, or inside a container where environment variables are the natural injection point. Getting precedence, validation, and hot-reload boundaries right is what lets us ship one binary with one mental model instead of a per-platform configuration zoo.

## Prior art

The systems in our brief expose options in three recognizable styles, and each teaches a specific lesson.

[...]

PR #226, opened 2026-06-06
feat(storage): I1 to I4 resilience invariant checkers with negative fixtures (#120)

ELares:
## What

Adds `ironbus_storage::invariants`, the reusable I1 to I4 checker functions that #120 makes "the contract the simulation, corpus, fuzz, and property layers all check against, so it must be the single source of truth for what 'recovered correctly' means." This completes the checker half of #120 (the schema, emission, and I3 enforcement landed in #223 / #224 / #225).

## The checkers (pure functions over observable recovery output)

- `check_no_acked_loss` (**I1**): every acknowledged-durable offset is present in the recovered log (membership by offset, so it does not assume I2).
- `check_longest_valid_prefix` (**I2**): the recovered records are a contiguous run of offsets `0..n` with matching sequences, so recovery never reordered, left a gap, or read past a torn tail.
- `check_bounded_loss` (**I3**): the structured loss report is within the per-event and global caps (delegates to `LossReport::check_caps`).
- `check_pure_recovery` (**I4**): two recoveries from identical durable bytes produce identical records.

Each returns a typed `InvariantViolation`.

## Negative fixtures (the buggy-checker guard)

Per #120's failure consideration ("A buggy checker that always passes is guarded by negative fixtures"), **each checker has a known-bad input it must reject**: a lost acked offset (I1), an offset gap and a broken sequence (I2), an over-cap loss (I3), and diverging payloads / differing lengths (I4). Plus positive fixtures.

## Invoked by the layers

- The crash-recovery **property sweep** (`power_loss_recovers_the_durable_prefix`) now calls `check_longest_valid_prefix` (I2) and `check_no_acked_loss` (I1) on every recovered case, so the checkers run against thousands of real recoveries.
- A new test (`recovery_is_a_pure_function_of_the_durable_bytes`) verifies **I4** exactly as the issue specifies: recover twice from two **independent** disks loaded with identical durable bytes (captured from a torn-tail image) and assert the records match via `check_pure_recovery`.

## Verification

- Full `cargo test --workspace --all-features --locked` green; the crash-recovery sweep (now invoking the checkers per case) passes at `PROPTEST_CASES=20000`.

## Remaining under #120

The loss report asserted by corpus fixtures (ties to the corpus issue #123) and making the caps configurable (ties to the config issue #14) remain, so #120 stays open and refs.

refs #120, #21


ELares:
## Independent adversarial review: SOLID (8/8); no false-pass, no clippy cast risk

A reviewer audited the checkers for the two failure modes that matter most for a verification utility, a checker that FALSE-PASSES on a violating state, and a clippy cast-lint CI failure, and returned SOLID on every point:

1. **I1 (`check_no_acked_loss`):** membership-by-offset catches a lost acked offset and cannot false-pass; it deliberately checks presence not content (documented; content is covered by I2/I4), and cannot false-fail on a valid recovery. **SOLID**
2. **I2 (`check_longest_valid_prefix`):** `offset == seq == index` catches gaps, reorders, duplicates, and broken sequences; cannot false-pass. The only false-fail risk (offsets starting > 0 after retention) does not exist today and is a noted future concern, not a current flaw. **SOLID**
3. **I4 (`check_pure_recovery`):** zip-then-length-check catches diverging fields AND a strict-prefix length mismatch; `OwnedRecord` derives `PartialEq` so all fields compare; cannot false-pass or false-fail on identical input. **SOLID**
4. **I3 (`check_bounded_loss`):** faithful pass-through of `check_caps`, no logic change. **SOLID**
5. **I4 integration test:** genuinely recovers from two INDEPENDENT `InMemoryFs` disks built from the same captured torn-tail bytes (snapshot is the live image), non-vacuous (`!first.is_empty()`, 7 appends minus a 3-byte tear keeps 6 records). **SOLID**
6. **Property-test wiring:** `(0..durable)` is the correct ack history, the inline per-record asserts are preserved (not weakened), and the checkers run on every swept case. **SOLID**
7. **API/clippy:** `pub mod invariants`, rustdoc with `# Errors`, `InvariantViolation` Display+Error+PartialEq; the `as u64` casts are `usize -> u64` (non-truncating), and a clippy run confirmed no pedantic warnings. **SOLID**

Both critical questions answered NEGATIVE (no false-pass; no clippy CI failure). CI is 13/13 green on `1114006`; the crash-recovery sweep (now invoking the checkers per case) passes at PROPTEST_CASES=20000. This lands the I1 to I4 checkers; the corpus-fixture assertion (#123) and configurable caps (#14) remain, so #120 stays open. Merging.

PR #227, opened 2026-06-06
feat(server): expose per-reason recovery-loss on /metrics (#16, refs #120)

ELares:
## What

`/metrics` exposed total recovery loss as `ironbus_recovery_truncated_bytes` (a gauge) but not **why** recovery dropped bytes. Now that recovery emits a structured loss report (#224), surface it as a labelled gauge series so an operator sees both how much and why. Advances #16 (loss counters) and the metrics side of #133 step 7 ("the Prometheus loss counter ... agree").

```
# TYPE ironbus_recovery_loss_bytes gauge
ironbus_recovery_loss_bytes{reason="torn_tail"} 0
ironbus_recovery_loss_bytes{reason="corrupt_record_header"} 0
ironbus_recovery_loss_bytes{reason="corrupt_record_body"} 0
ironbus_recovery_loss_bytes{reason="corrupt_segment_header"} 0
ironbus_recovery_loss_bytes{reason="sequence_gap"} 0
```

The per-reason totals sum to the existing `ironbus_recovery_truncated_bytes`.

## How

- `loss`: `ReasonCode::ALL` (enumerate in code order), `ReasonCode::metric_label` (frozen lower-snake-case labels), `LossReport::bytes_skipped_for(reason)`.
- `engine`: `Engine::loss_report` delegate.
- `health`: a `recovery_loss: [u64; 5]` field in the `Copy` `MetricsSnapshot`, computed under the same engine lock as the other gauges, rendered by `recovery_loss_lines`.

**No new dependency**: the metric is rendered directly as Prometheus text (no serialization), consistent with the rest of the endpoint and the dev-only `serde_json` decision.

## Tests

- The `metrics_exposes_engine_gauges` test now asserts the per-reason series is present and zero on a clean start.
- A loss unit test pins that the per-reason totals sum to the grand total and that `ReasonCode::ALL`'s labels are frozen and distinct.

## Verification

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features` clean; `cargo build -p ironbus-server` clean under `-D warnings`.

refs #16, #120


ELares:
## Independent adversarial review: SOLID (7/7); the metric does not lie

A reviewer's central task was to confirm the per-reason series provably sums to the existing total gauge (else the metric lies). Verified SOLID on every point:

1. **Per-reason totals sum to `ironbus_recovery_truncated_bytes`:** recovery pushes exactly ONE `LossEvent` per truncation with `bytes_skipped = len - valid_end = recovered_truncated_bytes`; `ReasonCode::ALL` covers every possible reason; `bytes_skipped_for` filters + saturating-sums. So the per-reason sum always equals the total gauge (and both are 0 on a clean recovery). A unit test pins the sum invariant. **SOLID**
2. **`ALL`/`metric_label`/`code` mutually consistent and compiler-enforced:** `ALL` is `[ReasonCode; 5]` and `metric_label` is a non-wildcard match, so a 6th variant breaks the compile in both places; labels are distinct lower_snake_case. **SOLID**
3. **Prometheus format well-formed:** single HELP/TYPE `gauge`, `name{reason="label"} value` per line via `writeln!` (newline-terminated), the prior block ends with `\n`, `gauge` is correct for a last-recovery snapshot; the test anchors full lines. **SOLID**
4. **Lock discipline:** the per-reason array is computed under the same engine lock as the other gauges; `MetricsSnapshot` stays `Copy` (`[u64; 5]` + all-Copy fields); the 5x `bytes_skipped_for` over 0-1 events is trivial. **SOLID**
5. **`Engine::loss_report` delegate** correct, `#[must_use]`, documented, `LossReport` imported. **SOLID**
6. **clippy clean** (ran it); `use std::fmt::Write` already in scope. **SOLID**

CI is 13/13 green on `ef8105d`. The metric does not lie. Merging.

PR #228, opened 2026-06-06
ci: pin GitHub Actions to commit SHAs (#142)

### LINKED ISSUE #142: [TASK] Pin GitHub Actions to commit SHAs (supply-chain)
Third-party GitHub Actions in `.github/workflows/` are pinned to mutable tags (`actions/checkout@v4`, `dtolnay/rust-toolchain@stable` / `@1.78.0`, `Swatinem/rust-cache@v2`), so a compromised or force-pushed tag could run arbitrary code in CI. For a project that intends to ship reproducible, signed releases with an SBOM, CI actions should be pinned to full commit SHAs.

## Scope
- Pin every third-party action to a full commit SHA (with a trailing comment noting the human-readable version).
- Note: `dtolnay/rust-toolchain` has no major-version tag, so SHA pinning is the only option there.
- Add Dependabot (or a scheduled job) to bump the pinned SHAs so they do not rot.

## Acceptance criteria
- [ ] All third-party actions across all workflows are SHA-pinned.
- [ ] A bump mechanism (Dependabot `github-actions` ecosystem) is configured.
- [ ] CI stays green.

Found during the review of #141. Relates to #102, #128.

ELares:
## What

Resolves #142. Every GitHub Actions `uses:` in `ci.yml` and `nightly.yml` is now pinned to a full commit SHA, with the human-readable tag kept as a trailing comment. A mutable tag (`@v4`, `@stable`) lets a compromised or retagged action run with the workflow's permissions; a SHA is immutable.

| Action | Pinned to | was |
|--------|-----------|-----|
| `actions/checkout` | `34e1148...` | `@v4` |
| `dtolnay/rust-toolchain` | `29eef33...` | `@stable` / `@1.78.0` |
| `Swatinem/rust-cache` | `e18b497...` | `@v2` |
| `taiki-e/install-action` | `56545b3...` | `@v2` |
| `actions/upload-artifact` | `ea165f8...` | `@v4` |

## The `dtolnay/rust-toolchain` subtlety

That action selects the Rust toolchain from the git ref (`@stable`, `@1.78.0`), so a bare SHA pin would lose the toolchain selection. It is pinned to a single SHA (the `stable` branch, whose `action.yml` defaults `toolchain` to `stable`), and:
- the bare `@stable` uses keep that `stable` default (no input change needed);
- the **MSRV** job passes `toolchain: "1.78.0"` explicitly (verified against the action's `action.yml`), so the 1.78 selection no longer rides on a mutable ref;
- the existing `components: rustfmt` / `components: clippy` inputs are preserved (only the `uses:` line changed).

## Verification

- YAML validated; no unpinned `@vN` / `@stable` / `@1.78.0` remain.
- **The PR's own CI run is the real gate**: for a `pull_request` event GitHub runs the workflow from the PR head, so all 13 jobs here execute the pinned actions. If any SHA were wrong the action would fail to resolve and the job would go red.

Note: SHA-pinning trades automatic minor updates for supply-chain safety; a periodic bump (or Dependabot, a follow-up) keeps them current.

Closes #142


ELares:
## Independent adversarial review: SOLID (7/7); MSRV gate verified intact

A reviewer's central task was to confirm the MSRV gate still builds with 1.78.0 (not silently stable) and that every SHA is a real commit of its action. SOLID on all points:

1. **Every pinned SHA matches its tag** (verified via `gh api repos/<a>/commits/<tag>`): checkout v4, rust-cache v2, install-action v2, upload-artifact v4, and dtolnay stable branch all match. **SOLID**
2. **MSRV is correct, the gate is NOT broken:** the msrv job's `with: toolchain: "1.78.0"` is present and correctly indented under that step; the pinned `action.yml` has `inputs.toolchain` (`default: stable`), so the explicit input overrides the default and installs 1.78.0, not stable. **SOLID**
3. **Bare `@stable` uses** keep the action's `stable` default; the `components: rustfmt`/`clippy` inputs are preserved. **SOLID**
4. **YAML valid**, the MSRV `with:` block associates with the dtolnay step (not the next), no duplicate `with:`. **SOLID**
5. **Completeness:** all 25 `uses:` across both files are 40-hex SHAs with tag comments; none missed. **SOLID**
6. **No functional drift:** only the SHA pins, the MSRV toolchain input, and the CHANGELOG. **SOLID**

The end-to-end proof is in CI itself: 13/13 green on `f5a2610`, and the `msrv (1.78)` job ran the SHA-pinned `dtolnay/rust-toolchain` with `toolchain: "1.78.0"` and built successfully. Merging.

PR #229, opened 2026-06-06
feat(storage): short-read and read-error fault injection in the sim fs (#151)

ELares:
## What

Adds the **read-side** fault modes #151 calls for. The `FaultFs` simulation seam already injected fsync failures, clean write failures, and one-shot torn writes; this adds two more deterministic faults on the same shared `FaultControl`:

- `set_fail_read(bool)`: every `read_at` returns an injected IO error (a bad block).
- `set_short_read(limit)`: every `read_at` returns at most `limit` bytes even when more are available (a partial read), while leaving real EOF intact (a short read never invents a spurious zero-byte read).

No ambient randomness, armed explicitly, so runs stay replayable.

## Why it matters (invariants under read faults)

Two recovery tests assert the storage invariants hold under the new faults:
- `recovery_recovers_the_full_prefix_under_persistent_short_reads`: with every read capped to **one byte**, recovery's `read_exact_at` loops over the partial reads and still recovers the whole durable prefix, so no acked record is lost to a partial read.
- `recovery_fails_cleanly_under_an_injected_read_error`: an injected read error during recovery surfaces as a clean `StorageError::Io`, never a panic or a silent partial recovery.

Plus three `fault.rs` unit tests (read error fails reads only; a short read returns fewer bytes yet `read_exact_at` still fills; a short read at the end still signals real EOF).

## Scope

This lands the read-fault **primitives** + invariant tests. The remaining half of #151, a **seeded fault scheduler** driving all the primitives (torn writes, short reads, errors) from one PRNG, stays a linked follow-up, so this `refs #151` rather than closing it.

## Verification

refs #151, #21, #7, #8


ELares:
## Independent adversarial review: SOLID (5/5); no spurious-EOF, tests non-vacuous

The reviewer's two central questions were answered:

(a) **Can a short read ever return a spurious 0 (EOF) when data remains?** NO: `cap = min(limit, buf.len())` is >= 1 whenever `buf` is non-empty and `limit > 0`, so the inner `read_at` on a non-empty sub-slice returns >= 1 with data available, never a spurious 0; at true EOF it returns 0 legitimately. So no recovery corruption and no `read_exact_at` infinite loop. **SOLID**
(b) **Are the recovery tests non-vacuous?** YES: the 1-byte cap forces `read_exact_at` to loop ~64x just on the segment header, and `assert_prefix` checks payloads, so a corrupting loop would fail; the injected-read-error test fires on the first header read and asserts a clean `StorageError::Io`. **SOLID**

Also verified: the new faults touch only `read_at` (read error checked before short read; writes/syncs/dir ops unaffected, proven by `arming_read_failure_fails_reads_only`); atomics use SeqCst consistent with the existing fields and `0` is a sound "off" sentinel; rustdoc + module/struct docs updated; clippy clean; no em/en dashes.

CI is 13/13 green on `662313b`. This lands the read-fault primitives + invariant tests; the seed-driven fault scheduler stays a #151 follow-up. Merging.

PR #230, opened 2026-06-06
test(storage): seed-driven fault-recovery invariant property test (#151)

### LINKED ISSUE #151: [TASK] Fault injection in the simulation file (torn writes, short reads, injected IO errors)

ELares:
## What

Completes the #151 fault-injection contract. #229 added the read-side fault **primitives** (short reads, read errors) alongside the existing fsync/write/torn faults; this adds the **seeded driver**: a proptest that, from its seed, picks an arbitrary workload and one of the five injected faults and applies it during recovery, asserting the resilience invariants hold.

## The invariant under any fault

`recovery_under_an_arbitrary_seeded_fault_holds_the_invariants` asserts recovery under any injected fault (fsync EIO / clean write failure / read error / short read / torn write):
- **never panics** (reaching past `Log::open` at all proves it), and
- on success, is a contiguous valid prefix from offset 0 (**I2**, `check_longest_valid_prefix`) that preserves every durably acked record (**I1**, `check_no_acked_loss`), and
- otherwise fails **closed** with a clean typed `StorageError` (the accepted outcome).

A buggy recovery that lost an acked record below its durability level, read past a torn tail, or panicked on an IO fault fails here. The fault and the workload come from the proptest seed, so any counterexample replays exactly.

## Why this closes #151

#151's two acceptance criteria are now both met: (1) a fault-injecting wrapper produces torn writes, short reads, and injected errors deterministically from a seed (the primitives in #229, driven here from the proptest seed); (2) recovery/durability tests assert the invariants hold under these faults (this property test plus the targeted tests in #229).

## Verification

- Full `cargo test --workspace --all-features --locked` green; the new property test passes at **`PROPTEST_CASES=20000`** across all five fault modes and both segment configs.

Closes #151


ELares:
## Independent adversarial review: SOLID (meaningful, not vacuous); coverage finding addressed

A reviewer probed the central risk for a property test, vacuity, and returned SOLID on all 7 points:

1. **Not vacuous:** the read faults (`FailRead`, `ShortRead`) fire on EVERY recovery (recovery always reads via `read_exact_at`). The sync/write/torn faults fire only when recovery itself writes/syncs. **SOLID (with the coverage nuance below)**
2. **I1 sound and not too strong:** `durable = flushed_offset`; recovery reads the live disk and recovers `>= durable`, so `check_no_acked_loss(0..durable)` holds and never wrongly fails; the post-recovery read happens after `disarm_all`. **SOLID**
3. **I2 correct:** recovery always yields a contiguous prefix from offset 0 (no retention yet). **SOLID**
4. **No-panic genuinely asserted:** a panic in `Log::open` unwinds and fails the proptest case; the recovery lib path has no unwrap/expect on the IO-fault path. **SOLID**
5. **Deterministic/replayable:** fault and workload from the proptest seed, `ManualClock`, no ambient randomness. **SOLID**
6. **Complementary, not redundant** with `a_sync_fault_at_any_point_loses_no_acked_record` (that injects during the WRITE phase + power loss; this injects during RECOVERY). **SOLID**
7. **clippy clean, no dashes.** **SOLID**

**Comment addressed:** the reviewer noted the write/sync/torn arms are largely no-ops on a clean-image recovery (those paths only run on a torn-tail truncation `sync_all` or a sealed-highest roll-forward header write, which a clean workload seldom produces). Rather than overclaim, I made the test's doc honest about exactly which faults reliably fire, and filed **#231** to add targeted tests that force recovery's write paths under the write/sync/torn faults (the reviewer's suggested follow-up). This PR's value, robust read-fault coverage plus the I1/I2/no-panic invariants under any fault, stands on its own.

Re-running CI on the doc commit, then merging (closes #151).

PR #232, opened 2026-06-06
test(storage): recovery fails closed when a fault strikes the roll-forward (#231)

### LINKED ISSUE #231: [TASK] Exercise write/sync/torn faults during recovery's write paths (truncation sync, roll-forward header)
Follow-up from the #151 seed-driven fault property test (`recovery_under_an_arbitrary_seeded_fault_holds_the_invariants` in #230).

## Gap

That property test recovers a CLEAN synced image under a seeded fault. The independent review found that the **read** faults (`FailRead`, `ShortRead`) fire on every recovery (recovery always reads), but the **write/sync/torn** faults rarely fire, because recovering a clean image does no `write_all_at` and no `sync_all`:
- recovery only `sync_all`s when it truncates a torn tail (`Log::recover`), which a clean workload does not produce;
- recovery only `write_all_at`s a header when it rolls a SEALED highest segment forward (the crash-between-seal-and-create state), which a clean workload does not produce.

So in the property test, `FailWrite`/`FailSync`/`TornWrite` are mostly no-ops (recovery succeeds because the fault never triggers). The invariants asserted are still correct; the coverage of fault-DURING-recovery's write paths is the gap.

## Scope

Add targeted (or property) tests that force recovery's write paths and inject the write/sync/torn faults there:
- **Truncation `sync_all` under `FailSync`:** construct a torn tail (an unsynced record above the flushed offset, chopped) so recovery must `set_len` + `sync_all`, then arm `FailSync`; recovery must fail closed (no acked record below durability lost). Partly covered by `recovery_is_idempotent_after_a_fault_during_recovery`; generalize.
- **Roll-forward `start_segment` under `FailWrite` / `TornWrite` / `FailSync`:** construct a sealed highest segment with no successor (a crash between seal and create), so recovery rolls forward and writes + syncs a new segment header, then arm the fault; recovery must fail cleanly.

## Acceptance criteria
- [ ] A test forces recovery's truncation `sync_all` and asserts a clean fail-closed under `FailSync`.
- [ ] A test forces recovery's roll-forward header write and asserts a clean fail-closed under `FailWrite` / `TornWrite`.
- [ ] No acked record below its durability level is lost in any case; no panic.

Relates to #151, #21, #7.

ELares:
## What

Closes #231, the coverage gap filed from the #151 review. The seed-driven fault property test (#230) recovers a CLEAN image, which does no `write_all_at` and no `sync_all`, so its write/sync/torn fault arms rarely fired. This adds the missing case: a fault DURING recovery's write path.

## How

`recovery_fails_closed_when_a_fault_strikes_the_roll_forward` seals the highest segment and leaves **no successor** (a crash between a roll's seal and the next create), so recovery must roll forward, creating, writing, and syncing a fresh segment header. For each write-side fault, recovery must fail **closed** with a clean typed error:
- `FailWrite` (the header `write_all_at` fails cleanly),
- `TornWrite` (a torn header write),
- `FailSync` (the header `fsync` fails),

each must yield `StorageError::Io`, never a panic or a silent partial recovery. The acked records in the sealed segment are untouched on disk, so the broker simply refuses to start (fail-closed, retriable).

The sealed-highest state is built with `SegmentWriter::resume(...).seal()` over a scanned segment, so the construction is faithful (a real footer at the record boundary that `scan_recovery` trusts as sealed).

## Why this closes #231

#231's two criteria: (1) truncation `sync_all` under `FailSync` is already covered by `recovery_is_idempotent_after_a_fault_during_recovery` (it tears a tail and faults the truncation sync); (2) the roll-forward header write under a write fault is this new test. Both fault-during-recovery write paths are now exercised; no acked record is lost below durability and recovery never panics.

## Verification

- Full `cargo test --workspace --all-features --locked` green; the new test asserts all three faults fail closed.

Closes #231


ELares:
## Independent adversarial review: SOLID after addressing the vacuity finding

The reviewer probed the central risk, that the constructed sealed-highest state might not genuinely trigger a roll-forward (making the test vacuous), and rated 5 of 6 points SOLID:

1. **Roll-forward genuinely triggered:** `SegmentWriter::resume(.., valid_end, record_count, last_seq).seal()` writes the footer at exactly the record boundary with matching `record_count`/`last_seq`, so `scan_recovery`'s `ends_at_footer && body_matches` trusts it as sealed and `recover` takes the `footer.is_some()` roll-forward branch. **SOLID**
2. **Fault fires on the roll-forward, not earlier:** the scan reads complete first (the three write/sync faults do not affect reads); then `start_segment`'s header `write_all_at` (FailWrite/TornWrite) or `writer.sync()` (FailSync) faults; `create_new` delegates without faulting. **SOLID**
3. **Could it pass for the wrong reason?** Flagged WEAK: the `StorageError::Io(_)` assertion alone did not prove the state genuinely rolls forward. **Addressed**: added an anti-vacuity sanity check that the same construction, with NO fault, rolls forward to segment 1 carrying both records (on its own fresh sealed disk, since rolling forward creates segment 1). The faulted cases now provably exercise the roll-forward write path.
4. **Sealing correctness:** `resume`/`seal` consistent, footer at the boundary, no spurious unwrap failure for a 2-record log. **SOLID**
5. **Fail-closed semantics:** recovery erroring is correct (the broker refuses to start); the sealed segment's acked records stay untouched on disk for a clean retry. **SOLID**
6. **clippy clean, no dashes.** **SOLID**

CI was 13/13 green on the original commit; re-running on the sanity-check commit, then merging (closes #231).

Issue #139, opened 2026-06-05
[META] Issue coherence and vision-fit audit


ELares:
A fresh-eyes coherence pass ran over all 22 parent issues (#1 vision, #2 prior-art, #3..#22 subsystems) against the committed README (the single source of vision truth) and the per-issue resolved decisions. Result: 12 of 22 issues are clear (no implementer-blocking ambiguity) and 17 of 22 are vision-fitting (no drift from the frozen tenets and the frozen model). The 5 vision-fit breaks are #1, #4, #6, #10, #13, and the 10 clarity gaps are #1, #6, #7, #8, #9, #11, #14, #15, #19, #21. The audit is not surfacing many independent bugs: it is surfacing a small number of root drifts that propagate. Three root drifts account for almost every contradiction, and resolving them at the owning issue clears the downstream noise. This issue records the canonical values, the resolution table, and the few items that still need a human decision before code starts.

## Root drift 1: the separate-WAL ghost (the most serious)

The frozen model is log-is-WAL: a publish is exactly one framed, CRC32C-checksummed, record-aligned append to the active segment, and that append IS the durable record. There is no separate WAL file, and the offset index is the only derived, rebuildable artifact. #4 froze this correctly and the README headlines it ("The active segment IS the write-ahead log: there is no separate WAL file to keep in sync"). But #6's resolved text still describes a two-tier pipeline: "the segmented main log is fed from the WAL asynchronously and is NOT on the ack-critical path; only the WAL fdatasync gates the ack" with "WAL truncation/checkpoint" after main-log durability. That phantom architecture then leaks by reference into #1, #2, #3, #7, and #19 (each of which quotes or inherits #6's durability sentence). Every one of these must be restated as: a record is durable when its bytes are in the active log segment and that segment has been fdatasync'd; no second file, no async hand-off, no WAL checkpoint. The fix belongs in #6; #1 must restate log-is-WAL; #2, #3, #7, #19 only need their one durability clause re-anchored to the active segment.

## Root drift 2: the resync primitive (record-format contract)

[...]

Issue #18, opened 2026-06-05
[DESIGN] Security model: authentication, authorization, TLS, and edge threat model


ELares:
IronBus runs where the threat model is hostile and the operator is absent: a single static binary on an unattended edge device, often on a shared or untrusted LAN, frequently with physical exposure, and with limited ability to push credential rotations. This issue specifies the security model: an explicit edge threat model, authentication options (bearer token, username/password, mTLS), authorization scopes (publish, subscribe, admin), transport TLS, optional encryption at rest, secret handling, and the secure defaults that make all of this safe out of the box. The guiding principle is a minimal attack surface with no unsafe network defaults, borrowing the auth ergonomics of NATS and MQTT and the transport hygiene of TLS best practice. Security here is a configuration concern as much as a protocol concern, so this issue is co-designed with #11 and #14.

## Why this matters for IronBus

Security ties directly to four of the five tenets. **Edge First** is the dominant one: unlike a datacenter broker behind a VPC and an IAM boundary, an IronBus node may sit on a factory VLAN, a retail store Wi-Fi, or a customer-premises network the operator does not control, and the device itself can be unplugged and carried away. **Resilient** extends past power loss to adversarial and accidental misuse: a misconfigured client, a stolen token, or a hostile peer on the LAN must not be able to read or corrupt the queue. **Simple** means security must be on by default and obvious to configure correctly; a broker that is only secure after a long hardening checklist is insecure in practice. **Cross Platform** constrains the implementation: auth and TLS must work identically on aarch64/armv7 Linux musl targets with a small, auditable dependency tree, not depend on a platform secret service that exists only on one OS. The single-binary, supply-chain-auditable posture (cargo-auditable SBOM, cargo-deny advisory gating) is itself a security property: a smaller dependency tree is a smaller attack surface.

## Prior art

**NATS** authenticates with several mechanisms layered on its line-oriented `CONNECT` handshake: token (a bearer string), user/password, TLS client certificates, and its nkey/JWT system where an account-signed JWT plus an ed25519 nkey challenge proves identity without sending a shared secret over the wire. Authorization in NATS is subject-level allow/deny lists for publish and subscribe per user, which is exactly the scope granularity IronBus needs but specialized to a subject tree. NATS reserves a `$` prefix (`$SYS`, `$JS`) for privileged system subjects, a useful pattern for separating admin operations from data operations.

[...]

Issue #12, opened 2026-06-05
[DESIGN] Compression and encoding


ELares:
IronBus ships built-in compression because edge uplinks are metered, edge flash is finite, and the payloads are overwhelmingly tiny, similar telemetry records that compress badly in isolation. This issue specifies the codec, the default level, the unit of compression (per-message vs per-batch vs per-segment), zstd dictionary training for small similar messages, where the compressed marker lives in #5, and how all of it stays configurable per #14 with a CPU budget for ARM cores. The leading proposal is zstd at a low default level with optional trained per-message-type dictionaries, compressed per-batch, with the codec/level/dictionary-id self-describing on disk so old data stays readable after defaults change.

## Why this matters for IronBus

Compression touches four of the five tenets. **Edge First**: a trained dictionary takes a few-hundred-byte JSON telemetry record from roughly 2.8x to nearly 7x (Meta's GitHub-user-record example went 846 KB to 122 KB), which directly cuts cellular uplink bytes and SD/eMMC writes, and fewer bytes written is less flash wear. **HyperScale**: storing and forwarding an already-compressed batch end-to-end (Kafka's model) means brokers never recompress, so CPU and bandwidth both drop as message volume grows. **Resilient**: the compression unit sets the corruption blast radius, so the choice has to isolate failure (a corrupt compressed batch loses the whole batch) and integrate with #8. **Simple**: one codec (zstd) spans the entire speed/ratio range through a single frame format, so changing levels never changes the wire format or the decoder. **Cross Platform** also benefits: the same self-describing frame decodes identically on an x86 dev box and an aarch64 device.

## Prior art

**Kafka** compresses whole record batches, not individual messages, precisely because cross-message redundancy (shared keys, headers, schema) is where the savings live. The compressed batch is stored and transferred as-is and only decompressed by the consumer; `compression.type` defaults to `none` with gzip/snappy/lz4/zstd available (zstd since 2.1), and per-codec level knobs exist (KIP-390). Confluent's own docs state "compression is of full batches of data... more batching means better compression." **Cloudflare** measured an 8x size drop simply from batching about one second of logs before compressing, and per-batch ratios of Snappy 2.35x, LZ4 1.81x, gzip 3.58x, zstd L6 about 4.5x, with gzip costing roughly 10x the user CPU of Snappy.

[...]

PR #233, opened 2026-06-06
chore(spec): record #139 decisions, never-recycle invariant + pure-Rust default (#139)

ELares:
## What

The maintainer resolved the four #139 open decisions. This PR lands the two with a concrete repo artifact; all four are recorded as clarifying comments on #139 and the owning issues (#4, #12, #14).

## The four decisions

1. **Segment recycling: NEVER recycle in v1** (defer to #5, the format owner). → this PR adds a regression test.
2. **Compression: lz4_flex pure-Rust default, zstd opt-in only** (no vendored C on the default path). → this PR corrects the stale `deny.toml` comment.
3. **Config: amend #14 to add `[observability]` + `[auth]`** sections. → spec item, recorded on #14; no code yet (no TOML schema implemented).
4. **Protocol doc: README uses `ironbus tap`/`ironbus wire`, not netcat.** → already satisfied in the committed README; recorded for completeness.

## Changes

- **`segment_ids_increase_monotonically_and_are_never_recycled`** (storage test): asserts segment ids are a contiguous, monotonic run that is never reused, across rolls AND a restart (id 0 and every prior id retained). The storage layer already enforces this (recovery rolls forward to `last_id + 1`, `start_segment` only mints higher ids); the test pins the invariant so a future change reintroducing recycling fails. This keeps #18's at-rest GCM nonce safe (a `segment_id` is never reused under a fixed key).
- **`deny.toml`**: the comment claiming `zstd-sys` is the only vendored-C crate on the default path is corrected to record that the default path is pure Rust (lz4_flex) and zstd (vendored C) is opt-in only.

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-storage --lib` clean; `deny.toml` is valid TOML.
- Full `cargo test --workspace --all-features --locked` green; the new test passes.

refs #139, #4, #12, #18


ELares:
## Independent adversarial review: SOLID (6/6); the test genuinely catches recycling

The reviewer's central question, would the test FAIL if recycling were reintroduced, was answered definitively: YES, on four independent assertions:
1. `segment_ids == 0..=max_before` (fails on a duplicate or gap),
2. `active_segment_id() > max_before` after restart (fails if a recycled id <= max_before),
3. every old id in `0..=max_before` still present (fails if id 0 is overwritten),
4. `ids_after == 0..=active` (fails on a gap or duplicate).

A naive recycling implementation fails all four, so the test is a real guard, not vacuous. Other findings:
- **`small_config` (128-byte cap) reliably rolls** 16 ~64-byte records into 3-4 segments, so `max_before >= 2` is robust, not flaky. **SOLID**
- The invariant (contiguous, monotonic, never-reused) is the correct encoding for v1 (no retention yet); a future retention feature would need the test to evolve, a noted concern, not a current bug. **SOLID**
- Uses only public API (`segment_ids`, `active_segment_id`, `filesystem`), observing on-disk behavior, not internals. **SOLID**
- **deny.toml is a comment-only change** (no `[bans]`/`[licenses]` policy change), TOML valid, cargo-deny green, accurately reflects the lz4_flex-pure-Rust decision, no dashes. **SOLID**
- Scope correct (`refs #139`, decisions 3/4 recorded on the owning issues; README already uses `ironbus tap`/`wire`, no netcat). **SOLID**
- clippy clean, no dashes. **SOLID**

CI is 13/13 green on `25d3555`. Merging.

PR #234, opened 2026-06-06
feat(core): make AckCursor persistable for the durable consumer-state store (#60)

ELares:
## What

The foundation-first first slice toward the consumer-groups / fan-out keystone (#9, #133 step 4). The work-group commit cursor (`AckCursor`) already implements the in-memory sparse acked-ahead set fully (the hard mechanism of #61), with extensive proptests. But #61's restore criterion ("after restart, committed_offset AND the sparse set are restored exactly") and the parent #60 (durable consumer-state store) need the cursor to be **persistable**: today the engine drops the ahead set on restart and redelivers all in-flight work, including the acked-ahead offsets.

## How (IO-free)

- **`ahead_ranges() -> &[(u64, u64)]`**: snapshot the sparse set as run-length `[start, end)` ranges (sorted, disjoint, non-adjacent, above the watermark).
- **`resume_with_ahead(committed, ranges) -> Result<AckCursor, AckCursorError>`**: restore a cursor from such a snapshot, **validating the shape** (a single `start > prev_end` check enforces non-empty, sorted, disjoint, non-adjacent, and strictly-above-committed, where `prev_end` starts at `committed`), and rejecting a corrupt or torn snapshot with a typed `AckCursorError`, so the caller can fall back to `resume()` (drop the ahead set and redeliver) rather than trust bad state. A rejected snapshot never yields a half-built cursor.

So a restart can redeliver **only genuinely unacked offsets** instead of the acked-ahead ones too. The byte framing and CRC of the on-disk snapshot remain the storage layer's concern (#60); this is the pure data foundation it builds on. `resume()` (drop the ahead set) is retained, so the engine's current behavior is unchanged until #60 wires the store.

## Tests

- `resume_with_ahead_restores_the_sparse_set_exactly`: a gap snapshot round-trips to an identical cursor; the real gaps redeliver, the acked-ahead survives.
- `resume_with_ahead_rejects_a_malformed_snapshot`: empty/reversed, unsorted, overlapping, adjacent, and at-the-watermark ranges are all rejected; a well-formed one is accepted.
- `no_offset_below_committed_is_ever_unacked` (proptest): the #61 no-silent-skip invariant under any ack order.
- `snapshot_then_resume_with_ahead_round_trips` (proptest): a cursor's own snapshot always restores identically.

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-core --all-targets` clean; `cargo build -p ironbus-core` clean under `-D warnings`; ironbus-core stays IO-free.
- Full `cargo test --workspace --all-features --locked` green; the cursor proptests pass at `PROPTEST_CASES=20000`.

refs #60, #61, #9


ELares:
## Independent adversarial review: SOLID (7/7); the validator exactly enforces the invariant

The reviewer's central question, can `resume_with_ahead` accept an illegal state that would corrupt later `ack`/`advance`, was answered NO definitively:

1. **Validation exactly matches the cursor invariant.** The single `start > prev_end` check (with `prev_end` seeded to `committed`, plus the separate `start < end`) enforces non-empty, sorted, pairwise-disjoint, **non-adjacent** (strict `>`, so `[(5,7),(7,9)]` is rejected), and strictly-above-committed (the first range must clear the watermark). The crucial adjacent case is rejected, so `advance`'s "at most one range can be contiguous" assumption holds. **SOLID**
2. **Round-trip exact:** `ahead_ranges()` returns the Vec unchanged, restore sets exactly `{committed, ahead}`, `PartialEq` covers both; a live cursor's `insert`/`advance` always maintain precisely the invariant the validator checks, so no cursor can produce a snapshot it cannot restore (proptest at 20000 confirms). **SOLID**
3. **Redelivery semantics correct:** post-restore, acked-ahead offsets stay acked and only the real gaps redeliver. **SOLID**
4. **`AckCursorError`** is Copy/Debug/PartialEq + Display + Error; `std::error::Error` does not violate the IO-free rule (the lint forbids fs/net/process/tokio, and the IO-free CI job passed). **SOLID**
5. **No behavior change:** `resume()` retained, the engine is untouched, this is pure additive foundation (wiring deferred to #60). **SOLID**
6. **Tests non-vacuous,** covering empty/reversed, unsorted, overlapping, **adjacent**, and at-the-watermark rejections plus the no-silent-skip and round-trip properties. **SOLID**
7. **clippy clean, no dashes.** **SOLID**

CI is 13/13 green on `f73b491` (including ironbus-core is IO-free). This lands the persistable-cursor foundation for the consumer-state store; #60 wires the on-disk framing next. Merging.

Issue #127, opened 2026-06-05
[TASK] Community and security docs: CONTRIBUTING, CODE_OF_CONDUCT, SECURITY, GOVERNANCE, DCO


ELares:
Parent: #22

Author the baseline community and security documents so a contributor can clone, read, and participate in minutes, and so vulnerabilities are reported and embargoed responsibly. This fills two gaps the design survey left open: contributor sign-off provenance and maintainer succession/tie-break. Scope is the documents and the sign-off mechanic; CI wiring for the DCO check is shared with the CI sibling task.

## Approach
Add `CODE_OF_CONDUCT.md` as Contributor Covenant v2.1 verbatim with a real enforcement contact. Add `SECURITY.md` routing reports privately via GitHub Security Advisories plus a `security@` alias, with a 72-hour acknowledgement SLA, a 90-day coordinated-disclosure embargo from triage, a commitment that fixes ship as a patch release, and a RUSTSEC advisory filing; this feeds the #18 threat model. Add `CONTRIBUTING.md` covering the clone-build-test loop, the RFC link, and the merge rules. Use DCO with required `Signed-off-by` trailers (no CLA), enforced by a CI check; copyright held by "The IronBus Authors". Add `GOVERNANCE.md` naming initial maintainers, a lazy-consensus model with a maintainer-majority tie-break on deadlock, and a succession note.

## Acceptance criteria
- [ ] `CODE_OF_CONDUCT.md` is Contributor Covenant v2.1 with a working contact.
- [ ] `SECURITY.md` documents the private channel, 72-hour ack, 90-day embargo, and patch+RUSTSEC flow.
- [ ] `CONTRIBUTING.md` documents the build/test loop, DCO sign-off, and the RFC link.
- [ ] DCO `Signed-off-by` is required and enforced by a CI check.
- [ ] `GOVERNANCE.md` names maintainers, the tie-break rule, and succession.
- [ ] No CLA is required; copyright is "The IronBus Authors".

## Failure considerations
A security report posted as a public issue or left past embargo could expose edge fleets before a patch exists; private GHSA intake, a 72-hour ack, and a 90-day coordinated window get durability and corruption bugs (#6, #8) fixed before disclosure, and templates discourage public exploit detail. A missing sign-off makes the provenance of a durability-critical contribution unprovable; the enforced DCO check blocks unsigned commits. Single-maintainer deadlock risks stalling the project; the GOVERNANCE.md tie-break and succession note bound that bus-factor risk.

PR #250, opened 2026-06-07
Add a cargo-fuzz harness for the untrusted-byte parsers

ELares:
## What

A cargo-fuzz harness that continuously soaks the four parsers that read untrusted or possibly corrupt bytes, under libFuzzer and AddressSanitizer. This makes good the README's existing claim that "the record and segment parsers are continuously fuzzed" (refs #21, #123).

A detached `fuzz/` crate (its own empty `[workspace]` table, so the stable, merge-blocking workspace is never built with the nightly sanitizer toolchain) defines four targets:

| Target | Parser | Path it guards |
| --- | --- | --- |
| `record_codec` | `codec::decode` / `decoded_len` | recovery and delivery |
| `frame_decode` | `frame::decode_frame` | bytes off the client socket |
| `cursor_snapshot` | `AckCursor::decode_snapshot` | the durable checkpoint payload |
| `segment_scan` | `SegmentReader::open` / `scan` / `scan_recovery` | every startup, over on-disk bytes |

Each target asserts the contract the proptests assume on structured inputs: on **any** byte string the parser only returns a typed error or a valid view, it never panics and never reads out of bounds. `segment_scan` is the most security-critical, since recovery runs it on every startup over bytes a power cut may have corrupted.

## Nightly soak

A new `fuzz` job in `nightly.yml` installs a pinned `cargo-fuzz@0.13.1` and soaks every target for three minutes under ASan (about 12 minutes total). A crash exits cargo-fuzz non-zero, fails the run, and uploads the offending input as an artifact so it can be minimized into a permanent regression seed. The job is SHA-pinned and actionlint-clean; `RUSTFLAGS` is cleared at the job level so cargo-fuzz's sanitizer flags are not clobbered by the inherited `-D warnings`.

## Hygiene

- Build outputs (`target/`), discovered corpora (`corpus/`), and crash artifacts (`artifacts/`) are git-ignored; `fuzz/Cargo.lock` is committed for reproducible soaks.
- SPDX headers on all four targets; the SPDX CI check (Rust-files-only) covers them.
- `fuzz/README.md` documents each target and how to run a soak locally.

## Verification

Built and soaked locally on macOS aarch64 (nightly + cargo-fuzz 0.13.1):

- `cargo +nightly fuzz build` builds all four targets, confirming every API call typechecks and links.
- 15-second soak per target: 6.2M (`segment_scan`) to 21M (`frame_decode`) executions, **0 crashes**, exit 0.
- `actionlint` (with `SHELLCHECK_OPTS=--exclude=SC2015`, matching CI) passes on the workflow.
- `rustfmt --check` clean on the targets; no em or en dashes in the prose.

The detached crate does not change the main workspace: `cargo metadata` still lists exactly the six product crates, so the stable CI matrix is untouched.

refs #21, #123


ELares:

An independent review (not the author) verified every fuzz target against current public APIs and checked the workflow, hygiene, and prose. Confirmed sound: all four targets call real public APIs (`codec::decode`/`decoded_len`, `frame::decode_frame`, `AckCursor::decode_snapshot`, `SegmentReader::open`/`scan`/`scan_recovery`); `InMemoryFile::write_all_at` takes `&self` so the un-`mut` binding compiles; the empty `[workspace]` table correctly detaches the crate (parent `cargo metadata` still lists exactly the six product crates); SPDX headers present on all targets; the soak loop runs under `bash -e` so the first crash fails the job; `taiki-e/install-action` `cargo-fuzz@0.13.1` is a valid spec; no panics/unwraps in the targets; no em or en dashes in the prose.

One finding raised and addressed:

- **Toolchain selection in the soak loop.** The loop ran `cargo fuzz run`, relying on `nightly` being the default. `dtolnay/rust-toolchain` does run `rustup default nightly` (action.yml line 99), so the happy path worked, but that step is `continue-on-error: true` (their issue #127), leaving a fallback-to-stable failure window where the `-Zsanitizer=address` build would error. Fixed in 7b329e9 by selecting nightly explicitly via the rustup proxy (`cargo +nightly fuzz run`), which is unambiguous and matches the documented local command in `fuzz/README.md`. No-op on the happy path.

`segment_scan` necessarily fails fast on inputs shorter than the 64-byte segment header (a typed `SegmentError::Truncated`, never a panic), which is the intended contract; richer seeding to drive deeper into the scanner can follow.

PR #251, opened 2026-06-07
Gate the both-slots-torn checkpoint crash class end-to-end

ELares:
## What

An engine-level crash gate for the #164 "both-slots-torn checkpoint" class (refs #21), closing the last block-layer item that issue left open now that the durable checkpoint exists (#235).

The cursor checkpoint is a two-slot file: a single torn write can only damage one slot, so recovery falls back to the other (the `Checkpoint` unit tests already pin that byte-level behavior). What was untested is the **engine's consumption** of a file where **both** slots are corrupt at once, the media-rot / double-fault case.

## The test

`a_both_slots_torn_checkpoint_falls_back_to_redeliver_from_zero` (in `engine.rs`'s test module):

1. Drives a queue until both slots hold a real, distinct snapshot (ack two and checkpoint at committed 2, then ack two more and checkpoint at committed 4, so seq 1 and seq 2 land in different slots).
2. Flips a crc32c-covered payload byte in each slot of the on-disk `cursor.ckpt`, then reopens the engine.
3. Asserts the engine fails **safe**: `Engine::open` never errors or panics, the unreadable cursor is discarded, `committed_offset()` resets to 0, and all four durable records redeliver from zero exactly once as a fresh first delivery (`deliveries == 1`). At-least-once: duplicates are allowed, a lost ack is not.

A **control branch** runs the identical lifecycle with an intact checkpoint and asserts it resumes at the committed head (4) and redelivers nothing, so the test cannot pass vacuously.

## Verification

- `cargo test -p ironbus-server`: 84 tests pass (83 lib + 1 determinism), including the new gate.
- **Mutation check** for non-vacuity: neutralizing the corruption (XOR `0x00`) makes the gate fail (`committed` stays at 4, the intact-checkpoint value), proving the assertion is driven by the corruption.
- `cargo fmt --all --check` clean; `cargo clippy -p ironbus-server --all-targets -W clippy::pedantic -D warnings` clean.
- No production code change: the engine was already safe (`Checkpoint::open` returns `None` when both slots are unreadable; `resume_cursor_from_snapshot` resumes at committed 0). This makes that safety falsifiable per-PR.
- No em or en dashes in the prose.

refs #21, #164


ELares:

An independent review (not the author) verified the gate is rigorous and non-vacuous, re-deriving each claim from the source: both checkpoint slots are genuinely torn (bytes 10 and `slot+10` land inside the crc32c-covered payload region of each 78-byte slot for the 13-byte committed=2 snapshot, since the slot CRC covers `[0, payload_start + len)`); two successive `checkpoint_cursor()` calls write different slots via the seq%2 alternation; `Checkpoint::open` returns `Ok((_, None))` on the (None, None) branch so the engine resumes at committed 0 without error or panic; the corruption is visible to the reopen (InMemoryFile shares one `Arc<Mutex<State>>`, `write_all_at` then `snapshot` read the same live bytes, `sync_all` makes it durable); the control branch uses a separate engine and filesystem so the two branches cannot interfere; the frozen ManualClock makes it deterministic (the 30ns visibility never fires at now=0, every record delivers once). SPDX intact, no em or en dashes, diff is test-only plus the CHANGELOG line. Verdict: SHIP, no findings.

PR #252, opened 2026-06-07
Cover consumer-group fan-out in the golden path (#133 step 4)

ELares:
## What

Extends the golden-path acceptance test (`crates/ironbus-cli/tests/golden_path.rs`) to cover **#133 step 4, consumer-group fan-out**, end to end over the real `ironbus` binary. The consumer-groups stack is now complete (engine groups, wire `Sub`/`Unsub`, `ironbus sub --group`, durable per-group cursors #248), so this proves it as one scenario through the CLI.

## The scenario

One durable log of four records (`m0..m3`), fanned out two ways:

- **Broadcast group `bcast`** (its own cursor): drains all four, asserts it sees the whole batch.
- **Competing group `work`** (two members sharing one cursor): member 1 takes a bounded credit of 2 and acks; member 2 drains the rest. Asserts the two members' deliveries are **disjoint** and **together cover the batch exactly once** (none dropped or doubled), the per-group at-least-once property.
- **Independence**: a fresh fetch on each group then sees nothing, proving neither group's progress consumed the other's view of the same log.
- **Durable resume**: restart the broker; both groups resume past their acks (`fetched 0`), exercising the durable per-group cursors (#60, #248) over the wire.

## Why it is deterministic (not flaky)

- `--checkpoint-interval 1` checkpoints each group **synchronously per ack** (server routes `maybe_checkpoint_group` to the session's group), so the restart resume races nothing.
- Sequential `sub` calls: each consumer fully acks before the next runs, so the competing-member split is stable. Verified across **5 consecutive runs** (~0.8s each, all green).

## Non-vacuity

- A broken groups implementation (shared cursor) would make `bcast` consume the batch and leave member 1 with 0 messages, failing `m1.len() == 2`.
- Broken durable per-group cursors would redeliver after restart, failing the `fetched 0` resume assertions.

## Verification

- `cargo test -p ironbus-cli`: 36 tests pass (34 unit + 2 golden-path), including the existing `golden_path_produce_consume_restart_resume`.
- `cargo fmt --all --check` clean; `cargo clippy -p ironbus-cli --all-targets -W clippy::pedantic -D warnings` clean.
- No production code change (test + CHANGELOG only). SPDX intact. No em or en dashes.

The remaining #133 steps (overload spill, the simulated power cut, the loss report, the installer) stay open.

refs #9, #133


ELares:

An independent review (not the author) verified each claim against the implementation: competing-group semantics are real (ack_in advances only the subscribed group's cursor, so member 1 acking 0,1 starts member 2 at 2; the FLOW credit loop delivers at most max), broadcast independence is real (each named group is its own WorkGroup in the BTreeMap with an independent AckCursor), and the durable restart resume is deterministic, not racy (server.rs runs maybe_checkpoint_group synchronously per ack before the response is written, and Checkpoint::write fsyncs, so the group cursor is durable before the sub subprocess returns and before the SIGKILL). ChildGuard::drop kills then waits synchronously; the data dir is PID-unique so the two golden_path tests do not collide. Non-vacuity confirmed for both failure modes: a shared cursor would make member 1 get 0 (fails m1.len()==2), and broken durable cursors would redeliver after restart (fails the fetched-0 resume). No hangs (sub returns promptly on Poll::Idle; 30s connection timeout). SPDX intact, no em or en dashes, diff is test plus CHANGELOG only.

One non-blocking note: the test-only payloads() helper would mis-split a key containing the literal 'payload=', which cannot occur here (empty keys, fixed m0..m3 inputs). Verdict: SHIP.

PR #253, opened 2026-06-07
Wire an in-band dead-letter advisory to the consumer (#63)

ELares:
## What

Wire an **in-band dead-letter advisory** to the fetching consumer (refs #63, #11). When a message exceeds `MaxDeliver` the engine parks it (committed past, never redelivered), but until now the consumer was told nothing and just silently never saw that offset. `session.rs` even said so: "the dead-letter advisory + DLQ write (#63) is not yet wired, so the consumer is not told here."

## How

- **proto**: a new `DeadLetter` frame, wire tag **17** (the next free tag; the frozen-tag test pins it). Body: the dead-lettered `offset` (8-byte LE u64) plus a one-byte `reason` (`0` = max-deliver exhausted; the byte leaves room for future causes). New `DeadLetterBody` + `encode_dead_letter`/`decode_dead_letter` (fixed 9-byte layout, trailing bytes rejected).
- **server**: `session.rs` emits a `DeadLetter` frame in the Flow batch for each `Poll::Parked` offset, instead of silently swallowing it. It does not count toward the delivered total.
- **client**: `Client::fetch` now returns `Fetch { messages, dead_letters }`. The credit caps the **total** frames a server may stream back (deliveries plus advisories), so a buggy or hostile server cannot stream either unboundedly.
- **cli**: `ironbus sub` prints a `dead-letter offset=<n> reason=max-deliver` line for each advisory.

The durable DLQ topic write and idempotent reconciliation remain the larger #63 work; this is the in-band notification half. The `/metrics` `ironbus_dead_lettered_total` counter and `ironbus_last_dead_lettered_offset` gauge already existed; this adds the consumer-facing signal.

## Tests

- proto: a `DeadLetterBody` round-trip and short/overlong-body rejection; the frozen-tag test pins tag 17 and the unknown-tag proptest range moves to `18..=255`.
- client: `fetch_surfaces_a_dead_letter_advisory` drives a scripted server batch and asserts `fetch` returns the advisory in `dead_letters`, separate from `messages`.
- cli: the existing end-to-end poison test (`a_poison_message_is_dead_lettered_after_exceeding_max_deliver`) now asserts the consumer **is** told (`dead-letter offset=0 reason=max-deliver`), where it previously asserted silence.

## Verification

- `cargo test --workspace`: all green (incl. the new + updated tests).
- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -W clippy::pedantic -D warnings` clean.
- No em or en dashes. A stray proptest regression seed from a transient dev failure was removed and the directory gitignored (the repo tracks none; second commit).

refs #63, #11


ELares:

An independent review (not the author) verified the frozen wire contract (tag 17 is free; as_u8/from_u8 are a consistent bijection; ALL_TYPES bumped to 17; both frozen-tag tests updated and the unknown-tag proptest range moved to 18..=255), the DeadLetterBody codec (fixed 9-byte layout, Truncated on short and TrailingBytes on overlong, bounds-safe), the server emit (correct parked offset, advisory excluded from the delivered count, batch keeps draining), and the DoS bound (the server loop is bounded by credit and the client rejects once messages+dead_letters reaches the credit on BOTH arms, so a hostile server cannot stream unbounded advisories). The CLI prints the advisory and counts only messages; the client test asserts the advisory is surfaced and the end-to-end poison test now asserts the exact advisory line (non-vacuous). Hygiene clean: no em or en dashes, the transient proptest regression seed removed and the dir gitignored, full suite (516 tests) plus clippy pedantic and fmt all green. Verdict: SHIP, no findings.

PR #254, opened 2026-06-07
Cover offline inspection agreeing with recovery in the golden path (#133 step 9)

ELares:
## What

Extends the golden-path acceptance test with one new `#[test]` for #133 step 9: with the broker STOPPED, the offline inspection verbs agree with what recovery would resume. The whole test drives the REAL `ironbus` binary end to end (same `start_broker`, `run`, `ChildGuard` helpers and `#![cfg(unix)]` gate as the existing scenarios).

The new test:
1. Boots the broker and produces N=4 records via `ironbus pub`, asserting each pub returns its durable offset.
2. Cleanly stops the broker (drops the kill-guard).
3. With the broker STOPPED, runs `dump --data-dir` and `peek --data-dir` and asserts the offline reader shows exactly offsets 0..N, in order, exit code 0. It asserts there is no phantom record past the durable high-water mark (no `offset=N`) and no loss note on a clean directory. It asserts `peek --limit 2` shows a bounded subset (offsets 0 and 1) and `peek --from-offset 2 --limit 10` reads only the remaining durable records (offsets 2 and 3, never a phantom offset N).
4. Asserts `dump --json` is valid NDJSON: exactly one JSON object per record, offsets 0..N in order, each a record object (`crc:ok`, no `loss` object), and nothing past the high-water mark.

A small `offline_offsets` helper parses the `offset=` token from each record line, skipping any trailing loss note.

## Why it is non-vacuous

The assertions compare the EXACT set of offsets the offline reader emits (`assert_eq!` on a `Vec<u64>`), not a loose `contains`. A reader that dropped a record, doubled one, added a phantom past the high-water mark, or reordered the log would fail. Records are fsynced before `pub` returns the offset, so the data directory is durable after the broker exits and the offline reader sees the committed log without a server running.

## Verification

- `cargo test -p ironbus-cli --test golden_path`: PASS, run 3 times to confirm it is not flaky (it spawns real processes); all three runs green.
- `cargo clippy -p ironbus-cli --all-targets -- -W clippy::pedantic -D warnings`: clean.

Touches only `crates/ironbus-cli/tests/golden_path.rs` and `CHANGELOG.md`. No em or en dashes in the diff.

refs #133


ELares:

An independent review (not the author) verified the durability contract end to end: `Engine::produce` calls `log.sync()` (fdatasync) BEFORE returning the offset, so the four records are durable before `pub` returns and a SIGKILL via ChildGuard (which kill+waits) cannot leave a torn tail; the OfflineReader's durable head is the segment-chain high-water mark and it never yields a record at or above it (a torn tail past it is reported as loss, never read). Non-vacuity confirmed: the test asserts the EXACT offset set/order (`assert_eq!`, not a loose contains), no phantom offset N, and no loss note on a clean directory, so a dropped, doubled, reordered, or phantom record fails it. peek windowing (`--limit`, `--from-offset`) and `dump --json` NDJSON (exact line count, per-line object, no loss object) are asserted with correct flag names. Not flaky (3 runs). Scope clean: only golden_path.rs + CHANGELOG, SPDX intact, no em or en dashes, the CHANGELOG merge keeps both this and the #253 bullets. clippy pedantic and fmt clean. Verdict: SHIP, no findings.

PR #255, opened 2026-06-07
Pin the checkpoint single-byte-corruption invariant with a proptest

ELares:
## What

Hardens the durable cursor checkpoint decoder (`crates/ironbus-storage/src/checkpoint.rs`) with a property test. The checkpoint is a crash-safe two-slot crc32c-protected file; on recovery `Checkpoint::open` reads both slots, validates each, and returns the higher-sequence valid slot's payload or `None`. It stores the consumer cursor, so this is durability-critical.

## The invariant

For an arbitrary short run of writes (each a short arbitrary payload within `MAX_PAYLOAD`) followed by a single flipped byte at an arbitrary offset anywhere in the 156-byte file, `Checkpoint::open` MUST:

- never panic, and
- return either `None` OR a payload byte-equal to one of the payloads actually written, never a torn, partial, or garbage value.

In other words: an arbitrary one-byte corruption can only LOSE the cursor (regress to an earlier durable value or `None`) or return a genuinely-written value, never fabricate one. The existing tests only pin a handful of fixed torn-slot cases (`a_torn_newest_slot...`, `both_slots_torn...`, `a_corrupted_length...`, `a_corrupted_sequence...`); this proptest generalizes them over the whole input and offset space, so a regression that wrongly accepted a corrupt slot would fail.

The test drives the real `Checkpoint::open` over an `InMemoryFile`, mirroring the existing `fresh()` helper and the fixed tests' corruption pattern (`snapshot` -> flip -> `set_len(0)` -> `write_all_at` -> `sync_data`).

## Non-vacuity

The 156-byte file is two 78-byte slots: `seq[0..8]`, `len[8..10]`, `payload[10..74]`, `crc[74..78]`. The flip offset is uniform over `0..156`, so it lands in the payload region for some inputs and the seq/len/crc region for others, reaching the corruption-tolerance decode path across the input space. I confirmed this empirically with a throwaway 200000-iteration probe (NOT committed): the flip reached both the payload region (164072) and the seq/len/crc region (35928), and both outcomes occurred (`Some` returning a written value 188134 times, `None` 11866 times), so the `Some`-branch assertion is exercised non-trivially. The invariant held on every iteration.

## Verification

- `cargo test -p ironbus-storage checkpoint` passes (run twice, stable); also a `PROPTEST_CASES=20000` deep sweep passes.
- `cargo fmt --all -- --check` clean.
- `cargo clippy -p ironbus-storage --all-targets -- -W clippy::pedantic -D warnings` clean.
- No em/en dashes in the touched files; no `proptest-regressions/` artifact committed. Touches only `checkpoint.rs` and `CHANGELOG.md`.

refs #21


ELares:

An independent review (not the author) verified the proptest drives the real `Checkpoint::open` over an InMemoryFile and asserts the recovered payload is byte-equal to one ACTUALLY written (accepting an older slot's value, rejecting any torn, partial, or fabricated value). Non-vacuity confirmed: writes are 1..6 (a slot is always populated), the flip is uniform over all 156 bytes so it reaches the crc-covered seq/len/payload region, the crc bytes, AND the unused padding, and both Some and None outcomes are exercised. The subtle correctness point holds: a flip in a slot's unused padding (beyond the meaningful len, which the crc does not cover) leaves the slot valid and returns the original payload, and the `any(|p| p == &recovered)` check correctly accepts that, so there are no false-positive failures. Genuinely new (the only prior arbitrary-corruption proptest targets the segment log in crash_recovery.rs, not the checkpoint). Clean scope: only checkpoint.rs + CHANGELOG, no committed proptest-regressions, SPDX intact, no em or en dashes, the CHANGELOG merge keeps all three recent bullets. Stable at PROPTEST_CASES=20000 across 3 runs; clippy pedantic and fmt clean. Verdict: SHIP, no findings.

PR #258, opened 2026-06-07
Cover power-cut recovery and the loss report in the golden path (#133 steps 6-7)

ELares:
## What

Drive **#133 steps 6 and 7** (power-cut recovery + the loss report) end to end over the real `ironbus` binary, in `crates/ironbus-cli/tests/golden_path.rs`.

## The scenario

1. Produce four durable records (fsynced before `pub` returns), then stop the broker.
2. **Simulate the power cut**: append a torn tail (`0xFF` bytes, which cannot start a valid record since they are not the `0x4942` record magic) to the active `seg-*.log`, past the last durable record. Segments grow (no pre-allocation), so the torn bytes land immediately after the durable head.
3. **Offline** (broker stopped): `dump --data-dir` shows exactly offsets `[0,1,2,3]` and a torn-tail loss note of the expected byte count, never reading past the durable head.
4. **Online**: restart the broker (with `--health-addr`). Recovery truncates the torn tail and exposes the dropped bytes on `/metrics` as `ironbus_recovery_truncated_bytes`. The test asserts this **equals** the offline `dump` loss, so the online counter and the offline reader agree on the byte count.
5. The durable prefix is fully consumable (all four via `sub`), and the log continues cleanly from the truncated head (`pub` returns offset 4).

## Invariants exercised

- I1 (no acked record lost): the four durable records all survive.
- Recovery never reads past a torn tail: offline and online both stop at the durable head.
- Loss is bounded and reported: the same byte count appears in `dump` and on `/metrics`.

## Non-vacuity / not flaky

- The byte-count assertions are **exact** (`dump` loss == `/metrics` truncated == the appended torn length). If recovery failed to truncate, failed to report, or dropped a durable record, the test fails. `dump_loss_bytes(...).expect(...)` requires the loss note to be present, so a no-op corruption would fail.
- Stable across **3 consecutive runs**.

## Helpers added

- `start_broker_with_health`: boots with `--health-addr 127.0.0.1:0` and parses both the wire and health addresses from stdout.
- `http_get` + `metric_value`: a minimal HTTP/1.0 GET to read a named gauge from the `/metrics` exposition.
- `append_torn_tail`: appends bytes to the active segment.

## Verification

- `cargo test -p ironbus-cli --test golden_path`: all 4 tests pass.
- `cargo fmt --all --check` clean; `cargo clippy -p ironbus-cli --all-targets -W clippy::pedantic -D warnings` clean.
- No production code change (test + CHANGELOG only). No em or en dashes.

Only the overload-spill and installer steps of #133 remain after this.

refs #133, #7, #8, #16


ELares:

An independent review (not the author) verified the test across seven dimensions and returned SHIP. Confirmed: segments are not pre-allocated (the 0xFF torn tail lands immediately past the durable head; 0xFFFF cannot be misread as a record since the magic is 0x4942), records are fsynced before pub returns, the offline reader never mutates the dir (so the tail is still present for the online recovery), and the byte-count agreement is real, not coincidental: both the offline `total_bytes_skipped` and the online `recovered_truncated_bytes` are `physical_len - valid_end`, so both equal the appended length. Non-vacuity confirmed for every failure mode (lost record, no truncation, no loss report, disagreeing counts), assertions are exact, and the test is stable across 3 runs with all children guarded and timeouts on every wait.

Two items raised and addressed in e978e83:
- The module-level doc still said the power cut and loss report were not built. Updated to list all four scenarios and note only the overload spill (step 5) and installer (step 1) remain.
- The reviewer noted a vanishingly-unlikely window between the health listener binding and its accept loop. `http_get` now retries briefly (the connect already queued in the listen backlog, so this only removes any read-timing window on a slow runner).

Will merge on green CI for the new commit.

Issue #13, opened 2026-06-05
[DESIGN] Retention, compaction, and segment lifecycle


ELares:
Retention is how IronBus reclaims disk: it decides which records age out, when whole segments are deleted, and what happens when the disk fills. This issue specifies three retention modes (time, total size, and message count), a cheap whole-segment delete path that never touches individual records, an optional key-based log compaction mode for state-snapshot topics, and an explicit disk-full policy that interlocks with #10. Tiered offload to object storage is called out as a non-goal for v1 but is designed for, so the segment lifecycle does not have to be re-cut later. Because IronBus targets small-flash edge devices, retention defaults are conservative and bounded, and everything here builds directly on the segment model in #4.

## Why this matters for IronBus

Retention is where four tenets meet. **Resilient**: a queue that never reclaims space wedges the device when the disk fills, so a correct disk-full policy is a durability feature, not an afterthought. **Edge First**: edge nodes have tiny disks (SD/eMMC measured in single-digit GB), so the default must bound consumption tightly and prefer the cheapest possible reclamation, which is deleting a sealed segment file (one `unlink`) rather than rewriting records. **Simple**: one knob set with safe defaults, not a matrix of surprising interactions. **HyperScale**: the same segment-delete mechanism that keeps an edge box bounded also lets a beefier node hold a large retained window without per-record bookkeeping. Tie-ins: limits and their defaults live in #14; the flash-wear, fsync-cost, and brownout pressures that shape conservative defaults are detailed in #20; the disk-full decision is a backpressure decision shared with #10.

## Prior art

[...]

PR #259, opened 2026-06-07
Shed an over-cap produce at a durable-log byte cap (refs #10, #13)

ELares:
## What

Implements the **SHED** half of the README-frozen overflow policy "spill-then-shed (block opt-in only)" for the single durable queue. The log already spills by default (a produce becomes durable log depth on disk); this adds the bounded backstop.

## Semantics (drop-new shed at a byte cap)

- **Configurable cap.** New `LogConfig::max_total_bytes` (a `u64`). `0` means **unlimited** and is the **default**, so existing behavior is unchanged: an operator opts in. Set it with the new `LogConfig::with_max_total_bytes` builder; `LogConfig::new` still defaults it to unlimited.
- **What is capped.** The total durable **RECORD** bytes across every segment, the same quantity recovery sums as `durable_bytes` (per segment, `valid_end - SEGMENT_HEADER_LEN`, so segment headers and footers are excluded).
- **Drop-new rejection.** In `Log::append`, BEFORE writing: if the cap is set and the current total is at or over it, return the new `StorageError::AtCapacity { durable_bytes, cap }` and write nothing. No offset or sequence advances. The check is **at-or-over** like `max_segment_bytes` (the log overshoots by at most the last record), and a record on an **empty** log is always written (an oversized first record is not wedged out).
- **O(1) accounting.** A running `sealed_record_bytes` total is initialized at `open`/`recover` (reusing the existing `durable_bytes` computation) and updated on each roll; the live total is `sealed_record_bytes + (active.write_pos() - SEGMENT_HEADER_LEN)`. No O(segments) scan per append. Exposed as `Log::durable_record_bytes()` / `Engine::durable_record_bytes()`.
- **Non-fatal.** `AtCapacity` is a normal, recoverable shed, NOT the fatal `WriterFrozen` freeze. `EngineError::is_fatal()` returns FALSE for it, the writer stays live, and a later produce succeeds once retention (#13) frees space. This is the "never a silent drop and never an indefinite hang" requirement: a prompt, typed rejection, not a hang and not a silent loss.

## How the producer is told + observability

- `Engine::produce` returns the error unchanged (wrapped via `EngineError::Storage`) and increments a new `Counters::produce_rejected`. A rejected produce does NOT move `produced`/`produced_bytes`.
- The session (`handle_pub`) matches the error kind (via `EngineError::is_at_capacity` / `StorageError::is_at_capacity`, not the message text) and replies a distinct, stable `Err` frame body `at capacity`. Every other non-fatal produce error keeps `produce failed`. The connection stays open, so the producer can keep going.
- `/metrics` renders `ironbus_produce_rejected_total` (a counter, with HELP/TYPE, contiguous with the other produce families).
- `ironbus-client` already maps an `Err` frame to `ClientError::Server(msg)`, so `at capacity` surfaces to a client and the CLI with no new client variant.

## Tests

- **Storage** (`log.rs`): over-cap produce returns `AtCapacity`, is non-fatal, advances nothing, and is absent after a reopen (durable head unchanged); under-cap appends succeed and the writer stays live for more; byte accounting is exact across a roll (sealed + active) and across a reopen; an oversized first record on an empty log is written; `max_total_bytes = 0` preserves unlimited behavior.
- **Engine**: a rejected produce increments `produce_rejected` and does not move `produced`/`produced_bytes` or the flushed offset; the writer never freezes (a second over-cap produce is still a counted shed); an under-cap produce works.
- **/metrics**: `ironbus_produce_rejected_total` is rendered and increments (0, then 2 after two sheds; the one success is counted once).
- **Session wire**: an over-cap `Pub` returns an `Err` frame whose body is `at capacity` (distinct from `produce failed`), the session does not end (process returns Ok), a follow-up `Ping` is answered, and the shed is repeatable.

## Verification

`cargo test --workspace`, `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings`, and `cargo build -p ironbus-core` all pass. A Python scan over the changed files reports no em or en dashes.

Unblocks golden-path #133 step 5.

refs #10, #13


ELares:

An independent review (not the author) rigorously verified the durability-critical pieces. Byte accounting is correct: at recovery, sealed_record_bytes = durable_bytes minus the active segment's record bytes (no double-count); on roll it gains exactly the sealed segment's record bytes; durable_record_bytes() = sealed + (active.write_pos - SEGMENT_HEADER_LEN), proven by a concrete 2-sealed-plus-1-active walkthrough, with saturating_sub guarding any underflow. Drop-new never loses or corrupts data: the at-or-over check returns AtCapacity BEFORE any write/roll/offset advance, a reopen test confirms the rejected record is absent and the durable head unchanged, and the path only rejects NEW produces (nothing acked is touched). Non-fatal is enforced by a typed predicate (is_at_capacity), not a string, so it is never confused with the fatal WriterFrozen; the writer stays live and the session stays open. Edge cases all covered (empty-log always writes, overshoot at most one record, 0 = unlimited, no u64 overflow). No regression to the crash gates or the byte-identical determinism test (21 + 1 pass; the test changes are mechanical max_total_bytes: 0 additions). The produce_rejected counter is saturating, increments only on at-capacity, leaves produced/produced_bytes/flushed untouched, and renders contiguously on /metrics. No panics in library paths, all new APIs rustdoc'd, ironbus-core untouched, no em or en dashes, clippy pedantic + fmt clean. Verdict: SHIP, no findings.

PR #260, opened 2026-06-07
Expose the durable-log byte cap on serve and cover golden-path #133 step 5

ELares:
## The flag

Adds a `serve --max-total-bytes <bytes>` flag, exposing the durable-log byte cap (#10) on the CLI. It is parsed like the existing numeric `serve` flags: default `0` (unlimited, the current behavior), accepts any `u64` (including `0`). It threads into the `EngineConfig`'s `LogConfig` via `LogConfig::new(max_segment_bytes).with_max_total_bytes(max_total_bytes)`, so BOTH the segment cap and the total cap are honored. The USAGE string now lists `[--max-total-bytes <bytes>]` on the `serve` line. An existing `serve` invocation without the flag is byte-for-byte unchanged (the cap stays `0` = off). Two unit tests mirror the existing flag tests: a non-numeric value is a usage error, and an explicit `0` (unlimited) parses and validates (never `EXIT_USAGE`).

## The step-5 scenario (overload spill then shed)

Adds one `#[test]` to `crates/ironbus-cli/tests/golden_path.rs` completing #133 step 5, driving the real `ironbus` binary with the existing helpers (`http_get`, `metric_value`, `run`, `payloads`). It generalizes `start_broker_with_health` into `start_broker_with_health_args(data_dir, extra)` so the broker can boot with both `--max-total-bytes` and `--health-addr` (the old helper delegates to the new one, so existing callers are unchanged), and adds a `run_err` helper that also returns stderr (so a shed pub's distinct `at capacity` message is asserted).

The split is deterministic: each `pub` (empty key, empty headers, 2-byte payload) is 46 bytes on disk (36 header + 2 payload + 8 trailer), and a 100-byte cap accepts exactly the first three records (record 0 always writes on an empty log; 1 and 2 fit under 100; from 3 on the log is over the cap). Verified directly against the binary: 3 accepted, 7 shed, `ironbus_produce_rejected_total 7`.

## Non-vacuity (the load-bearing checks)

- At least one pub SPILLED (accepted) and at least one was SHED (rejected), so the cap actually engaged.
- The client shed count EQUALS `/metrics ironbus_produce_rejected_total` EXACTLY (the client rejections match the server's shed counter, no silent drop, no phantom rejection).
- After the shed the broker is STILL ALIVE: a further pub is promptly shed again (never a hang, never a silent success), and the live counter advances by exactly one.
- Every accepted record is durably consumable via `sub --max 100 --ack`, exactly the accepted prefix, in order, nothing extra. We never consume between pubs (acking does not free cap space; retention is #13).

## No flake

The new test was run 3x in isolation, all pass (it spawns the real broker process per run). The whole `ironbus-cli` suite passes (36 unit + 5 golden-path). `cargo fmt --all -- --check`, `cargo clippy -p ironbus-cli --all-targets -- -W clippy::pedantic -D warnings`, and a dash scan over the changed files are all clean.

refs #133, #10


ELares:

An independent review (not the author) verified the flag and the step-5 test across five dimensions, all clean. The flag is parsed as a u64 (default 0 = unlimited, non-numeric rejected as a usage error), threaded via LogConfig::new(max_segment_bytes).with_max_total_bytes(...) so BOTH the segment cap and the total cap are honored, the USAGE line is updated, and a serve without the flag keeps the cap off. The step-5 assertions are non-vacuous and relationship-based, not hardcoded counts: it requires at least one spill AND at least one shed, the client shed count equals /metrics ironbus_produce_rejected_total exactly, a further pub after the shed is promptly rejected again with the counter advancing by exactly one (no hang, no silent success), and the accepted prefix is durably consumable as exactly the accepted payloads in order. Acking does not free cap space (#13), and the test consumes only at the end, so the spill-then-shed split is deterministic. Not flaky (2 runs plus the full suite). No unwrap/expect/panic in the library path, SPDX intact, module doc updated (only the installer step of #133 remains), no em or en dashes, clippy pedantic + fmt clean. Verdict: SHIP, no findings.

Issue #84, opened 2026-06-05
[TASK] Consumer-safe retention knob, truncation signal, and clock-source semantics


ELares:
Parent: #13

Resolve the two seams the design left implicit: how retention interacts with the slowest consumer cursor in a fan-out queue, and what clock retention's `max_age` trusts. Both are correctness contracts that #9, #5, and #16 depend on.

## Approach

Add `retention.consumer_safe` (default `false`). When `false`, retention is bounded purely by time/size/count and MAY delete a sealed segment below a lagging consumer's committed cursor; that consumer then receives a single explicit truncation/gap signal (defined with #9) and resets to the new durable log-start offset. When `true`, the reaper will not delete a sealed segment whose range is at or below the minimum committed cursor across registered DURABLE consumers, and the disk-full case is handed to #10. Default is `false` because `consumer_safe=true` lets one dead consumer fill a tiny flash disk, violating the prime Edge First bound; the truncation signal makes the loss loud, not silent. For clocks: `max_age` compares against the segment footer's highest BROKER append timestamp (from #5), never a producer-supplied timestamp, clamped `max(footer_time, now)` so a backward NTP step cannot make old data look young, and a clock-anomaly warning is emitted to #16 on a large jump. A consumer-lag metric and a truncation-event counter are exported.

## Acceptance criteria

- [ ] `consumer_safe` exists, defaults to `false`, and is documented with both behaviors.
- [ ] With `false`, a lagging consumer whose data is reaped receives exactly one explicit truncation signal and resets to log-start (test with a deliberately stalled consumer).
- [ ] With `true`, the reaper never deletes a sealed segment at or below the minimum durable committed cursor, and disk-full is handed to #10.
- [ ] `max_age` uses broker footer time, never producer time, clamped against backward jumps.
- [ ] A large clock jump emits an anomaly warning to #16 and does not mass-reap by age.
- [ ] Consumer-lag and truncation-event metrics are exported.

## Failure considerations

`consumer_safe=true` plus a dead consumer fills the disk; the default `false` plus loud truncation, and the #10 handoff when `true`, bound the device. A producer-supplied age basis would let a misconfigured producer force premature deletion of everyone's data; using broker footer time prevents it. An RTC-less board's forward boot step could mass-reap by age; the clamp and anomaly handling prevent it. A silent truncation would look like a bug to consumers; the explicit single signal plus metric make it observable.

ELares:
[...]

Issue #83, opened 2026-06-05
[TASK] Optional key-based log compaction with atomic MANIFEST swap


ELares:
Parent: #13

Implement opt-in Kafka-style key-based log compaction for changelog/state-snapshot topics: keep at least the last value per key, honor null-value tombstones, never reorder records or rewrite offsets, and never compact the active segment. Off by default because it costs CPU and flash writes an edge core cannot spare; when enabled it is rate-limited and yields to appends.

## Approach

The cleaner reads N adjacent dirty sealed segments, builds a key-to-latest-offset map, writes survivors (each keeping its ORIGINAL offset, producing a sparse offset range) into a fresh segment file with a new monotonic segment id, fsyncs it, then performs a SINGLE atomic MANIFEST entry replacement pointing the covered offset range at the new file, then unlinks the originals using rename-then-unlink so open readers drain rather than reading freed data. A crash before the swap leaves the originals authoritative (discard the orphan new segment on recovery); a crash after the swap leaves unreferenced originals that recovery unlinks. Compaction is gated by `min_dirty_ratio` (default 0.5). Tombstones are retained for `tombstone_ttl` (default 24h) so offline consumers observe deletes. `compact_and_delete` is supported: the cheap age/size reaper runs FIRST so we never compact a segment we are about to delete. The read path and sidecar index must tolerate sparse (non-contiguous) offsets; coordinate the reader gap semantics with #9.

## Acceptance criteria

- [ ] Compaction preserves at least the last value per key, honors tombstones for `tombstone_ttl` (default 24h), and never reorders records or rewrites offsets.
- [ ] The active segment is never compacted.
- [ ] Survivors keep original offsets; the index and readers tolerate sparse offsets and skip gaps.
- [ ] Crash before/after the atomic MANIFEST swap recovers via longest-valid-prefix with no compaction-specific repair (fault-injection test).
- [ ] Open readers never read freed data during swap/unlink (rename-then-unlink or drain).
- [ ] Cleaner is rate-limited, yields to the append path, and is off by default.
- [ ] `compact_and_delete` runs the size/age reaper before the compactor.

## Failure considerations

[...]

Issue #82, opened 2026-06-05
[TASK] Disk-full policy and bounded block handoff to backpressure


ELares:
Parent: #13

Specify and implement the overflow policy that governs what happens when retention cannot free enough space to admit a new append: `block` (default), `drop_oldest`, and `refuse`. This is a shared flow-control decision with #10; this task owns the retention-side trigger and the precise, documented loss semantics of each mode, and hands the actual producer stall to #10.

## Approach

Default `disk_full = block`, but block is BOUNDED: when the disk-full condition holds (the reaper has reaped every eligible sealed segment and still cannot admit the append, because remaining bytes are the active segment or cursor-pinned), the append future does not resolve until space frees OR a `block_timeout` (default 30s) elapses, at which point it escalates to `refuse` and returns an explicit error to the producer. This prevents an unbounded hang that would be indistinguishable from a crash on an edge box. `drop_oldest` deletes the oldest sealed segment even though it is past no bound, accepting bounded REPORTED loss, intended for freshness-sensitive telemetry. `refuse` rejects the append immediately with no wait. The handoff to #10 is defined precisely: retention signals "cannot admit" and #10 owns the producer-facing stall/shed mechanics. Each mode's loss semantics are documented. Startup validation in #14 rejects `max_bytes` smaller than two segment roll sizes so the bound is always satisfiable by reaping.

## Acceptance criteria

- [ ] Three modes implemented with documented loss semantics; default is `block`.
- [ ] `block` is bounded by `block_timeout` (default 30s) and escalates to `refuse` rather than hanging forever.
- [ ] `drop_oldest` loss is bounded and emitted as an observable event to #16, never silent.
- [ ] The retention-to-#10 handoff is specified with one clear owner for the producer stall.
- [ ] Startup config validation rejects `max_bytes` below two segment roll sizes.
- [ ] A full-disk integration test exercises all three modes and the timeout escalation.

## Failure considerations

Unbounded `block` hangs the producer and looks like a dead process; the `block_timeout` escalation to `refuse` surfaces an error instead. `drop_oldest` could silently lose data; mandating an observable loss event prevents silent loss. A too-small `max_bytes` makes the bound permanently unsatisfiable and thrashes; fail-fast startup validation catches it. Two subsystems both implementing disk-full logic would diverge; the explicit single-owner handoff to #10 prevents overlapping code paths.

ELares:
[...]

PR #261, opened 2026-06-07
Reclaim disk with consumer-safe size retention (refs #13, #80)

ELares:
## Summary

Implements size-based, **consumer-safe RETENTION** (issue #13, task #80, the consumer-safe rule from #84): a reaper that reclaims disk by deleting whole OLD SEALED segments when the durable log exceeds a configured byte bound, but NEVER deletes a segment any consumer still needs. This is the drain side of the overflow policy that complements the just-merged byte-cap shed (#259): the cap sheds new produces when full, retention frees space as consumers drain. **Default OFF** (`0` = unlimited), so existing behavior is unchanged.

## Reaper semantics (`Log::reap_to_size(max_retained_bytes, protect_below_offset)`)

Whole-sealed-segment unlink only, looping from the OLDEST sealed segment. `segments[0]` is deleted only while ALL hold:
- `max_retained_bytes != 0` (a zero bound is UNLIMITED: nothing reaped, the default);
- `durable_record_bytes() > max_retained_bytes` (still over the bound);
- more than one segment exists, so the ACTIVE segment (the last slot) is never reaped;
- `segments[1].base_offset <= protect_below_offset` (the consumer-safety rule, below).

The loop stops as soon as the bound is satisfied, the next oldest segment is not fully consumed (`segments[1].base_offset > protect_below_offset`), or only the active segment remains. A whole segment is unlinked or left untouched; a record inside a segment is never rewritten or partially removed.

**Crash-safe ordering:** the segment file is unlinked and the directory fsynced (so the removal is durable) BEFORE the slot leaves `segments` and `sealed_record_bytes` is decremented. A crash after some unlinks leaves a shorter contiguous chain with a non-zero start, which recovery already accepts (the base-continuity check is skipped for the first segment); a crash mid-unlink is fine because unlink is atomic and the in-memory state was not yet changed. On any IO error the in-memory state is left consistent with disk (the failing segment is never orphaned), and the error propagates (never swallowed into silent corruption).

## Byte-accounting source

Each reaped segment's record bytes are read as `valid_end - SEGMENT_HEADER_LEN` from a streaming recovery scan (`SegmentReader::scan_recovery`, no payload materialization). This is EXACTLY the per-segment term `sealed_record_bytes` accumulated on each roll: a sealed segment's `valid_end` is the start of its footer, so it excludes both the 64-byte header and the 32-byte footer. After a reap the running total still equals a fresh reopen's recomputed value (cross-checked by a test). Chosen over `file_len - SEGMENT_HEADER_LEN` because the physical length includes the 32-byte footer, which would over-decrement.

## Consumer-safety rule

`protect_below_offset` is the floor below which records are safe to drop: every consumer has committed at least this far. Since segments are contiguous, `segments[0]` ends exactly where `segments[1]` begins, so deleting `segments[0]` only when `segments[1].base_offset <= protect_below_offset` guarantees every record in it has offset `< segments[1].base_offset <= protect_below_offset`, needed by no consumer.

## Engine wiring (min-across-groups)

`EngineConfig.max_retained_bytes` (`0` = unlimited, the default; every existing config literal updated to default it to 0). After a successful produce the engine calls `reap_to_size(max_retained_bytes, protect_below_offset)` where `protect_below_offset` is the **MINIMUM committed offset across ALL consumer groups** (the default group `""` always exists, so the min is well-defined), so the slowest group's unconsumed records are never reaped. A reap IO error propagates (best-effort, but real; the produce already succeeded and was counted, so the error does not undo a durable record).

## Metric and flag

- Counter: `Counters.segments_reaped` (saturating), rendered on `/metrics` as `ironbus_segments_reaped_total` following the existing counter-family pattern.
- Flag: `serve --max-retained-bytes <bytes>` (default `0` = off), parsed like the existing numeric flags and added to the USAGE string.
- The `Filesystem::remove` + dir-sync seam already existed (with power-loss semantics), so no fs-trait change was needed.

## Tests

**Storage (`log.rs`):** reaps the oldest fully-consumed segments until under the bound; never the active segment; never a segment above the protect floor (a low protect offset reaps nothing); stops at the first not-fully-consumed segment; `max_retained_bytes = 0` and an oversized bound reap nothing; after a reap a REOPEN recovers the remaining contiguous chain from a non-zero start (head/offsets correct, reaped records out of range, survivors read back); byte accounting matches a fresh reopen; crash-safety (a power loss after a reap resurrects nothing and breaks no chain); plus a real-`StdFs`-directory reap+recovery.

**Engine:** reclaims old consumed segments as the log grows (live `durable_record_bytes()` drops, `segments_reaped` increments); a SLOW group pins the protect floor and prevents reaping the segments it still needs, then unblocks once it catches up; the default unlimited config reaps nothing.

**`/metrics`:** the reaped counter renders and increments.

## Verification

`cargo test --workspace`, `cargo fmt --all -- --check`, and `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings` all pass. The crash-recovery and determinism gates pass unchanged. A python scan over the changed files reports no em/en dashes.

## Follow-ups (out of scope)

Time/count retention modes (#80), compaction (#83), the disk-full drop-oldest policy (#82), and the below-earliest truncation signal (#84) remain follow-ups.

refs #13, #80


ELares:

An independent review (not the author) rigorously verified the durability-critical pieces, all correct. Byte accounting is exact: roll accumulates write_pos - SEGMENT_HEADER_LEN and reap subtracts valid_end - SEGMENT_HEADER_LEN, both the record region (excluding the 32-byte footer); the cross-check test (sealed_record_bytes after a reap equals a fresh reopen) would fail by 32 bytes on any footer drift, so it is non-vacuous. Crash-safety ordering is correct: unlink then dir-sync then the in-memory slot removal and counter decrement, so no crash leaves memory ahead of disk (recovery rescans the directory), and an IO error propagates as a typed StorageError leaving segments/sealed_record_bytes unchanged. Consumer-safety holds: a segment is reaped only when segments[1].base_offset <= protect_below_offset where protect_below is the minimum committed offset across all groups (default group always included, a new lagging group sits at 0 and pins the floor), and committed is exclusive so the segment containing protect_below is never reaped (slow-group-blocks-then-resumes tested). The active segment is never reaped, 0 = unlimited off, a bound larger than the log reaps nothing, and recovery after a reap reopens the remaining contiguous chain from a non-zero start. No regression (crash gates 21/21, determinism 1/1; the EngineConfig field ripple to client/server/session/determinism configs and the run_serve/open_disk_engine refactor are mechanical and behavior-preserving). Reaping is O(1) per produce and skipped entirely when off. No panics in library paths, all new APIs rustdoc'd, ironbus-core untouched, no em or en dashes, the metric renders contiguously, clippy pedantic + fmt clean. Verdict: SHIP, no findings.

PR #262, opened 2026-06-07
Cover golden-path #133 step 2: health endpoints come up on loopback

ELares:
Adds the golden-path #133 step-2 test, end to end over the real `ironbus` binary, to `crates/ironbus-cli/tests/golden_path.rs`. Step 2 (per #133): boot zero-config and assert `/healthz` and `/readyz` come up, report healthy, and are bound to loopback only.

The new test `golden_path_health_endpoints_come_up_on_loopback` reuses the existing `start_broker_with_health` and `http_get` helpers (no broker-spawn or HTTP logic duplicated, no src changed). It asserts:

- **Loopback only**: the parsed health address host is `127.x` or `::1` (the serve flag pins `127.0.0.1`), so a regression that bound `0.0.0.0` or a routable address (which would leak the no-auth-on-loopback endpoints) fails here.
- **/healthz**: status line exactly `HTTP/1.1 200 OK` AND body exactly `ok` (the healthy-liveness marker from `health.rs`).
- **/readyz**: status line exactly `HTTP/1.1 200 OK` AND body exactly `ready` (the healthy-readiness marker; a frozen writer would answer `503 Service Unavailable` "writer frozen"). A fresh boot's writer is live, so it is ready immediately; the existing `http_get` retry converges any brief readiness lag without an unbounded loop.
- **Unknown path**: `GET /nope` is a real `HTTP/1.1 404 Not Found` with body `unknown endpoint`, proving the router is real and the healthy assertions are non-vacuous (not a blanket 200).

Each assertion is anchored to the EXACT status line and the EXACT body marker `health.rs` returns (a small `split_status_and_body` helper splits the response at the `\r\n\r\n` header boundary), so the test cannot pass on an empty or wrong response.

The module-doc lists step 2 as covered; the CHANGELOG gets a terse top bullet under `### Added`.

Verification (worktree root):
- `cargo test -p ironbus-cli --test golden_path`: all 6 pass.
- The new test run 3x in isolation: not flaky (3/3 pass).
- `cargo clippy -p ironbus-cli --all-targets -- -W clippy::pedantic -D warnings`: clean.
- Python dash scan over the two changed files: no em/en dashes.

refs #133, #16


ELares:

An independent review (not the author) verified the assertions match health.rs byte-for-byte: /healthz returns 200 'ok' (health.rs:127), /readyz returns 200 'ready' when the engine is healthy (a fresh boot is immediately healthy, so no readiness race; health.rs:129-134), and an unknown path returns 404 'unknown endpoint' (health.rs:167). The split_status_and_body helper splits on the correct CRLFCRLF boundary and checks the status line, not the body. Non-vacuous: exact status and body markers, and the 404 assertion proves the router is real (a blanket-200 server would fail it). The loopback assertion reads the broker's actually-bound health address (would catch an accidental 0.0.0.0 bind). Not flaky (3 runs; http_get retry is bounded to ~2s). Scope is test plus CHANGELOG only, SPDX intact, module doc updated, no em or en dashes, clippy pedantic + fmt clean. Verdict: SHIP, no findings.

Issue #130, opened 2026-06-05
[META] ADR index: catalog every decision resolved in the second pass


ELares:
The second pass resolved roughly 130 explicit decisions across the 22 parent issues (durability default, offset semantics, segment lifecycle, checksum algorithm, runtime, target triples, TLS floor, and many more). These decisions are buried in issue bodies and will rot or be silently re-litigated. #22 already mandates an in-tree `rfcs/NNNN-slug.md` record per design issue, but there is no flat, queryable index of the atomic decisions themselves with their status and the cross-issue dependencies between them. This issue owns that index.

### Deliverable
A single `docs/adr/INDEX.md` (and a machine-readable `adr.yaml` the table renders from, mirroring the #2 claims.yaml pattern) listing every resolved decision as one row:

`ADR-id | issue (slug) | decision (one line) | status (accepted/superseded/conflicted) | supersedes | depends-on | resolving RFC`

### Seed rows (representative)
- ADR-0007 #6 default level = batch group-commit fdatasync, no proactive linger. depends-on ADR for sizing table. status: CONFLICTED (group-commit ms differs in #1/#2/#14).
- ADR-0011 #4 adopt log-is-WAL single append path. status: CONFLICTED (#6 and #1 assume a separate WAL).
- ADR-0019 #5 CRC32C only, checksum_algo byte = 0x1, reject others. status: CONFLICTED (#8 adds xxh3-64).
- ADR-0024 #5 do NOT recycle segments in v1. status: CONFLICTED (#4 and #20 recycle with generation tags).
- ADR-0031 #3 pure monotonic u64 offset, never reused, time as separate field. status: accepted.
- ADR-0048 #1 runtime = tokio, io_uring feature-gated Linux-only. status: accepted; depends-on #17 cross-build gate.
- ADR-0052 #18 TLS 1.3 only, no 1.2 fallback, rustls bundled. status: accepted.
- ADR-0060 #22 dual MIT OR Apache-2.0; MSRV 1.78. status: accepted.

### Why this is M0
The `status: CONFLICTED` column is the actionable output: it is the authoritative list of the cross-issue contradictions surfaced in this review (WAL architecture, sizing numbers, checksum stack, segment recycling, group-commit timing). Each conflicted ADR must be driven to a single accepted decision before the dependent code-phase issues start, so this index is the gate the planning phase exits through.

### Acceptance criteria
- Every decision: line in every parent issue digest has exactly one ADR row.
[...]

Issue #125, opened 2026-06-05
[TASK] RFC process scaffolding, CODEOWNERS, and back-filled design records


ELares:
Parent: #22

Stand up the RFC process so every design decision is vetted and recorded in-tree, and solve the cold-start problem where the process governs a repo that currently has no `rfcs/` directory, template, owners, or quorum rule. The GitHub issue is the working surface, but the canonical, mirrorable record is a Markdown file. This is the discipline that keeps the issue-driven design trail honest before any crate exists.

## Approach
Create `rfcs/0000-template.md` defining the required sections (summary, motivation, design, alternatives, drawbacks, unresolved questions) and the RFC states: draft, accepted, superseded, rejected. Back-fill each of the 22 parent designs as `rfcs/0001..0022-slug.md` so the trail survives independent of GitHub; the file wins on any conflict with an issue body. Add a `CODEOWNERS` file listing initial maintainers. Document the merge rule in CONTRIBUTING (sibling task references it): an RFC PR requires approval from at least one CODEOWNER plus a minimum 72-hour comment window; superseded RFCs are marked, never deleted. Make RFC-0001 ratify the workspace crate boundaries, dogfooding the process. A CI status check enforces the 72-hour window on `rfcs/` PRs.

## Acceptance criteria
- [ ] `rfcs/0000-template.md` exists with required sections and the four states.
- [ ] All 22 parent designs are back-filled as numbered RFC files.
- [ ] `CODEOWNERS` lists initial maintainers and covers `rfcs/`.
- [ ] The RFC lifecycle (open, review, accept, record, supersede) is documented end to end.
- [ ] RFC-0001 ratifies crate boundaries and is referenced from #1.
- [ ] A status check enforces the minimum comment window on RFC PRs.

## Failure considerations
Without a defined approver set and quorum, the first RFCs could be rubber-stamped by one person or stall forever; the CODEOWNERS + one-approval + 72-hour rule prevents both unilateral merges and deadlock. An issue body is mutable and platform-locked, so relying on it loses the design history; the in-tree file as canonical record makes the trail auditable and offline-readable. Back-filling 22 files is upfront cost, but it is paid once and buys a permanent, vendorable design record.

ELares:
[...]

Issue #22, opened 2026-06-05
[DESIGN] Project governance: license, repo structure, RFC process, versioning


ELares:
This issue specifies how IronBus is run as a project before any code lands: the license and its reasoning, the intended Cargo workspace layout, a lightweight RFC process so each design issue is vetted and recorded, SemVer and Minimum Supported Rust Version (MSRV) policy, and the baseline community and CI documents (CONTRIBUTING, CODE_OF_CONDUCT, SECURITY). The repository is intentionally code-free for now. It is issue-driven: every design here (#1 and its children) is a spec to be reviewed and merged as a document before a single line of the corresponding crate is written. Governance is the scaffolding that keeps that discipline honest.

## Why this matters for IronBus

Governance is where the Simple and Cross Platform tenets show up first, before the storage engine or wire protocol exist. Simple means a contributor can clone, read `CONTRIBUTING.md`, and understand the workspace in minutes; it also means one permissive license with no legal ambiguity for downstream embedders. Cross Platform and Edge First mean the license and supply-chain story must let IronBus be vendored into proprietary edge firmware and audited on-device (see #17 for the static-musl, `cargo-auditable` SBOM angle), which a copyleft or source-available license would obstruct. Resilient applies to the process itself: an RFC trail and a strict SemVer plus MSRV contract are how we avoid silently breaking the durability and recovery guarantees that the rest of these issues spend so much effort establishing. Getting governance wrong is cheap to fix now and expensive to fix after the first release.

## Prior art

We model the process on how the systems in our research brief actually run, not just how they store bytes. The two cleanest references are Rust's own ecosystem and the data-infrastructure projects.

License: the Rust toolchain, `tokio`, `bytes`, `crc32fast`, `lz4_flex`, and most of the crates IronBus would depend on ship under MIT OR Apache-2.0 dual licensing. Pulling a crate licensed otherwise (for example GPL) into a static binary would force a license change, so `cargo-deny` is the standard gate that enforces an allowlist of acceptable dependency licenses at CI time. Apache-2.0 specifically adds an explicit patent grant that bare MIT lacks, which matters for an infrastructure project.

Process: Rust, Kafka (KIPs such as KIP-98 and KIP-33 cited below), and many CNCF projects record non-trivial design decisions as numbered, reviewed documents before implementation. KIP-98 (exactly-once) and KIP-33 (the time index) are exactly the kind of vetted-spec-then-code artifact IronBus wants for #6, #3, and #5.

[...]

PR #263, opened 2026-06-07
Add ADR scaffolding and three Accepted records (refs #125)

ELares:
Adds a lightweight in-tree decision-record scaffold and records three decisions that were already made in the tree. The owning GitHub issues and the top-level README stay canonical; the ADRs just record those decisions next to the code with a pointer back to the issue.

## Files

- `docs/adr/README.md`: the ADR process (what an ADR is, four-digit numbering, the Proposed/Accepted/Superseded lifecycle, and that the owning issues and the README remain the source of truth; the ADR index is #130, under governance #22).
- `docs/adr/template.md`: a short Title / Status / Owning issue / Context / Decision / Consequences template.
- `docs/adr/0001-log-is-wal.md` (Accepted, owning #4 storage, #6 durability).
- `docs/adr/0002-segments-never-recycled-in-v1.md` (Accepted, owning #18 security, decided #139).
- `docs/adr/0003-default-compression-lz4-zstd-opt-in.md` (Accepted, owning #12 compression, decided #139).
- `CHANGELOG.md`: one terse bullet under the first `### Added`.

`.github/CODEOWNERS` already existed with the correct `* @ELares` rule, so it was left untouched (no gratuitous change).

## Source verified for each ADR

- **0001 log-is-WAL**: `README.md` line 60 ("The active segment **is** the write-ahead log: there is no separate WAL file to keep in sync"), the architecture diagram ("this IS the WAL"), the #4 issue-table row ("the active segment is the WAL"), and the "Key decisions already committed" table ("Log-is-WAL ... No separate WAL file. The offset index is derived and rebuildable").
- **0002 segments never recycled in v1**: the storage test `segment_ids_increase_monotonically_and_are_never_recycled` in `crates/ironbus-storage/src/log.rs`, whose comment records the #139 decision and that it keeps the #18 at-rest nonce safe ("a segment_id is never reused under a fixed key").
- **0003 lz4 default, zstd opt-in**: the `deny.toml` `[bans]` comment ("Per #139, the default path is pure Rust ... The default compression codec is lz4_flex (pure Rust); zstd (zstd-sys, vendored C) is opt-in only, behind an explicit feature, never on the default path"). The ADR notes this refines the older README #12 row ("zstd default, lz4 fallback").

## Checks

- Non-Rust docs only, no code/workflows touched. `git status` shows only `docs/adr/` and `CHANGELOG.md` changed.
- A Python scan over all created/changed files reports no em dashes or en dashes.

refs #125


ELares:
### Independent review (ADR accuracy)

An independent review verified each of the three ADRs against its cited source in the tree: 0001 (log-is-WAL) matches README.md line 60 and issues #4/#6; 0002 (segments never recycled in v1) matches the test segment_ids_increase_monotonically_and_are_never_recycled and its #139-decision comment, with the #18 nonce-safety rationale; 0003 (lz4_flex default, zstd opt-in) matches the deny.toml comment and the #139 decision, and correctly notes the README #12 row is stale while citing deny.toml as the current decision. The ADR process doc correctly keeps the owning issues and README canonical; statuses are Accepted; scope is docs plus CHANGELOG only; CODEOWNERS already existed on main and was rightly left out; no em or en dashes. Verdict: SHIP.

PR #264, opened 2026-06-07
Add Miri and coverage jobs to the nightly lane (#21)

ELares:
Adds two jobs to `.github/workflows/nightly.yml`, matching the existing lane's structure and SHA-pinning style. The existing `deep-sweep` and `fuzz` jobs are untouched, `permissions: contents: read` is preserved, and the workflow stays actionlint-clean.

## `miri` (IO-free and in-memory paths)
Runs Miri (the MIR interpreter) to catch undefined behavior the native test run cannot see: out-of-bounds reads, use-after-free, invalid pointer arithmetic, and data races in the codecs and the in-memory simulation.

Miri interprets the program, so it cannot execute a test that spawns a process, touches a real file, or does real networking. The lane is scoped to exactly the paths Miri can run cleanly:

- `cargo +nightly miri test -p ironbus-core --lib`: `ironbus-core` is fully IO-free (a CI gate forbids fs/net/process there), so its whole lib test set runs.
- `cargo +nightly miri test -p ironbus-storage --lib` with the real-IO tests skipped by name. The `InMemoryFile` / `InMemoryFs` simulation lib tests do no real IO and run under Miri with isolation ON (the tests drive a `ManualClock`, so no real time either). The real-IO tests are excluded with `--skip`:
  - `--skip std_file_tests::` (the `std_file_tests` module in `io.rs`)
  - `--skip std_tests::` (the `std_tests` modules in `fs.rs` and `naming.rs`)
  - `--skip stdfs_read_across_segments`
  - `--skip rolling_and_recovery_on_a_real_directory`
  - `--skip reap_and_recovery_on_a_real_directory`

  These open real files through foreign functions Miri cannot interpret. Per the issue guidance, this narrows the test selection by name rather than disabling Miri isolation globally; the exclusion and its rationale are documented in a comment on the step.

The process-spawning integration tests (which boot the real `ironbus` binary, e.g. the golden path) are out of scope because the job runs only `--lib`. No `MIRIFLAGS` were needed: the in-memory tests use `ManualClock` (no real clock) and `crc32c` falls back to its software path under Miri.

## `coverage` (informational)
Produces a workspace LCOV report with `cargo-llvm-cov` and uploads it as a build artifact:

- Installs `cargo-llvm-cov` via the pinned `taiki-e/install-action`, plus the `llvm-tools-preview` component.
- `cargo llvm-cov --workspace --all-features --locked --lcov --output-path lcov.info`.
- Uploads `lcov.info` via the pinned `actions/upload-artifact`.

This lane is informational: no threshold gate, and no Codecov upload or any new external service or token. It succeeds as long as coverage is collected.

## Reused action SHAs (no new or unpinned actions)
Every `uses:` is a full commit SHA already pinned elsewhere in the repo's workflows, with the human tag as a trailing comment:

- `actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5` # v4
- `dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8` (nightly via the `toolchain:` input for Miri; stable for coverage)
- `Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32` # v2
- `taiki-e/install-action@56545b37b57562edd73171cb6c62cc509db4c34e` # v2 (`tool: cargo-llvm-cov`)
- `actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02` # v4

## Verification
- `SHELLCHECK_OPTS=--exclude=SC2015 actionlint .github/workflows/nightly.yml` passes.
- A Python scan over the two changed files reports no em dashes or en dashes.
- Touches only `.github/workflows/nightly.yml` and `CHANGELOG.md`; no code or other workflows changed.

Miri and coverage are slow and were not run to completion locally; the toolchain, components, install-action invocations, and skip filters are syntactically verified and the action SHAs match those already used in the repo.

refs #21


ELares:
### Independent adversarial review and resolution

The independent review ran the Miri job's exact commands and caught a real BLOCKER: proptest's default failure-persistence calls getcwd, which Miri rejects under its default isolation, aborting the test binary, so the nightly Miri lane would have been perpetually red with no PR check to surface it. Everything else (the skip list excluding only real-IO/foreign-function tests, the coverage job, the reused action SHAs, actionlint, scope) was confirmed correct.

Fix: MIRIFLAGS=-Zmiri-disable-isolation (the standard remedy for running proptest under Miri; the error was literally 'getcwd not available when isolation is enabled', which the flag eliminates). Safe here because the job runs only --lib over the IO-free core and the in-memory storage simulation, with the real-IO tests skipped, so no test does real host IO.

Empirically confirmed locally with the fix: cargo +nightly miri test -p ironbus-core --lib passes 103/103, and -p ironbus-storage --lib (with the skip list) passes 156/156, no Miri-detected UB, the 16 real-IO tests correctly filtered. The BLOCKER is resolved and verified.

Merging on green CI (15/15) plus this review and the local Miri confirmation.

PR #265, opened 2026-06-07
Proptest the wire codecs in ironbus-proto

ELares:
Property tests pinning the wire codecs in \`ironbus-proto\`, adding ONLY the invariants not already covered by the existing unit tests, proptests, and cargo-fuzz targets. No production code changed; \`proptest\` was already a dev-dependency (no \`Cargo.toml\` change).

## Properties added

**Frame envelope (\`frame.rs\`)**
- \`decode_frame_on_arbitrary_bytes_never_panics\`: \`decode_frame\` on ARBITRARY bytes always returns a typed \`Ok(Frame)\` / \`Ok(Incomplete)\` / \`Err\`, never panicking or reading out of bounds (the property-level complement to the \`frame_decode\` fuzz target). When it frames, \`consumed\` stays within the input and equals \`LEN_PREFIX + 1 + body.len()\`; when Incomplete, it needs strictly more than the input it was given.
- \`decode_frame_with_cap_rejects_an_over_cap_length\`: any declared length strictly above the effective cap \`min(max_len, MAX_FRAME_LEN)\` is rejected as a typed \`FrameError::FrameTooLarge\`, never an allocation blowup, regardless of how few body bytes are present (the decoder reads only the 4-byte prefix).

**Message bodies (\`message.rs\`)**
- \`any_dead_letter_round_trips\`: a \`DEAD_LETTER\` body round-trips for any offset and reason, and is always the fixed 9-byte layout.
- \`any_sub_round_trips\`: the whole-body-is-the-group SUB case round-trips for any byte string, and the encoded body is exactly the group bytes (no framing of its own).
- \`decoding_arbitrary_bytes_never_panics\` extended to also cover \`decode_dead_letter\` (and asserts \`decode_sub\` recovers the bytes), so all four fallible body codecs plus the infallible SUB are swept.
- \`any_deliver_round_trips\` now generates ARBITRARY headers (it previously pinned headers to \`b""\`), so the DELIVER round-trip exercises the headers length field.

## Skipped as already covered (not duplicated)
- Frame envelope round-trip (req 1): \`any_frame_round_trips\` already covers it (type tag + arbitrary body + \`consumed == buf.len()\`).
- The unknown-type-tag forward-compat proptest: \`an_unknown_type_tag_still_frames\` exists; not touched per the task.
- Truncated-valid-frame Incomplete: \`a_truncated_frame_is_incomplete\` exists.
- \`pub\` and \`ack\` round-trips: \`any_pub_round_trips\` / \`any_ack_round_trips\` already cover them. Fixed-size / trailing-byte invariants for \`ack\` and \`dead_letter\` already have unit tests; the fixed-9-byte invariant is additionally pinned in the new \`dead_letter\` proptest.
- \`pub\`/\`ack\`/\`deliver\` arbitrary-bytes never-panic: already in \`decoding_arbitrary_bytes_never_panics\` (extended, not duplicated).

Default proptest case counts kept (the nightly deep sweep raises them); round-trip strategies generate VALID inputs, never-panic strategies use \`any::<Vec<u8>>()\`-style arbitrary bytes.

## Verification
- \`cargo test -p ironbus-proto\`: 37 passed, stable across two runs (no real codec bug found; nothing to report).
- \`cargo fmt --all -- --check\`: clean.
- \`cargo clippy -p ironbus-proto --all-targets -- -W clippy::pedantic -D warnings\`: clean.
- \`cargo build --workspace\`: clean.
- Em/en-dash scan over the three changed files: none.
- No \`proptest-regressions/\` file created or committed; \`git status\` clean of it.

refs #21, #11


ELares:
### Independent review

An independent review verified the four new proptests are non-vacuous and meaningful (round-trip strategies generate valid in-bounds inputs; the never-panic sweeps feed arbitrary bytes to the real decoders decode_frame/decode_frame_with_cap/decode_dead_letter/decode_sub and assert a typed result; the cap test asserts FrameTooLarge with no allocation), found no codec bug across 3 runs including PROPTEST_CASES=20000 and 50000, and confirmed the skipped properties (frame/pub/ack round-trips, unknown-tag, truncated-Incomplete) are genuinely already covered. No committed regression file, SPDX intact, no em or en dashes, clippy pedantic + fmt clean. Verdict: SHIP.

PR #267, opened 2026-06-07
Complete the segment reaper with time and count retention modes (refs #13, #80)

ELares:
Completes the consumer-safe segment reaper (#80) by adding the remaining two of its three retention modes. Size landed in #261; this adds TIME and COUNT to the same whole-segment, oldest-first, consumer-safe, crash-safe reaper.

## Semantics
- **Time (`max_age_ms`)**: a sealed segment is deletable when its MAXIMUM record timestamp is older than `now - max_age` (i.e. EVERY record in it has aged out). The max (not the min/last) is used because producer timestamps are not necessarily monotonic, so a segment is never deleted while it still holds a record newer than the bound. `now` is read from the Log's clock seam (`self.clock.now_unix_millis()`), so the deterministic simulation drives it (no `SystemTime::now`).
- **Count (`max_messages`)**: while the log's TOTAL durable record count exceeds the bound, delete oldest sealed segments until under it.
- **Composition**: all three bounds compose. A sealed segment is deleted if ANY enabled bound trips (over size OR over count OR aged out), it is the oldest, it is consumer-safe (`segments[1].base_offset <= protect_below_offset`), and it is NOT the active segment. Oldest-first. Each bound at `0` is off; all three at `0` is the default, so #261's behavior and its tests are unchanged.

## Per-segment metadata + recovery recomputation
The reaper needs each sealed segment's RECORD COUNT and MAX TIMESTAMP, tracked O(1) with no hot-path rescans:
- `SegmentWriter` already tracked a running record count; it now also tracks a running MAX timestamp, updated on each `append` (the record header carries `timestamp_ms`).
- `SegmentSlot` gains `record_count` and `max_timestamp_ms`, frozen from the active writer's running totals when the segment is SEALED on a roll.
- At RECOVERY they are recomputed from the streaming `scan_recovery` (`RecoveryScan`/`BodyWalk` gain `max_timestamp_ms`, accumulated record-by-record without materializing payloads), so the reaper behaves identically after a reopen.
- A running `total_record_count` on the `Log` mirrors `sealed_record_bytes` (init at recovery, `+=` on append, `-=` the reaped segment's count on reap), so the count bound is O(1); exposed as `Log::durable_record_count()` / `Engine::durable_record_count()`.

## Accounting cross-check
A reaped segment decrements BOTH `sealed_record_bytes` (by `valid_end - SEGMENT_HEADER_LEN`) and `total_record_count` (by the scan's `record_count`) by exactly that segment's amounts, read from the same streaming recovery scan and cross-checked against the slot's stored metadata. Tests assert both running totals equal a fresh reopen's recomputed values after a reap. The unlink ordering is unchanged from #261: unlink, dir-sync, then in-memory removal and total decrements.

## Clock seam
The age check reads `now` once per reap pass from the Log's clock (`self.clock.now_unix_millis()`), which is the same clock instance the engine wires in, so the engine clock drives `now` with no `SystemTime`/`Instant::now` anywhere. The underflow case (`now < max_age_ms`) is handled by comparing `max_timestamp_ms + max_age_ms < now` (nothing old enough yet).

## API + flags + metric
- `Log::reap(RetentionBounds { max_bytes, max_age_ms, max_messages }, protect_below_offset)` generalizes the reaper; `reap_to_size` is kept as a thin wrapper over it (the #261 size path and tests are unchanged).
- `EngineConfig.max_age_ms` and `EngineConfig.max_messages` (both `0` default); every existing `EngineConfig` literal across the workspace (engine/session/server/client/cli/health/determinism) updated to `= 0`, keeping existing tests unchanged. The post-produce reap passes all three bounds and the min-committed protect offset.
- `serve --max-age-ms <ms>` and `--max-messages <n>` (default `0` = off), parsed like the existing numeric flags, threaded into `EngineConfig`, documented in USAGE. Milliseconds for age to avoid a duration-parser dependency.
- Reuses the existing `ironbus_segments_reaped_total` counter (it counts reaped segments regardless of which bound triggered).

## Tests
- **Storage**: count reaps oldest until under the bound (never the active, never below protect); time reaps a fully-aged segment, a segment with one recent record is NOT reaped (max not min), advancing the clock makes more eligible, recovery recomputes max timestamp identically; each bound independently triggers; all-off reaps nothing; composed bounds still gate on consumer-safety; byte+count accounting after a reap matches a fresh reopen; writer tracks max-not-last and `scan_recovery` reconstructs it.
- **Engine**: count- and time-based reclamation end to end; a slow group blocks reaping below its cursor under the count and age bounds, then unblocks on catch-up.
- **/metrics**: the reaped counter increments under the count bound.
- **CLI**: the two new flags parse, reject non-numerics, and appear in USAGE.

## Verification
`cargo test --workspace` (490 passed, 0 failed), `cargo fmt --all -- --check`, and `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings` all clean. Crash-recovery and determinism gates pass (also at `PROPTEST_CASES=4000`). A Python scan over the changed files reports no em or en dashes.

This completes the #80 segment reaper (size + time + count). Compaction (#83), the disk-full drop-oldest policy (#82), and the below-earliest truncation signal (#84) remain separate.

refs #13, #80


ELares:

An independent review traced the durability-critical core by hand and ran every gate, all correct: the running total_record_count is +1 per append, initialized over every segment at recovery, and decremented by exactly the reaped segment's scan record_count, so it provably equals a fresh reopen (cross-checked by count_retention_after_a_reap_matches_a_fresh_reopen); seal-time and recovery-time per-segment metadata agree (both compute the MAX timestamp, not the last, proven by writer_tracks_the_max_timestamp_not_the_last and assert_scans_agree); the age check is underflow/overflow-safe (saturating_add(max_age) < now) and reads now only from the clock seam (ManualClock in tests); consumer-safety structurally gates EVERY bound (slow-group-blocks tests under both count and age, plus the composed test); composition and off-by-default are correct (all-zero reaps nothing, #261 size behavior unchanged); crash-safe ordering (unlink, dir-sync, then both decrements) is preserved; crash-recovery 21/21 and determinism 1/1 pass. The only findings were doc drift: the ironbus_segments_reaped_total HELP and two engine doc/comment spots still said 'size retention' though the counter now covers all three bounds. Fixed in 3f37eb9 (the byte-bound field docs correctly keep 'size'). Verdict: SHIP.

Merging on green CI plus this review.

PR #269, opened 2026-06-07
Add the disk-full drop-oldest policy and the below-earliest truncation signal (#82, #84)

ELares:
## Summary

Adds the opt-in disk-full **drop-oldest** overflow policy (#82) and the below-earliest **truncation** signal (#84), which together unblock golden-path #133 step 8.

Today the durable-log byte cap (#259, `EngineConfig`/`LogConfig::max_total_bytes`) sheds NEW produces (drop-new) when the cap is reached, and consumer-safe retention (#261/#267) reclaims old sealed segments only when fully consumed by every group. This PR adds the **opt-in** drop-oldest policy: when the cap is hit, instead of rejecting the producer, the engine reclaims the OLDEST sealed segment even if a slow consumer has not consumed it, then accepts the produce. **Default behavior is UNCHANGED** (drop-new shed); every existing `EngineConfig` literal defaults to `DropNew`.

## Policy semantics

- `EngineConfig.disk_full_policy: DiskFullPolicy` (default `DropNew`). Variants: `DropNew` (the existing shed: reject the over-cap produce; default) and `DropOldest` (reclaim space then accept).
- `Block`/`Refuse` are **out of scope** for v1: the README froze `block` as opt-in-only and this PR does not add a producer stall. The enum is `#[non_exhaustive]` so a future block variant is not breaking.
- The policy has no effect unless `max_total_bytes` is set (with no cap, no produce is ever over-cap).

## Forced reap (crash-safety + accounting)

- New storage primitive `Log::reap_oldest_forced() -> Result<Option<ReapOutcome>, StorageError>`: unlinks the OLDEST sealed segment IGNORING consumer-safety (it may delete records below a slow group's cursor), never the active segment, returns `None` when only the active segment remains.
- **Identical crash-safe ordering and dual-accounting to #261/#267**: unlink -> dir-sync (removal durable) -> in-memory slot removal + exact `sealed_record_bytes`/`total_record_count` decrement (read as `valid_end - SEGMENT_HEADER_LEN` and `record_count` from a streaming scan, the same terms the totals accumulated). A crash before the in-memory update leaves disk and memory consistent; a crash after leaves a shorter contiguous chain recovery already accepts. After a force-reap the running totals still equal a fresh reopen's recomputed values.
- In `Engine::produce` under `DropOldest`: on the `AtCapacity` condition, loop — first the normal consumer-safe reap (no data loss if retention is also configured), then `reap_oldest_forced()` if still over cap, then retry the append. **Bounded**: if only the active segment remains, fall back to the drop-new rejection (the wedge guard) so a single oversized in-flight set cannot wedge the log empty. Each forced reap increments the new `segments_force_reaped` counter, exposed on `/metrics` as `ironbus_segments_force_reaped_total`. Under `DropNew`, behavior is exactly as today (no forced reap).

## earliest_retained + truncation-once (#84)

- `Engine::earliest_retained_offset()` = the oldest segment's base_offset (0 when nothing reaped), backed by `Log::earliest_offset()`.
- On poll, BEFORE delivering, if the group's cursor is BELOW `earliest_retained_offset()` (its data was force-reaped out from under it): reset the cursor UP to earliest-retained (resuming at the oldest record still present, dropping the now-meaningless acked-ahead set and in-flight leases), and surface a new `Poll::Truncated { earliest_retained, skipped }` ONCE. After the reset `committed == earliest`, so a subsequent poll is no longer below earliest and never re-truncates the same gap. The next poll delivers normally.

## New wire frame (FROZEN, tag 18)

- `FrameType::Truncated` = tag **18** (the next free integer), pinned by the frozen-tag test; `ALL_TYPES` and the unknown-tag proptest range (`19u8..=255`) updated.
- Body: `earliest_retained: u64 LE`, then `skipped: u64 LE` (fixed 16 bytes). `encode_truncated`/`decode_truncated`; a short or overlong body is a typed `BodyError`, never a panic.

## Client + CLI surfacing

- `Client::fetch` returns a new `Fetch.truncations: Vec<Truncation>` (like `dead_letters`). The credit bound caps total frames (deliveries + dead-letters + truncations).
- The session emits one `Truncated` frame per gap; it consumes one credit slot.
- `ironbus sub` prints `truncated: resumed at offset <n>, skipped <k> record(s)`.
- `serve --disk-full-policy <drop-new|drop-oldest>` (default drop-new), validated as a usage error on a bad value.

## Tests

- **Storage**: `reap_oldest_forced` deletes the oldest sealed segment below any protect offset, never the active, updates both totals exactly (cross-checked vs a fresh reopen), crash-safe (power loss keeps it gone, reopen recovers the remaining chain); `earliest_offset` rises after a reap; repeated forced reaps stop at the last active segment.
- **Engine**: with `DropOldest` + the cap + a STUCK consumer, producing past the cap force-reaps its old segments, the force counter increments, and the produce SUCCEEDS (not rejected); under `DropNew` the same scenario rejects and reaps nothing (data preserved). The stuck consumer's NEXT poll returns a truncation (resets to earliest-retained), then delivers from there; a second poll does NOT re-truncate. A caught-up consumer never truncates. The wedge guard falls back to drop-new when only the active segment remains.
- **Proto**: `Truncated` body round-trips and rejects a short/overlong body; the frozen-tag test pins tag 18; the unknown-tag proptest moves past 18; the arbitrary-bytes never-panic sweep covers it.
- **Session/wire**: a truncated consumer gets exactly one `Truncated` frame, the connection stays open, a later fetch delivers without re-truncating.
- **Client**: `fetch` surfaces a truncation from a scripted batch, and a truncation counts against the credit bound.
- **/metrics**: `ironbus_segments_force_reaped_total` renders and increments.

## Verification

- `cargo test --workspace`: 512 passed, 0 failed (includes the determinism and crash-recovery gates, unchanged).
- `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings`: clean.
- Em/en-dash scan over all changed files (incl. CHANGELOG): none.

The golden-path #133 step-8 e2e test over the `ironbus` binary is a follow-up.

refs #82, #84


ELares:

An independent review traced the dangerous paths by hand and ran the full suite, all 8 categories clean. The cardinal safety property holds: the committed < earliest truncation gate is the UNCONDITIONAL first step of the single delivery path (poll_in, which every poll funnels through), so a consumer can never be silently advanced past force-reaped data; after the reset committed == earliest so the same gap never re-truncates, and the next poll delivers normally (verified by a_truncated_consumer_gets_exactly_one_truncation_then_delivers and a_caught_up_consumer_never_sees_a_truncation). DropNew is provably byte-for-byte unchanged (the determinism golden-image test passes, the field is purely additive, all 16 existing literals default to DropNew). reap_oldest_forced uses identical accounting and crash-safe ordering to the consumer-safe reaper (unlink, dir-sync, then decrement both running totals), cross-checked against a fresh reopen plus a power-loss test. The wedge guard provably terminates (each iteration removes a sealed segment, falls back to drop-new when only the active remains). Tag 18 is the next free integer with both frozen-tag tests and the unknown-tag range updated, a fixed 16-byte bounds-safe body. The client credit caps deliveries + dead_letters + truncations. 561 tests pass, crash gates 21/21, determinism 1/1, clippy pedantic + fmt clean, ironbus-core untouched, no em or en dashes. Two non-blocking NITs (an advisory skipped count that matches its documented earliest - old_cursor definition, and an unreachable defensive check). Verdict: SHIP.

PR #270, opened 2026-06-07
docs: correct stale README compression row (lz4_flex default, zstd opt-in) per #139

ELares:
## What

The README is the canonical vision source, but the #139 coherence audit found its **compression** subsystem row is stale. It still described the OLD position (zstd default, lz4 fallback) instead of the frozen #139 decision (lz4_flex default, zstd opt-in only). This PR corrects only that verified factual staleness.

## Before / after (README compression row)

Before:

```
| Compression | [#12](...) | zstd default, per-batch, lz4 fallback, dictionaries deferred to a later milestone |
```

After:

```
| Compression | [#12](...) | lz4_flex default (pure Rust), per-batch; zstd opt-in only behind a feature, never on the default path; dictionaries deferred to a later milestone (#139) |
```

## Source the correct decision was verified against

The frozen #139 decision was confirmed in three independent places, all in agreement:

- `deny.toml` `[bans]` comment: "Per #139, the default path is pure Rust ... The default compression codec is lz4_flex (pure Rust); zstd (zstd-sys, vendored C) is opt-in only, behind an explicit feature, never on the default path."
- `docs/adr/0003-default-compression-lz4-zstd-opt-in.md` (Status: Accepted): "The default compression codec is `lz4_flex`, which is pure Rust. zstd (via `zstd-sys`, vendored C) is opt-in only, behind an explicit feature, and is never on the default path." The ADR explicitly flags the old README row ("zstd default, lz4 fallback") as the text it refines.
- Keeping zstd off the default path keeps the default static musl binary pure Rust (no vendored C).

## Other potential contradictions scanned (per the audit brief)

I scanned the README for other concrete, verifiable contradictions with frozen decisions or the code. All other claims I checked are already consistent, so I changed nothing else:

- **Segment / record sizes** (README: "64 MiB segments (8 MiB on the edge profile)", "max record 16 MiB ... up to 1 GiB"): verified against `crates/ironbus-core/src/format.rs` (`DEFAULT_SEGMENT_BYTES = 64 MiB`, `EDGE_SEGMENT_BYTES = 8 MiB`, `DEFAULT_MAX_RECORD_BYTES = 16 MiB`, `MAX_RECORD_BYTES_CEILING = 1 GiB`). Consistent, left unchanged.
- **Log-is-WAL phrasing** (README: "the active segment IS the write-ahead log: there is no separate WAL file"): verified against `docs/adr/0001-log-is-wal.md`. The README is consistent everywhere (diagram, #4 row, decisions table); no "separate WAL" contradiction exists. Left unchanged.

Per the conservative brief, only the verified-stale compression row was edited; vision and tone are untouched.

## Verification

- Dash scan (python over README.md + CHANGELOG.md): no U+2013/U+2014. PASS.
- `git status`: only `README.md` and `CHANGELOG.md` changed.
- CHANGELOG: terse bullet added under the existing `### Changed` in `[Unreleased]`.

refs #139, #12


ELares:
### Independent verification (orchestrator, not the author)

Verified the change is accurate and minimal: the only README line changed is the #12 compression row, from the stale 'zstd default, lz4 fallback' to 'lz4_flex default (pure Rust), per-batch; zstd opt-in only behind a feature, never on the default path; dictionaries deferred (#139)'. This matches the frozen decision in deny.toml line 23 ('compression codec is lz4_flex (pure Rust); zstd ... opt-in only') and ADR 0003 ('lz4_flex is the default compression codec, zstd is opt-in only'). No other README row changed; the agent conservatively confirmed segment sizes and the log-is-WAL phrasing are already consistent. No em or en dashes introduced. SHIP.

PR #271, opened 2026-06-07
Add the golden-path #133 step-8 drop-oldest truncation e2e test

ELares:
## Scenario (#133 step 8, over the real binary)

The disk-full drop-oldest policy + truncation signal landed in #269. This adds the step-8 acceptance test `golden_path_drop_oldest_truncates_a_stuck_consumer` to `crates/ironbus-cli/tests/golden_path.rs`, driving the real `ironbus` binary end to end:

1. Boot `serve --disk-full-policy drop-oldest --max-segment-bytes 4096 --max-total-bytes 7000` (plus the helper's `--health-addr 127.0.0.1:0` and `--checkpoint-interval 1`).
2. Produce one record, consume ONLY it on group `stuck` and ack it, so the group's committed cursor is stuck at offset 1.
3. Produce 60 more records. Under drop-oldest the byte cap is hit repeatedly and the oldest sealed segments are force-reaped (including records the stuck group never consumed).
4. The stuck consumer fetches again: it gets exactly one truncation, resets up to the earliest retained offset, and resumes there.
5. A follow-up fetch on the same group does not re-truncate (the gap is closed).

## Cap / segment sizing (forces the reap deterministically)

- `--max-segment-bytes 4096` (the CLI floor). Each record carries a 956-byte payload, empty key/headers, for ~1000 record bytes on disk (36-byte header + 956 payload + 8-byte trailer), so a segment seals after ~4 records.
- `--max-total-bytes 7000` is above one sealed segment's record bytes, so at least one segment seals before the cap engages and a sealed segment is always available to force-reap (avoids the active-only wedge-guard fall-back to drop-new).
- Producing 60 more records makes force-reaping march the earliest-retained offset well past the stuck cursor on any timing. Empirically the resume offset lands around 55 with the force counter at 11.

## Assertions (exact where stable, relationships where timing-dependent)

- **No shed:** every one of the 60 produces exits 0 and lands at the next contiguous offset (a drop-NEW broker would shed and exit non-zero), proving the drop-oldest path engaged and nothing was lost.
- **Force happened:** `/metrics ironbus_segments_force_reaped_total` > 0.
- **Truncation once:** the resume fetch's stdout has exactly one `truncated: resumed at offset <n>, skipped <k> record(s)` line (`truncation_resume_offsets(...).len() == 1`), never zero or repeated.
- **Resume offset:** `<n>` > 1 (the stuck cursor), proving the reaped span `[1, n)` was skipped; and the FIRST delivered record's offset equals `<n>` exactly (it resumes at the earliest retained record, did not silently skip past it).
- **Progress + contiguity:** at least one record delivered after the truncation, the delivered offsets are contiguous from `<n>`, and the `fetched N message(s)` count matches.
- **No re-truncation:** a further fetch on the same group has zero `truncated:` lines (the one-time reset closed the gap).

The first, caught-up fetch (consuming offset 0) also asserts zero truncations, so the truncation assertions are non-vacuous (it only fires after the reap).

## Non-vacuity

The force-reap counter check and the no-shed check anchor the truncation: without an actual force-reap there would be no truncation to observe. The resume offset is tied to both the truncation advisory and the first delivered record, so a stub that printed a truncation without resetting the cursor would fail.

## Verification

- `cargo test -p ironbus-cli --test golden_path` passes; the new test was run 3x in isolation and the full suite 3x, all green (no flake) over the spawned broker.
- `cargo clippy -p ironbus-cli --all-targets -- -W clippy::pedantic -D warnings`: clean (the produce loop was extracted into a `produce_contiguous` helper to stay under the line cap; no `#[allow]`).
- Dash scan (em/en) over the two touched files: none.
- Only `crates/ironbus-cli/tests/golden_path.rs` and `CHANGELOG.md` changed; no production code. The module-doc remaining-steps list now reads "Only the installer and in-place upgrade steps, which need a release, remain."

refs #133, #82, #84


ELares:
### Independent review

An independent review confirmed the reap is forced deterministically (the 7000-byte total cap is well above ~4 records x ~1000 bytes per sealed segment, so a sealed segment always exists to force-reap below the stuck cursor; no active-only wedge fall-back), and re-ran the test 5x plus the full suite 3x with no flake/hang/leak. All assertions are non-vacuous and load-bearing: exactly one truncation line on the resume fetch, resume offset above the stuck cursor (a span was skipped), delivered[0] == resume (resumed exactly at earliest-retained, the no-silent-skip property), contiguous delivery after, no second truncation on the follow-up (gap closed), /metrics force-reap counter > 0, and all 60 produces exit 0 (drop-oldest engaged, not drop-new shed). Scope is test plus CHANGELOG only, SPDX intact, module doc updated, no em or en dashes, clippy pedantic + fmt clean. Verdict: SHIP.

PR #272, opened 2026-06-07
Add the durable DLQ sink and the crash-atomic exactly-once move (#63, #9)

ELares:
## What and why
Today a poison message (one that exceeds `MaxDeliver`) is only PARKED: the engine commits the group cursor past it, emits the in-band dead-letter advisory (#253), and counts `ironbus_dead_lettered_total`. The poison is then gone forever, never inspectable. This PR adds the missing durable DLQ **sink**: the poison message is written to a durable, recoverable dead-letter store, moved crash-atomically and **exactly once**.

## The DLQ sink design (reused segment Log)
The sink is a SECOND segmented `Log` rooted at a `dlq/` subdirectory of the data directory, via a new `Filesystem::subdir` (and a non-creating `subdir_exists`) implemented for the in-memory backend (a shared-store key prefix, so it shares the parent's power-loss image), `StdFs` (a real child directory), and a `FaultFs` passthrough that shares the same `FaultControl`. So DLQ records use the EXACT framed, CRC32C'd, recoverable segment format as the main log and are read back by the SAME `SegmentReader` / `OfflineReader`, with no second on-disk format.

### Dead-letter metadata encoding
Each DLQ record preserves the ORIGINAL record verbatim: its `key`, `payload`, and original enqueue `timestamp_ms` are the DLQ record's key, payload, and timestamp. The dead-letter metadata is a fixed, self-describing prefix of the DLQ record's `headers` blob, ahead of the original headers:

| bytes | field |
|---|---|
| `[0,4)` | magic `DLQ1` |
| `[4,12)` | source offset (u64 LE) |
| `[12,16)` | attempt / deliveries (u32 LE) |
| `[16,18)` | group name length g (u16 LE) |
| `[18,20)` | original headers length h (u16 LE) |
| `[20,20+g)` | group name |
| `[20+g,20+g+h)` | original headers |

## The crash-safe move and EXACTLY-ONCE idempotency
On dead-letter of `(group G, source offset O, attempt A)`:
1. **APPEND** the poison record to the DLQ sink and **FSYNC** it.
2. **THEN** commit group G's cursor past O (the existing park-commit).

The crash window is between (1) and (2): if the process dies after the DLQ fsync but before G's cursor commit becomes durable, on recovery the source message is still uncommitted in G, so it redelivers and is re-evaluated as poison and would re-append. We make the append IDEMPOTENT by the key `(group, source offset, attempt)`: a durable **per-group highest-dead-lettered-source-offset** high-water mark, **rebuilt at open from the DLQ records themselves** (the sink is the single source of truth, no sidecar to keep consistent). A `(group, offset)` at or below the group's high-water mark is already in the sink, so it is committed-past WITHOUT a second DLQ write. Source offsets only rise per group, so the high-water mark is a sound dedupe key. Net: the message ends up in EXACTLY ONE place across any crash, never double-written and never dropped.

The sink is opened LAZILY (first dead-letter), or eagerly by `Engine::open` when the `dlq/` subdir already exists (so the high-water mark is present before the first poison redelivers). A broker that never dead-letters never creates `dlq/`.

### The crash-window test
`a_crash_between_the_dlq_append_and_the_source_commit_yields_exactly_one_dlq_entry`: with a high `checkpoint_interval` (so the park's cursor advance is not durable), a message is poisoned (DLQ record fsynced, cursor advanced only in memory), then `simulate_power_loss` reverts the un-checkpointed source cursor while the fsynced DLQ record survives. On reopen the poison redelivers, is re-poisoned, and is committed-past with NO duplicate append. Asserts exactly one DLQ entry for `(group, offset, attempt)` and no loss. `a_second_crash_redelivery_still_adds_no_duplicate` repeats the cycle twice.

## Offline inspection
`ironbus dump --dlq --data-dir <dir>` streams the sink's dead-letter records (dlq_offset, source_offset, group, attempt, ts_ms, key/payload sizes; NDJSON with `--json`), reusing the read-only `OfflineReader`. It never mutates the directory (proved by a byte-for-byte tree snapshot before/after) and never creates `dlq/`. An empty or never-poisoned broker shows nothing. Frozen exit codes apply (a missing data dir is still not-found / exit 2).

## Metric
`ironbus_dlq_records_total` on `/metrics`: the DLQ depth (records present at open plus every append since), rebuilt from the durable sink so it survives a restart. The existing advisory and `ironbus_dead_lettered_total` are unchanged.

## Tests
- Durably written to the DLQ (reopen shows it with the original timestamp, group, and attempt) and committed-past in the source group.
- Crash mid-move yields exactly one DLQ entry and no loss; a second crash adds no duplicate.
- API-level idempotency: re-poisoning the same `(group, offset, attempt)` does not double-write.
- Offline `dump --dlq` read-only, empty-DLQ, and missing-dir (not-found) cases.
- `/metrics` DLQ counter renders and increments.
- The no-poison path never creates or touches `dlq/`.
- `Filesystem::subdir` / `subdir_exists` isolation, shared power-loss image, and unsafe-name rejection (in-memory and `StdFs`).

## Verification
`cargo test --workspace` (531 passed, 0 failed), `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings` clean. Crash-recovery and determinism gates pass unchanged. A python scan over the changed files reports no em/en dashes. SPDX on the new file.

## What remains of #63
The BackOff schedule, MaxDeliver config validation, and cumulative-ack rejection remain separate. The single-fsync group-commit optimization ("one durable consumer-state event under a single group-commit fsync") is a follow-up; the append-DLQ-fsync THEN commit-cursor ordering with the idempotency key already yields exactly-one-place.

refs #63, #9


ELares:

An independent review traced every crash point in the dead-letter move and confirmed the cardinal property (exactly one place across any crash) holds. append_poison does log.append then log.sync (fsync) BEFORE the source cursor is advanced, and the cursor advance is durable only via the separate checkpoint path, so a power loss after the DLQ fsync but before the cursor checkpoint reverts the cursor while the DLQ record survives; Engine::open eagerly opens the sink and rebuilds the per-group highest-dead-lettered-offset HWM BEFORE any poll, so the re-poison hits the suppression path, which skips the append AND still commits the source past (engine.rs:1135), so there is no double-write, no drop, and no infinite loop. The HWM rebuild decodes (group, source_offset) from every DLQ record via the same codec the writer uses, per-group max, multi-group safe, no off-by-one. The crash test is valid: InMemoryFs::subdir shares the parent's power-loss image (proven by a_subdir_shares_the_power_loss_image_with_the_parent), so the test exercises the real window and asserts exactly one entry (==1). The no-poison path never creates dlq/. The metadata headers prefix round-trips and decodes bounds-safe (typed skip on malformed, never a panic). C: Clone is satisfied by every shipped clock (atomic-load Clone, ironbus-core stays IO-free). dump --dlq is read-only (before/after dir snapshot byte-identical). 532 tests pass, crash gates + determinism unchanged, clippy pedantic + fmt clean, the #253 advisory + ironbus_dead_lettered_total intact. Two cosmetic NITs only. Verdict: SHIP, no blockers.

Issue #114, opened 2026-06-05
[TASK] Baseline comparison rig and CI regression gate


ELares:
Parent: #19

Build the apples-to-apples baseline comparison rig and the CI regression gate. The rig runs peer systems under identical workloads with durability modes explicitly matched and labeled, and the gate fails CI on real drift while tolerating the inherent noise of edge hardware.

## Approach
Baselines run on the same device, same message size, and same durability label as IronBus: NATS JetStream File (MaxAckPending=1000, default FileStore block sizes), Redis Streams in both appendfsync=everysec and appendfsync=always (both reported), and Mosquitto QoS 1 as primary with QoS 0/2 reported. Kafka and Redpanda are excluded from edge SLO gates (they are JVM/Seastar multi-node systems) and may appear only in an x86-ref informational appendix labeled `not an edge-class comparison`. Every comparison row must carry matching durability labels on both sides or the report build fails (schema lint), so we never compare our durable number against a peer's page-cache number. Also report Little's-Law-derived queue occupancy (L = lambda x W) at target throughput to show p99 stays within SLO at the chosen concurrency bound. Regression gate: compute a 7-day rolling median per device and fail CI when the throughput median drops > 10 percent, the p99 median rises > 15 percent, or the p99.9 median rises > 25 percent versus the last released tag; runs whose warm-up CoV check failed are advisory-only and a human ratifies edge regressions before they block a release.

## Acceptance criteria
- [ ] NATS JetStream File, Redis Streams (everysec and always), and Mosquitto (QoS 1 primary) run under identical workloads.
- [ ] Durability labels matched on both sides; mismatched comparison fails the report build.
- [ ] Kafka/Redpanda confined to an x86-ref informational appendix, clearly labeled.
- [ ] Little's-Law queue occupancy reported at target throughput.
- [ ] Rolling-median regression gate (10 percent throughput / 15 percent p99 / 25 percent p99.9) wired into CI.
- [ ] Noisy runs flagged advisory; edge regressions require human ratification.

## Failure considerations
The central trap is mislabeled durability turning a comparison into marketing; the both-sides label-match lint blocks it. Single-run percent gates would flap on noisy edge CI and train people to ignore them, so the gate uses a rolling median with wider p99.9 tolerance and an advisory escape hatch for warm-up-failed runs. Including cluster-class brokers as edge gates would be dishonest, so they are appendix-only.

ELares:
[...]

Issue #112, opened 2026-06-05
[TASK] criterion micro-bench suite for hot-path components


ELares:
Parent: #19

Build the `criterion` micro-benchmark suite that measures the per-component hot paths in isolation: WAL append, CRC32C checksum, record decode, and the zero-copy slice handoff that fans one buffer out to the durable log and every consumer. These complement (do not replace) the macro-bench: they catch per-component regressions early and localize a slowdown the end-to-end number only shows in aggregate.

## Approach
Use `criterion` because it handles warm-up, outlier detection, and statistically sound comparison for us. Bench each component against the record format from #5 and the log layout from #4: WAL append at representative batch sizes (1, 16 KiB, 1 MiB), CRC32C using the hardware instruction (and a portable fallback path, benched separately so the ARM target is honest), record decode, and the `bytes::BytesMut` freeze-to-`Bytes` plus O(1) refcounted slice handoff. Report per-message CPU cost in microseconds (informational, regression-tracked, not a hard gate per the parent). Run the suite on the target ARM cores, not only the x86 dev box, since absolute throughput on x86 lies about edge cost. Keep results in the same versioned archive format so micro and macro regressions are tracked side by side.

## Acceptance criteria
- [ ] criterion benches exist for WAL append, CRC32C, record decode, and zero-copy slice handoff.
- [ ] Warm-up and outlier handling are documented and use criterion defaults.
- [ ] Hardware-CRC and portable-CRC paths are benched separately.
- [ ] Per-message CPU microseconds are reported and regression-tracked.
- [ ] Suite runs on at least one ARM reference core, not only x86.
- [ ] Results archived in the shared versioned format.

## Failure considerations
The trap is overfitting a hot loop to a microbenchmark that does not reflect end-to-end behavior; this suite is explicitly scoped as a localization tool and the SLO gates live in the macro-bench. Benching CRC or compression only on x86 would hide ARM cost, so ARM runs are mandatory. In-process measurement cannot model real fsync cost or multi-consumer fan-out, which is exactly why this is paired with the macro-bench rather than trusted alone.

ELares:
[...]

PR #273, opened 2026-06-07
Add a criterion micro-bench suite for the IO-free hot-path codecs (#112)

ELares:
Establishes the criterion micro-bench infrastructure for `ironbus-core`'s IO-free hot-path codecs. These benches are **informational**: they run on demand via `cargo bench -p ironbus-core`, and are deliberately **NOT a per-PR CI gate** (not added to any workflow). The regression gate is tracked separately in #114 and is out of scope here.

Scope is intentionally narrow to avoid conflicting with in-flight server/storage work: this touches ONLY `crates/ironbus-core/Cargo.toml`, the new `crates/ironbus-core/benches/`, and `CHANGELOG.md`. No production code in any crate changes; `ironbus-core` stays IO-free.

## Benches added

`benches/codec.rs` (record frame codec):
- `codec/encode` of a representative record into a reused buffer, two cases: a small ~64-byte payload and a larger ~4 KiB payload (`Throughput::Bytes` = payload size).
- `codec/decode` of a pre-encoded frame (same two payload sizes).
- `codec/decode/decoded_len` the header-only length helper.

`benches/cursor.rs` (`AckCursor`):
- `cursor/ack_in_order`: a stream of strictly in-order acks (the watermark advances one at a time, the ahead set never grows).
- `cursor/ack_out_of_order`: ack all odd offsets first (the acked-ahead set grows to ~N/2 isolated runs), then fill the even gaps (runs merge and the watermark collapses), exercising the insert search/merge and advance under a deep ahead set.
- `cursor/snapshot`: `encode_snapshot` and `decode_snapshot` round-trip over a fragmented cursor (`Throughput::Bytes` = snapshot size).

All inputs are built from fixed values (no ambient randomness, deterministic runs) and wrapped in `criterion::black_box`. Bench groups have clear names.

## Wiring
- `criterion = "0.5"` added to `[dev-dependencies]` (criterion 0.5.1 resolved).
- Two `[[bench]]` tables with `harness = false`.
- Not added to CI.

## cargo-deny
`cargo deny check` PASSES (advisories ok, bans ok, licenses ok, sources ok), exit 0, with no change to `deny.toml`. Criterion's transitive tree introduces only `MIT`, `Apache-2.0`, and `Unicode-3.0` licenses, all already in the allow-list; `bans` reports no new policy-violating duplicate. The only deny output is informational `license-not-encountered` warnings for allow-listed licenses no crate happens to use.

## Verification
- `cargo build --workspace`: ok
- `cargo test --workspace`: ok (all suites pass)
- `cargo bench -p ironbus-core --no-run`: ok (both bench binaries compile; a short sample run executes cleanly, no panics)
- `cargo clippy -p ironbus-core --all-targets -- -W clippy::pedantic -D warnings`: clean (benches are an `--all-targets` target)
- `cargo deny check`: ok (no policy relaxation)
- em/en dash scan over changed files: none

refs #112


ELares:
### Independent review

An independent review confirmed the two key risks are clean. The benches are real and non-vacuous: they call the current public APIs (codec encode/decode/decoded_len, AckCursor ack/snapshot) with deterministic black-box'd inputs, and short runs show genuinely distinct hot paths (decode@4096 ~222 ns vs decoded_len@4096 ~8.5 ns, in-order vs out-of-order cursor acks measurably diverge, snapshot round-trips), both bench binaries compile and run without panic. cargo-deny passes at exit 0 with deny.toml UNTOUCHED; the Cargo.lock diff is purely additive (criterion's transitive tree, all licenses already allow-listed, no version bumps or removals of existing deps). Scope is the two new bench files plus core Cargo.toml, CHANGELOG, and Cargo.lock; benches are informational (not in CI). SPDX on both bench files, clippy pedantic + fmt clean, no em or en dashes, core stays IO-free. Verdict: SHIP, no findings.

Issue #65, opened 2026-06-05
[TASK] Per-consumer credit and in-flight window with redelivery accounting


ELares:
Parent: #10

Specify the credit-based delivery mechanism that bounds how many messages and bytes each consumer can have in flight, so one consumer's occupancy is provably bounded and isolated from peers. This is the core flow-control primitive every overflow policy and slow-consumer action depends on, mirroring MQTT Receive Maximum and JetStream MaxAckPending and exposed on the wire by #11.

## Approach
Each consumer has an independent cursor, pending set, and send quota. Defaults: `in_flight_window` = 64 messages and `in_flight_bytes` = 8 MiB per consumer; effective credit = `min(message credits remaining, byte credits remaining)`, hard floor 1, no representable unbounded value. The broker decrements quota on send, restores on ack (#9 owns the cursor and ack semantics), and stops sending at zero (warn-then-pause). Effective reliability is clamped to `min(producer intent, consumer capability)`. A redelivered message re-occupies exactly one credit and its bytes, paced by the same window; a `max_deliver` cap (default 5) routes poison records to the DLQ rather than holding a credit forever. Provide the worked Little's Law justification (10k msg/s x 5 ms service = 50 in-flight, so 64 gives ~28% headroom) and hand the byte budget to #20 for the RAM-ceiling check.

## Acceptance criteria
- [ ] Advertisement, decrement-on-send, restore-on-ack, and behavior-at-zero are fully specified.
- [x] Effective reliability defined as `min(producer, consumer)` with a worked example.
- [x] Redelivery consumes exactly one credit; `max_deliver` (default 5) routes to DLQ.
- [x] Per-consumer isolation proven on paper: a stuck consumer cannot touch a peer's quota, memory, or progress.
- [ ] Both message-count and byte budgets enforced; no `request(MAX)`-style opt-out is representable.
- [ ] Little's Law worked example committed and cross-checked with #19.

## Failure considerations
A poison message that always naks could inflate occupancy past the bound or live-lock a credit: counting redelivery as one credit plus `max_deliver` to the DLQ bounds both. A large-message consumer could blow RAM despite a small message count: the parallel byte budget caps it. A consumer that stops acking pins its own credits only; isolation keeps peers unaffected.

ELares:
Core delivered in #274 (merged, squash `ed6038a`).

**Delivered (boxes ticked):**
[...]

PR #274, opened 2026-06-07
Per-consumer credit and in-flight window with redelivery accounting (#65)

ELares:
Implements #65: a per-CONSUMER (per-connection) in-flight credit window with redelivery accounting, the consumer-side half of credit-based flow control (parent #10), so in a competing group one slow consumer cannot consume a peer's in-flight budget (the per-consumer isolation from #10). Today's bound is per-GROUP (`max_in_flight`); this composes a per-CONNECTION bound WITH it.

## Credit semantics (documented in rustdoc on `EngineConfig::consumer_credit` and `Session`)

- **Advertisement / default**: new `EngineConfig.consumer_credit` (the standing maximum un-acked messages one connection may hold), default **64** (NOT the MQTT absent-value 65535; memory-justified by Little's Law: 10k msg/s x 5 ms = 50 in-flight, so 64 leaves ~28% headroom). Floored to 1 by the engine. New `serve --consumer-credit <n>` flag (validated `>= 1`); the CLI default agrees with the engine default.
- **Decrement on send**: each Flow delivery occupies one of the connection's slots. The accounting is **derived** from the connection-scoped `leased` set (#175) rather than a separate counter, so it can never drift: `remaining = ceiling - leased.len()`.
- **Restore on ack / nack / term / expire**: an ack, a successful nack/term, the per-batch prune of committed offsets, and the start-of-Flow release of leases the engine no longer holds ACTIVELY all REMOVE from `leased`, restoring credit.
- **Behavior at zero**: a consumer at zero remaining credit gets zero deliveries from a Flow until it frees a slot, even if the group has messages and other consumers are draining.
- **Effective bound**: a Flow delivers `min(requested_credit, ceiling - already_held, whatever the group makes available)`, so the effective reliability bound is `min(producer-side group window, consumer credit)`. When the group window is smaller it stays the binding bound (no regression to the per-group window).

## Per-consumer isolation (the core property)

The credit and the `leased` set are per CONNECTION. One stuck consumer that fills its ceiling and stops acking pins ONLY its own slots; it never touches another connection's `leased` set or remaining credit, so it cannot reduce a peer's available deliveries.

**On paper**: consumer A's deliverable count is `ceiling_A - |leased_A|`; nothing A does mutates `leased_B`, and B's count `ceiling_B - |leased_B|` is independent of A. A's held leases are A's alone.

**In test**: `one_stuck_consumer_does_not_starve_a_peer_in_the_same_group` (session, deterministic manual clock) and `one_stuck_consumer_does_not_starve_a_peer_over_a_real_server` (client, end to end over TCP): two connections in the same competing group, each credit 2; A fills its credit and goes stuck, B still receives its full credit of 2 and keeps draining while A holds its slots; A and B hold disjoint offsets.

## Redelivery accounting

When one of a session's leases EXPIRES, the engine redelivers the message to whoever next polls under a fresh generation. New `LeaseTable::holds_active(token, now)` / `Engine::holds_active_lease_in(group, token)` treat an expired lease as inactive even when the generation still matches. At the start of each Flow, `Session::release_stale_leases` drops every `leased` entry the engine no longer holds ACTIVELY, which FREES the original consumer's slot; the redelivery is counted against whoever next claims it (the same or another consumer). A redelivery to the SAME consumer re-occupies exactly one slot (the offset key is overwritten), not two. No message is lost or double-counted; at-least-once is preserved.

**Tests**: `an_expired_lease_frees_the_original_consumer_and_is_recounted_on_redelivery` (A's expired lease redelivers to B, A's slot frees, A's stale token is fenced, B commits exactly once) and `a_redelivery_to_the_same_consumer_still_counts_as_one_slot`.

## Server-enforced vs wire-negotiated (and follow-ups)

This is the **server-enforced per-connection default** slice #65 allows as a first slice: the credit ceiling is a config/`serve` value the Flow respects, enforced per session. Deferred follow-ups: **wire-negotiating** the credit in the Connect/Info handshake (a capability), and the **parallel per-consumer BYTE budget** (`in_flight_bytes`). The `max_deliver`-to-DLQ poison cap already exists (`DeliveryConfig` + #63).

## Tests added
- `ironbus-core` lease: `holds_active_is_true_only_for_a_live_unexpired_owning_token`.
- `ironbus-server` session: cap at credit (huge requested capped), restore on ack, nack+term restore, per-consumer isolation, expire-and-redeliver accounting, same-consumer redelivery, effective `min(group, credit)`, default single-consumer path unchanged.
- `ironbus-client` end to end over a real server: fetch capped at credit, two-consumer isolation.
- `ironbus-cli`: `serve --consumer-credit 0` rejected as a usage error.

## Constraints / verification
- `cargo test --workspace`: all 15 test binaries pass (golden_path, determinism, crash_recovery unchanged).
- `cargo fmt --all -- --check`: clean. `cargo clippy --workspace --all-targets -- -W clippy::pedantic -D warnings`: clean.
- No unwrap/expect/panic in library paths; typed errors; `ironbus-core` stays IO-free; clock routed through the seam (`holds_active_lease_in` reads `now_monotonic`). Every `EngineConfig` literal updated and defaulted so existing tests are unchanged.
- Em/en-dash scan over changed files: none. No new .rs files (SPDX unchanged).

refs #65, #9, #10


ELares:
## Review resolution (SHOULD-FIX + NIT addressed) — b2965ae

The prior review's SHOULD-FIX (with `consumer_credit` defaulting to 64, `ironbus sub --max 100 --ack` silently capped at 64 because `cmd_sub` did a single `client.fetch(max)`) and NIT (hand-copied `DEFAULT_CONSUMER_CREDIT` literal) are fixed:

- `cmd_sub` re-fetches up to `--max` only for cursor-advancing dispositions (`Ack`, `Term`); acking/terming frees credit so the loop drains. `Peek`/`Nack` take a single window-bounded batch (re-fetching them only re-serves the same records, and an immediate nack would redeliver in a storm).
- `the_in_flight_window_bounds_each_delivery_batch` now drives the Flow layer directly (a raw `Client`), since `--ack` correctly drains across batches now; the per-batch window cap is asserted where it is observable (`fetch(10)` with credit 64 returns 2 = the `max_in_flight` window, non-vacuous).
- CLI `DEFAULT_CONSUMER_CREDIT` aliases `ironbus_server::engine::DEFAULT_CONSUMER_CREDIT` (single source of truth, full-path so the non-unix build is unaffected).

**Independent adversarial review (not the author): SHIP.** It traced the highest-risk item, a possible `max - total` u32 underflow, and proved it impossible at two independent layers: the server caps `delivered <= want` (advisories are separate frame types, never counted in `messages`), and the client `over_credit` guard rejects any over-limit batch before pushing. So `got <= want = max - total` every iteration and the subtraction never goes negative. Nack-storm prevention, infinite-loop safety, the non-vacuous window test, and the non-unix const path all verified. No blockers, no nits.

CI: all 15 checks green (incl. windows-latest, exercising the non-unix const path). Local: 544 workspace tests pass, clippy `-D warnings` clean, golden_path stable across 3 runs.

PR #276, opened 2026-06-07
Add the xxh3-64 large-payload checksum to the record frame (#146)

### LINKED ISSUE #146: [TASK] Place the xxh3-64 large-payload checksum in the frozen record frame

ELares:
## What and why

#5 froze the record trailer as exactly 8 bytes (`body_crc: u32` then `total_len: u32`) with no room for a second checksum, while #8 requires an independent xxh3-64 for large payloads stored in the record framing. This places that checksum without disturbing the frozen trailer.

## Frame layout decision

- A new header flag bit `RecordFlags::HAS_XXH3` (`0b0000_0100`) signals the presence of an extra checksum field.
- When set, an 8-byte little-endian xxh3-64 field is placed IMMEDIATELY BEFORE the 8-byte trailer, and it IS included in `total_len`.
- The trailer stays exactly 8 bytes, so existing 8-byte-trailer parsing and `total_len`-authoritative framing are unchanged for records without the flag. A sub-threshold record is byte-for-byte identical to before (the determinism disk image is unaffected).
- `HAS_XXH3` is DERIVED by the codec from the stored body size, never taken from the caller, and decode cross-checks that the flag agrees with the body reaching the threshold (a mismatch is `BadLength`).

## Threshold choice

- `XXH3_PAYLOAD_THRESHOLD = 64 * 1024` (a named const already in `format`).
- Measured on the STORED body (key + headers + payload, the bytes actually written, i.e. post-compression), because the checksum protects the bytes on disk and resync/verify operate on those same stored bytes. The xxh3-64 covers the same body byte range as `body_crc`.

## CRC32C stays resync-gating

CRC32C is verified first. A body corruption is reported as `BadBodyCrc`; a mismatch caught only by the independent xxh3-64 (including a flip in the xxh3 field itself) is a distinct, typed `DecodeError::BadXxh3`. When the flag is clear, decode behaves exactly as today.

## Dependency and supply chain

Adds `xxhash-rust` (xxh3 feature only, `default-features = false`): pure Rust, `no_std`, zero transitive dependencies, no IO and no exposed unsafe. It is BSL-1.0, which is permissive but was not yet in the allow-list, so `deny.toml` now allows `BSL-1.0` with a comment. `cargo deny check licenses bans sources advisories` passes. `ironbus-core` stays IO-free: the new dep pulls in no `std::fs`/`net`/`process` and no async runtime (verified via `cargo tree`), and adds exactly one crate to the graph (`rustc_version`/`semver` were already present via `crc32c`).

## Tests

In `ironbus-core::codec`:
- `sub_threshold_has_no_xxh3_field_or_flag`
- `at_threshold_emits_xxh3_field_and_verifies`
- `over_threshold_body_corruption_is_caught` (CRC32C catches it: `BadBodyCrc`)
- `over_threshold_xxh3_field_corruption_is_caught` (`BadXxh3`)
- `xxh3_field_byte_flip_via_proptest_style_full_sweep`
- proptests `roundtrip_straddling_threshold` and `arbitrary_bytes_near_threshold_never_panic`

In `ironbus-core::types`, the flag tests were extended to recognize `HAS_XXH3`. Full workspace tests pass, including `ironbus-storage` crash-recovery and determinism (records still parse, byte-identical disk image for small payloads). `cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, and `cargo deny` are all clean. The existing `record_codec` fuzz target already exercises the new branch (arbitrary bytes can set the flag) and needs no change.

refs #5, #8
Closes #146


ELares:
## Review resolution (SHIP) + NIT addressed — 9aa446b

Independent adversarial review (not the author): **SHIP**. It ran a full length-math trace and confirmed the parser is panic-free / OOB-free on untrusted bytes (the flag byte is inside the CRC32C-protected header range, so a flipped HAS_XXH3 bit is caught by the header CRC before any sizing, with the flag/threshold cross-check as a second layer), sub-threshold records are byte-identical to the prior layout (determinism image unchanged), CRC32C stays resync-gating with the distinct `BadXxh3`, and recovery is purely `total_len`-driven. License confirmed genuinely Boost (BSL-1.0), NOT the Business Source License; `xxhash-rust` is `no_std`, no IO, zero transitive deps, so ironbus-core stays IO-free.

The one optional NIT (no storage-level test round-tripped a >= threshold record through `scan`/`scan_recovery`) is addressed in 9aa446b: `scan_recovery_reads_an_over_threshold_xxh3_record` appends a `>= XXH3_PAYLOAD_THRESHOLD` record sandwiched between sub-threshold records and asserts both the buffered scan and the streaming recovery scan recover it intact and agree. Honors the "never under-test durability/recovery" bar.

CI: 15/15 green. Local: ironbus-core 111 + ironbus-storage 189 (was 188) pass.

Issue #277, opened 2026-06-07
[TASK] Idle named-group eviction / lifecycle policy (follow-up to #240)


ELares:
## Context

PR for #240 added the work-group CAP (`EngineConfig.max_groups`, `0` = unlimited) plus group-name validation, rejecting a new named group past the cap with `EngineError::TooManyGroups` before it allocates. It deliberately DEFERRED the eviction / lifecycle half of #240, because the wire cannot reach named groups yet (only the default group `""` is reachable today), so eviction is not urgent and inventing a policy now would be premature.

## Scope (the deferred half of #240)

Decide and implement an eviction / lifecycle policy for idle NAMED work-groups so a long-lived broker reclaims the per-group `AckCursor` + `LeaseTable` state, not just bounds it:

- Drop a named, NON-default, fully-caught-up group (its committed cursor is at the head, no acked-ahead set) with NO in-flight leases after an idle window, OR on an explicit `Unsub`.
- The default group `""` is NEVER evicted (it is durable, checkpointed to `cursor.ckpt`).
- An evicted/closed group can be recreated cleanly (resuming from its durable `cursor-<hex>.ckpt` if present, or at offset 0).
- Route any idle-window timing through the existing clock seam (no `Instant::now`).
- Decide the interaction with the durable per-group checkpoint files: whether eviction also removes the on-disk `cursor-<hex>.ckpt`, or keeps it so a re-`Sub` resumes.

This is a prerequisite of the #9 wire-routing slice once named groups are reachable.

refs #240, #9

ELares:
Implemented in #296 (branch `idle-group-eviction-277`). The idle named-group eviction / lifecycle policy is delivered: a fully-caught-up, lease-free, idle NAMED group is evicted on a clock-driven sweep (start of poll + after produce, via the `now_monotonic` seam, no thread), reclaiming its slot against the `max_groups` cap. `EngineConfig.group_idle_evict_ms` (default 0 = disabled; `serve --group-idle-evict-ms <n>`). The durable `cursor-<hex>.ckpt` is checkpointed at the head before drop and KEPT, so a re-subscribe resumes from it and redelivers nothing acked (`Engine::ensure_group`); behind groups and the default group are never evicted, so a committed position is never lost. Explicit `Unsub` of a now-idle named group reclaims it immediately (`Engine::evict_group_if_idle`). Checkpoint format unchanged; no background janitor (both deferred items stayed out of scope).

Issue #240, opened 2026-06-06
[TASK] Bound the number of named work-groups before the wire can name them (#9)


ELares:
## Context

PR #239 added named work-groups to the engine: `poll_in`/`ack_in`/`nack_in`/`progress_in` create a group lazily via `groups.entry(name).or_insert_with(...)`, and groups are never evicted. Today this is safe because the wire protocol does NOT expose group names (only the default group `""` is reachable, via the existing `poll`/`ack`/... delegation).

## The hazard (future)

Once the wire-level `Sub`/group routing lands (the next #9 slice), an unauthenticated client could create arbitrary group names, each allocating a `WorkGroup` (an `AckCursor` + a `LeaseTable`) that is never reclaimed. That is an unbounded-memory / DoS vector, and it violates the project's "no unbounded queues / bounded resources" tenet.

## Scope

Before (or in the same PR as) wiring group names to the wire:
- Add a configurable cap on the number of live groups per engine (e.g. `max_groups`, with a sane default), rejecting a new group past the cap with a typed error rather than allocating.
- Decide and implement an eviction / lifecycle policy for idle named groups (e.g. drop a named, non-default, fully-caught-up group with no in-flight leases after an idle window, or require explicit `Unsub`). The default group `""` is never evicted.
- A group name validation policy (length cap, charset) so a single client cannot exhaust memory with pathological names.
- Tests: the cap rejects the N+1th group cleanly; an evicted/closed group can be recreated; the default group is exempt.

## Why a separate issue

The engine seam (#239) is correct and safe as merged because the wire cannot reach named groups yet. This bound is a prerequisite of the wire-routing slice, not of the engine seam, so it is tracked here rather than absorbed.

refs #9, #10

ELares:
Opened #278: implements the CAP + NAME VALIDATION half of this issue. `EngineConfig.max_groups` (default 1024, `0` = unlimited, matching the other `max_*` bounds) rejects a new named group past the cap with the typed `EngineError::TooManyGroups` before allocating; the default group `""` is always exempt. Names are validated (1 to 128 graphic-ASCII bytes, `EngineError::InvalidGroupName`) before allocation. `serve --max-groups <n>` wires it through. The eviction / lifecycle half is deferred to #277. `refs #240` (not Closes), since the lifecycle policy is still open here.

ELares:
Cap + name-validation half delivered in #278 (merged, squash `64195a41`).

[...]

PR #278, opened 2026-06-07
Bound the number of named work-groups with a cap and name validation (#240)

ELares:
## Summary

Bounds named work-group memory before the wire can name them, closing the future unbounded-memory / denial-of-service vector flagged in #240. PR #239 created named work-groups lazily and never evicts them; once the #9 wire-routing slice lets an unauthenticated client name arbitrary groups, each would allocate an `AckCursor` + `LeaseTable` with no bound. This adds the CAP and NAME VALIDATION half of #240; the eviction / lifecycle half is deferred to a linked follow-up.

`refs #240` (not `Closes`): #240's scope also includes the eviction / lifecycle policy, tracked in the follow-up below.

## What changed

- **Cap (`EngineConfig.max_groups`).** New config field, default `1024`, `0` = UNLIMITED. The `0` = unlimited / off semantics match the existing `max_*` bounds (`max_retained_bytes`, `max_age_ms`, `max_messages`, `LogConfig::max_total_bytes`); the default is a non-zero defensible bound (a single edge broker is not expected to fan out to thousands of distinct groups; 1024 `AckCursor`/`LeaseTable` pairs is a few hundred KiB). The CLI default aliases `ironbus_server::engine::DEFAULT_MAX_GROUPS` so the two cannot drift. (This replaces the prior hard-coded `MAX_GROUPS` constant.)
- **Typed error on overflow.** A new NAMED group past the cap is rejected with the existing `EngineError::TooManyGroups { max }` BEFORE the `WorkGroup` is allocated. The default group `""` is ALWAYS exempt and never counted against the cap; recreating an already-live group is fine (it is not a new allocation). The cap also bounds the open-time recovery scan of `cursor-<hex>.ckpt` files, so a tampered data directory cannot exceed it.
- **Name validation.** A new group name must be 1 to 128 graphic-ASCII bytes (no empty, over-length, control, space, or non-ASCII bytes); an invalid name is rejected with `EngineError::InvalidGroupName` BEFORE allocation, at the same `poll_in` seam where a group is created. The default group `""` bypasses validation.
- **CLI flag.** `serve --max-groups <n>` maps to the config (`0` = unlimited), matching the existing flag-parsing pattern; documented in `USAGE`.

## Typed error variants

- `EngineError::TooManyGroups { max }` (cap reached)
- `EngineError::InvalidGroupName` (empty / over-length / non-graphic-ASCII)

## Tests

- **Engine cap:** filling the cap with distinct named groups succeeds; the N+1th distinct name is rejected with `TooManyGroups` and allocates NOTHING (group count does not grow, the name is absent); re-polling an already-live group does NOT count again; the default group `""` is exempt even at/over the cap (and even at a cap of 1); `max_groups == 0` creates far more than the default without rejection.
- **Engine name validation:** over-length, space, control-byte, and non-ASCII names are each rejected with `InvalidGroupName` and allocate nothing; a valid name and a name exactly at the 128-byte length boundary are accepted.
- **CLI:** `serve --max-groups` rejects a non-numeric value, accepts a valid value, accepts an explicit `0` (unlimited, mirroring the `--max-total-bytes 0` test), and is listed in `USAGE`.

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings` (pedantic), and `cargo test --workspace` are all green.

## Wire note

The cap and validation take effect once #9 wire-routing lets clients reach NAMED groups. Today only the default group `""` is wire-reachable (and it is exempt), so no wire-layer test is needed yet.

## Deferred follow-up

Idle named-group eviction / lifecycle policy is deliberately deferred (not urgent while the wire cannot name groups) and tracked in **#277** (`[TASK] Idle named-group eviction / lifecycle policy (follow-up to #240)`, milestone M1): drop an idle, non-default, fully-caught-up group with no in-flight leases, or on explicit `Unsub`; the default group is never evicted.


ELares:
## Review resolution — BLOCKER fixed (261d274), rebased on main (dd39ed5)

Independent adversarial review (not the author): SHIP-WITH-NITS, with one finding it rightly flagged as a data-loss BLOCKER.

**BLOCKER (fixed):** the group cap was applied to the OPEN-TIME recovery scan, not just to new-group creation. The reviewer reproduced that lowering `--max-groups` below the on-disk named-group count silently DROPPED the committed cursors of the excess groups (reset to offset 0, redelivering the whole already-acked log, orphaning their checkpoint files, with `open()` succeeding: silent data loss). Fixed in 261d274: recovery now loads every durable named group UNCONDITIONALLY; the cap gates only the `poll_in` new-group allocation path. Added regression test `lowering_the_group_cap_still_recovers_every_durable_group` (three groups committed to distinct offsets recover intact after reopening under a cap of 2). The reviewer also noted the wire (`Sub`/`Flow`) ALREADY reaches named groups, so this cap is load-bearing now; the CHANGELOG framing was corrected accordingly.

**Confirmed clean by the review:** create-path coverage is airtight (`poll_in` is the only create seam; `ack_in`/`nack_in`/`progress_in`/`term_in` fence on an unknown group, never create), cap off-by-one correct, `0` = unlimited, every `EngineConfig` literal updated, CLI flag + USAGE, typed errors, no panic/unwrap/unsafe in library paths.

**Acknowledged NITs (non-blocking, by design):** the default group counts toward the cap (so a cap of N permits N-1 named groups, documented in the CLI help); over-cap/invalid subscriptions surface as a generic "fetch failed" on the first Flow rather than leaking the limit; individual control bytes 0x00/0x7f rely on `is_ascii_graphic()` (tab exercises the control path). Eviction/lifecycle is deferred to #277.

CI: 15/15 green after the rebase. Local: ironbus-core 111 + ironbus-storage 189 + ironbus-server 121 + cli 51 + golden 7 pass.

Issue #19, opened 2026-06-05
[DESIGN] Performance targets, SLOs, and benchmarking methodology


ELares:
IronBus promises "HyperScale" and "Edge First" in the same breath, which is a contradiction unless we pin those words to numbers measured on the actual hardware. This issue proposes concrete, falsifiable performance targets (sustained append throughput, p99 end-to-end latency, steady-state memory ceiling, and the fsync-bound durable write rate), defines the reference devices we hold ourselves to, and specifies the benchmark harness, workload mixes, warm-up/steady-state rules, and baseline comparisons that decide whether a build meets those targets. No code exists yet, so the goal here is to agree on the numbers and the measurement discipline before anyone writes a hot loop and overfits to a microbenchmark.

## Why this matters for IronBus

Performance targets are where four of the five tenets are either honored or quietly abandoned. **HyperScale** is meaningless without a sustained (not burst) throughput floor and a tail-latency ceiling that holds under hours of load, not thirty seconds. **Edge First** means the target hardware is a Raspberry Pi 4 or an RK3399, not a 64-core x86 server, so a RAM ceiling and a per-message CPU budget are first-class SLOs, not afterthoughts. **Resilient** means a target must specify behavior under the durability and backpressure regimes that actually ship, not just a fsync-disabled hero number. **Simple** means the harness and its results are reproducible by anyone with `cargo bench` and a checked-in config, with the methodology documented so a number can be trusted. A target without a stated device, durability mode, message size, and consumer fan-out is not a target; it is marketing.

## Prior art

[...]

PR #279, opened 2026-06-07
Add the open-loop macro-bench harness with coordinated-omission proof (#111)

ELares:
## Summary

A new `ironbus-bench` workspace crate (a binary plus a small library): the OPEN-LOOP macro-bench harness that is the honest, coordinated-omission-free SLO instrument the parent (#19) depends on. It drives the SHIPPING `ironbus` binary over a real loopback socket through the real #11 client and measures end-to-end throughput and tail latency.

## Open-loop design

- **Separate sender and receiver, never closed-loop.** A dedicated SENDER drives one real #11 client; a separate latency-recording RECEIVER drives a second. They are never coupled into a request/reply loop, so a stalled broker cannot throttle the offered load and hide its own tail.
- **Constant arrival rate with Poisson jitter.** The sender produces on a schedule fixed before the run: a constant target arrival rate with exponential (Poisson process) inter-arrival jitter, seeded for reproducibility.
- **Overload workload (#10).** A `--max-total-bytes` cap drives the shed-not-OOM workload; a non-fatal `at capacity` shed is tolerated and counted, the same way the broker sheds rather than OOMs.

## Intended-send-time accounting (anti coordinated omission)

Each payload carries a token: its scheduled (INTENDED) send time plus a sequence number. The receiver reads the token back and records latency as `now - intended` (wrk2 style), NOT from when the sender actually got around to sending. When the broker stalls and the sender backs up, the drained backlog carries its original, now-old intended times, so the full stall lands in the tail. A single `CLOCK_MONOTONIC_RAW` clock (monotonic and NOT NTP-slewed; the same constant and `clock_gettime` call on both Linux and macOS) is used for both ends, with the default co-located sender + receiver on one host so there is no cross-host skew.

## HdrHistogram + raw archive

Latencies land in an `HdrHistogram` (1 us to 60 s, 3 significant figures). The RAW histogram is archived per run in the standard HdrHistogram V2+DEFLATE encoding, base64, so percentiles are recomputable and runs are mergeable across windows by any HdrHistogram library.

## Provenance JSON (every run)

Every run emits a versioned provenance record: `schema_version`, git SHA + dirty flag (stamped at build time by `build.rs`), build profile/target, host (os/arch/hostname/cpus), the clock source, the full config, the headline results, the raw base64 histogram, and a copy-pasteable reproduce command. Written to `--json <path>` or stdout.

## Metrics reported

msg/s, MB/s, p50/p99/p99.9 (and max), steady-state broker RSS (sampled mid-run, median of samples), and write amplification (data-dir bytes on disk vs payload bytes produced).

## The injected-stall self-test (the headline gate)

`tests/injected_stall.rs` is a real Unix `#[test]` that runs a healthy baseline and then a run in which a watcher thread `SIGSTOP`s the broker for ~200 ms mid-run and `SIGCONT`s it. It asserts the freeze pushes the recorded p99.9 AND max over a separation floor (0.6 of the stall = 120 ms) that the un-stalled baseline stays below. It FAILS if the tail does not move, which is exactly the coordinated-omission regression it exists to catch. The signal is wide and stable (healthy p99.9 ~30 ms, stalled p99.9 ~210 ms, floor 120 ms), so it is non-flaky: verified 10/10 locally. `SIGSTOP`/`SIGCONT` go through `libc::kill` (a foreign call, Unix-`cfg`-gated, isolated with safety comments).

## CI posture

The harness binary runs on demand (`cargo run -p ironbus-bench -- ...`). ONLY the injected-stall self-test runs in `cargo test`, off the per-PR CI critical path like the criterion micro-benches (#112). The crate compiles under the existing `cargo test --workspace` and clippy (pedantic, `-D warnings`).

## Dependencies and deny.toml

All bench-only, permissive (MIT or Apache-2.0), pure-Rust: `hdrhistogram` (with `serialization` => `base64`, `flate2`, `miniz_oxide`, `nom`, `byteorder`), `rand` + `rand_distr` (Poisson jitter), `serde` + `serde_json` (provenance), and `libc` (clock + signals, Unix only). The license allow-list already covers them; a `deny.toml` comment documents the surface. `cargo deny check` is clean.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo deny check`: all clean.
- `cargo test --workspace`: green, including the injected-stall self-test.
- Injected-stall self-test run 10/10 with no failures.
- No em/en dashes in the diff; SPDX headers on every new source.

refs #111 (harness delivered; reliable CI gating of the live self-test deferred to #284)


ELares:
## Review resolution — SHIP-WITH-NITS, addressed (03c99bf)

Independent adversarial reviewer (not the author) reviewed the real HEAD and verified the load-bearing concern: the connection-tolerance fix does NOT weaken the coordinated-omission proof. The histogram is only ever written with a true measured `now - intended` latency, so a graceful early end can never fabricate tail samples; an early receiver end fails SAFE (the missing stall makes `p999 >= floor` FAIL, not vacuously pass); `is_connection_ended` matches only `Closed` + `Io{ConnectionReset,ConnectionAborted,BrokenPipe,UnexpectedEof}`, so `Server`/`Frame`/`Body`/`BadResponse` and crucially `Io(TimedOut)` (a never-recovering wedge) stay fatal. Also confirmed: wrk2-style intended-send-time measurement (no closed-loop), both non-vacuity directions asserted (8/8 on a stable host), the HdrHistogram raw archive round-trips (V2+DEFLATE cookies verified), clippy/fmt/deny clean, no en/em dashes, deps permissive and ironbus-core stays dep-clean, MSRV-1.78-safe, SPDX + justified unsafe.

The one issue: `Closes #111` overstated completion because the headline self-test is `#[ignore]`d off per-PR CI. Resolved: the PR now uses `refs #111` (not Closes); the 5 deterministic criteria (open-loop intended-time generator, HdrHistogram + raw archive, real-#11-client against the shipping binary, the five reported metrics, provenance JSON) are delivered and gate per-PR CI; the live SIGSTOP proof is reliable on a stable host (passes repeatedly; flaky on shared GitHub runners where the freeze does not manifest in the tail) and runs on demand via `cargo test -p ironbus-bench -- --ignored`. #284 tracks wiring it into a reliable gate (a stable/self-hosted runner or an in-broker fault-injection seam). CHANGELOG corrected to state this honestly.

CI: 15/15 green.

PR #281, opened 2026-06-07
Make the ironbus-core IO-free check structural via an AST walk (#143)

### LINKED ISSUE #143: [TASK] Make the ironbus-core IO-free check structural, not regex-based
The `ironbus-core is IO-free` CI job currently enforces the IO-free invariant with a source grep plus a `cargo tree` dependency check. The grep was hardened in #141 (whitespace-robust, grouped imports, scans all `*.rs`, covers `std::{fs,net,process}` and async runtimes), but a regex can still be evaded in principle (macros, re-exports, creative pathing). The dependency-tree half is sound.

## Scope
Replace or back the source grep with a structural guarantee that cannot be evaded, for example:
- a `#![deny(...)]` lint via a custom dylint, or
- a dedicated compile of `ironbus-core` against a target/shim that makes `std::fs`/`std::net`/process spawning unavailable, or
- a test that introspects the crate for forbidden symbol references.

## Acceptance criteria
- [ ] The IO-free invariant for `ironbus-core` is enforced by a mechanism that does not rely on a source-text regex.
- [ ] The check fails on a deliberately planted violation and passes on clean code.
- [ ] Runs in CI as a required check.

Found during the review of #141. Relates to #16, #21.

ELares:
## Mechanism
Replaces the source-text **grep** half of the `ironbus-core is IO-free` CI job with a real **AST walk**, and keeps the `cargo tree` dependency half. A new dev-only workspace member, `tools/io-free-check`, parses every `crates/ironbus-core/src/*.rs` with `syn` (`full` + `visit` features) and walks the syntax tree with a `syn::visit::Visit` visitor, collecting forbidden references with a file/path/span. Comments never appear in the AST, whitespace and import grouping are irrelevant, and `use` trees are expanded to their real paths, so the evasions a line regex is vulnerable to (grouping, aliasing, globs, creative pathing) are all caught.

## What it forbids
- `std::{fs, net, process, os, io}` (and the same submodules under `core`/`alloc`). `std::fmt` / `core::fmt` are formatting-only and stay **allowed**.
- Async runtimes and low-level event/syscall crates: `tokio`, `async_std`, `smol`, `mio` (as a path root or a `use` root).
- `extern "C" { ... }` foreign-function (FFI) blocks, a raw escape hatch for IO/syscalls outside Rust's view.
- `use` trees are walked recursively, so `use std::{fs, net::TcpStream}`, `use std::fs as f`, and `use std::net::*` are each classified by their real, fully-expanded path. A leading `::`/`crate`/`self`/`super` qualifier is stripped so `::std::fs` cannot hide.

## Where it lives
`tools/io-free-check` (new), added to the workspace as `members = ["crates/*", "tools/*"]`. It is depended on by **no shipped crate**, so `syn`/`proc-macro2` never enter `ironbus-core`'s own dependency graph (the `cargo tree` step proves the graph stays runtime-free). The walker logic is in `lib.rs` (unit-tested) and a thin `main.rs` walks the filesystem and emits GitHub `::error::` annotations. `syn`/`proc-macro2` were already in `deny.toml`'s allow-list (MIT OR Apache-2.0), so no policy change was needed; `cargo deny check` passes.

## Dep-tree check retained
The AST source walk only sees `ironbus-core`'s **own** source, so IO that arrives by **re-export** from a dependency is invisible to it. The existing `cargo tree -p ironbus-core` check is kept as the necessary second half and catches exactly that. A comment in the CI job and the crate docs document that the two together (AST source walk + dep-tree) are the structural invariant, and note a compiler-level `dylint` driver as a possible future hardening. The job name `ironbus-core is IO-free` is unchanged so branch protection's required-check list still matches.

## Planted-violation proof
- **Unit tests** over in-memory snippets in `tools/io-free-check/src/lib.rs`: `plain_fully_qualified_path_is_caught` (`std::fs::read`), `simple_use_is_caught`, `grouped_use_is_caught`, `aliased_use_is_caught`, `glob_use_is_caught`, `leading_colon_colon_is_caught`, `nested_group_is_caught`, `async_runtime_use_and_path_are_caught`, `path_in_generic_argument_is_caught`, `extern_block_is_caught`, `std_fmt_is_allowed`, `violation_carries_a_span`, and `clean_source_passes` (pass-on-clean). All 13 run under `cargo test --workspace`.
- **Manual, reverted plant**: appending `use std::{fmt as _f, net::TcpStream as _s};` to `crates/ironbus-core/src/types.rs` made `cargo run -p io-free-check` report `std::net` with its span and exit 1 (the clean `fmt as _f` arm was correctly ignored); after revert it exits 0. The commit contains no planted violation.

## Acceptance criteria
- [x] Enforced by a non-regex (AST) mechanism.
- [x] Fails on a planted violation, passes on clean code.
- [x] Runs in CI as a required check (same stable job name).

Local gates: `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked`, `cargo deny check`, and `actionlint` all pass.

Closes #143


ELares:
## Review resolution — SHIP-WITH-NITS, SHOULD-FIX addressed (9c50f98)

Independent adversarial reviewer (not the author): SHIP-WITH-NITS. It empirically confirmed the `syn` AST walk catches every evasion claimed PLUS type aliases, trait impls, qualified-self paths, where-clauses, turbofish, `pub use` re-exports, and assoc-type bindings; that the required-check NAME (`ironbus-core is IO-free`) is preserved so branch protection's gate is not silently dropped; that `syn` stays out of ironbus-core's shipped graph (`cargo tree -p ironbus-core` unchanged); and it proved the gate fails on a planted `use std::net::TcpStream;` (exit 1 with file/path/span) then reverted it.

SHOULD-FIX (fixed in 9c50f98): the one residual false-negative, `std::*` IO hidden inside a local `macro_rules!` body, is caught by neither the AST walk (syn sees a macro body as opaque tokens) nor the dep-tree check (std::* needs no dependency). This is now explicitly documented in the io-free-check crate doc and the CI job comment as a review-only case (ironbus-core has no such macro today). No behavior change.

CI: 15/15 green, including the `ironbus-core is IO-free` job running the new structural check.

Issue #122, opened 2026-06-05
[TASK] loom scoped concurrency tests for commit index, ring-buffer handoff, and refcounts


ELares:
Parent: #21

Exhaustively permute the interleavings of the genuinely lock-free hot paths with loom, rather than hoping a stress test trips a race. Scope is deliberately tight: loom's state space is factorial, so only the three structures where a data race could violate I1 (lost ack) or corrupt the commit watermark are covered.

## Approach

Model exactly three structures as small bounded loom tests: (1) the commit-index publish/observe (producer publishes a monotonic commit watermark, consumer observes it), (2) the SPSC/MPSC ring-buffer producer-to-WAL handoff, and (3) the segment/buffer refcount handoff (reader pin vs writer recycle). Use loom's replacement atomics and cell types. Cap each test at 3 threads and 4 shared-memory operations per thread; when the unbounded space exceeds a 2-minute budget, apply bounded preemption (LOOM_MAX_PREEMPTIONS 2 to 3). Correctness arguments must hold under AcqRel: do not rest on SeqCst-only reasoning, per loom's documented modeling limits (SeqCst modeled as AcqRel can yield false positives, load-buffering is under-explored). Tests run under RUSTFLAGS="--cfg loom" and are scoped, not whole-broker. This pins the consumer-side handoff from #9 and the commit semantics feeding #10.

## Acceptance criteria

- [ ] loom tests cover the commit index, the ring-buffer WAL handoff, and the refcount handoff.
- [ ] Each test is capped at 3 threads and 4 operations per thread; bounded preemptions applied past a 2-minute budget.
- [ ] No correctness argument relies on SeqCst-only ordering; AcqRel suffices.
- [ ] Tests pass under RUSTFLAGS="--cfg loom" and run as a scoped PR gate.
- [ ] Documented scope limits and known loom false-positive/false-negative caveats are recorded.
- [ ] loom is a cfg-gated dev-dependency and does not inflate the shipped binary.

## Failure considerations

The dominant risk is state-space explosion: an unscoped loom test never finishes and the gate gets disabled, silently removing race coverage. The thread/op ceiling and bounded preemptions keep it tractable. A subtler risk is a false negative from loom's under-explored load-buffering, mitigated by not resting correctness on subtle memory-model arguments and by cross-checking the same paths in the simulation. False positives from SeqCst-as-AcqRel are avoided by banning SeqCst-only reasoning.

ELares:
[...]

PR #282, opened 2026-06-07
Document the shared subsystem invariants and the glossary (#131)

### LINKED ISSUE #131: [META] Shared invariants and glossary every subsystem must uphold
The same load-bearing terms (offset, seq, durable, ack, commit, torn tail, poison, segment generation, cursor, lease) are defined locally and slightly differently in #3, #4, #5, #6, #7, #8, and #9. And several invariants are assumed by many issues but owned by none (the checkpoint-lower-bound rule, the loss-report schema, the offset-continuity rule). This issue freezes one glossary plus one numbered invariants list that #21 checks (I1..I4 are already referenced there) and every other issue cites.

### Glossary (canonical definitions, abbreviated)
- **offset**: monotonic u64, +1 per accepted record, never reset, never reused, sparse (holes from skips/compaction are legal). Authoritative for ordering. Owner #3.
- **seq**: per-record durable counter inside a segment, in [base_seq, base_seq+record_count). Owner #5.
- **durable**: bytes covering the record have returned from fdatasync on the active segment (no separate WAL). Owner #6 (pending the WAL-architecture reconciliation).
- **ack**: producer-visible signal returned only after durable (default level). Carries the assigned offset. Owner #3 + #6.
- **canonical record timestamp**: broker-assigned monotonic-anchored wall-clock at append, one field, the only time source for retention age, lag-age, and the time index.
- **earliest_retained**: lowest offset still on disk; consumers below it are reset with one truncation event.

### Invariants (numbered, testable)
- **I1 durable-prefix**: at every write boundary, recovery yields a prefix of the accepted-offset sequence with no reordering and no hole below the durable head. (#7, #21)
- **I2 ack-implies-durable**: no ack is observable for a record not covered by a returned fdatasync (except the explicitly-labeled opt-in ack-on-buffer mode). (#6)
- **I3 bounded-reported-loss**: any skip loses at most one contiguous region per segment and one whole unreadable segment, never partial-within-a-record, never silent; emitted in (records_lost, bytes_lost, segments_affected) plus offset range and reason, capped per-event and globally (1% default) with hard-fail above. (#8, #7, #21)
- **I4 checkpoint-lower-bound**: a checkpoint names only (durable_seq, durable_offset) for which fdatasync has already returned; recovery treats it as a floor and re-validates forward. (#7)
- **I5 offset-monotonic-across-recovery**: the offset counter never regresses or reuses a value across a crash, recycle, or restart. (#3, #4)
- **I6 ordering-never-consults-wall-clock**: offset and per-producer seq alone order and dedup; no wall-clock read affects ordering. (#3, #20)
- **I7 integrity-before-transform**: CRC32C over the outermost on-disk bytes is verified before decrypt and before decompress. (#5, #18, #12)
- **I8 single-writer**: exactly one logical writer mints offsets and owns the active-segment fd. (#4, #6)

### Loss-report schema (versioned, single definition)
Freeze the structured RecoveryReport/SkipEvent schema once here (the fields #7, #8, #16, #21, and #15 all assert against) as `ironbus.loss-report.v1`.

### Acceptance criteria
- Every issue that uses a glossary term links to this definition; conflicting local definitions are removed.
- #21 invariant checkers I1..I8 reference these exact statements.
- The loss-report schema has one versioned definition all five consumers cite.

ELares:
## What this covers

Adds `docs/INVARIANTS.md`, the canonical cross-subsystem reference that #131 asks for: one numbered invariants list plus one glossary every other issue can cite. It is derived from and cross-checked against the actual code and the README (both canonical); the code wins where they differ, and the divergences are flagged inline.

- **Shared invariants I1 to I8** (the issue's list): log-is-WAL durable prefix, ack-implies-durable (`fdatasync` before ack, freeze-on-failed-fsync), bounded-and-reported loss (one segment or 64 MiB per event, 1% global, fail closed), pure recovery, offset-monotonic-across-recovery plus the checkpoint lower bound, ordering-never-consults-wall-clock, single-writer. Each invariant has a precise statement, the mechanism that holds it, and a citation to the file plus the test, property, or checker.
- **Other always-on invariants**: bounded resources / no unbounded queues, torn-tail truncation never replaying recycled bytes, at-least-once leases with a hard cap and generation fencing, and the exactly-once DLQ move.
- **Resilience invariant checkers (the code's I1 to I4)** in `invariants.rs`, with a mapping table to the shared invariants.
- **A glossary** of the load-bearing terms (offset, seq, segment, active vs sealed, the WAL, cursor / committed offset, lease / visibility timeout / fence token, work-group vs broadcast, default group, credit and the in-flight window, DLQ / poison / max-deliver, truncation signal, the IO / clock seams, MSRV, frozen-tag discipline, durable, ack, torn tail, loss report), each defined as the code uses it.

Every invariant cites where it is enforced or tested. Specified-but-unenforced items are in a clearly-marked "Specified, enforcement pending" section with the tracking issue.

## Where the issue diverges from the implemented reality (flagged in the doc)

- The issue's **I1 to I8 numbering does not match the code's separate I1 to I4** resilience checkers in `crates/ironbus-storage/src/invariants.rs` (code I1 = no-acked-loss maps to shared I2; code I2 = longest-valid-prefix maps to shared I1; I3 and I4 line up). A mapping table is included. Aligning the two schemes is open work under #120 / #21.
- The DLQ **idempotency check keys on `(group, source_offset)`** against the per-group high-water mark, while the issue and the `dlq.rs` module doc name the key `(group, source_offset, attempt)`. The attempt is stored in the DLQ record but does not participate in the dedup test; the enforced key is (group, offset).
- The issue's **I4 is "checkpoint-lower-bound"**, but the code's checker numbered I4 is "pure recovery". The checkpoint-lower-bound property is real and implemented (documented as I5b), just under a different number.

## Specified but NOT yet enforced in code (high-signal)

- **I7 decrypt/decompress halves**: CRC-before-transform is enforced (`codec::decode` verifies header then body CRC32C before the xxh3-64), but encryption-at-rest and compression are not implemented, so "before decrypt / before decompress" is not exercised (#12, #139, #18).
- **`none` / `interval` durability modes** (the opt-in ack-on-buffer exception to I2): only fsync-before-ack exists (#6).
- **Opt-in producer dedup** (part of I6): ordering-from-offset is enforced; the dedup window is not implemented (#3).
- **Full invariant-checker harness** wiring I1 to I8 against the corpus and sim as one suite, and aligning the numbering (#120, #21).
- **Loom concurrency model checking**: single-writer is held structurally by the `Mutex`, not yet proven under loom (#122).
- **Named, externally-versioned `ironbus.loss-report.v1` artifact**: the in-code `LossReport` schema (version 1, stable reason codes) is the source of truth, but the separately-frozen named schema is not cut yet (#8, #16).

## Scope

Docs only: `docs/INVARIANTS.md` (new), a link from `docs/README.md`, and a terse `CHANGELOG.md` Added entry. No crate source, `Cargo.toml`, or CI touched (conflict-free with parallel code PRs; does not touch the sibling byte-layout contracts doc). No en or em dashes. `cargo fmt --all --check` and `cargo test --workspace` are green.

Closes #131


ELares:
## Review resolution — SHIP-WITH-NITS, both addressed (fe33ced)

Independent reviewer (not the author) verified `docs/INVARIANTS.md` is accurate: every invariant statement and its cited enforcement location checks out against the source (the I1..I4 checker-numbering mapping in `invariants.rs`, the DLQ dedup key `(group, source_offset)`, the I2/I3/I5b/I6/I8 citations), and every "specified, enforcement pending" item (encryption-at-rest, none/interval durability modes, producer dedup, loom #122, the full checker harness #120) is genuinely unenforced in the tree.

- SHOULD-FIX (fixed): the new bullet had been fused with the pre-existing `max_groups` CHANGELOG entry, dropping that entry's opening clause. Resolved by rebasing on current `main` and reconstructing the `### Added` section so the INVARIANTS, CONTRACTS, and intact `max_groups` bullets are three separate entries.
- NIT (fixed): reworded the encryption-flag note so it no longer implies a named encryption bit exists (only `COMPRESSED` is defined; the flags byte has reserved space).

Docs-only, no en/em dashes, CI 15/15 green.

PR #283, opened 2026-06-07
Add key_shared work-group routing with per-key order preservation (#64)

### LINKED ISSUE #64: [TASK] key_shared routing and rebalancing with per-key order preservation
Parent: #9

Implement key_shared work-group routing: messages with the same record key always go to the same live member, preserving per-key order while still letting the group drain in parallel across keys. This is the optional ordering mode over a competing group; the default remains `key_ordering = none` (plain competing distribution). The hard part is rebalancing when membership changes without breaking the per-key order the mode exists to provide.

## Approach
Route by consistent hash (jump or rendezvous hash) of the record key over the current live-member set. Enforce per-key in-flight serialization: a member never receives key K's next message until K's prior message is acked or its lease expires, so per-key order holds even with parallel members. On member join or leave, only the affected key range remaps (consistent hash minimizes reshuffle versus sticky-range, which remaps half the keyspace on a single change). For keys whose owner changes, wait for in-flight on the moved keys to drain or expire before the new owner receives them, so an old in-flight message plus a new routed message cannot interleave out of order. Document the batching constraint: a fetched batch may span members because keys hash independently, so producers get a per-key order promise only, not cross-key.

## Acceptance criteria
- [ ] Same-key messages always route to the same live member; per-key delivery order is preserved.
- [ ] A member join or leave remaps only the affected key range, not the whole keyspace.
- [ ] Per-key order survives a mid-stream member churn, verified by a test that asserts ordering across a join and a leave.
- [ ] A key with an in-flight message is not delivered to a new owner until the prior message drains or its lease expires.
- [ ] `key_ordering = none` remains the default and is unaffected by this code path.

## Failure considerations
A member dying mid-key leaves an in-flight message that the drain-or-expire rule reclaims via the lease task before the new owner proceeds, preserving order at the cost of one redelivery (at-least-once). Rapid membership flapping could thrash remapping; the consistent-hash minimal-reshuffle property plus the drain guard bound the damage to affected keys only. A skewed key distribution can hot-spot one member; this is surfaced to #16 as per-member in-flight depth so operators can repartition.

ELares:
Implements `key_shared` routing (#64): optional per-key affinity over a competing work-group. A record's key routes to ONE live member, so same-key records always reach the same member and keep their order, while the group drains in parallel across keys. The default `key_ordering = none` (plain competing distribution) is the engine's unchanged default path.

## Mode-wiring choice: server-side config, not on the wire
For v1 the mode is **server-side per-group config**, NOT a wire-negotiated field. The frozen `Sub` frame body is exactly the group name (a proptested whole-body-is-the-group invariant); carrying a `key_ordering` flag would change that frozen shape. So no frame tag or flag bit changed in this PR. An operator declares groups with `serve --key-shared-group <name>` (repeatable); a session enables the mode and joins on SUB to a configured group. Wire-negotiated `key_ordering` on `Sub`/`Connect` is a noted follow-up.

## Hash function
**Rendezvous (highest-random-weight) hashing.** For a key, the live member whose `xxh3_64(member_id_le ++ key)` is largest owns it (ties break on the larger `MemberId`, so the choice is total and deterministic). Rendezvous hashing has the minimal-reshuffle property a sticky cursor lacks: a single join or leave moves only the keys that hash to (or away from) that member, never the whole keyspace. `xxh3_64` is already an `ironbus-core` dependency. Empty-keyed records have no affinity and keep plain competing distribution.

## Per-key serialization + drain-on-remap design
The `KeyRouter` tracks, per key, the offset of its currently delivered-but-unacked record (`in_flight: BTreeMap<key, offset>`). The poll consults `decide(member, key, offset)`:
- **KeyBusy**: the key holds an EARLIER in-flight offset, so no member may take a higher offset of that key yet (per-key serialization). The same offset is allowed through (it is the outstanding record's own redelivery on lease expiry, which the lease layer gates and which must precede the key's next record).
- **NotOwner / Deliver**: ownership under the current live-member set.

The lease layer decides whether the outstanding offset itself is redeliverable (in-flight while visibility holds, reclaimable once expired). Combined, this gives the **drain-or-expire guard across a rebalance**: if a key's owner changes while it has an in-flight record, the new owner cannot receive the key's next record until that in-flight record drains (acked, clearing the entry) or expires and is redelivered as the same offset, so an old in-flight and a newly-routed record can never interleave out of order. A departed member's in-flight records stay leased and drain or expire normally (at-least-once, one redelivery).

To route by key the poll must read a record's key before claiming; a new non-mutating `LeaseTable::is_claimable` peek lets the router decide routing without consuming a generation on an offset it then declines.

## Surface
- Core: `ironbus_core::keyshared` (`KeyOrdering`, `MemberId`, `KeyRouter`, `RouteDecision`) plus `LeaseTable::is_claimable`.
- Engine: `WorkGroup` gains an optional router; new seams `set_configured_key_shared_groups`, `set_key_ordering_in`, `join_member_in`, `leave_member_in`, `poll_in_member` (identical to `poll_in` for a `none` group), `route_decision_in`, `busy_keys_in`. ack / dead-letter clear the key's entry; the committed cursor prunes the per-key map.
- Session/server: a per-connection `MemberId`; SUB to a configured group enables the mode and joins; UNSUB / subscription switch / disconnect leaves (cleanup runs on every connection exit path). The Flow fetch routes through the member-aware poll.
- CLI: `serve --key-shared-group <name>` (repeatable), documented in USAGE.

## Deferred (follow-ups, not acceptance criteria)
- Wire-negotiated `key_ordering` on the handshake/`Sub` (this PR is server-side config).
- Per-member in-flight depth on `/metrics` for the #16 hot-spot signal; `busy_keys_in` already surfaces a cheap per-group count as the seam.

No new follow-up issue was filed: the deferred items are pre-existing follow-ups (#64 wire form, #16 metrics) and nothing is left as a bare TODO.

## Tests
Core `keyshared` (13): `an_empty_group_has_no_owner`, `join_and_leave_track_membership_idempotently`, `the_same_key_always_maps_to_the_same_owner`, `keys_spread_across_members_not_all_to_one`, `removing_a_member_remaps_only_its_keys_minimal_reshuffle`, `adding_a_member_remaps_only_the_keys_that_move_to_it`, `an_owner_gets_deliver_and_a_non_owner_gets_not_owner`, `an_empty_key_is_deliverable_to_any_member`, `a_busy_key_blocks_a_higher_offset_until_cleared`, `a_busy_key_blocks_its_next_record_for_a_new_owner_after_a_rebalance`, `retain_above_prunes_committed_keys`, `clear_offset_is_a_noop_for_an_unknown_offset`, `key_ordering_defaults_to_none`.

Lease: `is_claimable_is_a_non_mutating_peek_of_claim`.

Engine (8): `key_ordering_defaults_to_none_and_is_unchanged`, `setting_key_shared_attaches_a_router_and_reverting_drops_it`, `same_key_always_routes_to_the_same_live_member`, `a_busy_key_is_not_redelivered_to_a_new_owner_until_it_drains`, `per_key_order_survives_a_mid_stream_join`, `an_empty_key_keeps_plain_competing_distribution`, `a_non_owner_never_takes_a_key_even_when_polling`, `busy_keys_tracks_in_flight_and_clears_on_ack`.

Session (3): `key_shared_over_the_wire_routes_a_key_to_one_member_in_order`, `an_unconfigured_group_stays_plain_competing_over_the_wire`, `leaving_a_key_shared_group_drops_membership`.

CLI (3): `serve_accepts_repeated_key_shared_group_flags`, `serve_rejects_a_key_shared_group_without_a_value`, `usage_lists_the_key_shared_group_flag`.

`cargo fmt --all` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean (pedantic), `cargo test --workspace` fully green (determinism gate and golden path included). No em/en dashes. `ironbus-core` stays IO-free; no `unsafe`; no library unwrap/expect/panic.

Closes #64


ELares:
## Review resolution — SHIP-WITH-NITS, all three addressed (78defc4)

Independent adversarial reviewer (not the author) traced the load-bearing ordering properties and could not break them: per-key order across churn (full leave sequence), drain-or-expire on owner change (the router consults the non-mutating `is_claimable` peek with the same `now < deadline` threshold as `claim`), same-key affinity (deterministic rendezvous hash, total tiebreak), minimal-range remap, no double-delivery (whole `poll_in_member` under the engine lock, generation-fenced claim), and the `none` default path byte-for-byte unchanged (no proto/frame change). The leave case IS genuinely tested, so `Closes #64` is justified.

Addressed:
- **S1 (empty-key over-serialization):** the per-key gate had been applied to the empty key `b""`, throttling all empty-keyed traffic to one in-flight record across the group, contradicting the "empty key = plain competing, drains in parallel" contract. `decide` now bypasses the gate for the empty key (the lease layer alone bounds it per offset) and `mark_in_flight` is a no-op for it. Strictly safe before and after; new core test `empty_keys_drain_in_parallel_and_are_not_serialized`.
- **S2:** the engine empty-key test acked between polls so it could not detect S1. It now polls two empty-keyed records across two members with NO ack between and asserts both deliver as distinct offsets (fails under the old gate), then drains exactly once.
- **N1:** `per_key_order_survives_a_mid_stream_join` used a key whose owner never moved. Renamed to `..._that_moves_the_owner`; it now picks a key owned by m2 in the two-member set, drops m2, produces under sole-owner m1, then rejoins m2 so the join provably MOVES the owner, and asserts the drain guard blocks o1 and that o1 then delivers to the NEW owner m2.

ironbus-core stays IO-free / panic-free / unsafe-free; clippy `-D warnings` clean; no en/em dashes; full workspace + determinism + golden path green. CI: 15/15.

PR #285, opened 2026-06-07
Document the WAL behavior under load and the on-disk file lifecycle (#135)

### LINKED ISSUE #135: [DESIGN] WAL under high write rates and on-disk file cleanup (segment janitor and checkpointing)

ELares:
Adds `docs/WAL.md`, a design document on how the active-segment-is-the-WAL model absorbs a sustained high write rate and how the on-disk files are created, sealed, and retired. Derived from and cross-checked against the actual source (storage `log.rs` / `segment.rs` / `checkpoint.rs` / `dlq.rs` / `loss.rs` / `naming.rs`, core `format.rs`, server `engine.rs` / `server.rs`, the CLI knobs, the README, and ADRs 0001/0002). The code is treated as canonical where it diverges from the #135 text.

## What it covers (IMPLEMENTED)
- The log-is-WAL model: one append path, `fdatasync`-before-ack (`Engine::produce`), single-writer realized as `Mutex<Engine>` today.
- High write rate: size-cap segment roll (`max_segment_bytes`, default 64 MiB), the O(1) running totals (`sealed_record_bytes`, `total_record_count`, per-slot `record_count` / `max_timestamp_ms`) that keep append and the cap checks O(1), and the spill-then-shed overflow policy (`max_total_bytes` / `AtCapacity`, `drop-new` vs `drop-oldest`). No producer stall in v1.
- The on-disk file classes and lifecycle: active and sealed `.log`, the default `cursor.ckpt` and per-group `cursor-<hex>.ckpt` two-slot checkpoints, and the `dlq/` second segmented log (what each is, when created, when reaped).
- Recovery: torn-tail truncation to the longest valid prefix, the bounded-loss caps (per-event and 1% global), and the one-time `Poll::Truncated` signal a lagging consumer gets under `drop-oldest`.
- Retention (whole-sealed-segment deletion under size/age/count, consumer-safe protect floor) and the disk-full drop-oldest primitive, with crash-safe unlink-then-account ordering.
- The `serve` knobs in a table (`--max-segment-bytes`, `--max-total-bytes`, `--max-retained-bytes`, `--max-age-ms`, `--max-messages`, `--checkpoint-interval`, `--disk-full-policy`) with their real defaults.

Cross-references `CONTRACTS.md` (byte layouts) and `INVARIANTS.md` (I1 to I8) by name rather than duplicating their tables.

## Notable discrepancies / unimplemented (each verified absent in source)
- **Specified but not built:** index sidecars (`.index` / `.tindex`), a `current` pointer file, a manifest, `fallocate` preallocation and generation-stamped recycling (ADR 0002 forbids recycling in v1), a dedicated append actor + commit thread + group-commit + admission credits (`server.rs` calls these a follow-up), a separate durable-point deletion watermark, a 90% disk high-water forced reaper run + periodic janitor timer, a `retention.consumer_safe` toggle, compaction (#83) / an archival sink, and a persisted `quarantined` segment state.
- **Code-vs-intent divergences:** no "edge profile" switch (8 MiB `EDGE_SEGMENT_BYTES` is an unused constant; you pass `--max-segment-bytes` explicitly); segment roll is size-only (no 24h/age active-segment roll despite `DEFAULT_SEGMENT_ROLL_HOURS`); the lifecycle is two states (active/sealed) not six; deletion is gated by retention + slowest-consumer, two watermarks not three; and `ReasonCode::SequenceGap` is defined in the loss schema but never emitted by the recovery path (which instead hard-errors with `RecoveredSequenceMismatch`).

## Validation
- Docs only: one new file `docs/WAL.md`, a link from `docs/README.md`, and a terse `CHANGELOG.md` entry. No crate source, `Cargo.toml`, or CI touched; `CONTRACTS.md` / `INVARIANTS.md` untouched.
- `cargo fmt --all --check` clean; `cargo test --workspace` green (0 failures).
- No em/en dashes.

Closes #135


ELares:
## Review resolution (SHIP) + NIT addressed (ef4e545)

Independent reviewer (not the author) confirmed all six code-vs-intent findings against the source:
- Log-is-WAL, fdatasync-before-ack, the produce holds the engine Mutex across the fsync (head-of-line blocking); the append-actor / group-commit is an unbuilt throughput follow-up.
- Segment roll is SIZE-only; `EDGE_SEGMENT_BYTES` (8 MiB) and `DEFAULT_SEGMENT_ROLL_HOURS = 1` are unwired constants (no edge-profile switch, no active-segment time roll).
- Deletion is gated by retention bounds AND the slowest-consumer protect floor (`min_committed_offset`), two watermarks; no third durable-point watermark.
- `ReasonCode::SequenceGap` (code 5) is defined but never emitted (recovery hard-errors with `RecoveredSequenceMismatch` instead).
- The verified-absent list (index sidecars, `current` pointer, manifest, fallocate, generation-recycling, compaction, consumer_safe field, quarantined state, high-water janitor) is genuinely absent.
- All seven CLI knobs exist with the documented defaults/units.

NIT addressed: the doc now also notes `ReasonCode::CorruptSegmentHeader` (code 4) is likewise defined-but-unemitted, for symmetry with the SequenceGap discrepancy.

Docs-only, no en/em dashes, CI 15/15 green. Rebased on current main with the CHANGELOG and docs/README links resolved alongside the threat-model doc (#286).

Issue #195, opened 2026-06-06
Graceful shutdown: checkpoint the cursor on SIGTERM/SIGINT (needs a signal-handling deps/unsafe decision)


ELares:
## Problem

`ironbus serve` has no graceful-shutdown path. The serve loop polls a `shutdown` AtomicBool that is only ever flipped in unit tests; there is no signal handler. So SIGTERM/SIGINT kills the process abruptly with no final cursor checkpoint and no in-flight drain.

Consequence (surfaced by the #194 docs review): a consumer's acks are durable only via the cursor checkpoint, which fires on a clean client disconnect (the EOF path) or after `--checkpoint-interval` cumulative commits (default 1024). Signalling the broker is therefore an abrupt stop, and a long-lived consumer still connected at that moment gets its recently-acked messages redelivered after restart (up to `--checkpoint-interval`). This is safe at-least-once, but a clean operator shutdown should not redeliver: a graceful stop should flush the committed cursor.

## What is needed

On SIGTERM and SIGINT (and SIGHUP), flip the serve `shutdown` flag, stop accepting new connections, flush the committed cursor (`Engine::checkpoint_cursor`), optionally drain in-flight handlers within a bounded grace period, then exit 0.

## The dependency decision (the blocker)

Catching a signal in Rust needs either an external crate or `unsafe`:
- `ctrlc = "3"` is the standard minimal option. It passes `cargo deny check` (licenses/bans/advisories ok) and satisfies MSRV 1.78, BUT it pulls in `nix` + `libc` transitively. That is the first non-trivial runtime dependency beyond `crc32c`, and it sits against the README's "no external dependencies" tenet and the strict `deny.toml`.
- A hand-rolled handler via `libc::sigaction` needs `unsafe`, which the engineering rules gate behind a justification + safety proof + Miri; Miri cannot exercise OS signal delivery, so that ceremony is impractical here.
- `signal-hook` is similar to `ctrlc` (also pulls deps).

This is a deliberate architecture/supply-chain call (accept `ctrlc`/`nix`, or carve a narrowly-justified `unsafe` exception for a self-contained `sigaction` handler, or accept abrupt-stop for v1 and bound the window with a time-based periodic checkpoint instead). It should be decided, not made in passing.

## Acceptance criteria

- [ ] SIGTERM/SIGINT trigger a clean shutdown that checkpoints the cursor before exit, so a restart after a graceful stop does not redeliver acked messages.
[...]

PR #287, opened 2026-06-07
Graceful shutdown: checkpoint the cursor on SIGTERM/SIGINT via ctrlc (#195)

ELares:
## What & why

`ironbus serve` had no graceful-shutdown path. A signal killed the process abruptly with no final cursor checkpoint, so a consumer still connected at the moment of the signal got its recently-acked messages redelivered after a restart (up to `--checkpoint-interval`). This is at-least-once safe, but a clean operator stop should not redeliver: a graceful stop should flush the committed cursor (surfaced by the #194 usage-guide review).

## The dependency decision (resolved)

Use the `ctrlc` crate (the maintainer's call over a hand-rolled `unsafe` sigaction, whose unsafe-justification ceremony is impractical here because Miri cannot exercise OS signal delivery). Added `ctrlc = "^3.4"` with the `termination` feature (for SIGTERM/SIGHUP, not just SIGINT) as a **Unix-only** `ironbus-cli` dependency. It pulls `nix` + `libc` transitively; both, and the rest of the new surface (`bitflags`, `cfg-if`, `cfg_aliases`, and the macOS-only `objc2`/`block2`/`dispatch2`), are permissive (MIT or MIT OR Apache-2.0) and already covered by the `deny.toml` allow-list. `cargo deny check` passes; a `deny.toml` comment records the new dependency surface. The caret pin (not a wildcard) keeps `wildcards = "deny"` satisfied. `ironbus-core` stays IO-free and the `ironbus-server` lib never sees `ctrlc`.

## Signal wiring + accept-wake design

- `cmd_serve` (Unix) installs a handler via `ctrlc::try_set_handler` (fallible, not panicking) that flips the **same** `shutdown: Arc<AtomicBool>` the serve loop already polls. The handler body is a single async-signal-safe atomic store; the serve loop unwinds the rest.
- The accept wake is **already** in place: the serve loop uses a non-blocking `accept()` that re-checks the shutdown flag every `ACCEPT_POLL` (~50 ms). So flipping the flag breaks the accept loop within a bounded time, not only after the next connection. No `accept()` change was needed; the mechanism is documented at the call site.

## Checkpoint-on-shutdown

After the serve loop returns (it stops accepting once the flag is set), `cmd_serve` forces a final checkpoint of **every** live work-group's committed cursor via the new `Engine::checkpoint_all_groups` (the default group plus every named group; not interval-gated, unlike `maybe_checkpoint_group`), then exits 0. A long-lived consumer still connected at the signal does not get to run its own close-path flush (its handler thread is detached, not joined), so this server-side flush is what makes the clean stop non-redelivering.

## Test

`graceful_shutdown_on_sigterm_checkpoints_the_cursor_and_does_not_redeliver` (Unix integration test, real `ironbus` binary + real `ironbus-client`):

1. Boots the broker with a **large** `--checkpoint-interval` (so the per-ack interval checkpoint never fires).
2. Produces 3, fetches, and acks all 3 over **one still-connected** client (so the clean-disconnect close-path flush never fires either). The shutdown flush is therefore the only path that can persist the cursor.
3. Sends `SIGTERM`, asserts the broker exits **0**.
4. Restarts on the same data dir and asserts **no redelivery**; the durable log continues from offset 3.

Verified non-vacuous: with the shutdown flush disabled, the restart redelivers all three acked messages and the test fails. A new engine unit test (`checkpoint_all_groups_flushes_every_live_group_for_a_graceful_shutdown`) covers the named-group flush the integration test does not exercise.

## Acceptance criteria

- [x] SIGTERM/SIGINT trigger a clean shutdown that checkpoints the cursor before exit, so a restart after a graceful stop does not redeliver acked messages.
- [x] The dependency approach (ctrlc) is wired and recorded.
- [x] A real gating test shows a signalled-then-restarted broker resuming WITHOUT redelivery.

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace`, and `cargo deny check` all clean.


ELares:
## Review resolution — SHIP-WITH-NITS, ready (2bce16b)

Independent adversarial reviewer (not the author) verified the load-bearing property and proved it non-vacuous: disabling the `checkpoint_all_groups` flush made the test fail with exactly the redelivery `[0, 1, 2]`, confirming the flush is what prevents it.

- No-redelivery-after-graceful-stop: SIGTERM/SIGINT/SIGHUP -> ctrlc handler thread flips the flag -> the serve loop exits -> `cmd_serve` flushes EVERY live group's cursor (`checkpoint_all_groups`) before exit 0. The integration test produces+fetches+ACKs 3 over a live client with `--checkpoint-interval 1000000` (so only the shutdown flush can persist the cursor), sends a real `kill -TERM`, asserts exit 0, restarts on the same data dir, and asserts NO redelivery (then the log continues at offset 3). Not `#[ignore]`d; 3/3 non-flaky.
- Accept loop wakes on a 50 ms non-blocking poll (bounded shutdown latency, no busy-spin).
- `ctrlc::try_set_handler` (fallible, no panic; `MultipleHandlers` -> typed error); handler body is a single atomic store; the `termination` feature covers SIGINT+SIGTERM+SIGHUP.
- `ctrlc` is in `ironbus-cli` only under `[target.'cfg(unix)'.dependencies]`; `cargo tree -p ironbus-core -i ctrlc` does not match (core stays dep-clean, IO-free check passes); transitive deps permissively licensed; `cargo deny check` ok; caret pin.
- Clean exit 0 on signal, non-zero on error; per-connection cleanup unchanged. No panic/unwrap in library paths; clippy/fmt clean; no en/em dashes; MSRV-safe (ctrlc/nix are 1.69).

Nits: the only blocker was the CHANGELOG rebase conflict (resolved here); the optional in-flight drain is not implemented (correct at-least-once behavior, honestly noted). All three acceptance criteria are met, so this closes #195.

CI: 15/15 green.

Closes #195

Issue #288, opened 2026-06-07
Broadcast-consumer cumulative ack (ack-all-up-to-offset)


ELares:
Follow-up to #63.

The cumulative-ack work-group GUARD has shipped: `Engine::cumulative_ack` / `cumulative_ack_in` HARD-REJECT a cumulative ack (ack-all-up-to-offset, the JetStream AckAll trap) on any competing or key_shared work-group with the typed `EngineError::CumulativeAckOnWorkGroup`. That is the safe half: cumulative ack on a shared, out-of-order-draining cursor would commit past peers' still-in-flight messages and silently drop them.

This issue tracks the OTHER half: the actual cumulative-ack path for BROADCAST consumers. A broadcast consumer is a group of one that sees every message in order, so committing its single cursor up to an offset is well-defined and safe. IronBus has no broadcast-consumer cursor type yet (every group today is a competing or key_shared work-group), so the guard currently rejects every cumulative ack.

Scope:
- Introduce a broadcast-consumer cursor (a single-member group that sees every record in order).
- Implement `cumulative_ack_in` for that path: commit the broadcast cursor up to `up_to` (validated against the durable head and the earliest-retained offset), idempotent on a re-ack.
- A wire surface for cumulative ack (owned by the verb/framing work, refs #11) and CLI exposure.
- Keep the work-group rejection unchanged.

refs #63

ELares:
Implemented in #325 (branch `feat/broadcast-cumulative-ack-288`).

- Broadcast-consumer cursor: a `WorkGroup.broadcast` flag (group-of-one that sees every record in order), set/read via `Engine::set_broadcast_in` / `is_broadcast_in`, mutually exclusive with key_shared.
- `AckCursor::commit_up_to` commits the watermark to an exclusive offset, idempotent and monotonic (re-ack is a no-op, never regresses).
- `Engine::cumulative_ack_in` accepts the verb only for a broadcast group; validates `up_to` against the durable head and earliest-retained (typed `CumulativeAckOutOfRange`, never a panic). The work-group HARD-REJECT (`CumulativeAckOnWorkGroup`) is unchanged.
- Wire: append-only `FrameType::CumulativeAck` tag 19 (body = exclusive `up_to` u64 LE then the group name); both frozen-tag tests updated, unknown-tag property narrowed to 20..=255.
[...]

PR #289, opened 2026-06-07
MaxDeliver validation, the BackOff schedule, and the cumulative-ack work-group guard (#63)

ELares:
Finishes the remaining parts of #63. The durable, crash-atomic DLQ move and the basic ack/nack/term/progress vocabulary already shipped (merged earlier); this PR adds the config and policy knobs that were still unwired.

## Already done (prior work, unchanged here)
- The durable DLQ sink and the crash-atomic, exactly-once poison move (idempotent on `(group, source_offset, attempt)`), the MAX_DELIVERIES advisory, and `nack` / `term` / `progress` semantics.
- The IO-free `DeliveryConfig` core (max-deliver disposition, the backoff schedule, the typed `ConfigError`) and the engine driving the backoff via the clock seam.

## What this PR adds
- **MaxDeliver config validation + WARN.** `serve --max-deliver 0` (or `u32::MAX`), both meaning unlimited delivery, is now allowed only behind a new `serve --allow-unlimited-deliver` flag plus a startup WARN; without the flag it is a usage error before the broker opens. The typed `DeliveryConfig::new` `ConfigError` already enforced unlimited-needs-the-opt-in; the CLI now surfaces the opt-in and the loud startup WARN instead of hard-rejecting.
- **BackOff schedule wired to the CLI.** A new `serve --backoff-ms 100,500,2000` flag sets the escalating per-attempt nack/redelivery backoff schedule (milliseconds, indexed by attempt, clamped to the last entry, applied when a nack carries no explicit delay). Omitted, the built-in default schedule is used; a single `0` disables backoff. The engine already drove the schedule for redelivery timing; this exposes it to operators.
- **Cumulative-ack work-group guard.** New `Engine::cumulative_ack` / `cumulative_ack_in` HARD-REJECT a cumulative ack (ack-all-up-to-offset, the JetStream AckAll trap) on any competing or key_shared work-group with the typed `EngineError::CumulativeAckOnWorkGroup`. Cumulative ack is safe only for a broadcast consumer (a group of one that sees every message in order), and IronBus has no broadcast-consumer cursor yet, so the guard rejects every cumulative ack. This is a single typed, non-panicking rejection point rather than a bare TODO.
- **Bug fix found while testing:** a value-less flag (`--allow-unlimited-deliver`) must advance the parse index by one; the first cut advanced by zero and spun the flag loop forever. Fixed and covered by the parse test.

## Deferred / out of scope (with follow-ups)
- **Broadcast-consumer cumulative ack** (the "accepted on broadcast consumers" half of the #63 criterion): filed as #288. Needs a new broadcast-consumer cursor type and a wire surface; building it from scratch here would be a large new feature, so this PR ships the guard and defers the feature.
- **DLQ retention >= source validation:** not applicable. DLQ retention is not configurable: the sink never reaps (a poison record is durable evidence and must never expire), so there is no short-DLQ-retention misconfiguration to reject. Noted rather than inventing config.

## Tests
ironbus-server engine:
- `cumulative_ack_is_rejected_on_a_work_group` (default, named competing, and key_shared groups; the rejection commits nothing)
- `the_cumulative_ack_rejection_renders_a_distinct_message`
- (existing, still green) `nack_with_no_delay_applies_the_backoff_schedule`, `an_explicit_nack_delay_overrides_the_backoff_schedule`, `the_backoff_schedule_escalates_across_attempts`

ironbus-cli:
- `validate_rejects_unlimited_max_deliver_without_the_opt_in`
- `validate_accepts_unlimited_max_deliver_with_the_opt_in`
- `the_delivery_config_rejects_unlimited_without_the_opt_in` (asserts the typed `ConfigError`)
- `serve_parses_the_allow_unlimited_deliver_flag`
- `take_number_list_parses_a_comma_separated_schedule`
- `take_number_list_rejects_a_bad_element`
- `serve_rejects_a_bad_backoff_ms`

Determinism and golden-path gates stay green. `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings` (pedantic), and `cargo test --workspace --all-features --locked` all clean locally; no em/en dashes.

refs #63 (remaining: the broadcast-consumer cumulative-ack path, tracked as #288)


ELares:

Independent reviewer (not the author) verified:
- **Parse-loop hang FULLY fixed:** `--allow-unlimited-deliver` advances the index by exactly 1 (the bug was advancing by 0, spinning); it is the ONLY value-less flag in the serve loop, every other arm advances via `take_value`/`take_number`/`take_number_list` (+2) or returns Err, and a missing value returns a Usage error rather than looping. No input can spin it; verified empirically.
- **Cumulative-ack guard correct and unbypassable:** `cumulative_ack`/`cumulative_ack_in` unconditionally return the typed `EngineError::CumulativeAckOnWorkGroup` for the default, competing, and key_shared groups; no panic. It is currently engine-internal/dormant (no wire caller yet, a typed landing point for #288), which the PR honestly notes.
- **MaxDeliver validation:** `max_deliver` 0/u32::MAX rejected unless `--allow-unlimited-deliver`, then a startup WARN; typed `ConfigError`, default 5; both paths tested.
- **BackOff schedule wired AND used:** `--backoff-ms` parses (robust against bad/empty/overflow elements, saturating ms->ns), and `nack_in` computes `effective_nack_delay(attempt, ...)` against the clock seam; existing backoff engine tests pass.
- **Deferrals real:** broadcast cumulative ack needs a new consumer type (#288, filed); DLQ retention has nothing to validate (the sink never reaps). So `refs #63` is correct.
- Engineering bar met (typed errors, no library panics, clippy/fmt clean, no dashes, ironbus-core IO-free, determinism + golden-path green, MSRV 1.78).

Non-blocking nit acknowledged: the WARN is written to stdout (consistent with the existing "listening on" startup line); moving it to stderr is a fine future one-liner.

CI: 15/15 green.

Issue #275, opened 2026-06-07
[TASK] Per-consumer BYTE budget and wire-advertised credit (follow-up to #65)


ELares:
Parent: #65 (Per-consumer credit and in-flight window). Related: #10, #11, #19, #20.

#274 delivered the per-consumer MESSAGE-COUNT credit (`EngineConfig.consumer_credit`, default 64, floored to 1, no unbounded opt-out), redelivery accounting (an expired lease frees the original consumer's slot, recounted against the next claimer, no double-count), per-consumer isolation (a stuck consumer pins only its own slots), and the effective bound `min(producer-side max_in_flight window, consumer credit)`. The DLQ `max_deliver` routing landed in #63.

Two parts of #65 were deliberately deferred and are tracked here so they are not lost as bare TODOs.

## Scope
- **Per-consumer BYTE budget.** Add `in_flight_bytes` (default 8 MiB) alongside the message-count credit so effective credit = `min(message credits remaining, byte credits remaining)`, hard floor of at least one message. A large-payload consumer must not blow the RAM ceiling despite a small in-flight message count. Decrement bytes on send, restore on ack/nack/term/lease-release, mirroring the message-count accounting, derived (not a separately mutated counter) so it cannot drift. Hand the byte budget to #20 for the RAM-ceiling cross-check.
- **Wire-advertised / negotiated credit.** Carry the per-consumer credit (and byte budget) in the Connect/Info handshake (#11 owns the frame), so the effective credit is negotiated `min(server cap, client request)` rather than only a server-side default. No `request(MAX)`-style unbounded value is representable on the wire.
- **Little's Law worked example committed** and cross-checked with #19 once the byte budget exists (the message-count headroom example is already in #65; the byte side needs #19).

## Acceptance criteria
- [ ] `in_flight_bytes` enforced; effective credit = `min(message, byte)` budget, floor >= 1 message; derived accounting, no drift.
- [ ] Byte budget restored on ack/nack/term/expired-lease exactly like the message credit; tested incl. redelivery re-occupying its bytes once.
- [ ] Credit + byte budget advertised/negotiated in the Connect/Info handshake (#11); no unbounded value representable on the wire.
- [ ] Little's Law byte-budget example committed and cross-checked with #19; RAM ceiling cross-checked with #20.
- [ ] Per-layer tests (engine byte-cap, session restore, end-to-end over a real server, handshake negotiation).

ELares:
[...]

PR #290, opened 2026-06-07
Add the ADR index cataloging the design decisions (#130)

ELares:
## What this catalogs

A new `docs/adr/INDEX.md`: a flat catalog of every resolved IronBus design decision, one row each with an ADR id, title, status, owning issue, and a one-line summary. It covers:

- The three numbered ADRs (0001 log-is-WAL, 0002 segments-never-recycled, 0003 lz4_flex-default-zstd-opt-in), each with a link to its file.
- The frozen decisions that do NOT yet have a numbered ADR file, drawn from the real source and each marked "ADR file pending" so a pending row is never read as an existing file:
  - The #139 frozen decisions: the codec/C-FFI posture, the `LossReport` serde decision (#120/#223), the single resync predicate (#5/#8), the log-is-WAL durability re-anchoring, the no-recycle/at-rest-nonce decision, the binary-only wire-protocol wording (#11), and the forward `[observability]`/`[auth]` config-table decision (#14, still open).
  - The frozen wire/on-disk format decisions documented in `docs/CONTRACTS.md`: the `[len:u32][type:u8]` frame envelope, the frozen tag set (Connect 1 .. Truncated 18), distinct response frames (#179), the 36-byte header / 8-byte trailer, and the xxh3-64 large-payload checksum (#146).
  - The durability/recovery decisions: fdatasync-before-ack (#6), torn-tail truncation (#7), the bounded-loss caps and I1 to I4 checkers (#120), and the visibility-timeout lease + DLQ model (#9/#63).
  - The dependency/default/bound decisions: graceful shutdown via `ctrlc` (#195), per-consumer credit default 64 (#65), and the named-group cap default 1024 (#240).

## Convention followed

A new `docs/adr/INDEX.md`, not an extension of `docs/adr/README.md`. That is the established place: `docs/adr/README.md` already pointed at the index, and issue #130's deliverable is explicitly "a single `docs/adr/INDEX.md`". The process README owns the lifecycle/numbering prose; the new INDEX owns the flat catalog. Both `docs/README.md` and `docs/adr/README.md` now link to it.

## Verification

Every entry was verified against the ADR files (`ls docs/adr/`), the owning issues (`gh issue view`), and the code (constants, frame tags, `deny.toml`, `Cargo.toml`), with the code / README / owning issue kept canonical:

- Statuses reflect the ACTUAL issue state: #146, #179, #195 are Accepted (closed); #65, #120, #240, and the config-tables decision are marked Open/Proposed because their issues are genuinely still open (credit byte budget #275, eviction lifecycle #277, the full checker harness, and the unbuilt TOML config).
- The stale #130 seed numbering (ADR-0007 and up, status CONFLICTED) is flagged as predating the #139 resolution pass, so a reader is never sent to a number that does not exist.
- The index cross-references `docs/CONTRACTS.md` and `docs/INVARIANTS.md` rather than duplicating their tables.

Used `refs #130` (not `Closes`): the index catalogs the resolved decisions and correctly marks the pending ones, but #130's own acceptance criteria include a machine-readable `adr.yaml` and a CI gate that this prose-index PR does not deliver, so the issue stays open for that follow-up (noted in the index).

Docs only: no crate source, `Cargo.toml`, or CI touched. `cargo fmt --all --check` and `cargo test --workspace` are green; no em or en dashes.


ELares:
Independent review (not the author): SHIP. The index is accurate and honestly scoped: the three numbered ADRs (0001 log-is-WAL, 0002 segments-never-recycled, 0003 lz4_flex-default) carry the correct Accepted status and owning issues; the 'ADR file pending' decisions are catalogued without inventing ADR numbers; unresolved items (#65/#120/#240, the #14 config tables) are correctly marked Open/Proposed rather than Accepted; and it explicitly keeps the code/README/owning-issue canonical (the index owns no decision). It also flags that the #130 seed-row numbering (ADR-0007/0011/... CONFLICTED) is stale and that the committed files renumber from 0001. Docs-only, no en/em dashes, CI green. `refs #130` is correct (the machine-readable adr.yaml + CI gate from #130's criteria are not delivered here).

Issue #292, opened 2026-06-07
[TASK] Wire-advertised / negotiated per-consumer credit in the Connect/Info handshake (follow-up to #275)


ELares:
Parent: #275 (Per-consumer BYTE budget and wire-advertised credit). Related: #65, #11, #106.

#275's SERVER-SIDE byte budget shipped (`EngineConfig.consumer_credit_bytes`, default 8 MiB, `0` = unlimited, effective credit = `min(message credits remaining, byte credits remaining)` with a floor of one message, derived from the connection-scoped in-flight set so it cannot drift; `serve --consumer-credit-bytes`). This issue tracks the deferred WIRE half so it is not lost as a bare TODO.

## Scope
- Carry the per-consumer message credit AND byte budget in the `Connect`/`Info` handshake (#11 owns the frame), so the effective credit is negotiated `min(server cap, client request)` rather than only a server-side default.
- No `request(MAX)`-style unbounded value representable on the wire.

## Why it was deferred from #275
The `Connect`/`Info` frame bodies are currently EMPTY (no negotiation). Adding negotiated fields is a FROZEN-FRAME change tied to #11 (wire protocol) and #106 (auth/capability handshake), and is larger and decision-heavy (field layout, version/capability flags, the absent-value encoding). #275's byte budget was the clearly-buildable, decision-light half and shipped on its own; this is the protocol-design half.

## Acceptance criteria
- [ ] Credit + byte budget advertised/negotiated in the `Connect`/`Info` handshake (#11); no unbounded value representable on the wire.
- [ ] Frozen-tag tests updated for the new (non-empty) handshake bodies.
- [ ] Client requests its desired credit/byte budget; server clamps to `min(server cap, client request)`.
- [ ] Per-layer tests (proto round-trip, session negotiation, end-to-end over a real server).

Once both this and the byte budget (#275) are delivered, #275 can be closed.

Issue #106, opened 2026-06-05
[TASK] Specify connection-scoped authentication and the three-scope authorization model


ELares:
Parent: #18

Specify the three v1 authentication mechanisms and the three-scope authorization model at the handshake level, consistent with the #11 CONNECT handshake. This is the core auth contract every verb is gated by.

## Approach
Define three mechanisms only (nkey/JWT is deferred, not flagged): bearer token (>=256-bit random, stored SHA-256, constant-time compared), username/password (Argon2id at the edge profile m=19 MiB, t=2, p=1), and mTLS (chain-verified to a configured CA). For mTLS, identity is the SAN (first URI SAN, else first DNS SAN); CN is never used, and a cert that verifies but matches no configured identity is rejected, not given a default scope. Authentication is established once in the #11 CONNECT handshake and bound to the connection for its lifetime (MQTT's session-scoped lesson). Authorization is exactly three scopes granted explicitly per identity: publish, subscribe, admin (admin does not imply the others). Specify which verbs and which #15 diagnostic commands require admin (anything exposing stored data or mutating retention/identities). Credential validity is an additive set (token_hashes = [...]) so rotation is two deploys with no server-side timer. Define the on-the-wire error as a single Authorization Violation with no oracle distinguishing bad-credential from insufficient-scope.

## Acceptance criteria
- [x] Token, username/password, and mTLS specified at the handshake level, consistent with #11.
- [x] mTLS identity = SAN with the URI-then-DNS rule; CN explicitly excluded; no-match rejected.
- [x] publish/subscribe/admin specified, granted explicitly, with the admin-gated verb and #15 command list.
- [x] Constant-time token compare and Argon2id password parameters specified.
- [x] Additive credential-set rotation semantics specified (no expiry timer).

## Failure considerations
The sharp risks are a timing oracle in bearer-token comparison (mitigate with constant-time compare and a uniform Authorization Violation), and an authz bypass where a verifying cert gets a default scope (mitigate by requiring explicit SAN-to-identity match). A leaked token is contained by the additive set: the operator adds the new hash, deploys, then removes the old, with no clock dependence.

ELares:
Specified and delivered in PR #316 (merged, squash `a2e12541`) as `docs/AUTHENTICATION.md`, on a fully green gate plus an independent adversarial review that confirmed all five acceptance criteria and both sharp-risk mitigations (the bearer-token timing oracle and the verifying-cert authz bypass).

[...]

PR #293, opened 2026-06-07
Add the per-consumer in-flight byte budget (#275)

ELares:
Implements the server-side per-consumer BYTE budget half of #275 (the deferred half of #65): a per-connection in-flight byte budget alongside the existing message-count credit, so a large-payload consumer cannot blow the RAM ceiling despite a small in-flight message count.

## What

- **Effective credit = `min(message credits remaining, byte credits remaining)`**, with a hard floor of ONE message. The message-count credit (#65) is unchanged; the byte budget is an additional `min()`.
- **Floor-of-one**: a single message larger than the whole byte budget is still delivered when the connection holds nothing in-flight, so an over-budget message never wedges the consumer. After it is in flight the budget binds normally.
- **`EngineConfig.consumer_credit_bytes`** (default 8 MiB = `DEFAULT_CONSUMER_CREDIT_BYTES`). Unlike the message credit (floored to 1), the byte budget is NOT floored: `0` = unlimited (the budget is off, only the message credit binds), matching the `0` = off convention of the other byte bounds (`max_total_bytes`, `max_retained_bytes`, `max_groups`).
- **`serve --consumer-credit-bytes <n>`** (`0` = unlimited). The CLI default aliases `ironbus_server::engine::DEFAULT_CONSUMER_CREDIT_BYTES`, so the CLI and engine defaults are a single source of truth and cannot drift.

## Byte-accounting design (mirrors #65)

The byte budget is **DERIVED** from the connection-scoped `leased` set, never a separately-mutated counter, so it cannot drift any more than the message count can. The session's `leased` map went from `offset -> generation` to `offset -> Lease { generation, bytes }`, where `bytes` is the message's key+headers+payload length (matching the engine's produced-bytes accounting). The in-flight byte total is the sum of those `bytes`. A delivery occupies its bytes; an ack, a successful nack/term, the per-batch prune of committed offsets, and the start-of-Flow release of leases the engine no longer holds ACTIVELY (the redelivery-accounting seam) all restore them. A redelivered message re-occupies its bytes exactly once (the redelivery overwrites the same offset key). The check is per-Flow, before each poll (a poll that returns a message has already leased it in the engine), so the in-flight total overshoots by at most the one message that crosses the budget, which is the standard credit semantic and stays bounded.

## Deferred: the wire-advertised credit

The other half of #275, carrying the per-consumer credit and byte budget in the `Connect`/`Info` handshake, is a FROZEN-FRAME change (the handshake bodies are currently empty) tied to #11 (wire protocol) and #106 (auth/capability). It is scoped OUT of this PR and filed as the linked follow-up **#292**. The frozen `Connect`/`Info` frames are untouched here.

## Tests

- **Engine**: `consumer_credit_bytes_passes_through_unfloored`, `a_zero_consumer_credit_bytes_means_unlimited`, `the_default_consumer_credit_bytes_is_eight_mib`.
- **Session**: `the_byte_budget_binds_before_the_message_credit`, `the_floor_of_one_lets_a_single_over_budget_message_through`, `acking_restores_the_byte_budget`, `nack_and_term_each_restore_the_byte_budget`, `an_expired_lease_restores_the_byte_budget_on_start_of_flow_release`, `a_redelivery_re_occupies_its_bytes_exactly_once`, `the_effective_credit_is_the_min_of_message_and_byte_both_directions`, `a_zero_byte_budget_is_unlimited`, `the_default_byte_budget_does_not_bind_small_messages`.
- **End-to-end over a real server** (client crate): `the_byte_budget_binds_over_a_real_server`, `the_byte_budget_floor_of_one_over_a_real_server`.
- **CLI**: `serve_parses_the_consumer_credit_bytes_flag`, `the_consumer_credit_bytes_default_is_eight_mib`, `serve_accepts_a_zero_consumer_credit_bytes_meaning_unlimited`, `serve_rejects_a_non_numeric_consumer_credit_bytes`, `usage_lists_the_consumer_credit_bytes_flag`.

`cargo fmt --all` clean, `cargo clippy --workspace --all-targets -- -D warnings` (pedantic) clean, `cargo test --workspace` green, the determinism image is unchanged (the new field is `0` = off in that config), and `ironbus-core` stays IO-free.

refs #275, #65, #292


ELares:

Independent reviewer (not the author) verified the credit-accounting restructure is safe:
- **#65 message-count credit UNCHANGED:** the `HashMap<u64,u64>` -> `HashMap<u64,Lease{generation,bytes}>` change is behavior-preserving; every mutation site (ack fence, nack/term, progress, per-batch prune, truncation prune, stale-lease release, SUB/UNSUB clear) is structurally identical on a `Lease` instead of a `u64`, and `remaining = ceiling - leased.len()` is unchanged. All six #65 tests pass.
- **Byte accounting DERIVED, cannot drift:** `in_flight_bytes()` sums the map (`leased.values().map(|l| l.bytes)`, saturating); there is NO separate counter to drift. `lease_bytes = key+headers+payload` exactly matches the engine's `produced_bytes`. A redelivery overwrites the same offset key (stale entry freed first by `release_stale_leases`), so bytes re-occupy exactly once.
- **Floor-of-one never wedges:** the byte gate is `ceiling_bytes != 0 && !leased.is_empty() && in_flight >= ceiling`, so an over-budget message is always delivered when nothing is in flight, and blocked once something is.
- **Effective = min(message, byte) both directions;** `0` = unlimited (pure #65); the two floors compose (a zero message credit still delivers nothing). Overshoot bounded by one message; all arithmetic saturating.
- Default 8 MiB; CLI const aliases the engine const (no drift); no library panic/unwrap/unsafe; clippy pedantic + fmt clean; no en/em dashes; ironbus-core IO-free; determinism image unchanged; MSRV 1.78; frozen Connect/Info frames untouched. `refs #275` correct (wire-advertised half deferred to #292, justified).

Non-blocking NITs noted: the USAGE help hardcodes `8388608` (bounded by tests); the `#[allow(clippy::too_many_lines)]` on the flat flag-parse fn is a reasonable judgment call.

CI: 15/15 green.

PR #296, opened 2026-06-07
Idle named-group eviction completing the #240 lifecycle (#277)

### LINKED ISSUE #277: [TASK] Idle named-group eviction / lifecycle policy (follow-up to #240)

ELares:
Implements the deferred eviction / lifecycle half of #240 (tracked in #277): the #240 PR landed the work-group CAP; this lands the eviction so a long-lived broker RECLAIMS idle named groups instead of only bounding their count. The cap bounds how many groups live; this drops the idle ones, so per-group `AckCursor` + `LeaseTable` state does not accumulate for groups no consumer touches.

## Eviction rules (SAFE, conservative)
A group is evicted ONLY when it is:
- NAMED (the default group `""` is NEVER evicted; it is the durable wire group),
- PLAIN COMPETING (a `key_shared` group carries live-member state and is left to its membership lifecycle),
- FULLY CAUGHT UP: committed cursor at the durable head (`committed == flushed`) with NO acked-ahead set,
- LEASE-FREE: its `LeaseTable` holds no lease (no consumer mid-work),
- IDLE: no poll / ack / nack / progress / term touched it for at least the configured window.

A group BEHIND the head is NEVER evicted.

## Never-lose-committed-position invariant
- Only fully-caught-up groups are dropped, so re-creation at the head loses nothing and redelivers nothing acked. A behind group is, by definition, not idle in the meaningful sense and is kept.
- Before a caught-up group is dropped it is DURABLY CHECKPOINTED AT THE HEAD; a checkpoint-write failure KEEPS the group (never trades a committed position for a disk hiccup). The `cursor-<hex>.ckpt` is KEPT, never deleted.
- A re-subscribe RESUMES from that checkpoint: new `Engine::ensure_group` resumes a re-created group from its durable checkpoint (the in-process counterpart to open-time recovery), so an evicted-then-recreated group resumes at the head rather than resetting to offset 0 and redelivering the whole already-acked log.
- The durable checkpoint FORMAT is unchanged.

## Sweep seam (clock-driven, no thread)
The sweep runs from deterministic seams, never a background janitor:
- at the START of every poll (`poll_in` / `poll_in_member`), BEFORE the cap gate, so an evicted slot is freed in time to admit a new group on the same poll at a previously-full cap;
- after every successful produce.

It reads the existing `now_monotonic` clock seam (no `SystemTime`/`Instant::now`). Per-group last-activity is updated on poll/ack/nack/progress/term; the polled group is refreshed BEFORE the sweep, so a self-poll of an otherwise-idle group never evicts-and-recreates it. `Engine::sweep_idle_groups_now` is the caller-supplied-`now`-free public entry for the owning actor.

## Config flag + 0-semantics
`EngineConfig.group_idle_evict_ms` (default `0` = DISABLED / never evict, matching the `0` = off convention of the other bounds), exposed as `serve --group-idle-evict-ms <n>`, documented in USAGE. Off by default so an operator opts in deliberately; an unconfigured broker is byte-for-byte unchanged (the disk image / determinism sweep is unaffected).

## Unsub interaction
An explicit `Unsub` of a now-idle named group reclaims it IMMEDIATELY via `Engine::evict_group_if_idle`: it enforces every position-safety clause of the sweep (named, plain competing, caught up with no acked-ahead set, lease-free) but skips the idle-window wait (the explicit Unsub stands in for it). The session UNSUB path captures the leaving group and calls it, so a caught-up, lease-free group is freed at once; a group still holding an in-flight lease (the consumer unsubbed without acking) is KEPT until the lease drains or expires, then the natural sweep reclaims it.

## Tests (deterministic, via the clock seam)
Engine: idle caught-up group IS evicted (slot freed, re-creatable); in-flight-lease / behind / acked-ahead / default group are NOT evicted; eviction frees a slot so a new group is created at a previously-full cap; a re-subscribed evicted group resumes at head and redelivers nothing it had acked; `0` = disabled evicts nothing; polling your own idle group keeps it alive; explicit-unsub reclaim and its no-ops (behind / leased / default / unknown / disabled).
Session: UNSUB reclaims a caught-up named group end to end (re-sub resumes at head); UNSUB does NOT reclaim a group with an unacked in-flight lease.
CLI: `--group-idle-evict-ms` flag parse / default(0) / explicit-0 / non-numeric / usage.

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings` (pedantic), and `cargo test --workspace` are all green; determinism + golden-path unchanged.

Out of scope (no follow-up needed): the durable checkpoint FORMAT is untouched; behind groups are kept (not evicted); no background janitor thread.

Closes #277


ELares:

Independent reviewer (not the author) attempted to break the never-lose-committed-position invariant along every path and could not:
- **Durability ordering correct:** `evict_group` writes the checkpoint FIRST (`write_group_checkpoint` -> `Checkpoint::write` = `write_all_at` + `sync_all`, alternating slot) then `groups.remove` only after the `?` succeeds. A checkpoint WRITE FAILURE propagates before the remove, so the group is KEPT and keeps serving (sweep and Unsub paths both keep-on-failure). Never drop-then-fail.
- **Re-subscribe resumes at head:** evicted at head=10, +5 produced, re-subscribe via `ensure_group` -> `resume_cursor_from_snapshot` delivers 10..14, no acked redelivery, no loss (test `a_re_subscribed_evicted_group_resumes_at_head_and_redelivers_nothing_acked`).
- **Stale on-disk ckpt:** eviction checkpoints the LIVE cursor at head UNCONDITIONALLY before drop, so no stale-ckpt-to-head gap can redeliver.
- **Caught-up is race-free:** `flushed` is the durable high-water mark; the produce-seam sweep runs AFTER `log.sync()`, and the engine is single-writer with no interleave, so no group is ever dropped ahead of its durable ckpt.
- **Acked-ahead and leased groups excluded:** `is_evictable` requires `committed == flushed && ahead_ranges().is_empty() && leases.in_flight() == 0` (even an expired-unreclaimed lease keeps the group); key_shared groups and the default group are never evicted.
- **`ensure_group` is byte-identical to open-time recovery;** the sweep is clock-seam-driven (no thread, no SystemTime), deterministic, self-poll-safe; default `group_idle_evict_ms = 0` disables it and leaves the disk image byte-for-byte unchanged.

Non-blocking NIT noted: a structurally-dead `or_insert_with` belt-and-suspenders in `poll_in` (already commented; harmless).

CI: 15/15 green. `Closes #277` justified (the eviction lifecycle is complete and safe; checkpoint-format change, behind-group eviction, and a janitor thread are correctly out of scope).

Issue #297, opened 2026-06-07
[TASK] Edge-device ALICE crash-consistency + dm-flakey/dm-dust block-layer conformance run


ELares:
Parent: #21. Follow-up to #123.

The deterministic, in-tree corruption corpus and the single-bit-flip recovery sweep shipped in #123 (Tier A: hand-written canonical poison fixtures plus the exhaustive bit-flip sweep, gating every PR under `cargo test`). This issue tracks the HARDWARE-DEPENDENT remainder of #123 that cannot be a deterministic in-tree gate:

- The ALICE / CrashMonkey-style crash-prefix conformance run on the reference aarch64 EDGE DEVICE: record the real syscall/block trace of a produce/consume workload and replay every crash prefix (plus reorder) against the recovery checker, catching a missing parent-directory fsync the IO seam abstracts away.
- Block-layer fault injection on Linux: `dm-flakey` (dropped / intermittent writes on a timer) and `dm-dust` (per-block read errors / bad sectors), exercising the real IO stack and the corruption-skip path end to end. Linux-only and root-only, so these live in a privileged CI container and on the reference device, NOT the per-PR unit suite (per #20).
- The differential conformance gate: the SAME scenarios run on the deterministic sim and on the real impl; recovered state and the structured loss reports must MATCH, and divergence is release-blocking. This is the oracle that keeps the deterministic simulation honest against real hardware.
- Tier C of the corpus: real bit-flipped segments captured from a wear-stressed eMMC/SD device, refreshed per hardware generation, with a versioned manifest recording sha256, provenance, and the expected `LossReport` per fixture.

These run nightly and pre-release, not on the per-PR critical path, so the in-tree corpus stays deterministic and fast while the hardware lanes provide the real-device oracle.

refs #123, #21

ELares:
Add to this issue's corpus scope (surfaced by the #298 review): a fixture for a CORRUPT MIDDLE SEALED segment in a multi-segment chain (flipped body CRC / truncated body in a non-final sealed segment), pinning the 'no silent gap across a segment boundary' behavior. Recovery already handles it correctly today (fails closed with UnsealedPredecessor, verified by the #298 reviewer), so this is a coverage addition, not a bug. The #298 corpus only mutates segment 0; the multi-segment cases there corrupt structure (chain gap, unsealed predecessor), not a sealed middle body.

ELares:
[...]

PR #298, opened 2026-06-07
Add the storage corruption corpus and the bit-flip recovery sweep (#123)

ELares:
## What

The buildable, per-PR-gating tier of #123 (Tier A): a structured, named corpus of corrupt on-disk inputs plus a single-bit-flip recovery sweep, in a new `crates/ironbus-storage/tests/corruption_corpus.rs`. Each case builds a known-good segment, applies one precise mutation, then drives the REAL recovery path (`Log::open` / `SegmentReader::scan_recovery`) and asserts a CONCRETE outcome, not merely "does not panic". It runs per-PR under `cargo test`, deterministic (a fixed `ManualClock`, fixed payloads, no real clock or randomness).

## The corpus cases and their asserted outcomes

Each case asserts the recovered prefix length, the emitted `ReasonCode` / `LossEvent` span, the truncation offset (recovery never reads past the torn tail), and that loss stays within the bounded-loss caps, through the shared `invariants` checkers (I2 longest-valid-prefix, I3 bounded loss):

| Case | Test | Asserted outcome |
| --- | --- | --- |
| Torn tail, partial header | `torn_tail_partial_record_header` | keep n-1, `TornTail` span over the partial bytes, truncate to prior record end |
| Torn tail, partial body | `torn_tail_partial_record_body` | keep n-1, `TornTail`, truncate to frame start |
| Torn tail, partial trailer | `torn_tail_partial_record_trailer` | keep n-1, `TornTail`, truncate to frame start |
| Flipped record header CRC | `flipped_record_header_crc` | keep n-1, `CorruptRecordHeader` |
| Flipped record body CRC | `flipped_record_body_crc` | keep n-1, `CorruptRecordBody` |
| Flipped xxh3 field (over threshold) | `flipped_xxh3_field_on_over_threshold_record` | keep the small first record, `CorruptRecordBody` at the big frame |
| Bad record magic | `bad_record_magic` | keep n-1, `CorruptRecordHeader` |
| Unsupported record version | `unsupported_record_version` | keep n-1, `CorruptRecordHeader` |
| Unsupported segment-header version | `unsupported_segment_header_version` | fail closed, `SegmentError::UnsupportedVersion` |
| Unsupported segment-header checksum_algo | `unsupported_segment_header_checksum_algo` | fail closed, `SegmentError::UnsupportedChecksumAlgo` |
| Flipped segment-header CRC | `flipped_segment_header_crc` | fail closed, `SegmentError::BadCrc` |
| Short segment header | `truncated_short_segment_header` | fail closed, `SegmentError::Truncated` |
| Truncated footer | `truncated_footer_recovers_records_unsealed` | recover all records as unsealed (seal not trusted) |
| Footer/header segment-id mismatch | `footer_header_segment_id_mismatch` | fail closed, `FooterSegmentMismatch` |
| Recycled frame, stale sequence | `recycled_frame_with_a_stale_sequence` | fail closed, `RecoveredSequenceMismatch` |
| Segment-chain gap | `segment_chain_gap` | fail closed, `SegmentChainBroken` |
| Unsealed non-final predecessor | `unsealed_non_final_predecessor` | fail closed, `UnsealedPredecessor` |
| All-zeros record region | `all_zeros_record_region` | recover 0 records, `CorruptRecordHeader` over the zeroed bytes |
| All-zeros whole segment | `all_zeros_whole_segment_fails_closed` | fail closed, `SegmentError::BadMagic` |

## The bit-flip proptest

- `single_bit_flip_sweep_over_every_offset_never_panics`: the worst-case EXHAUSTIVE sweep, flipping every (byte, bit) of a known-good segment one at a time, asserting recovery is always a typed result (never a panic or an OOB read) and a successful recovery is always a valid prefix of the original (the I2 "recovery equals a valid prefix" invariant), with loss within the caps.
- `single_bit_flip_sweep_is_a_valid_prefix_or_clean_error`: the proptest twin (seeded, shrinking) the nightly deep sweep drives at `PROPTEST_CASES`. Both share one oracle helper, so a regression a fuzz target would find that this catches becomes a permanent regression here.

## Real recovery bug found+fixed

None. The exhaustive sweep and every taxonomy case pass against the EXISTING library: recovery already handles the full taxonomy correctly (the prior `scan_recovery`, the `LossReport` / `ReasonCode` taxonomy, the bounded-loss `check_caps`, and the chain/sequence/footer validation). The corpus is therefore a permanent regression net rather than a bug fix. The highest-value outcome would have been a real bug; here the prior hardening already holds, and the corpus pins it.

## Nightly block-layer addition

No new flaky job on the per-PR critical path. The existing nightly `deep-sweep` lane already runs `cargo test --workspace` at `PROPTEST_CASES=50000`, which now automatically drives the corpus's bit-flip proptest at depth; its comment is updated to say so. The heavier block-layer fault injection (dm-flakey / dm-dust) is hardware/root-dependent and part of the deferred edge work below, so it is NOT added here.

## Deferred edge-ALICE follow-up

The hardware-dependent remainder of #123 (the ALICE / CrashMonkey crash-prefix run and the dm-flakey / dm-dust block-layer fault injection on the reference aarch64 edge device, the sim-vs-device differential gate, and real bit-flipped eMMC/SD captures) is deferred to **#297**, since it needs a privileged Linux device, not a deterministic in-tree gate.

## Verification

`cargo fmt --all` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean (pedantic), `cargo test --workspace` green (21 corpus tests + the existing crash_recovery and determinism gates still pass). No `unsafe`, no em/en dashes, `ironbus-core` stays IO-free.

refs #123, #21


ELares:
## Review resolution — SHIP-WITH-NITS, addressed (d1260af)

Independent adversarial reviewer (not the author) verified this is a genuine, high-quality regression net: all 19 taxonomy fixtures assert CONCRETE outcomes (spot-checked 11 against the real `scan_body_streaming`/`recover`/`codec` source, every reason code / loss span / truncation offset / error variant CORRECT), they drive the REAL `Log::open` recovery path (no test backdoor), the bit-flip sweep is exhaustive (every byte x bit) and asserts the I2 valid-prefix oracle (not just no-panic) with a seeded proptest twin the nightly deep-sweep drives at PROPTEST_CASES=50000. The reviewer independently tried to break recovery (valid-CRC-length-past-segment OOB, corrupt/truncated middle segment) and found NO missed bug, confirming the author's honest "found none".

Addressed:
- SHOULD-FIX (footer assertion too weak): `truncated_footer_recovers_records_unsealed` now also asserts the single structured TornTail loss over the 28 leftover footer bytes (span, reason, bytes-skipped, truncation point) via `assert_single_loss`, so a regression that mis-reported that loss fails.
- SHOULD-FIX (corrupt-middle-segment fixture): deferred to #297 with a note. Recovery already handles it correctly (fails closed with UnsealedPredecessor, verified by the reviewer), so it is a coverage addition, not a bug; #297 (the corpus follow-up) now tracks it.
- NIT (DLQ high-water-mark reconstruction): noted; segment-level DLQ corruption is covered transitively (the `dlq/` sink is a second `Log` via the same `Log::open`), and the checkpoint torn-slot cases are covered by pre-existing tests.

`refs #123, #21` correct (the hardware edge-ALICE / dm-flakey run is deferred to #297). CI 15/15 green; storage suite 189 unit + 21 corpus + 21 crash_recovery + 1 determinism.

Issue #299, opened 2026-06-07
[TASK] Mutating admin actions surface (consumer reset, DLQ redrive, force-reap)


ELares:
Follow-up to #99 (the read-only `/admin` introspection endpoint).

#99 deliberately shipped READ-ONLY introspection only: `GET /admin` returns a JSON operational snapshot and no path can mutate engine state, sharing `/metrics`'s unauthenticated loopback/trusted-network trust model (the #105/#107 threat model). A mutating admin surface is a distinct, higher-risk concern (unauth-RCE-over-introspection) and was kept out of scope.

This issue tracks the MUTATING admin actions, which must NOT share the unauthenticated trust model and need their own design:

- Consumer reset (move a work-group's committed cursor / reset a group).
- DLQ redrive (re-inject dead-lettered records from the durable sink back onto the main log).
- Force-reap (operator-triggered segment reclamation).

Requirements to settle before implementing:
- Authentication / authorization (these MUST NOT be unauthenticated like `/metrics` and read-only `/admin`); ties to #106 (auth) and #107 (TLS).
- A separate route surface and verb (POST/PUT, not GET) so a mutating action can never be reached by the read-only path.
- Audit logging of every mutation.

refs #99

ELares:
PR #407 lands the **OFFLINE, auth-free subset** of this surface: `admin consumer-reset` and `admin dlq-redrive`, which operate on a STOPPED broker's `--data-dir` under the same exclusive data-dir lock `serve` holds (they refuse a running broker, exit 5). These have no network surface, so they need no auth: the operator already has filesystem access, and the safety boundary is the broker being stopped plus the dirlock.

- **consumer-reset** rewrites a work-group's durable cursor checkpoint to `--to <offset|earliest|latest>`, clamped to the durable range (out-of-range rejected), reusing the broker's exact dual-slot CRC checkpoint + `AckCursor` snapshot codecs so the broker resumes the group from there.
- **dlq-redrive** re-injects the dead-lettered records onto the main log crash-safely (append+fsync, then advance a durable `dlq-redrive.ckpt` watermark) and idempotently (a re-run after a completed redrive re-injects nothing).

[...]

Issue #105, opened 2026-06-05
[TASK] Write the enumerated edge threat model and the secure-defaults catalog


ELares:
Parent: #18

Produce the written, reviewable artifact that the parent's acceptance criteria require but only describe in prose: an enumerated edge threat model and a catalog of secure defaults where every default is justified against a specific threat. This is the foundation document the other security sub-issues implement against, so it lands first.

## Approach
Write an asset-threat-mitigation-residual table (STRIDE-flavored) with one row per concrete threat: physical device theft, untrusted-LAN passive eavesdrop, untrusted-LAN active MITM, stolen bearer token, credential brute force by reconnect, pre-authentication handshake/connection flood, malicious admin-scope misuse, and supply-chain crate compromise. Each row records detection, mitigation, and accepted residual risk; the at-rest-encryption caveat (key co-located with a stolen disk defeats it) is recorded explicitly as accepted residual risk, not buried in prose. Then enumerate the secure defaults as a catalog: localhost-only bind, network bind requires TLS plus at least one auth identity, TLS 1.3 only, no anonymous network access, fail-closed on unsafe secrets-file permissions, pre-auth DoS limits on by default. Each default cites the threat row it mitigates. Cross-reference #14 for where each default lives and #20 for the physical-access realities.

## Acceptance criteria
- [ ] Threat table with the eight rows above, each with detection, mitigation, and residual risk.
- [ ] At-rest-encryption theft limitation recorded as an explicit accepted residual risk.
- [ ] Secure-defaults catalog where every entry maps to a specific threat row.
- [ ] Pre-auth denial of service appears as its own threat row (not folded into brute force).
- [ ] Document reviewed and linked from the parent issue.

## Failure considerations
The main risk is a threat model that reads as comprehensive but silently omits a category (the first draft missed pre-auth DoS). Mitigate by deriving rows from the STRIDE categories and from each tenet boundary, and by requiring every later security sub-issue to trace its mitigations back to a row here, so an unmapped mitigation or an unmitigated row is visible during review.

Issue #99, opened 2026-06-05
[TASK] /admin read-only introspection contract and feature-gated tracing/OTLP


ELares:
Parent: #16

Define the read-only `/admin` JSON contract that #15 consumes to render segments, consumers, lag, and last-skip-offset from a single endpoint, and make tracing/OTLP export demonstrably optional with documented zero cost when off. The CLI must be able to drive its human views from `/admin` alone without parsing metric names.

## Approach

`/admin` is read-only JSON, versioned via `Accept: application/vnd.ironbus.admin.v1+json`, bound to localhost by default with the same widen-requires-auth rule as `/metrics`. It exposes sub-resources for segments, consumers (with incremental lag), config snapshot, and resilience state (last-skip-offset, frozen flag, skip totals). All mutation is explicitly out of scope and deferred to a separate actions surface owned by #18 and #14. The Rust `tracing` crate plus JSON log layer is compiled in by default; OTLP span export is feature-gated and off at runtime by default, with head-based sampling defaulting to 0.0 while ERROR/WARN events are always recorded. Export uses a bounded lossy queue that drops (and counts) spans rather than blocking the thread-per-core core. A `build-dist`-owned `edge-min` profile compiles export out entirely. Metric and label names are frozen as a stability contract under #22; the CLI never parses metric names for logic.

## Acceptance criteria

- [ ] #15 can render segments, consumers, lag, and last-skip-offset from `/admin` v1 JSON alone.
- [ ] `/admin` is read-only; no route mutates state, and the schema version is pinned in the Accept header.
- [ ] With export off, a benchmark shows no measurable CPU/RAM cost from tracing beyond plain JSON logging.
- [ ] OTLP export uses a bounded queue that drops and counts spans under pressure instead of blocking.
- [ ] The `edge-min` build omits OTLP export and the binary shrinks measurably; a golden test fails CI on any metric name/type/unit change without a contract bump.

## Failure considerations

A mutating admin surface risks unauth-RCE-over-introspection, so this task is strictly read-only. Enabling export on a constrained node could stall the core, so it defaults off with a lossy bounded queue and an edge-min compile-out. A silent metric rename would break dashboards and the CLI, so names are a #22-gated contract enforced by a golden test.

ELares:
PR #300 delivers the READ-ONLY introspection half of this issue: an opt-in `GET /admin` JSON endpoint on the embedded health server, off by default behind `serve --enable-admin`.

Acceptance criteria addressed by #300:
[...]

PR #300, opened 2026-06-07
Add the read-only /admin introspection endpoint behind a feature flag (#99)

ELares:
## Summary

Adds an OPT-IN, READ-ONLY `/admin` JSON introspection endpoint on the embedded health server (the same server as `/healthz`, `/readyz`, `/metrics`), for debugging operational state without any mutating capability.

This PR uses `refs #99` (not `Closes`): it delivers the read-only introspection contract + feature-flag half (the endpoint serves the segments/consumers/lag/last-dead-lettered data the CLI human views need from a single endpoint without parsing metric names), but #99 also covers the feature-gated tracing/OTLP export + `edge-min` compile-out + golden metric-name contract test, which stay open. So this references #99 rather than closing it.

## Exposed fields (one JSON snapshot, read under one engine lock)

- `schema_version` (pinned to `1`).
- `broker`: `healthy`, `flushed_offset` (durable head), `committed_offset`, `earliest_retained_offset`, `consumer_lag`, `durable_record_bytes`, `durable_record_count`, `segment_count`, `recovery_truncated_bytes`, and the counters `produced`, `produced_bytes`, `produce_rejected`, `delivered`, `redelivered`, `dead_lettered`, `acks`, `segments_reaped`, `segments_force_reaped`.
- `groups[]`: per-work-group `name` (the default group `""` included), `committed_offset`, `consumer_lag`, `in_flight`.
- `dlq`: `records` (durable dead-letter depth) and `last_dead_lettered_offset` (`-1` if none).
- `config`: the effective bounds echo, `max_total_bytes`, `max_segment_bytes`, `max_retained_bytes`, `max_age_ms`, `max_messages`, `max_in_flight`, `consumer_credit`, `consumer_credit_bytes`, `max_deliver`, `max_groups`, `group_idle_evict_nanos`, `visibility_nanos`, `hard_cap_nanos`, `disk_full_policy`.

Every field derives from an existing read-only engine accessor; the only new accessors added are a read-only `Log::segment_count`, `Log::config`, `Engine::segment_count`, and `Engine::config_snapshot`. JSON is hand-rendered (no serde dependency, matching the hand-rendered Prometheus text). `ironbus-core` is untouched and stays IO-free.

## Gating + trust model

- OFF by default. The operator enables it with `serve --enable-admin` (a bare boolean flag), which the health thread passes to `serve_health(..., admin_enabled)`. It needs `--health-addr` (the admin route is served by the health server); `--enable-admin` without it emits a startup WARN.
- When disabled, `/admin` is `404`, indistinguishable from any unknown path, so the surface is invisible unless turned on.
- READ-ONLY and UNAUTHENTICATED, sharing `/metrics`'s trust model (loopback or a trusted network, the #105/#107 threat model). It exposes no mutating action and no secret material.

## Strictly read-only scope

GET-only: a non-GET is `405` (the method check precedes the path match), and no route can mutate engine state. Mutating admin actions (consumer reset, DLQ redrive, force-reap) are explicitly out of scope and deferred to a separate mutating-admin surface, filed as #299 (which must NOT share the unauthenticated trust model).

## Tests

- `admin_returns_a_json_snapshot_with_correct_values` — produces 3, group `orders` acks one, group `billing` leases one; asserts the broker counters, per-group committed/lag/in-flight, DLQ sentinel, and the distinctive non-default config echo, with a JSON content type.
- `admin_is_404_when_the_flag_is_off`
- `admin_rejects_a_non_get_method` (POST/PUT/DELETE/PATCH all 405)
- `admin_reports_the_dlq_state_after_a_dead_letter`
- `admin_group_name_is_json_escaped`
- `serve_enable_admin_is_off_by_default_and_set_by_the_flag` (CLI flag parse)

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` (pedantic), and `cargo test --workspace` all green; determinism and golden-path unchanged.


ELares:

Independent reviewer (not the author) verified the load-bearing invariant and the security posture:
- **Strictly READ-ONLY:** the method gate (`!= GET` -> 405) precedes the `/admin` match; every accessor reachable from `admin_snapshot` (`flushed/committed/earliest offsets`, durable bytes/count, segment count, counters, `group_consumer_stats`, `config_snapshot`, DLQ depth/last-offset) is `&self`, non-mutating, no interior mutability, no reap/checkpoint/cursor-advance, and NO lazy group/DLQ creation (`groups.get(...).map_or(...)`, not `entry().or_insert`). Exact-path match after query-strip, so `/admin/...` falls to 404; no path can mutate.
- **OFF by default:** `--enable-admin` defaults false; the route is a match guard that falls through to the shared 404 arm when disabled (identical to any unknown path). Tested.
- **No secret leak / honest trust model:** only operational state + numeric config bounds (no path/token/payload/key); honestly documented as UNAUTHENTICATED, sharing /metrics's loopback/trusted-network trust model (cross-referenced #105/#107). Single-lock-consistent snapshot (locks once into an owned struct, renders lock-free), poison-safe.
- **JSON valid on hostile input:** `escape_json` correctly escapes `\\ " \n \r \t` and all control bytes; the `-1` DLQ sentinel and u64 counters render as bare numbers; field accuracy spot-checked against a known produce/ack state.
- Strictly read-only scope; mutating actions (consumer reset, DLQ redrive, force-reap) deferred to #299 with a note they must NOT share the unauthenticated trust model. `refs #99` correct (tracing/OTLP half remains).

Non-blocking NIT acknowledged: the tests assert on substrings rather than a full JSON parse (the crate deliberately avoids a serde runtime dep), and the escape test does not drive a hostile name end-to-end; the `escape_json` implementation is verified correct and group names are graphic-ASCII-validated, so JSON validity is not at risk.

CI green.

Issue #284, opened 2026-06-07
[TASK] Make the macro-bench injected-stall self-test a reliable CI gate


ELares:
Follow-up to #111. The macro-bench harness and its coordinated-omission self-test (`an_injected_sigstop_shows_up_in_the_recorded_tail`) landed in #111, but the LIVE self-test is unreliable on GitHub's shared CI runners and so is marked `#[ignore]` (run on demand with `cargo test -p ironbus-bench -- --ignored`).

## The problem
The self-test SIGSTOPs the shipping broker mid-run and asserts the ~200 ms freeze appears in the recorded p99.9 tail. On a stable host this is solid (it passes repeatedly locally). On `ubuntu-latest` / `macos-latest` shared runners it fails two ways:
- Sometimes the broker resets the client connection during the freeze (handled now: the harness ends gracefully and reports measured samples, see `is_connection_ended`).
- More fundamentally, the SIGSTOP freeze often does NOT manifest in the tail at all: a representative failing run recorded 106 samples with max latency only 2.6 ms (the 200 ms freeze produced zero high-latency samples), i.e. the freeze did not delay any client op on that runner. This is an environment/scheduling artifact of the shared runner, not a harness-logic bug.

## Scope
Make the injected-stall proof a RELIABLE gate, one of:
- Run it on a dedicated stable / self-hosted runner where SIGSTOP timing is deterministic, gated as a required check there.
- Or replace the OS-SIGSTOP freeze with an in-broker fault-injection seam (a test-only "freeze for N ms" hook) so the stall is deterministic and not subject to runner scheduling, while keeping the open-loop intended-send-time measurement honest.
- Or run it in the nightly workflow with a tolerance/retry policy if a stable runner is unavailable.

## Acceptance criteria
- [ ] The injected-stall coordinated-omission proof runs as a gating check that is NOT flaky (no spurious pass or fail across at least 20 consecutive runs).
- [ ] The `#[ignore]` is removed (or the test is wired into the chosen reliable gate).
- [ ] The proof still measures from the intended send time and still fails if the freeze is absent from p99.9 (non-vacuous).

PR #302, opened 2026-06-07
Document the SLO target table and the ratification process (#110)

ELares:
## What this covers

Adds `docs/SLO.md`: the service-level-objective target table for IronBus and the process by which a target becomes a ratified, versioned SLO. Derived from the README's Performance targets section and from issues #19 (the SLO and methodology parent) and #110 (ratify and version the table). Linked from `docs/README.md`, with a terse `## [Unreleased] / ### Added` CHANGELOG entry.

Each SLO metric is tabulated with its DEFINITION tied to the exact field the macro-bench harness (`crates/ironbus-bench/`, #111) actually emits, its TARGET quoted from the README or the issues, its STATUS, and the conditions/profile it applies to (default vs edge):

- Throughput (msg/s, MB/s) -> `results.msgs_per_sec`, `results.mb_per_sec`
- Latency p50 / p99 / p99.9 -> `results.{p50_us,p99_us,p999_us}` from the HdrHistogram (1 us to 60 s, 3 sig figs)
- Steady-state RAM ceiling -> `results.steady_rss_bytes` (the #115 `tiny` 64 MiB budget)
- Write amplification -> `results.write_amplification`
- The marquee figure (256 B, fan-out 1, group-commit `fdatasync`, Pi 4): `>= 60,000 msg/s`, `p99 < 6 ms`
- The three durability-mode rows side by side, with the page-cache row labeled `not power-loss safe`

## Targets are STATED, not measured (honest)

IronBus has no ratified, measured baselines. Every number is a STATED TARGET / requirement from the README and #19 / #110, NOT a measured result, and the STATUS column says "target, not yet ratified against a measured baseline" on every row. Where the README states a number it is quoted (the marquee `>= 60,000 msg/s` / `p99 < 6 ms`, the `tiny` 64 MiB ceiling); where it does not, the metric is marked "target TBD pending ratification." The #19 illustrative per-device sketch is reproduced only as the starting proposal for ratification, never as a target and never as a measurement. No SLO is claimed to be "met."

The reason there is no committed baseline: the macro-bench harness (#111) is the measurement instrument, but its live coordinated-omission self-test (`an_injected_sigstop_shows_up_in_the_recorded_tail`) is `#[ignore]`d on shared CI because the SIGSTOP freeze does not reliably manifest in the tail on GitHub's shared runners; it is reliable only on a stable host. #284 tracks making it a reliable gate.

## How ratification works

A "Ratification process" section documents how a target becomes a ratified, versioned SLO: run the #111 harness on the reference edge device under #19's warm-up / steady-state discipline, archive the raw HdrHistogram plus the versioned provenance JSON, confirm the coordinated-omission self-test passes on a stable host, set the committed gate as the measured floor (on-device p99 minus a 20% margin, per the README and #110), then version the table with the git SHA and device. A "Not yet measured" disclaimer states plainly that no row has completed this process. Cross-references the bench harness and #284.

## Scope

Docs only: one new file plus a `docs/README.md` link and a CHANGELOG entry. No crate source, `Cargo.toml`, or CI touched. `cargo fmt --all --check` and `cargo test --workspace` pass. No em/en dashes.

`refs #110, #19` (the table + versioning intent is delivered as a doc; the remaining part is RATIFYING it against measured baselines, which is blocked on running #111 on the reference edge device and on #284 making the self-test a reliable gate).


ELares:
Independent review (not the author): SHIP. The doc is accurate and honest. Verified the README genuinely states the quoted numbers (the provisional marquee target "at least 60,000 messages per second with p99 latency under 6 ms on a Raspberry Pi 4", "tens of thousands of small messages per second per core", "64 MiB segments (8 MiB on the edge profile)", and the "measured floor = on-device p99 minus 20 percent margin" rule), each quoted verbatim. Every SLO metric is defined against the exact field the #111 bench harness emits (msgs_per_sec, mb_per_sec, p50/p99/p999_us from the HdrHistogram, steady_rss_bytes, write_amplification). Crucially honest: NO target is presented as a measured result, every row's status is "Target, not yet ratified against a measured baseline", and the ratification path (run #111 on the reference device, archive the provenance + raw HdrHistogram, the self-test passes on a stable host per #284, then version) is documented. README numbers not stated are marked TBD. Docs-only, no en/em dashes, CI green. `refs #110, #19` correct (ratification against measured baselines is the remaining part, blocked on the edge run + #284).

PR #303, opened 2026-06-07
Freeze the versioned loss-report schema and the I1 to I4 invariant gate (#120)

ELares:
## Summary

Delivers the genuinely-unmet parts of #120: names and freezes the recovery loss report as the external, versioned `ironbus.loss-report.v1` contract, and consolidates the four resilience checkers (I1 to I4) into one explicit per-PR + nightly invariant gate.

## What already existed (unchanged by this PR)

- The I1 to I4 checker functions in `crates/ironbus-storage/src/invariants.rs` (`check_no_acked_loss`, `check_longest_valid_prefix`, `check_bounded_loss`, `check_pure_recovery`), each with a negative fixture.
- `LossReport` / `LossEvent` / `ReasonCode` in `crates/ironbus-storage/src/loss.rs`, with a `schema_version` field, the serde derive, the bounded-loss caps, and the frozen numeric reason codes (`reason_codes_are_stable_and_distinct`). `serde_json` is dev-only.
- The checkers already used ad-hoc as oracles in the crash gates, the corruption corpus (#298), and the determinism meta-test; the nightly deep sweep already drives the storage proptests at `PROPTEST_CASES=50000`.

## What this PR adds

**1. The versioned, externally-frozen `ironbus.loss-report.v1` schema artifact**

- A golden serialization test (`golden_loss_report_v1_serialization_is_frozen`) that pins the EXACT JSON shape and field order of a representative `LossReport`/`LossEvent` for `schema_version = 1`, byte for byte. The prior `json_round_trips_and_is_stable` only asserted three substrings via `contains`, so it would NOT catch a field rename, removal, reorder, or a stray `#[serde(rename)]`. The golden makes any such change a CI failure, forcing a deliberate `schema_version` bump.
- A second golden (`golden_reason_code_vocabulary_is_frozen`) pinning every `ReasonCode`'s numeric `code()`, `metric_label()`, and serde JSON variant name as one frozen triple, so the externally-frozen reason vocabulary can never silently change (the numeric codes were already pinned; this adds the JSON-name leg).
- A normative schema doc `docs/schemas/loss-report.v1.md` enumerating every field, its type, the bounded-loss caps, and the stable `ReasonCode` integer values + JSON names + metric labels, cross-referenced from `docs/CONTRACTS.md` and linked from `docs/README.md`.

**2. The consolidated I1 to I4 invariant gate**

- A new `crates/ironbus-storage/tests/invariant_gate.rs` that drives the real recovery path (`Log::open`) over a generated sweep of segment shapes (record count x segment cap x durability boundary x clean/power-loss/torn-tail crash) and runs ALL FOUR checkers over each recovery in ONE pass (`invariant_gate_holds_over_the_shape_sweep`), so "recovered correctly" has one gate rather than scattered ad-hoc calls.
- A proptest twin (`invariant_gate_holds_under_a_proptest_sweep`) that the nightly deep sweep (`cargo test --workspace`, `PROPTEST_CASES=50000`) drives at depth, with shrinking. No workflow change is needed: the nightly already runs the workspace tests at that depth.
- A colocated `negative_fixtures` test feeding each checker a deliberately-broken input (a lost acked write for I1, an offset gap for I2, a per-event and a cascade over-cap for I3, diverging recoveries for I4) and asserting each FAILS, proving the gate is non-vacuous. (The torn-tail crash is clamped never to truncate below the durable head, so it tears only unsynced bytes and the gate tests recovery, not a synthetic I1 violation.)

## Verification

- `cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean (pedantic); `cargo test --workspace` green.
- The consolidated gate holds at `PROPTEST_CASES=50000` (the nightly depth); the existing crash_recovery, determinism, and corruption_corpus gates stay green.
- No library-path `unwrap`/`expect`/`panic` added (the new asserts are all in `#[cfg(test)]`); `ironbus-core` untouched and IO-free; no em/en dashes.

## Scope and what remains (so this uses `refs`, not `Closes`)

This PR scopes to the checker-gate + the versioned schema artifact, the two genuinely-unmet parts. The remaining #120 acceptance criteria are out of scope here and tracked elsewhere:

- The deterministic multi-actor SIMULATION core that would drive the checkers at scheduler depth is **#119** and is not built here.
- I3 cap ENFORCEMENT at the recovery binary level (freeze read-only + non-zero exit on exceed) and making the caps configurable per **#14** are separate from the checker that detects the violation, which exists.

refs #120, #21, #119


ELares:
## Review resolution (SHIP) — independent adversarial review, verified by mutation

Independent reviewer (not the author) confirmed every claim by mutate-and-rerun, then reverted:
- **Schema genuinely FROZEN:** `golden_loss_report_v1_serialization_is_frozen` pins the exact pretty JSON of a two-event `LossReport` byte-for-byte and round-trips it; renaming a field (`bytes_skipped` -> `bytes_dropped`) made it FAIL. `golden_reason_code_vocabulary_is_frozen` pins all five `(code, metric_label, json_name)` triples and `ReasonCode::ALL.len() == 5`; renumbering a code made BOTH it and the pre-existing `reason_codes_are_stable_and_distinct` FAIL.
- **I1 to I4 gate NON-VACUOUS:** the consolidated gate runs all four checkers over a 100+ shape sweep through the REAL `Log::open` -> `read_from` recovery path (proptest twin run at PROPTEST_CASES=50000, ~121s). Each checker has a negative fixture asserting its exact violation (`AckedRecordLost`, `NotAPrefix`, `CapViolation::PerEvent`, `CapViolation::Global`, `Nondeterministic`). Stubbing `check_no_acked_loss` to always-Ok made the negatives FAIL; and injecting a REAL acked-loss bug (`out.pop()`) into `read_from` made the POSITIVE sweep FAIL (`I1: acked offset 4 is missing`) - the gate catches a genuine recovery regression on the live path, not just synthetic fixtures.
- The golden serializes the real `LossReport`/`LossEvent`/`ReasonCode` types recovery emits (not a shadow struct); `serde_json` is dev-only (absent from the shipped `cargo tree`); ironbus-core untouched/IO-free; `docs/schemas/loss-report.v1.md` matches the source; clippy/fmt clean; no en/em dashes.

`refs #120, #21, #119` correct: the deterministic SIM core (#119) is honestly deferred, so #120 stays open for it. CI 15/15 green (rebased on current main).

PR #304, opened 2026-06-07
Freeze the resilience-counter taxonomy and document the observability contract (#96)

### LINKED ISSUE #96: [TASK] Resilience-observability contract: count every skip, freeze, drop, and overflow with its offset

ELares:
Delivers the resilience-observability **contract** (#96): every resilience event the broker sheds, drops, skips, dead-letters, truncates, force-reaps, or loses on recovery increments a **stable-named, documented counter**, so no resilience event is ever silent and the taxonomy can never silently drift.

## Audit: every resilience event -> its counter

| Resilience event | Counter | Status |
|---|---|---|
| Produce shed (AtCapacity / drop-new) | `ironbus_produce_rejected_total` | already counted |
| Disk-full drop-oldest force-reap | `ironbus_segments_force_reaped_total` | already counted |
| Retention reap (consumer-safe) | `ironbus_segments_reaped_total` | already counted |
| Dead-letter (past MaxDeliver) | `ironbus_dead_lettered_total` + `ironbus_dlq_records_total` | already counted |
| Redelivery (lease expired) | `ironbus_redelivered_total` | already counted |
| Recovery torn-tail / corruption loss | `ironbus_recovery_truncated_bytes`, `ironbus_recovery_loss_bytes{reason}`, `ironbus_recovery_loss_records{reason}` | already counted |
| **Below-earliest truncation served to a consumer (`Poll::Truncated`, #82/#84)** | **none** | **GAP, now fixed** |

The audit found every resilience event already counted **except** the runtime below-earliest truncation: when a slow consumer's records are force-reaped out from under it, `poll` surfaces a one-time `Poll::Truncated` resetting its cursor up to `earliest_retained`, so the consumer loses the span `[old_cursor, earliest_retained)`. That skip incremented **no counter**, the exact silent loss this contract forbids. (`segments_force_reaped` counts the reclamation; nothing counted the consumer-side skip.)

## Counters added

- `ironbus_truncations_total` (counter): below-earliest truncation **events** served to a consumer (one per surfaced `Poll::Truncated`; the same gap never re-counts).
- `ironbus_truncated_records_total` (counter): **records** skipped by those truncations (the span sum).

Both are incremented at the exact `Poll::Truncated` event site in `poll_in` and `poll_in_member`, rendered on `/metrics` with `# HELP`/`# TYPE` lines, and echoed in the `/admin` JSON snapshot.

## Frozen-taxonomy test

`the_resilience_counter_taxonomy_is_frozen` (in `crates/ironbus-server/src/health.rs`) renders `/metrics`, extracts the complete set of `ironbus_*_total` counter names it emits, and asserts it equals the pinned `FROZEN_RESILIENCE_COUNTERS` set **exactly** (plus a `# HELP`/`# TYPE` line per counter). Modeled on the frozen wire-tag tests, so adding, removing, or renaming a resilience counter is a deliberate, test-gated change and the taxonomy can never silently drift.

## Contract doc

New `docs/METRICS.md`: the normative per-counter catalog (name, the event that increments it, and the shed-vs-drop-vs-skip-vs-dead-letter-vs-truncation meaning), the recovery-loss series, the operational gauges, and the **deliberately-uncounted** dispositions (nack-requeue, term, fenced ack, idle-group eviction) with rationale. Cross-referenced from `docs/USAGE.md` (metrics section), the docs index, and the threat model.

## Tests
- Truncation counters increment at the event site, asserted both at the engine layer (extended `a_truncated_consumer_gets_exactly_one_truncation_then_delivers`) and via `/metrics` (new `metrics_renders_and_increments_truncations`); the same gap never re-counts.
- The frozen-taxonomy test above.
- The existing metrics harness still validates the `# TYPE` lines for the new counters.

## Scope
Counting + a stable, frozen taxonomy + the contract doc. No new metrics backend, histograms (#97), or OTLP export. `ironbus-core` stays IO-free (counters live in the server engine; rendering in `health.rs`).

`cargo fmt`/`clippy -D warnings`/`test --workspace` all green; no em/en dashes.

Closes #96


ELares:
## Review resolution — SHIP-WITH-NITS, both addressed (c40d1fa)

Independent adversarial reviewer (not the author) confirmed `Closes #96` is justified: every resilience drop/shed/skip/dead-letter/truncation/force-reap/recovery-loss event is counted, and the one genuine gap (the runtime below-earliest truncation served to a consumer, the silent skip the contract forbids) is exactly what this PR closes via `ironbus_truncations_total` + `ironbus_truncated_records_total`, incremented once at each of the two `Poll::Truncated` sites (`poll_in`, `poll_in_member`, mutually exclusive so no double-count). The deliberately-uncounted dispositions (fenced ack/nack/term, idle-group eviction which is provably loss-free, key_shared remap, TooManyGroups/InvalidGroupName request rejections, credit/byte-budget backpressure) are correctly outside the loss taxonomy and documented. The frozen-taxonomy test is a true exact-set assertion (12 names; add/remove/rename fails).

Both NITs addressed:
- The three em dashes in `docs/METRICS.md`'s See-also section are replaced with colons (the project forbids em/en dashes; they slipped the authoring check).
- `the_resilience_counter_taxonomy_is_frozen` now matches any `ironbus_*_total` sample with a numeric value (not only `_total 0`), so a counter that is non-zero on a fresh broker can no longer silently escape the exact-set check.

CI: 15/15 green. ironbus-core untouched; no serve flag added.

PR #305, opened 2026-06-07
Add the design risk register and the adversarial flaw-hunt record (#138)

### LINKED ISSUE #138: [META] Design risk register: adversarial flaw hunt before coding
IronBus's design record is unusually thorough, but an adversarial pass across the seams between issues exposes one architecturally fatal contradiction and a cluster of correctness, resilience, and edge-failure holes that the per-issue decisions hide because no two issues are read against each other. The single most dangerous finding: the frozen "log-is-WAL, no separate WAL file" model in #4 and the README is directly contradicted by #6, #10, #2, #13, and #20, all of which describe a *separate* WAL stream that is fdatasync'd, drained asynchronously into the segmented main log, truncated/checkpointed, recycled, and block-framed with FULL/FIRST/MIDDLE/LAST fragments. These are two different storage engines with two different crash-recovery proofs. Until that is resolved, every durability, recovery, and corruption-skip guarantee rests on an undefined substrate. Below is a prioritized risk register. Severity reflects blast radius (a silently-lost acked write or a fleet-wide boot hang is critical); likelihood reflects how routinely an edge fleet hits the trigger.

## Risk register

| # | Risk | Where it bites | Severity | Likelihood | Why it is dangerous |
|---|------|----------------|----------|------------|---------------------|
| R1 | log-is-WAL vs separate-WAL contradiction | #4 vs #6, #10, #2, #13, #20 | critical | high | Two incompatible engines. #4/README freeze "the active segment IS the WAL, no second on-disk log." #6 says "the segmented main log is fed from the WAL asynchronously," "WAL truncation/checkpoint happens only after the corresponding main-log bytes are fdatasync'd." #10 adds an "admission credit released on fsync completion" over a WAL fd. #20 wants "recycle WAL/active segments." #5 freezes record-aligned, no block layer; yet #1 and #2 specify "32 KiB WAL block framing, FULL/FIRST/MIDDLE/LAST fragments." You cannot write the recovery proof or the corruption-skip resync without picking one. |
| R2 | Single append actor + commit thread is the global throughput ceiling and a head-of-line stall | #4 (single logical writer), #6 (one commit thread), #19 | high | high | Every producer serializes through one actor that mints offsets and one OS thread that owns the fd. A single slow fdatasync on an SD card (50-500 ms GC pause is normal on cheap eMMC/SD) stalls ALL producers, not just one, because there is exactly one sync in flight and credits release only on its completion (#10). The 60k msg/s p99<6ms marquee (#19) assumes sync latency stays bounded; SD garbage collection routinely violates that. No experiment is specified to measure tail-sync latency on worn flash. |
| R3 | Numeric contradictions across frozen knobs | #1, #5, #6, #10, #14 | high | high | max_record_bytes is 1 MiB (#1), 16 MiB (#5/README), ceiling 64 MiB (#1) vs 1 GiB (#5). Segment is 8 MiB (#1), 64 MiB (#5/#13), 128 MiB (#4). Group-commit cap is 256 KiB (#1) vs 1 MiB (#6); window is 5 ms (#1) vs 0 ms (#6). max_frame_size 1 MiB but max_record 16 MiB (#5/#11): a legal 16 MiB record cannot traverse the default wire. These are config defaults that gate startup validation (#5 requires max_record < segment.bytes); shipping inconsistent ones is an immediate boot failure or a silently truncated product. |
| R4 | Derived-index rebuild cost on a huge log at boot is unbounded in practice | #4, #7 | high | medium | The offset/time index is derived and rebuilt by scanning validated records (#4, #7). On a throughput-profile 16 GiB retention (#14) the boot path re-validates and re-indexes potentially the whole tail. #7 caps tail *scan* at 512 MiB but defaults to "continue past the ceiling, scan to completion." A power cut on a full 16 GiB log can mean a multi-minute boot CRC walk on an armv7 core, during which the queue is down and producers are dropping. No measured boot-time SLO exists. |
| R5 | RTC-less lease expiry voids the hard lease cap and can lose or double-deliver | #9, #20 | high | medium | #9 stores lease deadlines as "monotonic-derived duration plus wall-clock enqueue time," treats all leases as expired on restart, and recomputes the hard cap "from persisted wall-clock first-receipt." Edge boards without an RTC boot at epoch 0 or a stale time until NTP lands. A backward wall-clock jump makes the 5-minute hard cap (#9) compute a negative or huge age, so max_deliver-then-DLQ either fires instantly or never. #20's "reuse last emitted timestamp on regression" protects ordering but not the lease-cap arithmetic. |
| R6 | Bounded-loss 1% global cap is gameable and conflates two failure axes | #8, #21, #7 | high | medium | The cap is "1% of total durable bytes per recovery" plus "one segment or 64 MiB per event" (#21). A long log makes 1% large: at 16 GiB that is 160 MiB of silently-skippable acked data per boot, which violates the prime invariant "no acked write is ever lost." Worse, a corrupt segment *header* (#8 SegmentHeaderBad) loses the WHOLE segment (up to 64 MiB) and the resync magic lives in records, not the header, so one header bit-flip is maximally expensive yet counts as one "event." An attacker or a flaky controller that flips one header byte per segment can walk the log to the 1% ceiling and freeze the instance (DoS) or, below it, silently drop megabytes. |
| R7 | Spill tier shares the same disk as the log, so disk-full deadlocks both | #10, #13, #20 | high | medium | Spill-to-disk is the durable overflow path (#10), but it lands on the same filesystem as the segments. When the disk fills, retention's disk_full=block (#13) waits for the reaper to free a segment, but the reaper cannot run if every sealed segment is pinned below a lagging cursor (retention.consumer_safe default false mitigates, but the quarantine store, the consumer-state logs, and the spill tier all also consume that same finite space). The 5% quarantine cap, 512 MiB spill cap, and segment store are not jointly budgeted against the device size, so they can collectively wedge a 4 GB eMMC. |
| R8 | Consumer-state log is an unbounded second log with its own compaction | #9, #13 | medium | high | The PEL/cursor state is "a separate append-only consumer-state log, periodically compacted/snapshotted" (#9). Under high redelivery (every nack, attempt-increment, lease-grant is a record) a single hot poison message at max_deliver=5 across many consumers writes many state records per data record. Its compaction is hand-waved ("periodically"), has no size cap, no crash-recovery spec, and is not in the RAM/flash budget (#20). It can out-grow the data log on a churny workload and is itself a corruption-skip surface that #8 never covers (it only covers data segments). |
| R9 | key_shared per-key head-of-line blocks the single shared log | #3, #9 | medium | medium | Sticky key routing requires "the broker must not hand key K to member B while member A has an un-acked in-flight record for K" (#3, #9). With one shared log and no per-key queue, a single stuck key holds a credit and forces the dispatcher to skip-scan or stall. On membership change, remap "waits for in-flight on moved keys to drain or expire" up to the 5-minute cap (#9), so a crashed consumer freezes that key range for 5 minutes. This interacts badly with R5: if lease expiry is broken by clock issues, the drain never completes. |
| R10 | "Treat failed fsync as fatal, freeze read-only" turns a transient EIO into fleet-wide downtime | #6, #16 | medium | medium | The PostgreSQL fsyncgate lesson is correct, but on edge flash a transient EIO (controller hiccup, brownout-induced write error) is common and often self-clears. Freezing the whole writer (#6) on the first such error, with no retry/backoff and only a manual unfreeze (#16 /admin/actions), means one glitch takes the node's ingest down until a human or service-manager restart. No bounded-retry-before-freeze policy is specified, and restart re-runs full recovery (R4). |
| R11 | dedup default-off plus session-scoped window makes reconnect retries duplicate acked writes | #3, #11 | medium | high | Dedup is off by default and, even opt-in, is session-scoped unless the optional stable producer_id+epoch is used (#3). #11 makes reconnect-driven PUB retries "safe" only via the msg-id dedup window owned by #3. With the default (no msg-id), an at-least-once producer that times out and retries across a reconnect appends the record twice with two offsets. The "at-least-once" contract technically permits this, but the README's edge-telemetry framing implies dedup matters; the default silently does not protect it. |
| R12 | TLS 1.3-only plus mandatory-on-non-loopback bricks RTC-less first boot | #18 | medium | medium | TLS 1.3 is mandatory on any non-loopback bind, no insecure opt-in (#18). Cert chain verification needs a sane clock; an RTC-less board at epoch 0 fails validity-period checks, so mTLS (the recommended LAN mechanism) cannot establish until NTP lands, yet NTP may itself ride over the network the device is trying to bring up. No "clock-not-yet-set" grace path is defined. |
| R13 | Zero-fill tail rule collides with preallocation and a genuine zeroed-payload record | #5, #4 | medium | low | #5 defines a 16-byte all-zero window as end-of-data (preallocated zeros). But #4 preallocates the active segment to full size, so a torn write that lands AFTER a zero gap (write reordering, #21 class 2) leaves valid records beyond a zero window that recovery will truncate, silently losing acked data past the gap. The segment generation tag (#5) guards stale-life replay but not intra-life reordered writes. |

## Open decisions

- [ ] Pick ONE storage model and delete the other from every issue: either log-is-WAL (active segment is the only durable file, no async drain, no WAL truncation, no fragments) or a real separate recycled WAL feeding sealed segments. Reconcile #4, #5, #6, #10, #2, #13, #20 to the winner. (R1)
- [ ] Freeze a single authoritative table of max_record_bytes, max_record ceiling, segment size per profile, group_commit cap, group_commit window, and max_frame_size, and assert max_record <= max_frame_size <= segment_bytes at config validation. (R3)
- [ ] Decide whether tail-sync latency on worn flash is acceptable for the single-commit-thread design, or whether a second concurrent sync / per-producer pipelining is needed. Specify the experiment. (R2)
- [ ] Define a measured boot-time recovery SLO as a function of log size and a hard policy for logs above it (refuse vs background lazy index). (R4)
- [ ] Specify RTC-less and clock-jump behavior for lease expiry, the hard lease cap, and TLS cert validity, including a "time-not-yet-trusted" boot state. (R5, R12)
- [ ] Reframe bounded loss as "0% of ACKED bytes, ever" and make the 1% cap apply only to never-acked tail bytes; treat a corrupt segment header as resync-from-records, not whole-segment drop. (R6)
- [ ] Jointly budget segment store + spill + quarantine + consumer-state against the device free space with a single high-water enforcer. (R7, R8)
- [ ] Add bounded-retry-then-freeze for fsync EIO, with automatic unfreeze on a successful probe write. (R10)

## Acceptance criteria

- [ ] Exactly one storage/durability substrate is described identically across #4, #5, #6, #10; a grep for "separate WAL", "fed from the WAL", "WAL truncation", and "FULL/FIRST/MIDDLE/LAST" finds zero contradictions with #4.
- [ ] A single frozen size/limit table exists and the startup validator rejects any inconsistent combination, with a unit test enumerated.
- [ ] A deterministic-sim scenario (#21) demonstrates: an SD-style multi-hundred-ms fsync stall does not lose an acked write and surfaces as backpressure, not data loss (R2).
- [ ] A boot-time recovery benchmark on the reference armv7/aarch64 device publishes recovery seconds vs log GiB, gated against a stated ceiling (R4).
- [ ] A crash test with wall-clock set to epoch 0 and to a backward jump shows correct lease cap, max_deliver/DLQ behavior, and no double-fire (R5).
- [ ] A corruption test flips one byte in N consecutive segment headers and shows resync recovers per-record without whole-segment loss and without hitting the global freeze cap (R6).
- [ ] A disk-full soak with an idle lagging consumer, an active spill tier, and quarantine pressure shows the node sheds with a reported counter and never deadlocks (R7).
- [ ] A redelivery-storm test bounds consumer-state-log growth and proves it recovers cleanly after kill -9 (R8).

## Related issues

- #1 vision, #2 prior-art: the separate-WAL and 32 KiB-fragment language that contradicts the frozen model (R1, R3).
- #3 semantics, #9 consumers: key_shared head-of-line, dedup defaults, lease/clock model (R5, R9, R11).
- #4 storage-log, #5 record-format, #6 durability-fsync: the core R1/R2/R3/R13 contradictions.
- #7 recovery, #8 corruption-skip, #21 verification: boot cost and bounded-loss gaming (R4, R6).
- #10 backpressure, #13 retention, #20 edge-constraints: spill/disk-full/budget interactions (R7, R8).
- #16 observability, #18 security: freeze-on-EIO operability and TLS-clock coupling (R10, R12).

ELares:
A new `docs/RISK_REGISTER.md`: the post-implementation design risk register, an adversarial flaw-hunt over the SHIPPED design, grounded in the code, the issues, the merged PRs, and the CHANGELOG. It is the complement to the design-time hunt that ran before coding (the META issue #138): this register records the failure modes the engine faces today, what is mitigated, the real defects review caught, and what honestly remains open. The code / README / issues stay canonical; nothing is invented.

## Categories covered
durability, concurrency, resource-exhaustion/DoS, recovery, observability, security, operational, performance. Each risk has an id, a short title, a category, the failure mode, a status (MITIGATED / FIXED / OPEN), and a residual-risk note.

## Section 1: mitigated risks (mechanism + test cited)
Per-consumer message credit (#65) and byte budget (#275); named-group cap (#240) plus idle eviction (#277); the disk-bounding stack (byte cap + drop-new shed #10, retention reaper #13, drop-oldest #82); MaxDeliver to a durable DLQ (#63); CRC32C + xxh3 (#146) with bounded-loss fail-closed recovery (#120) and the corruption corpus (#123); torn-tail truncation (#7); graceful-stop cursor flush (#195); connection cap / slowloris timeout / frame cap (#105); the frozen resilience-counter taxonomy (#96). Each cites the engine/storage mechanism and a named test verified against the source.

## Section 2: real fixed defects (grounded in real merged fixes)
The highest-signal section. Every entry is a real merged fix verified against the CHANGELOG, the closed issue, or the merged PR (none invented):
- **RR-10** lowering `--max-groups` silently dropped durable cursors on recovery; fixed: recovery loads every durable group unconditionally (#240, CHANGELOG + test `lowering_the_group_cap_still_recovers_every_durable_group`).
- **RR-11** key_shared empty-key OVER-serialization; fixed: empty keys bypass the per-key gate and drain in parallel (PR #283 review thread + test `empty_keys_drain_in_parallel_and_are_not_serialized`).
- **RR-12** macro-bench crashed on a dropped connection under a hard stall; fixed: the harness tolerates the drop and reports measured samples (PR #279 commit).
- **RR-13** serve-flag parse INFINITE LOOP on a value-less flag; fixed: a bare boolean flag advances the index by one (PR #289 body + test `serve_parses_the_allow_unlimited_deliver_flag`).
- **RR-14** below-earliest TRUNCATION served to a consumer was UNCOUNTED; fixed: added `ironbus_truncations_total` / `ironbus_truncated_records_total` (#96).
- **RR-15** Windows dead-code build break from a unix-only serve field; fixed: the non-unix stub consumes `enable_admin` (PR #301).

## Section 3: honest open / unmitigated risks (each with its tracking issue)
No authentication (#106), no TLS (#107), no at-rest encryption (#108); the single-`Mutex` append path head-of-line block (a slow fsync stalls every connection, WAL.md caveat); the macro-bench injected-stall self-test flaky on shared CI (#284); no wire-advertised credit or version/capability negotiation (#292/#132); resilience counters not yet durable (#98, verified OPEN); no loom coverage yet because there are no lock-free structures (#122).

## Scope and verification
Docs only: one new file plus a link from `docs/README.md` and a terse CHANGELOG entry. No crate source, Cargo.toml, or CI touched. `cargo fmt --all --check` and `cargo test --workspace` are green. No em or en dashes (verified with `git diff | grep -nP '[\x{2013}\x{2014}]'` and a rendered-file scan). Cross-references THREAT_MODEL.md, INVARIANTS.md, WAL.md, and CONTRACTS.md rather than duplicating them.

Closes #138


ELares:
Independent review (not the author): SHIP. The register is accurate and honest. Spot-verified the 6 fixed-defect entries against the real merged work: RR-10 (#240/#278 lowering --max-groups silently dropped durable cursors on recovery, fixed by loading all groups unconditionally), RR-11 (#283 key_shared empty-key over-serialization, fixed via the review commit), RR-12 (#279 macro-bench crash on a dropped connection under a hard stall, fixed by tolerating it), RR-13 (#289 value-less-flag parse infinite loop, fixed by advancing the index), RR-14 (#96 below-earliest truncation served to a consumer uncounted, fixed with ironbus_truncations_total), RR-15 (#301 Windows dead-code from a unix-only enable_admin field). The agent correctly noted #283/#279/#289 are PR numbers with feature titles and verified each underlying bug from the review thread/commits rather than the title. The mitigated risks cite real mechanisms (per-consumer credit+byte budget, group cap+eviction, disk caps+retention, MaxDeliver->DLQ, CRC+xxh3+bounded-loss recovery+corpus, graceful-shutdown flush, frozen counter taxonomy), and the open section is blunt about what is not there (no auth/TLS/at-rest #106/#107/#108, flaky bench self-test #284, single-Mutex head-of-line blocking, no loom #122, wire-credit #292, durable counters #98). No overstatement. Docs-only, no en/em dashes, CI green.

PR #306, opened 2026-06-07
Make the resilience counters durable across a restart (#98)

ELares:
## What

Makes the resilience `Counters` (produced, dead-lettered, reaped, truncated, etc.) durable across a restart (#98). Today the `Counters` struct lives only in memory and resets to zero on every `Engine::open`; a restart zeroes the operational history. This persists them with the same durability discipline as the cursor checkpoint.

## Snapshot format and durability discipline

- New `counters.ckpt`: a CRC32C-protected, two-slot, alternating-write file, the SAME crash-safe machinery as `cursor.ckpt` (`checkpoint.rs`). A torn slot fails its CRC and the prior slot survives; on recovery the higher-sequence valid slot wins. The checkpoint was made const-generic over the per-slot payload cap so the slightly larger counters slot reuses the dual-slot logic without duplicating it; two concrete aliases (`Checkpoint` for the cursor, `CountersCheckpoint` for the counters) keep every existing cursor call site unchanged.
- Payload: `Counters::encode_snapshot` writes a 1-byte version then the fixed set of `u64`s little-endian. `decode_snapshot` is tolerant: a short payload zero-fills missing trailing fields, trailing bytes are ignored, a wrong version never errors. So a field can be appended later without breaking either direction.

## Wiring: recover-at-open + cadence + shutdown flush

- `Engine::open` opens (creating if absent) `counters.ckpt`, recovers the last snapshot, and seeds `self.counters` (all-zeros if missing or torn).
- `maybe_checkpoint` piggybacks the counters snapshot on the existing cursor-checkpoint cadence (best-effort: a counters write failure never fails the cursor checkpoint).
- `checkpoint_all_groups` (the #195 graceful-shutdown flush, called by `cmd_serve` on SIGINT/SIGTERM/SIGHUP) writes the final counters snapshot LAST, after every cursor flush, so a restart after a clean stop shows the final counts.

## Monotonic lower-bound semantics

The snapshot is NOT fsynced on every increment (that would kill throughput); it is written on the cadence and on shutdown. So the resumed counters are a monotonic LOWER BOUND: a crash between an increment and the next snapshot loses at most the increments since the last snapshot, acceptable for observability. The counters only ever increase, so Prometheus reads them as a counter, never a gauge that rolls backward.

## Corrupt-snapshot-never-blocks-recovery safety

CRITICAL: the counters file is strictly an OBSERVABILITY aid, never correctness state. A torn or missing `counters.ckpt` recovers as all-zeros and NEVER blocks `Engine::open`, fails recovery, or affects the durable log, cursors, or DLQ. This is made explicit in `open_counters_checkpoint` (the only errors it can return are genuine IO failures from creating/opening the file, the same failures that would already have failed opening the log; a corrupt-but-CRC-valid payload decodes to zeros, a torn slot is dropped by the checkpoint CRC fallback).

## Tests

- `counters_persist_across_a_reopen`: produce/deliver/ack/dead-letter to bump several counters, flush, reopen on the shared in-memory fs, assert the counters resumed (not zeroed).
- `a_missing_counters_file_opens_with_zero_counters`: delete `counters.ckpt`, reopen succeeds with zero counters and the cursor is unaffected.
- `a_torn_counters_file_opens_with_zero_counters_and_the_log_is_unaffected`: corrupt both slots, open succeeds (never a hard error), counters are zero, and the durable log + cursor recover exactly.
- `the_graceful_shutdown_flush_persists_the_latest_counts`: under the default interval (no cadence checkpoint fires), only the shutdown flush makes the counters durable; a restart shows the final counts.
- `the_increment_path_does_not_fsync_per_increment`: a burst of increments does NOT change `counters.ckpt` on disk; an explicit checkpoint does (non-vacuous).
- `the_counters_snapshot_round_trips_every_field`: encode/decode round-trips every field; empty/short/padded payloads decode tolerantly to the right value.
- New `CountersCheckpoint` slot tests in `checkpoint.rs` (full-size round-trip, over-cap rejection, torn-slot fallback, fresh recovers nothing).

Determinism and golden-path stay green (two identical runs produce a byte-identical disk image including the deterministic `counters.ckpt`; the frozen counter taxonomy is unchanged). Docs updated: `docs/METRICS.md` (durability + lower-bound semantics) and `docs/WAL.md` (the new file class).

`cargo fmt`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` are all green. No new CLI/`serve` flag, so the non-unix `cmd_serve` stub is unchanged.

refs #98 (operational counters made durable as a snapshot lower-bound; the skip-counter replay-reconciliation + strict cross-restart monotonicity criteria remain, tracked in the follow-up)


ELares:
## Review resolution — SHIP-WITH-NITS, addressed (90e9735)

Independent adversarial reviewer (not the author) verified both load-bearing risks PASS:
- **Cursor checkpoint UNCHANGED:** the const-generic refactor is byte-for-byte behavior-preserving for the cursor case (slot layout 78 bytes, CRC range, alternating slots, torn-slot fallback all identical); `checkpoint_cursor`/`checkpoint_group`/`maybe_checkpoint_group`/`write_group_checkpoint`/open-time recovery are 100% unchanged; all existing cursor tests (torn-slot, both-slots-torn, single-byte proptest, the both-slots-torn engine gate) pass unchanged.
- **Corrupt/missing counters NEVER bricks recovery:** every adversarial input traced (missing, torn, both-slots-torn, CRC-valid garbage, wrong version, short, over-length) yields zero counters and a successful open; the only propagating errors are genuine IO failures that would already fail the log open; the durable log/cursors/DLQ are untouched.
- No per-increment fsync (all 13 increment sites are plain `+= 1`; snapshot only on the cadence + shutdown); round-trip covers all 11 fields; determinism image stays byte-identical.

Addressed:
- **SHOULD-FIX (traceability):** switched to `refs #98` (not Closes). #98's literal criteria want a checkpoint-PLUS-replay reconciliation (strict cross-restart monotonic non-decreasing, `last_skip_offset = max(checkpoint, replay)`, a repair counter) for the SKIP/LOSS counter family; this PR delivers the snapshot lower-bound for the OPERATIONAL counters (which cannot be replay-derived, so lower-bound is the only viable design for them). The skip-family reconciliation is tracked in the new follow-up #307; #98 stays open for it.
- **NIT (latent trap):** bumped `COUNTERS_PAYLOAD` 96 -> 128 so a future 12th counter cannot overflow the cap and fail the graceful-shutdown flush (which propagates `Truncated`, unlike the swallowed cadence write). counters.ckpt is unreleased, so no compat impact.
- **NIT:** fixed the `checkpoint_counters` rustdoc cross-reference (`checkpoint_cursor` does not flush the counters).

CI: 15/15 green (rebased on current main).

Issue #89, opened 2026-06-05
[TASK] Env-var and flag mapping, data_dir lifecycle, and secret redaction


ELares:
Parent: #14

Implement the env-var and command-line-flag layers that mirror every config key, plus the startup data_dir lifecycle and secret redaction. This serves the Cross-Platform and Edge-First promises: the same key surface works as a foreground binary, a systemd unit, or a container with no file at all, and a freshly flashed device provisions itself safely.

## Approach

Env vars flatten to `IRONBUS_<SECTION>_<KEY>` uppercased, nested tables join with another underscore (`IRONBUS_NETWORK_TLS_ENABLED`), arrays are comma-separated in one var, and values reuse the shared TOML scalar grammar; the bare profile key is `IRONBUS_PROFILE`. Ship the generated canonical SECTION_KEY map and unit-test it exhaustively. Flags mirror the same keys (`--durability-level`, `--segment-size`). At startup, resolve storage.data_dir, create it with `mkdir -p` semantics at 0700 if absent, and verify writability by creating and fsyncing a probe file; a read-only or unwritable data_dir is a fatal startup error naming the path. TLS private keys are referenced by path only, never inlined. Secret-bearing values (env or future inline) are redacted as `<redacted>` in `CONFIG LIST` and any log dump. A config file that is world-readable is warned about.

## Acceptance criteria

- [ ] Every key has a documented env-var and flag mapping; the SECTION_KEY map is generated and fully unit-tested.
- [ ] Arrays are comma-separated; durations/sizes use the shared grammar in env and flags.
- [ ] data_dir is created 0700 if absent and probe-write-and-fsync verified; unwritable is fatal naming the path.
- [ ] TLS keys are path-only; no secret is ever inlined in the config struct.
- [ ] Secrets are redacted in CONFIG LIST and log output.
- [ ] A world-readable config file produces a warning.

## Failure considerations

A single `IRONBUS_CONFIG` blob would defeat per-key precedence; mitigated by per-key flattening. A missing data_dir crashes first boot while a read-only mount silently loses durability; mitigated by create-if-absent plus fatal-on-unwritable probe. A CONFIG LIST or crash log leaking a secret; mitigated by redaction. Ties to #18.

ELares:
PR #308 delivers the buildable, decision-light parts of this issue:

[...]

Issue #85, opened 2026-06-05
[TASK] Precedence resolver and immutable effective-config with atomic load/reload


ELares:
Parent: #14

Implement the core that turns the five layers (built-in defaults < config file < environment variables < command-line flags < runtime admin API) into one immutable effective config, resolved once at startup with no per-message lookup on the hot path. This is the foundation every other config sub-issue builds on, and it owns the Resilient promise that the config system never wedges the broker.

## Approach

Resolve per key: for each key the winning value is the highest layer that sets it. Materialize the result into an immutable `Config` struct behind a single `Arc<Config>` published via one atomic pointer swap. Load is read-whole-file then parse then validate then swap; never read the file incrementally. `CONFIG RELOAD` re-runs the same pipeline and swaps only on full success, leaving the live `Arc` untouched on any error. A named-but-broken file (flag, env, or present at the default path) is a fatal startup error with file path and line/column; an absent default file is the zero-config path that starts on the `balanced` profile. At startup, log the fully materialized effective config: every resolved key, its value, and its source layer.

## Acceptance criteria

- [ ] Per-key precedence resolves with a worked example showing one key set at all five layers and the winner.
- [ ] Effective config is an immutable `Arc<Config>` swapped atomically; no per-message config lookup.
- [ ] `ironbus` with no file/env/flags starts on the documented `balanced` profile.
- [ ] A named-but-malformed/unreadable file is fatal with path + line/column; absent default file is not an error.
- [ ] `CONFIG RELOAD` swaps only on full success and provably leaves the live config unchanged on failure.
- [ ] Startup logs the materialized effective config with each key's source layer.

## Failure considerations

A partial or incremental read during reload could expose a half-applied config; mitigated by whole-read + validate + single atomic swap. Falling back to defaults on a broken file would silently revert tuned edge boxes; mitigated by fatal-on-named-broken-file. Ties to #15 for the reload trigger and to #16 for the materialized-config log.

ELares:
Closing for M0: the precedence RESOLVER shipped, flag > env > (profile) > default resolves into one effective config materialized at startup with no per-message hot-path lookup, and the fully-resolved effective config is logged with its source on boot (main.rs, docs/CLI.md); the five-layer model and the immutable-Arc / atomic-reload design are specified in docs/CONFIG.md. The config-FILE (TOML) layer + parser + CONFIG RELOAD are blocked on a toml-parser dependency decision and tracked in #382.

PR #308, opened 2026-06-07
Add env-var config mapping, the data_dir lifecycle, and a single-broker lock (#89)

ELares:
Delivers the buildable, decision-light parts of the config system from #89: an env-var + flag mapping with a clear precedence, the data_dir lifecycle, and the single-broker lock. The full TOML config file, profiles, the precedence-resolver-for-a-config-document, and hot reload are SEPARATE, decision-heavy follow-ups (#85/#87/#88) and out of scope here.

## Env-var mapping with precedence (flag > env > default)

Every `serve` setting can also be supplied via `IRONBUS_<FLAG>`: the flag name minus its leading `--`, uppercased, with `-` replaced by `_`.

- `--max-total-bytes` -> `IRONBUS_MAX_TOTAL_BYTES`
- `--data-dir` -> `IRONBUS_DATA_DIR`
- `--addr` -> `IRONBUS_ADDR`
- ...and one per flag (the full table is in `docs/CLI.md`).

Precedence: an explicit command-line flag overrides the env var, which overrides the compiled default. The env layer is read through an INJECTED `Fn(&str) -> Option<String>` seam (`EnvFn`), never `std::env` directly, so the env precedence is tested deterministically with a fixed map and not against the racy global process environment; production passes a closure over `std::env::var`. A bad env value (non-numeric where a number is expected, an unknown `IRONBUS_DISK_FULL_POLICY`, a non-boolean `IRONBUS_ENABLE_ADMIN`) is a usage error (exit 1) that NAMES THE ENV VAR, exactly as a bad flag value names the flag. The repeatable `--key-shared-group` stays command-line only (a per-group list does not flatten to one scalar).

## data_dir lifecycle

On `serve`, before the broker opens: the `--data-dir` is created (parents too, mode `0700`) if absent; a path that exists but is not a directory is a usage error (exit 1) naming the path; and the directory is probe-write-and-fsync verified writable, so a read-only or unwritable mount is a fatal error (exit 70) naming the path rather than a silent loss of durability.

## Single-broker lock (the corruption guard)

`serve` takes an exclusive advisory lock (`flock(LOCK_EX|LOCK_NB)`) on a `LOCK` file in the data dir. Two `serve` processes writing the SAME segmented log would interleave appends and corrupt it; the lock makes the second `serve` FAIL FAST (exit 70, "another ironbus broker is already running on `<dir>`") instead of double-opening. The lock is released by `close(2)` on the held fd drop and unconditionally by the kernel on process exit (clean shutdown OR crash), so it never goes stale the way a mere lock FILE's existence would.

### unsafe justification

The flock call is one small `unsafe fn flock_ex_nb` (the textbook safe use of `flock`: a kernel advisory-lock op that reads only a valid open fd and touches no user memory) with a `# Safety` doc and a `SAFETY:` comment at the call site, `#[allow(unsafe_code)]` opted-in per the workspace lint convention. `libc` is now a direct Unix-only `ironbus-cli` dependency; it was already in the tree (pulled by `ctrlc`) and on the `deny.toml` allow-list, so `cargo deny check` passes. Unix-focused like `serve`; the non-unix `cmd_serve` stub still compiles (the lock and lifecycle are `#[cfg(unix)]`).

## Out of scope (deferred follow-ups)

The full TOML config file, profiles, the precedence-resolver for a config DOCUMENT, secret redaction in CONFIG LIST, and hot reload are the decision-heavy parts of #89's parent surface, tracked separately as #85/#87/#88. This PR is `refs #89, #14`, not `Closes`, since those remain.

## Tests

New tests (all deterministic, golden path stays green):

- env-var seam: `the_env_var_name_mapping_is_ironbus_uppercase_underscored`, `env_var_sets_a_value_when_no_flag_is_given`, `an_explicit_flag_overrides_the_env_var`, `the_default_applies_when_neither_flag_nor_env_is_given`, `an_invalid_env_value_is_a_typed_error_naming_the_env_var`, `an_env_supplied_bool_and_list_resolve_through_the_seam`, `an_env_supplied_disk_full_policy_resolves_and_a_bad_one_names_the_env_var`, `usage_and_cli_docs_document_the_env_var_mapping`
- lock primitive (in `dirlock`): `lock_is_acquired_and_a_second_acquire_fails_fast`, `lock_releases_on_drop_so_a_fresh_acquire_succeeds`
- data_dir lifecycle (in `dirlock`): `prepare_creates_a_missing_data_dir`, `prepare_creates_parents_too`, `prepare_accepts_an_existing_writable_dir_idempotently`, `prepare_rejects_a_non_directory_path`
- serve-level integration: `serve_creates_a_missing_data_dir`, `serve_on_a_non_directory_data_dir_is_a_usage_error`, `serve_fails_fast_when_the_data_dir_is_already_locked`

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` all clean; `cargo deny check` ok.

refs #89, #14


ELares:

Independent reviewer (not the author) verified both load-bearing items under adversarial scrutiny:
- **`unsafe flock` SOUND, MINIMAL, justified:** the unsafe is exactly one `libc::flock(fd, LOCK_EX | LOCK_NB)` call; the owning `File` is held for the call and then moves into the `DirLock` so the fd stays open for the lock's lifetime (Drop -> close releases the flock); the return is checked with errno read immediately (EWOULDBLOCK -> typed AlreadyLocked, else LockIo); `#[allow(unsafe_code)]` localized; `# Safety` doc accurate. Judgment: a crate is NOT warranted (a non-blocking flock has none of the async-signal-safety / global-state hazards that justified the ctrlc dependency for #195; Miri cannot run any FFI syscall, so the written justification + the held-File soundness proof suffice for this minimal case; adding fs2/fd-lock for one syscall is net cost, not net safety).
- **The lock truly prevents two brokers:** acquired in `cmd_serve` BEFORE `open_disk_engine`/`Engine::open` (the only log writer); the only pre-lock write is a probe file (benign), so there is no double-open window / no TOCTOU; `LOCK_NB` makes a second `serve` fail fast (exit 70, "another ironbus broker is already running"), never a blocking wait; the lock is on `<data_dir>/LOCK` (per-dir), released by close on drop AND by the kernel on exit (never stale). Tests genuinely exercise a real second-acquire failure and the release.
- Env precedence flag>env>default for all 21 knobs via the injected `EnvFn` seam (deterministic, no global env in the resolution path); a bad env value is a typed usage error NAMING the env var (never a silent default); `--key-shared-group` correctly CLI-only. data_dir lifecycle: create-with-parents-0700, typed not-a-directory / not-writable errors, validated before the lock. Non-unix stub consumes every field; libc cfg(unix)-only and deny-clean; no library panic; clippy/fmt clean; no en/em dashes; core untouched; MSRV 1.78.

Four non-blocking cosmetic NITs noted (dangling-symlink message, an umask clarifying comment, an NFS doc note, a dedicated already-running exit code); none gating. `refs #89` correct (the config-file/profiles/reload subsystem deferred to the real open issues #85/#87/#88).

CI: 15/15 green.

Issue #100, opened 2026-06-05
[TASK] Static musl cross-build matrix with per-target static verification


ELares:
Parent: #17

Establish the cross-compilation pipeline that turns the IronBus source into one statically linked musl binary per Tier 1 architecture, with a verification step proving each artifact depends only on the kernel. This is the foundation every other build-dist task builds on, so it lands first.

## Approach

Tier 1 targets: `x86_64-unknown-linux-musl`, `aarch64-unknown-linux-musl`, `armv7-unknown-linux-musleabihf`. `riscv64gc-unknown-linux-musl` is build-only (compiles and unit-tests, ships no asset). Use `cargo-zigbuild` as the shipping builder (one pinned Zig sysroot across all targets) and run `cross` in CI as a non-shipping cross-check for x86_64 and aarch64 to catch zig-specific divergence. Pin the Rust toolchain in `rust-toolchain.toml` and the Zig version in the build image. After each build, run `file` and `ldd` (or `ldd`-equivalent on the target via qemu) and fail the job if any dynamic dependency is reported. Add a qemu-user smoke test that actually executes the aarch64 and armv7 binaries (run `ironbus --version`) so host-sysroot leakage is caught, not just inspected. Keep the build command documented and copy-pasteable per the acceptance criteria.

## Acceptance criteria

- [ ] A documented single command produces a static musl binary for each Tier 1 target.
- [ ] `file` reports statically linked and `ldd` reports no dynamic dependencies for every Tier 1 artifact.
- [ ] aarch64 and armv7 binaries execute under qemu-user and print a version.
- [ ] riscv64 compiles and passes unit tests in CI but produces no release asset.
- [ ] `cross` cross-check runs for x86_64/aarch64 and any divergence from `cargo-zigbuild` output fails CI.

## Failure considerations

The dominant risk is a C-FFI dependency silently linking the host sysroot, passing a build-host smoke test but failing on-device; the per-target `ldd`/`file` gate plus qemu execution catches both dynamic leakage and host-sysroot leakage. A zig miscompile is caught by the `cross` cross-check. Toolchain drift breaking determinism is mitigated by pinning Rust and Zig versions in-repo. Coordinate the C-FFI policy with #12 so the default tree stays pure-Rust.

PR #309, opened 2026-06-07
Add a size-optimized reproducible release profile for the static binary (#101)

ELares:
Completes the build-config + documentation half of #101: makes the shipped `ironbus` static binary bit-for-bit reproducible. The size-optimized `[profile.release]` already landed (`dcb625d`); this PR adds the reproducibility inputs that pair with it.

## The gap this closes

`[profile.release]` sets `strip = true`, which removes the symbol table but NOT the absolute build paths the compiler bakes into rodata (`panic!`/`assert!` location strings, `#[track_caller]` callsites, `file!()`). The native release binary embedded **11 absolute paths** (`/Users/.../crates/ironbus-server/src/health.rs`, ...), so two operators rebuilding the same commit from different directories or under a different `$HOME` got byte-different binaries, defeating #101's load-bearing claim that an operator can independently rebuild a tag and confirm exactly what is on a device.

## What changed

- **`.cargo/config.toml`**: pins `CARGO_INCREMENTAL = 0`, the one determinism input whose value is location-independent (incremental codegen reorders output and is not bit-reproducible; it pairs with `codegen-units = 1`).
- **`RELEASING.md`**: documents the full reproducible-build invocation, replacing the prior "tracked under #101" deferral. Enumerates every determinism input (`--remap-path-prefix`, `SOURCE_DATE_EPOCH`, `CARGO_INCREMENTAL=0`, `codegen-units=1`, `--locked`, SHA-pinned toolchain, embed-SBOM-before-strip) and gives the exact `cargo auditable build` invocation plus the before/after size delta.
- **`Cargo.toml`** comment and **CHANGELOG** updated to point the byte-identical CI gate at the release workflow (#103) and reference the documented inputs.

The workspace/cargo-home remaps live in the documented invocation, not `config.toml`, because their values (`$PWD`, `$CARGO_HOME`) are only known at build time and cargo does not expand env vars inside that file.

## Profile settings (unchanged, for reference)

`opt-level = "s"`, `lto = "fat"`, `codegen-units = 1`, `panic = "abort"`, `strip = true`.

## Panic strategy

`panic = "abort"` is kept (binding, from the original profile): handlers are panic-free by construction (#7), so there is no unwinding path to preserve. It is NOT in scope to re-decide here. It does not break the test suite because only the `cross build --release` CI jobs use `[profile.release]`; the test jobs use the dev profile and no CI job runs `cargo test --release`.

## Reproducibility, verified

The documented remap was run locally:

```sh
RUSTFLAGS="--remap-path-prefix=$PWD=/ironbus --remap-path-prefix=$CARGO_HOME=/cargo" \
  cargo build --release -p ironbus-cli --locked
```

Embedded absolute paths dropped from **11 to 0** (now stable `/ironbus/...` prefixes), with **no size change** (536832 bytes before and after), confirming the remap is behavior- and size-neutral.

## Binary size

Native release `ironbus` (host aarch64) is 410 KiB / 528 KiB on this host; the original profile measurement was 703 KiB -> 410 KiB (41% smaller). The determinism inputs are byte-rewrites only and do not change size.

## Verification

- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: clean (pedantic)
- `cargo test --workspace --all-features --locked`: green, 0 failures
- `ironbus --help` works; the binary still builds and runs
- No em/en dashes in the diff
- `.cargo/config.toml` only sets `[env]` (no `[build]` target overrides), so the `cross build --release` musl CI job and its static-link assertion are unaffected

## Scope / cross-references

Build-config + reproducibility documentation only. The byte-identical two-runner CI gate, the release workflow, and signing are **#103**. The static musl cross-build matrix with per-target static verification is **#100**. The `opt-level = "z"` throughput re-measure is gated on **#19**. Using `refs #101, #100` (not `Closes`) because the in-CI two-runner enforcement and the throughput gate remain in those follow-ups.


ELares:
Independent review (not the author): SHIP. Build-config + docs only (.cargo/config.toml, a Cargo.toml comment, RELEASING.md, CHANGELOG), no behavior change, CI 15/15 green including all three musl cross-builds + the static-link assertion (so the new .cargo/config.toml does not break "cross build --release"). The find is correct and valuable: strip=true does NOT remove the absolute build paths rustc bakes into rodata via panic / track_caller / file!() callsites (11 embedded paths), so a rebuild from a different directory was byte-different, defeating reproducibility. The fix is sound: CARGO_INCREMENTAL=0 is committed (a location-independent determinism input, robust against an env that flipped the default), while the path remaps are correctly documented as invocation-time in RELEASING.md (cargo does not expand env vars in config.toml, and their values are build-location-dependent so cannot be committed). Verified the remap drops embedded paths 11 to 0 with identical binary size. The full reproducible invocation (remap-path-prefix, SOURCE_DATE_EPOCH, locked, SHA-pinned toolchain, embed-SBOM-before-strip) is documented. No en/em dashes. refs #101, #100 correct: the in-CI two-runner byte-identical gate (#103) and the opt-level throughput re-measure (#19) remain.

(Re-posted: the prior comment was mangled by an unescaped-backtick shell-substitution on my side.)

Issue #104, opened 2026-06-05
[TASK] Distribution artifacts: .deb package, distroless image, and atomic upgrade/rollback


ELares:
Parent: #17

Ship the remaining distribution channels (a Debian package and a minimal container image) over the same verified static binary, and add safe in-place upgrade and rollback for a binary that is also a running broker with an open WAL. The draft treated distribution as install-once; this task covers the lifecycle on an unattended edge node.

## Approach

Generate a `.deb` (cargo-deb or nfpm) from the static binary, bundling a systemd unit and the default config from #14; `.rpm` is deferred behind the same nfpm spec until a dnf fleet exists. For the container, default to `distroless:static` (it provides CA certs, tzdata, and a nonroot user) rather than `FROM scratch`, document the required writable volume mount for the WAL/segment directory, and set a non-root `USER`; a `FROM scratch` minimal variant is reserved for the case where no TLS or tz dependency exists. For upgrade, perform an atomic swap: write the new binary to a temp path on the same filesystem, fsync, rename over the old path (atomic on POSIX), then restart the service; retain `ironbus.prev` for one-command rollback and fall back to it after N failed starts. `ironbus --version` emits the exact commit and build provenance. On-disk WAL/segment formats stay forward/backward compatible within a major version (#4, #5); a format bump is gated behind an explicit `migrate` subcommand, never silent.

## Acceptance criteria

- [ ] A `.deb` installs the binary, a systemd unit, and a default config from #14.
- [ ] A `distroless:static` image runs the binary as a non-root user with a documented writable WAL volume.
- [ ] Upgrade uses atomic write-temp + fsync + rename and never overwrites the live binary in place.
- [ ] `ironbus.prev` is retained and a one-command rollback works; the service falls back after N failed starts.
- [ ] `ironbus --version` emits commit and build provenance; format compat within a major is documented with explicit `migrate` gating.
- [ ] All four channels are documented: installer, .deb, container image, checksummed GitHub releases.

## Failure considerations

Power loss mid-upgrade leaving a truncated binary is the primary hazard; atomic rename keeps the old binary intact until the new one is fully on disk, and `ironbus.prev` plus fail-N-then-fallback recovers an unreachable node. A `FROM scratch` image missing CA certs or a passwd entry silently breaking TLS or non-root execution is avoided by defaulting to distroless. A silent on-disk format bump stranding WAL data is prevented by explicit `migrate` gating per #4.

ELares:
[...]

PR #311, opened 2026-06-07
Smoke-test each static musl binary on its target architecture (#100)

### LINKED ISSUE #100: [TASK] Static musl cross-build matrix with per-target static verification

ELares:
## What

The `cross-build` matrix in `.github/workflows/ci.yml` built each static musl binary and asserted it links static (no `PT_INTERP`, no `NEEDED`) but never EXECUTED it. A binary that links clean yet crashes on startup under musl/qemu (a relocation the loader cannot apply, a bad TLS or stack-guard setup, host-sysroot leakage) would pass the link check and ship broken. This adds a per-target SMOKE step that runs the freshly-built binary right after the static-link assertion, in the SAME `cross-build` matrix job (the job name is unchanged, so the required-check match holds).

## Smoke approach per target

| Target | How it runs | Why |
|--------|-------------|-----|
| `x86_64-unknown-linux-musl` | natively on the ubuntu x86_64 runner | the host IS the target arch |
| `aarch64-unknown-linux-musl` | `qemu-aarch64-static <bin>` | qemu-user from apt `qemu-user-static` |
| `armv7-unknown-linux-musleabihf` | `qemu-arm-static <bin>` | qemu-user from apt `qemu-user-static` |

The emulators come from apt `qemu-user-static` and are invoked explicitly (no Docker, no binfmt registration), so the job adds **no new pinned `uses:` action**. `qemu-user-static` is installed only on the two cross legs (`if: matrix.target != 'x86_64-...'`).

## What the smoke asserts (deterministic, offline, no socket/thread)

- `ironbus --version` exits 0 and prints a line starting with `ironbus ` (the new version verb).
- `ironbus --help` exits 0 and the output contains the usage banner (`USAGE:` ... `ironbus serve`).
- `ironbus dump --data-dir <empty dir>` exits 0: exercises the storage-open + offline-reader scan path with no record to decode, catching a startup crash that only the IO/storage init path triggers (which neither `--version` nor `--help` reach).

Each is pure exit-0 + stdout-substring, so it is fast and non-flaky under qemu-user.

## Deliberately limited

A networked produce/consume smoke is **not** run under qemu: sockets and threads under qemu-user are flaky, and the issue explicitly warns against it. The native `test` job already covers the full wire path on x86_64, so no coverage is lost. This is the only deliberate limitation; all three targets are smoked (version + help + offline dump), so the cross-arch execution gate the issue asks for is met on every Tier 1 target.

## Supporting CLI change

`ironbus --version` / `-V` / `version` did not exist (it exited 1). Added a single broker-free `ironbus <crate-version>` line via `CARGO_PKG_VERSION`, so the smoke has a deterministic version print and issue #100's "binaries execute under qemu-user and print a version" criterion is literally satisfiable. Covered by a unit test (`run_dispatches_version_without_a_server`) and documented in `docs/CLI.md` and the usage banner.

## Why this matters

This proves each static binary actually RUNS on its target architecture, not merely that it links statically: a musl/qemu startup fault that the readelf link assertion alone misses now fails the job.

## Scope

Does not touch the release workflow (#103, just merged) or add packaging (#104), both cross-referenced.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, and `cargo test --workspace --all-features --locked` all green locally.
- `actionlint` (with shellcheck) clean on `ci.yml`.
- The smoke commands confirmed exit 0 against the host binary, including the native empty-`$qemu` path under `set -euo pipefail`.
- The PR's own `cross-build` matrix runs the real per-target smoke under qemu.

Closes #100


ELares:
Independent review (not the author): SHIP. CI 15/15 green with the smoke step running on ALL THREE musl legs (the real verification, on GitHub runners): native on x86_64, qemu-aarch64-static / qemu-arm-static for the cross targets. The smoke is deterministic and offline (no sockets/threads under qemu-user): it asserts ironbus --version prints a single 'ironbus <semver>' line and exits 0, --help exits 0 with the expected USAGE substring, and dump on an EMPTY data dir exits 0 (exercising storage-open + the offline reader, catching a startup crash that --version/--help would miss). The networked produce/consume path is deliberately left to the native x86_64 test job (qemu-user sockets are flaky), honestly documented. The new ironbus --version/-V/version handler (env!(CARGO_PKG_VERSION), broker-free, exit 0) has a unit test. actionlint + shellcheck clean per the agent; the job name is unchanged so branch protection still matches; no en/em dashes. This proves each static musl binary RUNS, not just links statically (Done-v1 'static musl binary for 3 triples'). Closes #100 justified (all three targets smoked).

Issue #88, opened 2026-06-05
[TASK] Hot/cold/coupled reload engine and runtime admin API


ELares:
Parent: #14

Implement the reload engine that honors the hot/cold/coupled tags and the runtime admin API surface that reads and mutates the effective config. This realizes the Resilient guarantee that a bad live change can never tear down the running queue or corrupt the log, and it is the operator-facing companion to #15.

## Approach

Admin API: `CONFIG GET <key>`, `CONFIG SET <key> <value>`, `CONFIG LIST [section]`, `CONFIG DIFF` (effective vs file), `CONFIG RELOAD` (re-read file), `CONFIG SAVE` (explicit, atomic, comment-preserving). `SET` writes the in-memory highest-precedence layer only, is volatile, and is lost on restart; it never rewrites the file. `SAVE` writes a temp file in the same directory, fsyncs it, renames over the target, then fsyncs the directory. Reloads and SETs validate the proposed whole effective config (including coupled sets) before a single atomic `Arc` swap. A reload or SET touching a cold key (storage.segment_size, data_dir, network.*, record-format keys from #5) is rejected as a unit with a clear message and leaves the running config unchanged. Hot keys (backpressure.policy, retention limits, compression.level, codec as hot-forward to new segments only) apply live. `CONFIG DIFF` makes any divergence between live and on-disk config explicit so a restart never surprises the operator.

## Acceptance criteria

- [ ] Admin verbs GET/SET/LIST/DIFF/RELOAD/SAVE implemented; SET is volatile and does not touch the file.
- [ ] SAVE is atomic (temp + fsync + rename + dir-fsync) and preserves comments.
- [ ] A reload/SET touching a cold key is rejected atomically; live config provably unchanged.
- [ ] Hot keys apply live; compression.codec applies hot-forward to new segments only with old segments still readable.
- [ ] `CONFIG DIFF` reports every key where effective differs from file.
- [ ] Rejected reloads return an error naming the offending key.

## Failure considerations

A fat-fingered live change reverting on restart surprises the operator; mitigated by volatile SET + DIFF + explicit SAVE. A power loss during SAVE truncating the file; mitigated by atomic write. A partially applied coupled set; mitigated by whole-config validation before one atomic swap.

ELares:
[...]

Issue #87, opened 2026-06-05
[TASK] Compiled-in versioned profiles and materialized-config logging


ELares:
Parent: #14

Define the three named profiles (`edge-tiny`, `balanced`, `throughput`) as compiled-in, versioned presets that set coherent groups of knobs, applied first and then overridden by any explicit key. This delivers the Edge-First convenience of stamping one key on a fleet while defusing silent behavior drift across upgrades.

## Approach

Profiles are compiled into the static binary (offline edge devices need them with no external fetch) and carry a `profile_schema_version`. Each profile sets exactly the durability triple, storage.segment_size, retention triple, compression codec+level, and backpressure triple; it never sets storage.data_dir or any network/TLS key. Values: edge-tiny = fsync-batch/100ms/1024, 8MiB segments, size+time 256MiB/24h, lz4, drop-oldest, ring 1024, spill 64MiB. balanced (default) = fsync-batch/50ms/4096, 64MiB, size+time 2GiB/72h, zstd/3, spill-then-shed, ring 8192, spill 512MiB. throughput = fsync-batch/200ms/16384, 256MiB, size+time 16GiB/168h, zstd/1, spill-then-shed, ring 65536, spill 4GiB. Apply profile, then layer explicit keys on top (profile-then-override). Any change to a profile's contents is a documented breaking change in #22 and the CHANGELOG.

## Acceptance criteria

- [ ] Three profiles defined with exact knob values and a one-line rationale each.
- [ ] Profile-then-override semantics: explicit keys always win over the profile.
- [ ] Profiles never set data_dir or TLS keys.
- [ ] `profile_schema_version` is recorded and surfaced in the materialized-config log.
- [ ] All profiles default durability to batched fsync, never OS-flush-only.
- [ ] A profile content change is captured as a breaking-change entry.

## Failure considerations

Unversioned profiles silently shift a whole fleet on upgrade; mitigated by versioning + materialized-config logging + changelog discipline. A profile presuming data_dir or TLS would mis-provision environment-specific paths; mitigated by excluding those keys. Coupled values within a profile must satisfy the validators from #6, #13, #10, #12.

PR #312, opened 2026-06-07
Document the edge hardware-constraint to knob mapping (#117)

### LINKED ISSUE #117: [TASK] Hardware-constraint-to-knob mapping table and RTC-less clock model
Parent: #20

Produce the documented constraint table that maps each edge hardware limit to the specific IronBus knob that honors it, and specify the RTC-less clock model so ordering and retention survive a backward clock jump. The draft describes both informally but never delivers the table as an acceptance artifact, and it leaves the clock model as an open decision. This is primarily a design-document deliverable plus the clock-handling spec that #13 and #15 depend on.

## Approach

Build a table with one row per hardware limit (flash endurance, RAM ceiling, CPU, power loss, clock/RTC, thermal) and columns for the limit, the governing IronBus knob, the `tiny` default, and the owning issue (#6, #13, #10, #4). For the clock model, make a monotonic-plus-logical counter the sole ordering authority: record IDs never consult wall-clock, and on regression the last emitted timestamp is reused (Redis-style) so IDs never invert. Separately persist a periodically-checkpointed wall-clock anchor (last CLOCK_REALTIME plus the monotonic reading at that instant) used only for human-readable display and best-effort time retention. On an NTP step, re-anchor and emit a `clock_step` event. Document that time-based retention is best-effort and tolerant of a jumping clock.

## Acceptance criteria

- [ ] A constraint table maps every hardware limit to its governing knob, `tiny` default, and owning issue.
- [ ] Record IDs are monotonic-plus-logical and never invert under clock regression.
- [ ] A wall-clock anchor is persisted and used only for display and best-effort retention, never ordering.
- [ ] NTP steps emit a `clock_step` event and re-anchor; time retention is documented as best-effort.
- [ ] The table names the owning issue for each row so responsibility is unambiguous.

## Failure considerations

The main failure is a clock regression (NTP syncs and steps the wall-clock) inverting IDs or mis-expiring time-retained data; using the monotonic-plus-logical counter as sole ordering authority and reusing the last timestamp on regression prevents inversion, while the separate anchor keeps human-readable times usable. A secondary risk is the table drifting from implementation as knobs change; tie each row to its owning issue so reviews catch divergence.

ELares:
Adds `docs/EDGE_TUNING.md`: the operator-facing table that maps each edge hardware constraint to the specific shipped `serve` knob(s) that honor it, with a recommended edge value and the reasoning. Linked from `docs/README.md`. Docs only, no source / `Cargo.toml` / CI changed.

## Constraints covered
- **Limited RAM** (the 64 MiB ceiling, #115): `--max-segment-bytes` (8 MiB edge), `--consumer-credit`, `--consumer-credit-bytes`, `--max-in-flight`, `--max-groups`, `--max-total-bytes`, with a section on how they compose to keep RSS under the ceiling.
- **Flash wear / slow storage** (#19): group-commit `fdatasync`, `--checkpoint-interval`, retention (`--max-retained-bytes` / `--max-age-ms` / `--max-messages`) and `--max-total-bytes`, write amplification, never-recycle.
- **Limited CPU**: single-writer (no thread knob), lz4_flex (cheap, default) vs zstd (opt-in, off the default path), `--max-connections`.
- **Intermittent power**: always-on `fdatasync`-before-ack, `--disk-full-policy`, graceful shutdown (#195), and the bounded/reported recovery.

## Values cited to real flags
Every recommended value is a real, currently-accepted `serve` flag value, cited to its default and range in `docs/CLI.md`, which in turn cites the `main.rs` / `engine.rs` constants. No flag is invented or duplicated.

## The manual-not-auto finding (honesty)
The edge profile is NOT auto-selected. There is no `--profile` flag and no auto-selected "edge"/"tiny" profile in the shipped binary. `EDGE_SEGMENT_BYTES` (8 MiB) exists in `crates/ironbus-core/src/format.rs` but is **unwired** (confirmed in `docs/WAL.md`); you get the 8 MiB segment size only via `--max-segment-bytes 8388608`. The doc states these are RECOMMENDED MANUAL settings and cross-references the unshipped follow-ups: the auto RAM budget (#115) and named-profile selection (#87, with #85 / #88 for the TOML config and hot reload).

## Scope note
This delivers the issue's hardware-constraint-to-knob mapping table. The issue's separate RTC-less clock-model spec (monotonic-plus-logical ordering authority, persisted wall-clock anchor, NTP `clock_step` re-anchor) is intentionally out of scope here and left for a follow-up, so this is `refs #117`, not `Closes #117`.

Cross-references CLI.md, WAL.md, SLO.md, and THREAT_MODEL.md.

refs #117, #19


ELares:
Independent review (not the author): SHIP. Docs-only (EDGE_TUNING.md + README link + CHANGELOG), CI green, no en/em dashes. Every recommended edge value is a real serve flag cited to the main.rs/engine.rs consts: --max-segment-bytes 8388608 (8 MiB edge), --consumer-credit / --consumer-credit-bytes, --max-in-flight, --max-groups, --max-total-bytes for RAM; --checkpoint-interval + the retention trio for flash wear; --disk-full-policy + graceful shutdown for power loss. Honest manual-not-auto finding confirmed: there is NO --profile flag and no auto edge/tiny profile; EDGE_SEGMENT_BYTES (8 MiB) exists in format.rs but is unwired, so you get 8 MiB only via --max-segment-bytes 8388608 (the auto-profile is the #87/#115 follow-up). Also honest that on-disk compression has not landed (codec=none) and the interval/none fsync modes are not CLI-exposed. refs #117, #19 correct (the RTC-less clock-model spec is out of scope). Note for the record: the agent observed installer_verify::a_matching_binary_is_accepted (#103) can flake under PARALLEL cargo test (it shells out to install.sh + system sha256); passes 6/6 in isolation; worth a test-isolation fix in a follow-up but unrelated to this docs PR.

PR #314, opened 2026-06-07
Quarantine corrupt segments on recovery, copy-not-move and capped (#134)

### LINKED ISSUE #134: [TASK] Build quarantine store: copy-not-move, capped, accounted, with metadata-only fallback
Parent: #8

Build the quarantine subsystem that captures corrupt bytes for forensics without losing evidence, mutating good state, or filling a small edge disk. The draft proposes a quarantine directory but leaves move-versus-copy, the cap, the drop policy, and the interaction with disk accounting open.

## Approach

COPY offending bytes to a quarantine directory (never move); the source segment stays read-only and untouched until #14 reclaims it. Cap quarantine at min(256 MiB, 5% of the data-directory filesystem capacity). On reaching the cap, switch to metadata-only capture (offset range, byte count, stored CRC, segment id, reason code) and drop the captured bytes, evicting oldest forensic blobs first (FIFO). Account quarantine bytes as first-class storage against the data-directory budget and expose ironbus_quarantine_bytes to #11 and #14, so a bit-rotting node's forensic copies show up as real disk pressure. Let #14 reclaim a source segment once its quarantine copy or metadata record exists. Forensic blobs are named by segment id plus offset range for off-device pull via #16.

## Acceptance criteria

- [x] Copy-not-move capture; source segment never mutated by quarantine.
- [ ] Cap = min(256 MiB, 5% of FS); enforced before write.
- [ ] Metadata-only fallback on cap, FIFO eviction of oldest blobs.
- [ ] Quarantine bytes accounted against the data-directory budget.
- [x] ironbus_quarantine_bytes exposed to #11 and #14.
- [ ] #14 may reclaim a source segment once a copy or metadata record exists.
- [x] Forensic blob naming supports off-device pull via #16.

## Failure considerations

Moving instead of copying risks losing the only evidence if a crash interrupts the move, and mutates possibly-good source state; copy avoids both. An uncapped quarantine could fill a tiny eMMC and starve the live queue; the percentage-aware cap with metadata-only fallback bounds it. Double-counting quarantine plus the retained source as separate pressure could invisibly trip backpressure; first-class accounting plus letting retention reclaim the source keeps the disk math honest. Coordinate the cap default with #15.

ELares:
## What

Builds the forensic QUARANTINE store (#134): when recovery skips a CORRUPT region, COPY (not move) the corrupt bytes into a capped `quarantine/` subdirectory for offline forensics, without ever blocking recovery or being re-read as live data.

## Trigger and naming

Fires on the active-segment recovery corruption skip: a `ReasonCode` event OTHER than a clean `TornTail` (`quarantine::is_corruption_skip`). A clean torn tail is the expected power-loss case and carries no forensic value, so it is never quarantined. Each blob is named `q-<segment-id:016x>-<reason>-<start>-<end>.bin` (e.g. `q-0000000000000003-corrupt_record_body-4096-8192.bin`), so an operator pulling `quarantine/` off-device knows which segment and which bytes each blob holds without opening it.

## Copy-not-move

The corrupt span is read READ-ONLY from the live segment BEFORE the truncation that drops those bytes (the capture is wired into `Log::recover` ahead of `file.set_len`). The source is never mutated by quarantine; recovery's own truncation is what trims the live segment, exactly as before. A crash mid-copy can at worst leave a partial or absent blob, never lose the evidence.

## Cap policy

New `LogConfig::max_quarantine_bytes` (default `DEFAULT_MAX_QUARANTINE_BYTES` = 256 MiB; `0` = unlimited, matching the repo's `0`-as-off convention for the other byte caps). On reaching the cap the store evicts OLDEST blobs first (FIFO by sorted segment-then-span order). A single corrupt span larger than the whole cap is SKIPPED rather than partially written (the metadata-only fallback for that case is deferred). The issue's `min(256 MiB, 5% of FS)` ceiling is the server config layer's job, since the IO-free storage core cannot stat the host filesystem; deferred.

## Never blocks recovery

Opening the store, reading the source, and writing each blob are all best-effort. A quarantine IO failure is swallowed and can NEVER fail `Log::open` or touch the live log, exactly like the resilience-counters snapshot (a forensic aid, not correctness state). The store lives in a subdirectory recovery never enumerates as live data, and the `q-`/`.bin` name shape can never be parsed as a `seg-...log`.

## Counter

Exposed on `/metrics` as the `ironbus_quarantine_bytes` GAUGE (not a `_total`), so the frozen resilience-counter taxonomy is untouched and `the_resilience_counter_taxonomy_is_frozen` stays green. Documented in `docs/METRICS.md`. Plumbed `Log::quarantined_bytes()` -> `Engine::quarantined_bytes()` -> the `/metrics` snapshot.

## Tests

- `crates/ironbus-storage/tests/quarantine_recovery.rs`: a corrupt segment is copied to quarantine while recovery recovers the valid prefix and the live log is unaffected; the cap skips an over-cap span and (in unit tests) evicts oldest-first; an injected failing quarantine fs (`FaultFs`) does not fail recovery; a clean torn tail is never quarantined; the byte total survives a reopen.
- Quarantine module unit tests: verbatim copy, empty-span no-op, reopen accounting, FIFO eviction, over-cap skip, the recovery-loss helpers.
- `metrics_exposes_the_quarantine_gauge` pins the gauge and asserts it is not a `_total`.
- crash_recovery, determinism, and the corruption corpus (21/21) stay green; full `cargo test --workspace` green; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt` clean; no em/en dashes; ironbus-core stays IO-free.

## Deferral

The scrubber and source-segment reclaim (#14), and the percentage-aware `5% of FS` cap ceiling, are a linked follow-up.

Closes #134


ELares:
## Independent adversarial review: SHIP (with nits)

An independent reviewer (not the author) adversarially traced the two load-bearing safety properties and the full bar. Verdict: **SHIP-WITH-NITS**, no blockers.

**Never blocks recovery (highest risk) -- PASS, all paths.** `quarantine_corrupt_span` returns `u64`, never `Result`, and the call site in `Log::recover` assigns it with no `?`. Every internal IO op (subdir/open, list, `read_exact_at`, `create_new`, `write_all_at`, `sync_all`, `sync_dir`, evict `remove`) is reduced to `0`/`false`/`None` and swallowed. After any quarantine failure the code falls through to the unchanged `file.set_len(valid_end)?; file.sync_all()?`, so the live log still recovers the same valid prefix. Structurally there is no path by which a quarantine failure can fail `Log::open`.

**Copy-not-move, live recovery unchanged -- PASS.** Diffed against base `44f5353`: the `loss_report.push -> set_len -> sync_all` sequence is byte-for-byte intact. The capture block is inserted between the push and the truncation and reads `[valid_end, len)` (exactly the loss event span) from `&file` read-only before `set_len` trims it. `read_exact_at` is positional, so the read cannot disturb the handle later used for `set_len`.

**Cap + FIFO, quarantine-never-live, gauge-not-counter, blob crash-safety, bar -- all PASS.** Saturating arithmetic, over-cap single span skipped (not partial/panic), cap reconstructed on reopen via `scan_bytes`; `StdFs::list` + `parse_segment_file_name` doubly exclude the `quarantine/` subdir and `q-*.bin` blobs from live enumeration; the `ironbus_quarantine_bytes` gauge has no `_total` suffix so it is excluded from the frozen resilience-counter taxonomy (that test stays green); blob write is `write_all_at + sync_all + sync_dir` with rollback-and-return-0 on any failure; ironbus-core untouched, no new deps, MSRV 1.78, clippy `-D warnings` clean, no em/en dashes, corpus + crash_recovery + determinism + invariant gates green.

**Deferred to follow-up (not blockers):**
- SHOULD-FIX: the gauge reflects bytes captured at *this* recovery, not the persisted on-disk footprint that survives a clean restart (after a clean reopen it reads 0 while blobs remain on disk). The HELP text honestly documents the narrow "at the last recovery" scope, so this is a semantic gap, not a correctness bug; recovery, the live log, and the cap are unaffected. The `scan_bytes` infra to surface the persisted total already exists. Tracked as a follow-up under #134.
- NIT: FIFO eviction sorts by blob file name; within one segment the unpadded decimal start/end sort lexically, not by true offset. Best-effort forensic ordering only; the cap is strictly enforced regardless, so disk can never be exhausted. Folded into the same follow-up.

Merging on a fully green gate (15/15, fail=0, pending=0) plus this independent review. `refs #134` (the FS-percent cap, metadata-only fallback, budget accounting, and reclaim command remain deferred per the issue checklist).

Issue #109, opened 2026-06-05
[TASK] Specify secret handling, redaction, security audit events, and crypto SBOM posture


ELares:
Parent: #18

Specify how secrets are sourced, protected, and never leaked, the structured security audit events, and the crypto dependency posture that keeps the single binary auditable. This closes the operator-footgun and supply-chain surface.

## Approach
Secrets (token/password hashes, TLS private key, at-rest key) load from a file or environment variable referenced by #14, never inline in config. At startup, stat every secret-bearing file and refuse to start (fail closed) if it is group/world-readable (mode & 0o077 != 0) or wrong-owner, matching OpenSSH StrictModes; on non-POSIX platforms, skip with a logged notice. Wrap all secret types in a redacting newtype whose Debug/Display emit a fixed placeholder, zero key material after use where possible, and route every security event through one emitter that takes the identity name (safe), never the credential. Define structured, secret-free audit events exported via #16: authn outcome (identity, mechanism), authz denial (scope, verb), rate-limit and lockout trips, secrets-permission refusal at boot, and identity config-reload; each carries a monotonic sequence number plus wall-clock so ordering survives clock jumps. For supply chain, enumerate the crypto dependency surface (rustls, the AEAD and Argon2 crates), show it fits the cargo-auditable SBOM embedded in the #17 binary, and gate advisories with cargo-deny. The #15 admin data-introspection commands must redact stored credentials.

## Acceptance criteria
- [x] Secret sourcing from file/env via #14; no plaintext secrets inline; permission check refuses to start.
- [x] Redacting newtype for all secret types; no secret appears in any log, error, or #15 dump (redaction unit test).
- [x] Security audit event schema defined and exported via #16, with monotonic sequence plus wall-clock.
- [x] Crypto dependency surface enumerated and shown to fit the cargo-auditable SBOM and cargo-deny gating in #17.
- [x] Secret material zeroed after use where the language allows.

## Failure considerations
The primary risk is a secret leaking into logs or diagnostics; redaction-by-construction (newtype) plus a unit test asserting known secrets never appear confines this to a review-catchable boundary. A wrong-owner-but-tight-permissions file is caught by the owner check. On an unattended box, attacks are invisible without audit events, so the audit-event requirement is part of the security boundary, not an add-on.

ELares:
[...]

Issue #108, opened 2026-06-05
[TASK] Specify optional at-rest AEAD encryption and its interaction with checksums and recovery


ELares:
Parent: #18

Specify optional encryption at rest over segment payloads, including the AEAD choice, nonce construction, key sourcing, and the precise interaction with #5 checksums and #7/#8. This is the device-theft mitigation, with its honest limits documented.

## Approach
AEAD is AES-256-GCM where the runtime reports hardware AES (aarch64 crypto extensions, x86 AES-NI), ChaCha20-Poly1305 as the portable constant-time fallback, selected at startup by CPU feature detection and recorded in the segment header so reads are unambiguous. The 96-bit GCM nonce is segment-id (64-bit) concatenated with a per-segment monotonic record counter (32-bit), making nonce reuse structurally impossible without trusting an RNG that may be weak at early boot. Encrypt in place within existing record framing so the #6 and segment-rolling contracts are unchanged. Checksums (#5) are computed over ciphertext so corruption detection works without the key. Store a key-id (never the key) in the segment header; on read, a key-id matching no loaded key, or an AEAD tag mismatch, is a distinct REPORTED error class routed through #8 (bounded, reported loss), separate from plain checksum corruption. Key source priority: raw key file via #14, Argon2id-derived from passphrase, or TEE-sealed where present. Rotation is new-segments-only; history is never re-encrypted. Document the honest limit: a key co-located on a stolen device defeats at-rest encryption.

## Acceptance criteria
- [x] AEAD runtime selection (AES-256-GCM vs ChaCha20-Poly1305) and segment-header recording specified.
- [x] Deterministic nonce = segment-id || record-counter specified, with the no-reuse argument.
- [x] Key-id in segment header; rotation is new-segments-only; key-source priority defined in #14.
- [x] Tag mismatch / unknown key-id is a distinct REPORTED class via #8, not silent.
- [x] Device-theft limitation documented; #6 contract shown unchanged.

## Failure considerations
The critical failure is a wrong/rotated key yielding undecryptable data with a passing checksum; the key-id header plus tag-mismatch-as-reported-error makes this visible rather than silent. GCM nonce reuse is catastrophic and is prevented structurally by the deterministic nonce. Coordinate the header field with #4 and #5 to avoid framing drift.

ELares:
[...]

Issue #107, opened 2026-06-05
[TASK] Specify TLS 1.3 transport, the localhost-default bind invariant, and pre-auth DoS defenses


ELares:
Parent: #18

Specify the transport security layer, the bind-default invariant that keeps an unconfigured node off the network, and the pre-authentication denial-of-service defenses the first draft omitted. These are the network-boundary protections for an exposed edge node.

## Approach
TLS 1.3 only, no 1.2 fallback, cipher suites limited to the TLS 1.3 AEAD set; the binary bundles rustls so TLS capability is a property of IronBus, not the host OS. Server certificate verification is mandatory on clients; mTLS is the optional strongest mode. The bind invariant: default bind is 127.0.0.1; any non-loopback bind requires both TLS material and at least one configured auth identity, validated at startup before the listener opens, rejected with an actionable error otherwise, with no override flag into network-plaintext or network-anonymous. Plaintext exists only on loopback (silent). Pre-auth DoS defenses, all on by safe defaults for non-loopback binds and exempt on loopback: a cap on concurrent half-open (pre-auth) connections (default 128) past which accepts are dropped, a per-source-IP connection rate limit (default 10/sec, token-bucket), and a failed-auth lockout (close after N=5 failures with an escalating accept delay from that source, implemented as a fixed delay not a verify-time difference so the auth compare stays constant-time). All limits are #14 keys.

## Acceptance criteria
- [x] TLS 1.3-only with the restricted AEAD cipher set documented and reflected in #14.
- [x] Default bind localhost; non-loopback bind without TLS+auth rejected at startup with a clear error and no override.
- [x] Half-open connection cap, per-source connection rate limit, and failed-auth lockout specified with defaults, loopback exempt.
- [x] Each limit exposed as a #14 key with a safe default.
- [x] Defenses traced to the pre-auth-DoS threat row in the threat model.

## Failure considerations
The accidental-public-broker mistake (network bind, no auth) is prevented by fail-closed startup validation. Over-aggressive rate limits could lock out legitimate long-lived clients; mitigate by setting conservative defaults (connections are long-lived and reconnect rarely) and exempting loopback so local #15 use is never throttled.

ELares:
Specified and delivered in PR #317 (merged, squash `13e80808`) as `docs/TRANSPORT.md`, on a fully green gate plus an independent adversarial review that confirmed all five acceptance criteria and the exact defaults.

[...]

PR #316, opened 2026-06-07
Specify connection-scoped authentication and the three-scope authorization model (#106)

ELares:
## What this is

The design spec for connection-scoped authentication and the three-scope authorization model: a new normative document `docs/AUTHENTICATION.md`. It is the core auth contract every verb is gated by, the child of the security epic (#18) and sibling of #107 (TLS), #108 (at-rest encryption), and #109 (secret handling).

This is a `design`-label spec for features NOT YET IMPLEMENTED. The doc states that clearly at the top ("Status: specified, not yet implemented; tracked by #106") and never claims code exists that does not. It cites where it constrains the existing #11 CONNECT handshake (the frames it extends), the #14 config system (the keys it adds), and the #15 diagnostics surface (the commands it gates).

- [x] Token, username/password, and mTLS specified at the handshake level, consistent with #11 (auth established once in `Connect`, bound to the connection for its lifetime).
- [x] mTLS identity = SAN with the URI-then-DNS rule; CN explicitly excluded (never a fallback); a verifying cert that matches no configured identity is rejected, not given a default scope.
- [x] `publish` / `subscribe` / `admin` specified, granted explicitly per identity, `admin` does NOT imply the others; with the verb-to-scope mapping and the admin-gated verb and #15 diagnostic list (the `/admin` snapshot, online stored-data/DLQ readers, mutating admin #299, identity management).
- [x] Constant-time bearer-token compare over SHA-256 digests, and Argon2id password parameters (m = 19 MiB, t = 2, p = 1, per-credential salt in PHC encoding).
- [x] Additive credential-set rotation semantics (credential validity is a set, rotation is two deploys, no server-side expiry timer, no clock dependence).

Also specifies the single uniform "Authorization Violation" `Err` with no bad-credential-vs-insufficient-scope oracle (and no username-enumeration or timing oracle), and addresses the two sharp risks #106 names (the token timing oracle and the verifying-cert default-scope authz bypass).

## Also in this PR

- `docs/README.md`: a one-paragraph entry for the new doc in the "Reference docs" list, in the house style, ending with `(#106)`.
- `docs/THREAT_MODEL.md`: a one-line cross-reference from the specified-controls table's #106 row to the new doc. Minimal; the threat model is otherwise unchanged.
- `CHANGELOG.md`: a single terse bullet under `### Added` (refs #106).

Docs only. No Rust touched (`git status` is the four files above). No em/en dashes (`grep -nP "[\x{2013}\x{2014}]"` over the changed files returns nothing). Markdown files in this repo carry no SPDX header (only the SPDX CI check, which gates Rust sources only), so the new doc follows the majority of the reference docs and carries none.

## Deferred (not blocking this spec)

- The exact populated `Connect`/`Info` byte layout is owned by #11 and frozen there; this spec gives the semantics the bytes must encode, not the offsets.
- The online stored-data/DLQ reader verbs and the mutating-admin surface (#299) are named in the admin-gated list but their wire shapes are their own follow-ups.
- nkey/JWT account auth is explicitly out of v1 scope (deferred, not flagged).

refs #106


ELares:
## Independent adversarial review: SHIP (nits addressed)

An independent reviewer (not the author) verified the spec against issue #106. Verdict: **SHIP-WITH-NITS, recommend merge**.

- All five acceptance criteria satisfied with the doc text quoted: the three mechanisms (bearer token SHA-256 + constant-time compare, username/password Argon2id m=19 MiB/t=2/p=1, mTLS with first-URI-then-DNS SAN identity, CN excluded, verifying-but-unmatched cert rejected) specified at the #11 handshake level; the three explicit scopes with admin NOT implying the others, the admin-gated verb and #15 diagnostic list, and the offline peek/dump readers correctly carved out to filesystem permissions; additive credential-set rotation with no expiry timer.
- Both sharp risks correctly mitigated: the bearer-token timing oracle (constant-time compare plus a single uniform Authorization Violation, tag 12, with no bad-credential-vs-insufficient-scope or username-enumeration oracle) and the verifying-cert authz bypass (explicit SAN-to-identity match required, no-match rejected, chain verification grants no scope).
- Cross-doc consistency verified: the frame tags match CONTRACTS.md, the verb-to-scope table covers the real client-request set, and the THREAT_MODEL.md table retains BOTH the #106 (AUTHENTICATION.md) and #107 (TRANSPORT.md) rows after the #317 merge with no duplicate or dropped row. Honesty bar met (clearly specified-not-implemented, no shipped-control claims); no em/en dashes.

The two NITs the reviewer raised are fixed in `52383ca`: I6 is now cited by its real scope ("ordering never consults the wall clock," in both citations), and `Pong` is added to the enumerated server-to-client response frames. Merging on the green gate (15/15, fail=0, pending=0). `refs #106`.

PR #317, opened 2026-06-07
docs: spec TLS 1.3 transport, the bind invariant, and pre-auth DoS defenses (#107)

ELares:
## What

Adds `docs/TRANSPORT.md`, the normative design spec for IronBus' network-boundary security controls, plus the supporting reference-list, threat-model cross-link, and CHANGELOG entries. This is a DESIGN spec (label `design`, milestone M1), not code: the doc is marked **Status: specified, not yet implemented; tracked by #107** near the top and never claims a control exists in the binary. Docs only, no Rust touched.

Cites #11 (handshake / wire protocol) where it constrains the handshake, #14 (config system) where it adds keys, and #105 (threat model) where it closes threat rows.

## Acceptance criteria covered (from #107)

- [x] **TLS 1.3-only with the restricted AEAD cipher set documented and reflected in #14.** Section 1: floor and ceiling both pinned to TLS 1.3, no 1.2 fallback, the full and only TLS 1.3 AEAD suite set tabulated, rustls bundled in the binary, ChaCha20-Poly1305 preferred where no AES hardware. TLS material supplied via #14 keys.
- [x] **Default bind localhost; non-loopback bind without TLS+auth rejected at startup with a clear error and no override.** Section 2: default `127.0.0.1:7777`; loopback classification (wildcard counts as non-loopback); a non-loopback bind requires BOTH TLS material AND at least one auth identity, validated during startup config validation BEFORE the listener opens, fail-closed with an actionable, representative error; no `--insecure`/`--allow-plaintext`/`--allow-anonymous` override; plaintext loopback-only and silent.
- [x] **Half-open connection cap, per-source connection rate limit, and failed-auth lockout specified with defaults, loopback exempt.** Section 3: half-open pre-auth cap (128, distinct from and below `--max-connections`), per-source-IP token-bucket rate limit (10/sec), failed-auth lockout (N=5) with a FIXED escalating accept delay (not a verify-time difference, so the auth compare stays constant-time). All on by default for non-loopback, exempt on loopback; per-source tracker state itself bounded/idle-evicting.
- [x] **Each limit exposed as a #14 key with a safe default.** Section 3.4: a flag + `IRONBUS_<FLAG>` env var + default table (`--max-preauth-connections` 128, `--preauth-rate-per-ip` 10, `--auth-failure-lockout` 5), matching CLI.md's `serve` convention.
- [x] **Defenses traced to the pre-auth-DoS threat row in the threat model.** Section 4: a traceability table mapping each control to THREAT_MODEL.md's T7 (pre-auth flood), plus T2/T3 (TLS) and T6 (lockout) and the bind note. THREAT_MODEL.md's T7 row and specified-controls row are cross-linked back to TRANSPORT.md.

## Also

- New one-paragraph entry in the `docs/README.md` "Reference docs" list, ending `(#107)`.
- Minimal THREAT_MODEL.md edits: forward-reference the new spec from the T7 residual-risk cell and the #107 specified-controls row.
- Terse single-bullet CHANGELOG entry under `### Security`, ending `(refs #107)`.
- No em/en dashes anywhere (the doc is pure ASCII).

## Deferrals (candidate follow-ups, not gaps in this spec)

Items the spec intentionally defers to their owning issues rather than re-specifying: the exact #14 key spellings, TLS file formats, and the private-key file-permission rule (fixed with #14 and #109); the credential shapes and the three authorization scopes and the constant-time compare itself (#106); at-rest encryption (#108). The THREAT_MODEL.md row-flips and secure-defaults-catalog additions happen when #107 is implemented, not in this spec.

refs #107


ELares:
## Independent adversarial review: SHIP

An independent reviewer (not the author) verified the spec against issue #107. Verdict: **SHIP**, no blockers.

- All five acceptance criteria satisfied with the doc text quoted: TLS 1.3-only with the three AEAD suites as #14 keys; default loopback bind with a fail-closed non-loopback rejection and no override flag; the half-open cap (128), per-source token-bucket rate limit (10/sec), and failed-auth lockout (N=5) each with a default and loopback-exempt; each limit a #14 key; all traced to real THREAT_MODEL.md rows (T2/T3/T6/T7).
- Defaults match the issue exactly, and the lockout is correctly specified as a FIXED escalating accept-time delay (not a verify-time difference), so the auth compare stays constant-time.
- Cross-doc consistency clean: the THREAT_MODEL.md back-links point at genuine rows, the global `--max-connections`=256 and `DEFAULT_ADDR` match CLI.md, and the present-tense README "secure by default" claims are reconciled as specified-not-implemented.
- Honesty bar met: clearly marked "specified, not yet implemented; none of the controls exist in the binary today," no sentence implies a shipped control.
- Style bar clean: no em/en dashes in any changed file; markdown follows the majority no-SPDX-header convention; all 15 CI checks pass (0 pending, 0 failing).

One NIT (the CHANGELOG bullet is dense, but consistent with its neighbors and in the correct `### Security` category); not blocking. Merging on the green gate plus this review. `refs #107`.

Issue #102, opened 2026-06-05
[TASK] Supply-chain gate: cargo-deny policy, auditable SBOM round-trip, C-FFI ban


ELares:
Parent: #17

Make IronBus binaries auditable and the dependency tree policy-enforced, so an operator can recover a trustworthy SBOM from a shipped binary long after `Cargo.lock` is gone, and so no surprise license or C component slips into the single-binary distribution. This turns the draft's supply-chain aspirations into hard CI gates.

## Approach

Run `cargo-deny` as a required CI gate with: advisories = deny on any RUSTSEC vulnerability (and unmaintained, warn-then-deny after a documented grace window); licenses = an explicit allowlist (MIT, Apache-2.0, BSD-2/3, ISC, Unicode-DFS, MPL-2.0) with GPL/AGPL denied (allowlist fails safe when a new license appears); bans = deny multiple versions for build-critical crates and deny any cc-built or build-script-network-fetching C-FFI crate not on an explicit allowlist; sources = crates.io plus a pinned git allowlist only. Embed `cargo-auditable` in every shipped binary and additionally attach a syft-generated CycloneDX SBOM to the release. Close `cargo-auditable`'s blind spot (vendored C, build-script blobs, `dlopen`) with the C-FFI ban above. Verify the SBOM by recovering it from the built binary with `rust-audit-info` and diffing the crate set against `Cargo.lock`; a mismatch fails CI. Coordinate the C-FFI allowlist with #12 (lz4_flex pure-Rust default; zstd FFI opt-in only).

## Acceptance criteria

- [ ] `cargo-deny` passes as a required CI gate with the advisories, license allowlist, bans, and sources policy above.
- [ ] `cargo-auditable` embeds an SBOM recoverable by `syft`/`grype`/`rust-audit-info` from the shipped binary.
- [ ] CI recovers the embedded SBOM and diffs it against `Cargo.lock`; drift fails the build.
- [ ] Any non-allowlisted C-FFI or build-script-fetching crate fails CI.
- [ ] A CycloneDX SBOM is attached to each release.

## Failure considerations

[...]

PR #318, opened 2026-06-07
docs: specify secret handling, redaction, security audit events, and crypto SBOM posture (#109)

ELares:
## What

Adds `docs/SECRETS.md`, the normative (specified, not yet implemented) spec for how IronBus sources its secrets, protects them in memory and on disk, keeps them out of every log and diagnostic, the structured security audit events that make an attack on an unattended edge box visible, and the crypto dependency posture that keeps the single binary auditable. It is the child of the security epic #18 and the sibling of the auth (#106), TLS (#107), and at-rest encryption (#108) specs. The doc is honest that there is no auth and no TLS in the binary today, so the three load-bearing mechanisms (the redacting newtype, the fail-closed StrictModes boot check, and the audit-event emitter) are specified, not yet code.

Doc-only change: `docs/SECRETS.md` (new), plus the cross-link from the existing `docs/THREAT_MODEL.md` #109 row, a `docs/README.md` Reference-docs entry, and one `CHANGELOG.md` Security bullet. No source changed; `git status` is doc + CHANGELOG only.

- [x] **Secret sourcing from file/env via #14, never inline; permission check refuses to start.** Secrets are referenced by a file path or env-var name owned by #14, never carrying the value. At startup the broker stats every secret-bearing file and fails closed on `mode & 0o077 != 0` (group/world-readable) OR wrong-owner, matching OpenSSH StrictModes; on non-POSIX it skips with a logged notice. The wrong-owner-but-tight-permissions case is called out explicitly.
- [x] **Redacting newtype for all secret types; no secret in any log/error/#15 dump (redaction unit test).** A newtype wraps token/password hashes, the TLS private key, and the at-rest key; `Debug`/`Display` emit a fixed placeholder, raw bytes are reachable only through one explicit accessor, and it does not `Serialize` to the raw value. One security-event emitter takes the identity name and never the credential. A mandatory redaction unit test asserts a known sentinel never appears in any log, error, or #15 dump (and that the placeholder IS present where a secret field rendered).
- [x] **Security audit event schema defined and exported via #16, with monotonic sequence plus wall-clock.** A structured, secret-free event set (authn outcome, authz denial, rate-limit and lockout trips, secret-permission boot refusal, identity config-reload), each carrying a per-process monotonic `u64` sequence (the authoritative ordering, survives a clock jump) plus a wall-clock timestamp (for correlation, not ordering). Aligned with the METRICS.md "no silent event" philosophy and the frozen-taxonomy/`reason`-label discipline; the event-name set is frozen by a test.
- [x] **Crypto dependency surface enumerated and shown to fit the cargo-auditable SBOM and cargo-deny gating in #17.** rustls (+ provider), the RustCrypto AEAD crates (`aes-gcm`, `chacha20poly1305` over `aead`), `argon2`, `sha2`, `subtle`, and `zeroize`, each shown to be in the embedded `cargo-auditable` SBOM, advisory-gated by `deny.toml` (`yanked = deny`, `unmaintained = all`), and on the permissive license allow-list. The one knowing relaxation of the #139 pure-Rust rule (the TLS provider's vendored asm/C, off the record path) is recorded against the `deny.toml` #102 C-FFI note.
- [x] **Secret material zeroed after use where the language allows.** The newtype zeroes on `Drop` via a non-elidable write (`zeroize`/`Zeroizing`), with the honest Rust limits recorded (no defense against a core dump, swap, or a compiler-made copy). The #15 admin data-introspection commands (`dump`/`peek`/`/admin`) must redact stored credentials; the offline readers already print sizes only, and any config echo renders credential fields as the placeholder.

## Notes

- Nothing is deferred from the issue's acceptance criteria. The hot-reload audit event (identity config-reload) is documented as exercised only once config hot reload lands (#14); today config is read once at startup, which the doc states plainly. That is a downstream-feature note, not a deferred sub-part of this spec.

refs #109


ELares:
## Independent adversarial review: SHIP (SHOULD-FIX addressed)

An independent reviewer (not the author) verified the spec against issue #109. Verdict: **SHIP-WITH-NITS**, no blockers.

- All five acceptance criteria satisfied with the doc text quoted: file/env secret sourcing via #14 (no inline plaintext) and the fail-closed StrictModes boot check (`mode & 0o077`, wrong-owner, non-POSIX skip-with-notice); the redacting newtype (Debug AND Display redacted, single `expose_secret` accessor, no raw Serialize) with the mandatory redaction unit test; the secret-free audit-event schema exported via #16 with a monotonic per-process sequence (authoritative ordering, clock for correlation only) over the six events; the crypto surface enumerated and shown to fit the #17 cargo-auditable SBOM under cargo-deny (verified against the real deny.toml and RELEASING.md); zeroize-after-use and #15 redaction.
- Redaction boundary sound: newtype-by-construction forecloses Debug/Display, the single name-not-credential emitter cannot receive a credential, and ordering keyed on the sequence survives a clock jump.
- Cross-doc consistency holds: the THREAT_MODEL.md #109-row edit cites the real row; the #106/#107/#108/#109 rows are intact and unduplicated; METRICS.md naming and the "no silent event" discipline align. Honesty bar met (specified-not-implemented, no shipped-control claims); no em/en dashes.

The one SHOULD-FIX is fixed in `bba9873`: the `/admin` snapshot's output shape is documented in AUTHENTICATION.md (not CLI.md), and the verbatim "no secret material" phrase appeared in neither doc, so the reference is re-pointed to AUTHENTICATION.md and the snapshot described accurately (stored offsets and operational state, the config echo, never key or payload bytes); CLI.md is kept only for the accurate dump/peek sizes-only citation. Merging on the green gate (15/15). `refs #109`.

PR #319, opened 2026-06-07
docs: specify optional at-rest AEAD encryption (#108)

ELares:
## What

Adds `docs/AT_REST_ENCRYPTION.md`, the normative spec (specified, not yet implemented) for IronBus's optional, off-by-default authenticated encryption of segment record bodies. This is the device-theft mitigation (THREAT_MODEL.md threat row T1). Docs only, no source changed.

The new header fields reuse bytes the v1 segment header already reserves (the `flags` u16 and the `[44, 60)` reserved range), so the #4 record framing, the #5 checksum scope, the #6 publish path, and the segment-rolling contract are all unchanged. The doc is explicit that nothing here exists in the binary today.

- [x] **AEAD runtime selection and segment-header recording.** AES-256-GCM where the runtime reports hardware AES (aarch64 crypto extensions, x86 AES-NI), ChaCha20-Poly1305 as the portable constant-time fallback, chosen once at startup by CPU feature detection and recorded in the segment header (`aead_suite` in the reserved bytes) so a reader honors the recorded suite, not its own CPU.
- [x] **Deterministic nonce + no-reuse argument.** 96-bit nonce = `segment_id` (64-bit, never recycled per ADR 0002) `||` per-segment monotonic record counter (32-bit, derived from record position, not stored). The no-reuse argument is structural (unique within a segment, unique across segments, collision-free across a key) and needs no RNG, so a weak early-boot `/dev/urandom` cannot defeat it.
- [x] **Key-id in header; new-segments-only rotation; #14 key-source priority.** `key_id` (never the key) in the header; rotation rolls a new key onto new segments only and never re-encrypts history; key sourcing is TEE-sealed > raw key file (fail-closed perms via #109) > Argon2id-from-passphrase (the m=19 MiB / t=2 / p=1 edge parameters from AUTHENTICATION.md), the strongest available custody wins, no silent downgrade.
- [x] **Tag mismatch / unknown key-id is a distinct REPORTED class via #8.** Two new appended `ReasonCode`s (`UnknownKeyId`, `AeadTagMismatch`) routed through the `LossReport` bounded-loss path under the I3 caps, each distinct from the existing #5 `CorruptRecordBody`. The verify order is CRC-over-ciphertext first (I7), then AEAD decrypt, so a wrong/rotated key is never silent, never a passing checksum, and never looks like corruption.
- [x] **Device-theft limitation documented; #6 contract shown unchanged.** A co-located plaintext key defeats at-rest encryption (recorded plainly, with TEE-sealed custody as the one local mitigation that resists theft). The #6 eight-step publish path and the fdatasync-before-ack durability boundary are unchanged: encryption is an in-place body transform, the tag is a fixed 16-byte body delta, lengths describe the plaintext.

Also: #5 checksums computed over ciphertext (works without the key); cross-linked from the existing THREAT_MODEL.md #108 spec-table row (no new row invented); listed in the docs/README.md "Reference docs" list; one terse CHANGELOG Security bullet.

## Constraints

- No em or en dashes anywhere in the changed files (verified with `grep -rnP "[\x{2013}\x{2014}]"`, no matches).
- No SPDX header: the new doc starts with `# Title`, matching the THREAT_MODEL / CONTRACTS / INVARIANTS / WAL majority.
- `git status` shows only the new doc plus three edited docs/CHANGELOG files.

## Candidate follow-ups (not part of this doc)

- The implementation itself (AEAD crate, record `ENCRYPTED` flag bit, header `aead_suite`/`key_id`, the two new reason codes, the key loader) is tracked by #108 under the security epic #18.
- The full #14 key-config schema (config-key names, loader precedence resolver) and the #109 secret-handling fail-closed checks are owned by #14 / #109; this spec fixes only the priority and the never-store-the-key rule.
- Encrypting the cursor/counters checkpoints, the quarantine store, and the DLQ metadata is explicitly deferred from v1.

refs #108


ELares:
## Independent adversarial review: SHIP (SHOULD-FIX items addressed)

An independent reviewer (not the author) verified the spec against issue #108, leading with the cryptographic and on-disk-format claims. Verdict: **SHIP-WITH-NITS**, no blockers.

- Nonce no-reuse is sound: `segment_id` unique-and-never-recycled (ADR 0002, whose rationale exists for this AEAD nonce), rotation new-segments-only, and the 32-bit counter cannot wrap.
- The frozen `ironbus.loss-report.v1` schema (#120) is respected: `UnknownKeyId` / `AeadTagMismatch` are treated as append-only with NO prescribed integer values (so no collision or renumber), the v1 set is listed correctly, and adding a variant does not bump `SCHEMA_VERSION`.
- No framing drift: the new header fields reuse genuinely-reserved `flags` bits and reserved bytes already inside the `header_crc` coverage; header size and CRC coverage unchanged. Checksums-over-ciphertext is I7-consistent (CRC-first, then decrypt), so corruption is detected key-free and a tag failure is a genuine authenticity event. All five acceptance criteria covered; honesty bar met; no dashes.

The two SHOULD-FIX accuracy items and the strong no-wrap NIT are fixed in `b3b018c`:
- The old-reader protection was misdescribed as refuse-on-unknown. Unknown flag bits are preserve-and-ignore, so a pre-encryption reader is actually stopped by the 16-byte AEAD tag inflating the record length past its declared plaintext lengths, which the codec total-length self-check rejects as `BadLength` (corruption class). Refuse-on-unknown is now correctly scoped to an unknown `aead_suite` VALUE seen by a future encryption-aware reader. Both paths are stated.
- "only `COMPRESSED` flag defined today" contradicted CONTRACTS.md; corrected to the real `KNOWN` set (`COMPRESSED`, `HAS_KEY`, `HAS_XXH3`).
- The nonce no-wrap argument now cites the `record_count: u32` ceiling (`SegmentWriter::append` refuses at `u32::MAX`), making it airtight regardless of `max_segment_bytes`.

Merging on the green gate (15/15). `refs #108`.

Issue #307, opened 2026-06-07
[TASK] Counter replay-reconciliation and strict cross-restart monotonicity (follow-up to #98)


ELares:
Follow-up to #98. PR #306 made the OPERATIONAL resilience counters durable via a crash-safe snapshot (recovered at open, flushed on the checkpoint cadence + graceful shutdown, corrupt-never-blocks-recovery). That gives a MONOTONIC LOWER BOUND: on a crash the resumed value can be lower than at crash time by at most the increments since the last snapshot. The #98 review flagged that #98's literal acceptance criteria go further than a snapshot-only design.

## Remaining (the deferred criteria)
- **Checkpoint-plus-replay reconciliation** for the SKIP/LOSS counter family (records_skipped, bytes_skipped, last_skip_offset): on open, reconcile the snapshot value with what the durable LOG / loss report implies, so the counter is "not lower than before the crash" even across a kill -9. Today the recovery-loss gauges (`ironbus_recovery_loss_bytes/records`) are already repopulated from the durable `LossReport` on open (a separate, log-derived mechanism in health.rs), which partially covers this; this issue is to make the reconciliation explicit and unified.
- `ironbus_last_skip_offset` = max(checkpoint, replay).
- A strict cross-restart MONOTONIC NON-DECREASING property test for the reconciled counters.
- `ironbus_counter_checkpoint_repair_total` (counts a reconciliation/repair on open).

## Note
The OPERATIONAL counters (produced, delivered, acks, dead_lettered, reaped, ...) cannot be replay-derived from the log, so for them the snapshot lower-bound from #306 is the only viable design and is complete. This issue is specifically the skip/loss-family reconciliation (co-owned with #7), not the operational counters.

refs #98, #7, #16

ELares:
Delivered in PR #322 (merged, squash `367e97be`), on a fully green gate plus two independent adversarial review rounds (the first caught a real semantic mismatch and a vacuous test; both were fixed and re-verified).

All of this issue's named acceptance criteria for the skip/loss family are met:
- Checkpoint-plus-replay reconciliation for `records_skipped`, `bytes_skipped`, `last_skip_offset`: on open each becomes `max(snapshot, replay-from-durable-LossReport)`, so the value is never lower than before a crash, even across a `kill -9`. `records_skipped` is now a dedicated recovery-loss counter derived from `loss.total_records_lost_estimate()` (the original draft had mistakenly aliased it to the consumer-truncation counter).
[...]

PR #322, opened 2026-06-07
Reconcile the skip/loss counter family on open via checkpoint-plus-replay (#307)

ELares:
Follow-up to #98/#306. PR #306 made the resilience counters durable via a crash-safe snapshot, giving a monotonic **lower bound** (a `kill -9` between the last cadence snapshot and the crash loses the post-snapshot increments). This PR implements the deferred criteria for the **SKIP/LOSS counter family only** (the operational counters cannot be replay-derived and are complete as-is under #306).

## The reconciliation (max of checkpoint and log-replay)
On `Engine::open`, after seeding the counters from `counters.ckpt` and recovering the durable log, the broker reconciles each skip/loss value to `max(snapshot, replay)`, where the *replay* value is what the durable log / loss report implies:
- `records_skipped` (the existing `ironbus_truncated_records_total`) -> the loss report's total records lost,
- `bytes_skipped` (new gauge) -> the loss report's total bytes skipped (the same total the `ironbus_recovery_loss_bytes` gauges are repopulated from),
- `last_skip_offset` (new gauge) = `max(checkpoint, replay)`, the recovered head a torn-tail recovery landed on (`0` when the log is clean).

This is the **explicit, unified** form of the lower bound #306 left implicit, restoring a strict cross-restart **monotonic non-decreasing** property for the skip/loss family even across an arbitrary crash point including `kill -9`.

## Never lowers a counter, never blocks recovery
Reconciliation is a **pure `max`** over already-recovered in-memory values, so it can only **raise** a value, never lower one (preserving #306's lower bound). It is **best-effort on the read side**: an absent or malformed loss report degrades to an empty report (replay all-zeros) and the snapshot value stands, exactly like #306's corrupt-never-blocks-recovery on the write side, so it can never fail `Log::open`/recovery.

## The new counter
A reconciliation that actually raises a skip/loss value above its snapshot increments the new `ironbus_counter_checkpoint_repair_total`, added to the `FROZEN_RESILIENCE_COUNTERS` exact-set so `the_resilience_counter_taxonomy_is_frozen` stays green. The two reconciled values render as the `ironbus_bytes_skipped` and `ironbus_last_skip_offset` **gauges** (not `_total`), so the frozen `_total` set grows by exactly the repair counter. The snapshot codec appends the three new fields LAST (forward/backward compatible: a pre-#307 `counters.ckpt` decodes with them zero, then reconciliation re-derives them from the durable log).

## The property test
`reconciled_skip_loss_counters_never_resume_lower_across_an_arbitrary_crash` sweeps a grid of crash points (record count x torn-tail byte size, standing in for a `kill -9` that lost post-snapshot increments) and asserts the reconciled family on reopen is never lower than before the crash. Other tests: reconciliation raises and counts a repair; `last_skip_offset == max(checkpoint, replay)` in both orderings; no repair when the snapshot dominates; a missing/corrupt loss report degrades to the snapshot and does not fail recovery; plus the `/metrics` rendering and the snapshot round-trip.

## Verification
`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-server -p ironbus-storage -p ironbus-core` all pass locally (the frozen-taxonomy and crash-recovery gates included). `docs/METRICS.md` documents the new counter and the reconciliation semantics; a terse CHANGELOG bullet added.

## Review follow-up (honest semantic split)
Addressing the adversarial review. The reconciliation no longer maxes the **consumer-truncation** count `truncated_records` (a force-reap-driven, transient, consumer-cursor quantity that is NOT in the durable `LossReport` and NOT replay-derivable) against the recovery torn-tail loss estimate, which conflated two different quantities. The reconciliation now covers ONLY the genuinely replay-reconstructable **recovery-loss** family:
- a new `records_skipped` counter (the `ironbus_records_skipped` gauge) reconciled to `max(snapshot, loss.total_records_lost_estimate())`,
- `bytes_skipped` reconciled to `max(snapshot, loss.total_bytes_skipped())`,
- the recovery-head contribution to `last_skip_offset = max(checkpoint, replay)` (an upper bound on the highest skipped offset).

For that recovery-loss family the cross-restart **monotonic non-decreasing** guarantee genuinely holds even across a `kill -9`. The consumer-truncation counters (`truncated_records`, `truncations`, and the consumer-truncation contribution to `last_skip_offset`) keep #306's **snapshot-only lower bound**, like the operational counters, and are documented as such in `docs/METRICS.md` and the CHANGELOG.

The property test now has real teeth: it seeds a non-zero recovery-loss snapshot, loses a post-snapshot crash so the next reopen's replay is empty and strictly below the snapshot, and asserts the `max` preserves it (a `.min()` reconciliation provably fails the test, verified by flipping it). A separate test asserts only the weaker snapshot-lower-bound property for the consumer-truncation count, driven through the real force-reap path. The `COUNTERS_PAYLOAD` cap is bumped 128 -> 256 (snapshot is now `1 + 8*15 = 121` bytes).

**Residual limitation / `refs` not `Closes`:** the consumer-truncation reconciliation is genuinely **not log-derivable** (the count is a transient runtime quantity absent from the durable loss report), so it cannot be replay-reconciled and retains the snapshot-only bound. This is the correct resolution of the issue's partly-flawed premise. Because a real (non-closable) gap remains for the consumer-truncation family, this is `refs #307` and the issue stays open; the orchestrator can decide whether to file a follow-up or annotate #307.

refs #307


ELares:
## Independent adversarial review: SHIP-WITH-NITS (sending back to earn `Closes #307`)

An independent reviewer (not the author) verified the reconciliation against issue #307. The core is correct and has no blockers: each reconciled value is a true `max(snapshot, replay)` that can only raise (never lowers, preserving #306's lower bound), the runtime `last_skip_offset` watermark is monotonic, reconciliation never fails `Log::open` (an absent/malformed loss report degrades to the snapshot), the snapshot codec is backward-compatible (an old 11-field #306 snapshot decodes with the 3 new fields defaulting to 0), and the frozen taxonomy holds (`ironbus_counter_checkpoint_repair_total` is in the exact-set; the two new gauges are correctly non-`_total` and excluded). But two findings mean `Closes #307` is not yet earned:

1. **SHOULD-FIX (semantic mismatch).** `truncated_records` is incremented by consumer below-earliest truncations (a force-reap-driven quantity that is NOT in the durable `LossReport` and NOT replay-derivable from the log), yet the reconciliation maxes it against `loss.total_records_lost_estimate()` (a recovery torn-tail quantity). These are different quantities. The consequence: a `kill -9` that loses post-snapshot consumer-truncation increments still resumes lower than before the crash, exactly the regression #307 set out to fix. The honest position (and the issue's own premise is partly flawed here): the RECOVERY-LOSS family is genuinely checkpoint-plus-replay reconcilable (the `LossReport` is durable), but the CONSUMER-TRUNCATION counts share the operational counters' snapshot-only lower bound. Reconcile only the replay-derivable portion, and document the consumer-truncation portion's weaker guarantee, rather than maxing two different quantities.
2. **SHOULD-FIX (vacuous test).** `reconciled_skip_loss_counters_never_resume_lower_across_an_arbitrary_crash` only produces records and never triggers a truncation, so every pre-crash skip/loss value is 0 and every assertion reduces to `after >= 0` (trivially true for `u64`). It would still pass if reconciliation used `.min()`. The named regression is not exercised. (The max property IS tested elsewhere via non-zero snapshots, so this is a test-rigor gap, not a missing guarantee.)
3. **NIT (cap headroom).** The counter snapshot is now `1 + 8*14 = 113` bytes against the 128-byte `COUNTERS_PAYLOAD` cap, so only one more `u64` field fits; a 16th field would silently return `Truncated` and fail the graceful-shutdown checkpoint flush. The headroom comment still says "89 bytes ... room for several more." Bump the cap to leave real headroom and correct the comment.
4. **NIT (documented).** `last_skip_offset` replay uses the whole recovered head (`flushed`) whenever any torn tail exists, an upper bound on the highest skipped offset rather than the exact offset. Monotonic and documented, acceptable, but METRICS.md should describe it honestly.

Sending back to split the reconciliation honestly (recovery-loss reconciled vs consumer-truncation snapshot-only), give the property test real teeth, and bump the payload headroom. `Closes #307` will become `refs #307` with the residual consumer-truncation gap tracked, unless the split fully earns it.

ELares:
## Re-review of the fix: SHIP

A second independent reviewer verified the fix delta (`4027fb78`). All four findings are resolved with no new defect:

- **Semantic split correct.** `reconcile_skip_loss_counters` now maxes against the durable `LossReport` ONLY the genuinely replay-derivable family: `records_skipped` (a new recovery-loss gauge with NO runtime increment, from `loss.total_records_lost_estimate()`), `bytes_skipped`, and the recovery-head component of `last_skip_offset`. It no longer touches `truncated_records`. The consumer-truncation counters (`truncated_records`/`truncations`) stay incremented at the force-reap sites and keep #306's snapshot-only lower bound. No double-count with the existing recovery-loss gauges.
- **Property test now has teeth.** It seeds a non-zero recovery-loss snapshot, reopens against a now-durably-truncated clean log so replay is strictly below the snapshot, and asserts the resumed value is not lower; a `.min()` reconciliation provably fails it, and an `exercised >= 30` guard prevents a vacuous re-emptying. A separate test asserts only the weaker snapshot-lower-bound for the consumer-truncation counter.
- **Taxonomy + typing.** `ironbus_records_skipped` is a gauge (non-`_total`), excluded from `FROZEN_RESILIENCE_COUNTERS`; the only `_total` addition remains `ironbus_counter_checkpoint_repair_total`; `the_resilience_counter_taxonomy_is_frozen` passes.
- **Payload headroom safe.** `COUNTERS_PAYLOAD` 128 to 256; snapshot is 121 bytes (15 fields) with real headroom; an old #306 89-byte snapshot still decodes (tolerant short-read), the comment is corrected. No on-disk format break.
- Docs/CHANGELOG honestly distinguish the reconciled (cross-restart monotonic) recovery-loss family from the snapshot-only counters, and describe `ironbus_last_skip_offset`'s replay value as the recovered head (an upper bound).

Merging on the green gate (15/15). `refs #307` in the body; closing #307 separately with the rationale that its named skip/loss family is now fully reconciled and the consumer-truncation counter is inherently not log-derivable.

PR #323, opened 2026-06-07
Append actor + group commit: stop holding the engine lock across fsync (#177)

### LINKED ISSUE #177: [TASK] Append actor + group commit: stop holding the engine lock across fsync

ELares:
## What

Replaces the `Arc<Mutex<Engine>>`-held-across-fsync produce path with a single **append-actor thread** that owns the `Engine`. Connection handlers fan in over a **bounded `std::sync::mpsc::sync_channel`** and send commands instead of locking the engine, so no handler holds a lock across an `fdatasync`. The actor **group-commits** a drained batch of produces with ONE `fdatasync` and acks the whole batch only after it. This removes the head-of-line block (a stalled disk no longer freezes every connection) and amortizes the fsync.

Closes #177. Also folds in the #176 O(n^2) re-decode fix and refs #1.

## Design

- **Engine split (`engine.rs`):** `produce` now delegates to two new primitives so they are the single source of truth:
  - `append_no_sync` writes the record and records the pre-sync produce stats, NO fsync.
  - `commit_batch` issues the ONE covering `fdatasync`, advances the durable head, and runs the once-per-batch retention reap + idle-group sweep.
- **Append actor (`actor.rs`):** `EngineHandle` (cloneable, one per connection) with `produce` / `with(job)` / `shutdown`. The run loop blocks for one command, drains everything else available into the same pass, appends each `Produce` (no sync), then issues ONE `commit_batch` covering the drained batch and releases the parked replies. A non-produce job flushes the pending batch first (consistent durable head + total order). A closed channel is a typed `ActorGone` (never a panic). An `EngineAccess` trait lets the session dispatch run over the real actor in production and a direct `&mut Engine` in unit tests, so the dispatch logic is written once.
- **Session (`session.rs`):** `Session::process` takes an `EngineAccess` and returns a `Progress` (`consumed`, `needed`, `committed_progress`). Pings/Connect are answered in the handler WITHOUT the actor. Acks/flow/sub run as `with` jobs. The #176 fix uses the frame parser's `Incomplete { needed }` hint (surfaced as `Progress.needed`) so the read loop does not re-decode a trickled near-cap frame until it is whole.
- **Server / health / CLI / client:** all drive the actor through the handle. `cmd_serve` graceful shutdown asks the actor to drain (flush the pending batch + checkpoint every group), preserving #195.
- **Fault fs (`fault.rs`):** adds a condvar **sync gate** (stall `sync_data` without erroring) and a **sync counter**, for the acceptance tests. No wall-clock sleep.

## How each invariant is preserved

- **I2 (ack-implies-durable):** a `Produce` reply is sent only AFTER the covering `commit_batch` returns; with group commit the whole batch is acked only after the single `fdatasync`. Proved by `an_ack_is_sent_only_after_the_covering_fsync_completes_i2` (the reply is not available while the gate stalls the fsync).
- **Single total durable order:** the actor appends serially in arrival order and assigns contiguous offsets (asserted in the group-commit test).
- **No deadlock / no lost replies:** every command gets exactly one reply; a closed channel returns `ActorGone` rather than hanging (`a_closed_actor_is_a_typed_error_never_a_hang`). The bounded channel backpressures producers without deadlock.
- **Pings not blocked by a stalled produce fsync:** pings are answered in the handler and a ping-only pass never sends a command to the actor (it skips the interval checkpoint), so a stalled produce fsync cannot block another connection's ping.
- **Graceful shutdown drains:** the shutdown command flushes the pending batch and checkpoints every group before the actor exits; reopening shows the committed cursor persisted with no acked loss (`graceful_shutdown_drains_the_batch_and_checkpoints_with_no_acked_loss`).

## Acceptance tests

- `server::tests::a_stalled_produce_fsync_does_not_block_another_connections_ping` (the explicit acceptance): producer A's `sync_data` is gated mid-fsync via the fault fs; meanwhile B's ping returns `Pong`. No wall-clock sleep (a condvar waits for A to be parked).
- `actor::tests::a_batch_of_concurrent_produces_issues_one_fdatasync_not_n`: a drained batch counts exactly ONE sync via the instrumented fault fs.
- Plus I2, ordering, no-deadlock, the shutdown drain, the single-produce-one-sync path, and a session-level `needed`-hint test for the #176 fix.

## O(n^2) re-decode fix (#176)

The read loop no longer re-runs `Session::process` over the whole input buffer on every byte: `process` returns `Progress.needed` (the partial trailing frame's full framed length, from the parser's `Incomplete { needed }`), and the loop skips re-decoding until the buffer reaches it.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and the full test suite (`ironbus-server` 199, `ironbus-storage` 126 + crash-recovery/determinism gates, `ironbus-core` 205, plus `ironbus-cli`/`ironbus-client`) all pass. `ironbus-core` stays IO-free (AST check). No new resilience counters, frozen wire protocol unchanged, no em/en dashes. Docs updated (WAL.md, RISK_REGISTER.md RR-19 -> FIXED, README).


ELares:
## Independent adversarial review: SHIP

An independent reviewer (not the author) traced all ten safety invariants of the append-actor + group-commit model against the code and could not break it. Verdict: **SHIP**, no blockers.

- **I2 ack-implies-durable HOLDS.** The log advances `flushed_offset` only after the `fdatasync` returns `Ok` (and sets `active=None` and returns `WriterFrozen` BEFORE advancing on a sync error); `commit_batch` syncs first, and the actor releases every parked produce reply only AFTER `commit_batch` returns `Ok`. No error/shutdown/batch-boundary path acks before durability.
- **Failed-fsync semantics correct.** On a sync error the whole drained batch gets `Fatal`/`WriterFrozen`, the head never advances, and the un-synced records become a recoverable torn tail. No false ack.
- **No deadlock / exactly-one-reply.** The actor never sends into the command channel; one-shot `sync_channel(1)` reply channels never block the actor; a closed channel maps to a typed `ActorGone`, never a panic or hang; the bounded channel backpressures without deadlock.
- **Single total durable order** (serial append, FIFO drain, monotonic offsets; one connection's two pubs can never co-batch because each awaits its reply), **crash-consistency preserved** (the produce split did not move the durability boundary; the 21 crash-recovery + determinism sweeps pass), **actor lifecycle sound** (single owner, no aliasing, joined on shutdown), **graceful shutdown drains** (final flush + checkpoint, exit 0, no acked loss), and the **#176 `Incomplete { needed }` re-decode fix** is a pure delay gate that cannot drop or misframe a trickled frame.
- Pings are answered without the actor, so a stalled produce fsync does not block another connection's ping (the explicit acceptance test). The honest caveat that acks serialize behind a stalled produce fsync is inherent to the single-writer model and documented (RR-19). ironbus-core stays IO-free; clippy pedantic clean; no panic in lib paths; frozen wire protocol and resilience taxonomy unchanged; the #322 merge is clean.

One Windows-only test flake surfaced after the #322 rebase: `a_batch_of_concurrent_produces_issues_one_fdatasync_not_n` raced the actor's initial drain (it could fsync a partial batch before the burst finished enqueuing). Fixed in `7752ee5` with a gated-primer barrier so the burst provably lands in one drained batch, verified deterministic 60/60 locally and now green on Windows and macOS. Production group-commit behavior is unchanged.

Merging on the full green gate (15/15). `Closes #177`.

PR #324, opened 2026-06-07
Fix flaky installer_verify temp-dir collision on parallel macOS CI

ELares:
## Problem

`crates/ironbus-cli/tests/installer_verify.rs` has flaked on `test (macos-latest)` as a spurious `verify_checksum` failure (exit 2). It last flaked PR #323 and is a pre-existing intermittent failure on the parallel runner.

Root cause is a test-harness race, not an installer bug. The private `mod tempdir`'s `TempDir::new(prefix)` built the directory name from `prefix-{process::id()}-{SystemTime::now().as_nanos()}`. The six tests in this file run in **parallel under one test process**, so `process::id()` is identical for all of them, and `SystemTime::now()` is coarse enough on the macOS CI runner that two fixtures created in the same tick collide on the same path. With the idempotent `create_dir_all` they then **share one directory**, and whichever `TempDir` drops first runs `remove_dir_all` and deletes the sibling's directory mid-run, which surfaces as the spurious `verify_checksum` failure.

## Fix

Only the `tempdir` module changes:

- Add a process-wide `static COUNTER: AtomicU64` and append `fetch_add(1, Relaxed)` to the directory name, so every `TempDir::new` call gets a guaranteed-unique path even with the same `pid` and clock tick.
- Switch `create_dir_all` to `create_dir`, so any future name collision fails loudly instead of silently sharing a directory.
- Add a comment explaining why (parallel tests, same pid, coarse clock, sibling-deletion-on-drop).

The six test functions, the `Fixture`, `run_verify`, `sha256_hex`, and `scripts/install.sh` are all unchanged, so the fail-closed tamper-rejection contract (tampered binary, unlisted asset, empty/malformed `SHA256SUMS`, missing binary all rejected; matching binary accepted) is preserved exactly.

This is a `#[cfg(test)]`-only test-harness determinism fix. No shipped crate, lib path, or installer behavior changes. MSRV 1.78 safe (`AtomicU64`/`fetch_add`).

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy -p ironbus-cli --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-cli --test installer_verify` run 10x in a loop: all 6 tests pass every run, deterministically.


ELares:
## Independent review: SHIP

An independent reviewer (not the author) confirmed the fix: the process-wide `static AtomicU64` counter makes every `TempDir::new` name unique regardless of identical pid or coarse clock, so two parallel fixtures can never share a directory (and `create_dir` makes any residual collision fail loudly instead of silently sharing). A full-file scan found no other cross-test shared state (no global env/cwd mutation; every fixture is isolated to its own temp dir; `scripts/install.sh` is read-only and untouched). The six tests, `Fixture`, `run_verify`, and the fail-closed tamper-rejection contract are byte-identical. Verified green 15/15 including under `--test-threads=16` stress; clippy pedantic clean; test-only; no dashes.

Merging on the green gate.

PR #325, opened 2026-06-07
Broadcast-consumer cumulative ack (ack-all-up-to-offset) (#288)

### LINKED ISSUE #288: Broadcast-consumer cumulative ack (ack-all-up-to-offset)

ELares:
Implements the BROADCAST half of the cumulative-ack verb (#288), the safe counterpart to the work-group HARD-REJECT that shipped in #63. The work-group rejection is left UNCHANGED.

## Broadcast cursor + validated, idempotent commit
- A `WorkGroup` gains a `broadcast` flag (a single-member group that sees every record in order). `Engine::set_broadcast_in` / `is_broadcast_in` mark/read it; it is mutually exclusive with `key_shared` (each mode-set clears the other), so a competing cursor can never wear the broadcast flag.
- `AckCursor::commit_up_to(up_to)` is the new cursor primitive: it commits the watermark to an EXCLUSIVE offset in one move, **idempotent and monotonic** (an `up_to` at or below the current commit is a no-op success, the watermark never moves backwards), clipping or absorbing any acked-ahead runs the new watermark subsumes.
- `Engine::cumulative_ack_in` accepts the verb ONLY for a broadcast group. It validates `up_to` against the durable head and the earliest-retained offset: past the head or below earliest is the new typed `EngineError::CumulativeAckOutOfRange` (never a panic), and the cursor is left untouched on rejection. A successful or idempotent commit advances the single cursor.

## Work-group rejection unchanged (the safety trap)
A competing or `key_shared` group (and an unknown group) still returns `EngineError::CumulativeAckOnWorkGroup`. Only a genuine broadcast group-of-one commits cumulatively, so a shared out-of-order cursor can never be committed past peers' still-in-flight messages.

## New wire tag 19 (append-only, frozen tests updated)
- `FrameType::CumulativeAck = 19`, the next free value. NO existing tag renumbered or reused.
- Body layout (`CumulativeAckBody`): the exclusive `up_to` offset as a `u64` LE (8 bytes), then the work-group name as the remainder of the body (empty = default group), the same whole-tail-is-the-name shape as `SubBody`. Documented in `docs/CONTRACTS.md`.
- Both frozen tests (`type_tags_are_a_stable_bijection`, `type_tags_have_their_exact_frozen_wire_values`) updated to pin tag 19, and the unknown-tag property range narrowed from `19..=255` to `20..=255`.

## CLI + client surface
- `Client::cumulative_ack(group, up_to)`.
- A session handler maps the verb to the engine: `Ok` on commit (or idempotent no-op), a typed `Err` with the engine reason on rejection, end-session on a fatal storage error.
- `ironbus cumulative-ack --group <name> --up-to <offset>` subcommand drives the engine path; `serve --broadcast-group <name>` (repeatable) marks broadcast groups server-side.

## Tests
- Cursor: `commit_up_to_advances_the_watermark`, `commit_up_to_is_idempotent_and_never_regresses`, `commit_up_to_subsumes_and_clips_the_ahead_set`, `commit_up_to_into_a_gap_does_not_absorb_a_non_contiguous_run`, `commit_up_to_then_advance_over_a_contiguous_run`, property `commit_up_to_interleaved_with_acks_holds_invariants`.
- Engine: `broadcast_cumulative_ack_commits_the_cursor_up_to`, `broadcast_cumulative_ack_is_idempotent_and_never_regresses`, `broadcast_cumulative_ack_rejects_past_the_durable_head`, `broadcast_cumulative_ack_rejects_below_the_earliest_retained`, `marking_broadcast_clears_key_shared_and_vice_versa`, and the existing `cumulative_ack_is_rejected_on_a_work_group` (now also covering an unknown group).
- Proto: `cumulative_ack_round_trips`, `cumulative_ack_rejects_a_short_body`, `any_cumulative_ack_round_trips`, plus the two updated frozen-tag tests and the narrowed unknown-tag property.
- Session: `cumulative_ack_over_the_wire_commits_a_broadcast_group_and_rejects_a_work_group`.
- Client: `a_cumulative_ack_ok_reply_succeeds`, `a_cumulative_ack_error_reply_is_surfaced`.
- CLI: `cumulative_ack_subcommand_drives_the_engine_path`.

Local `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-proto -p ironbus-server -p ironbus-cli -p ironbus-core -p ironbus-client` all pass (the new tests, the frozen-tag tests, and the crash-recovery gates). `ironbus-core` stays IO-free; no new resilience counter (so the frozen taxonomy is untouched); the `cfg(not(unix))` `cmd_serve` stub consumes the new `broadcast_groups` so Windows `-D warnings` stays clean.

Closes #288


ELares:
## Independent adversarial review: REQUEST-CHANGES (two safety BLOCKERS)

The cursor math, validation, wire evolution, and bar are all correct, but the feature's defining safety property is documented, not enforced, and that reopens the JetStream AckAll silent-drop trap #63's guard exists to prevent. `Closes #288` is not justified as written.

**BLOCKER 1: the "group of one" premise is unenforced; a multi-consumer broadcast group silently drops in-flight messages.** The `broadcast` flag (engine.rs ~2244) is only a permission bit gating `cumulative_ack_in`. It does not restrict membership or change delivery, which still goes through the unchanged competing claim path (`poll_in`), so two connections can both SUB the same broadcast group and hold concurrent in-flight leases. Reachable loss with only shipped verbs:
- `set_broadcast_in("g", true)` (or `serve --broadcast-group g`)
- A: SUB g, FLOW -> leases offset 0 (in-flight, unacked)
- B: SUB g, FLOW -> leases offset 1
- B: ACK 1 -> cursor ahead set `[1,2)`, committed = 0
- CUMULATIVE_ACK g up_to=2 -> `commit_up_to(2)` absorbs `[1,2)`, committed jumps to 2 while A still holds offset 0 unacked
- A's lease expires (or A nacks): the next `poll_in` skips offset 0 via `is_acked(0)` (`0 < committed=2`); ack/nack of offset 0 are no-ops. Offset 0 is permanently, silently dropped, never processed, never redelivered.

**BLOCKER 2: flipping a live group to broadcast does not quiesce the lease table.** `set_broadcast_in` clears `router` but leaves `g.leases` intact, so: SUB competing, accrue out-of-order in-flight leases, flip to broadcast, cumulative-ack past them => the same silent drop. Clearing the router does not make a populated competing group a genuine group-of-one.

**PASS (verified, no change needed):** `AckCursor::commit_up_to` is monotonic, idempotent, clips a straddling ahead range, and never absorbs across a gap (the six cursor tests + the interleaved proptest pin this); the engine `up_to` range validation (reject past-head / below-earliest with typed `CumulativeAckOutOfRange`, inclusive boundaries) is correct; the work-group HARD-REJECT is preserved for competing/key_shared/unknown groups; the frozen wire evolution is append-only and correct (tag 19, both frozen tests updated, unknown range narrowed, unambiguous body codec, short body rejected); and the bar is met (IO-free core, no panic in lib, Windows stub consumes `broadcast_groups`, taxonomy unchanged, no dashes, docs match). The cursor primitive is sound; the defect is the engine committing it under an unverified precondition.

The safety distinction: cumulative-acking past your OWN in-flight leases (a genuine group-of-one) is the consumer's explicit choice and is safe; acking past ANOTHER consumer's in-flight message is the loss bug. The fix must ENFORCE the group-of-one invariant in code (reject a second active subscriber to a broadcast group, and reject/guard the broadcast flip so the group cannot carry competing multi-member in-flight state), so a cumulative ack can never commit past a different consumer's in-flight message, with regression tests reproducing both sequences above and asserting a typed rejection (no silent drop) while the legitimate single-consumer cumulative ack still works.

Sending back. `Closes #288` should become `refs #288` until the safety gap is closed in code.

ELares:
Addressed both safety BLOCKERS by ENFORCING the group-of-one invariant in code (commit 83c931e), not docs.

**Enforcement design**
- Per-group active-subscriber set in the engine (keyed by the connection `MemberId`). A broadcast group accepts AT MOST ONE active subscriber; the session registers on SUB (`Engine::subscribe_in`) and deregisters on UNSUB / subscription switch / disconnect (`Engine::unsubscribe_in`). New typed error `EngineError::BroadcastGroupBusy { group }` (no panic, no silent behavior).
- `set_broadcast_in(group, true)` now refuses (same typed error) to flip an existing group that carries competing in-flight state: live in-flight leases, a non-empty out-of-order acked-ahead set, or more than one active subscriber. A fresh / quiescent group (the configure-time `--broadcast-group` path) flips freely.
- `subscribe_in` never creates the group (first FLOW still does), so the "SUB alone does not create the group" invariant holds; a broadcast group always exists already (created by `set_broadcast_in` at configure time), so the cap binds every one.

**BLOCKER 1 (multi-subscriber broadcast)** — Exploit A step 3 (the second SUB) is rejected with `BroadcastGroupBusy`, so the competing in-flight leases that the cumulative ack would commit past can never accrue. Engine test `a_broadcast_group_rejects_a_second_concurrent_subscriber` + `exploit_a_no_offset_is_silently_dropped_under_the_subscriber_cap`; over the wire `exploit_a_a_second_sub_to_a_broadcast_group_is_rejected_over_the_wire` (the second SUB answers a typed `Err`, every produced record is delivered to the lone consumer, none skipped).

**BLOCKER 2 (flip a live competing group)** — Exploit B step 2 (the flip) is rejected with `BroadcastGroupBusy` because the group carries an out-of-order ahead set + in-flight lease, so the unsafe cumulative ack is never reachable and the group stays a work-group (still HARD-REJECTs cumulative ack). Tests `exploit_b_flipping_a_live_competing_group_to_broadcast_is_rejected` and `flipping_a_group_with_in_flight_leases_to_broadcast_is_rejected`.

**Legitimate case preserved** — `a_lone_broadcast_consumer_cumulative_acks_past_its_own_in_flight_leases`: a single broadcast consumer that leased offsets 0,1,2 in order cumulative-acks up to 3 and the next poll delivers offset 3 (no redelivery). The `up_to` range validation, idempotency, and the #63 work-group reject are unchanged.

I verified each new test FAILS pre-fix and PASSES post-fix by neutering each guard in turn (the second-subscriber cap: second SUB returns `Ok` instead of `Err`; the flip guard: `set_broadcast_in` returns `Ok` instead of `BroadcastGroupBusy`).

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-core -p ironbus-proto -p ironbus-server -p ironbus-cli` all green (incl. the frozen-tag and crash-recovery gates). Core stays IO-free, no panics in lib paths, no new resilience counter, the `cfg(not(unix))` `cmd_serve` stub still consumes `broadcast_groups`, no em/en dashes. Docs (CONTRACTS.md / CLI.md / USAGE.md) and CHANGELOG record the new typed rejection.

Both exploits are provably closed in code with the legitimate case working, so `Closes #288` is now earned. Not self-merging; leaving this for re-review.

ELares:
## Re-review of the safety fix: both BLOCKERs closed for named groups; one residual silent-loss path to close before merge

The fix (`83c931e`) genuinely closes both prior BLOCKERs for the named-group surface: the second SUB to a broadcast group is rejected (`BroadcastGroupBusy`) atomically with registration in one actor turn (no TOCTOU), the `set_broadcast_in` flip guard rejects on `in_flight() != 0 || !ahead_ranges().is_empty() || subscribers.len() > 1` (all three competing-state forms), the deregister is correctly wired on disconnect / UNSUB / subscription-switch with a unique per-connection `member_id` (no false busy, idempotent re-SUB), the legitimate single-consumer cumulative-ack still works, and the regression tests have teeth (neutering either guard breaks them). The cursor math, tag-19 wire evolution, range validation, and the #63 work-group HARD-REJECT are unchanged.

Two items, both worth closing before merge given the no-silent-loss invariant:

**SHOULD-FIX 1 (residual silent-loss bypass, default group).** `set_broadcast_in` skips name validation for an already-existing group, and the default group `""` is pre-created at open, so `--broadcast-group ""` flips the default group to broadcast while empty (the flip guard passes). The subscriber cap never binds the default group because its consumers never SUB (`handle_sub` only registers a non-empty group), so two connections can both poll the default subscription, hold competing in-flight leases, and a wire `cumulative_ack` with an empty group name commits past a peer's in-flight offset, the same silent drop, for the default group. It requires the operator to pass an empty `--broadcast-group ""`, so it is odd input, but it is a reachable silent-loss path and IronBus's invariant is zero silent loss. Fix: `set_broadcast_in` must reject the default/empty group with a typed error (a group whose consumers cannot be capped must never be broadcast), and document that `--broadcast-group` marks a NAMED group only.

**SHOULD-FIX 2 (missing liveness test).** The disconnect-deregister is correct by inspection, but the only slot-release test calls `unsubscribe_in` directly; there is no over-the-wire test that an abrupt DISCONNECT (the common case) frees the broadcast slot. A future refactor dropping the disconnect cleanup would brick the group `BroadcastGroupBusy` forever with no failing test. Add: A SUBs the broadcast group, A's connection drops, B SUBs and succeeds.

**NIT 3.** Deregister is a best-effort plain call, so a panic unwinding out of the connection loop would leak the registration (same exposure as the pre-existing `leave_current_key_shared` cleanup, no panic source in those lib paths, not a regression). Worth a one-line comment.

Sending back to fold in 1 and 2 so there is zero residual silent-loss path before merge. `Closes #288` will be fully earned once the default-group bypass is closed.

ELares:
## Folded in the two residual re-review items: zero residual silent-loss path

Both SHOULD-FIX items and the NIT are addressed in `8167061`. The default-group bypass is closed in code with a test, so `Closes #288` is now fully earned.

**SHOULD-FIX 1 (default-group bypass) — closed.** `set_broadcast_in(group, true)` now REJECTS the DEFAULT/empty group (`""` / `DEFAULT_GROUP`) with a new typed `EngineError::BroadcastGroupNotNamed` (distinct Display, no resilience counter, frozen taxonomy untouched). It is rejected at the very top of `set_broadcast_in`, before the `contains_key` / name-validation branch that the pre-created default group was slipping through. `serve --broadcast-group ""` surfaces it as a clean `CliError::Usage` startup error, not a panic (the configure-time loop already maps the engine error to `Usage`).

  No other uncapped-group cumulative-ack path remains: `set_broadcast_in` is the ONLY writer of `broadcast = true` in lib code (`WorkGroup::new` always starts `false`; `set_key_ordering_in` only ever clears it), and `cumulative_ack_in` strictly requires `is_broadcast_in(group)`. So the default group can never enter broadcast mode and its cumulative ack always hits the #63 work-group HARD-REJECT (`CumulativeAckOnWorkGroup`), committing nothing.

**SHOULD-FIX 2 (missing liveness test) — added.** `a_disconnect_frees_a_broadcast_groups_group_of_one_slot_over_the_wire` (session layer, real SUB frame path): A SUBs a named broadcast group, A's connection drops via the exact per-connection cleanup the server runs on every exit (`leave_current_key_shared` + `leave_current_subscription`), then B SUBs the same group and SUCCEEDS, and B drains every record. It has teeth: neutering the engine deregister (`unsubscribe_in`) trips it at `assertion ... the disconnect freed the slot`.

**NIT 3 — done.** One-line comment at the disconnect deregister in `server.rs` noting it is a best-effort plain call: a panic unwinding out of the connection loop would leak the registration (same exposure as the adjacent `leave_current_key_shared` cleanup, no panic source in those lib paths today), so a future panic-prone refactor must keep it on every exit path.

**Also.** Engine + CLI reject tests; docs (CONTRACTS.md / CLI.md / USAGE.md) now state `--broadcast-group` marks a NAMED group only (the default group cannot be broadcast); terse CHANGELOG bullet. All existing enforcement is intact: the named-group subscriber cap, the flip guard, the cursor math, the tag-19 wire verb, the `up_to` range validation, the #63 HARD-REJECT, and the legitimate single-consumer cumulative ack are unchanged.

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-core -p ironbus-proto -p ironbus-server -p ironbus-cli` are all green (incl. the frozen-tag, taxonomy-freeze, and crash-recovery gates, plus the new tests).

**Merge note (squash body):** the branch history carries `Closes #288` in this commit plus two earlier (premature) commit bodies. Since the squash aggregates commit bodies, please trim the squash commit message to exactly ONE `Closes #288` (this commit's) so the auto-close fires cleanly once.

Not self-merging; leaving for re-review.

ELares:
## Re-review of the round-2 fix: SHIP

The default-group silent-loss bypass is closed. Verified by direct inspection of `set_broadcast_in`: the new guard `if broadcast && group == DEFAULT_GROUP { return Err(BroadcastGroupNotNamed) }` is the FIRST statement, before any group lookup or mutation; `g.broadcast = true` occurs only at the end of that single function; `WorkGroup::new` starts `broadcast: false` and `set_key_ordering_in` only ever clears the flag. So `set_broadcast_in` is the sole writer of `broadcast = true`, the default/empty group can never reach it, and `cumulative_ack_in` (gated on `is_broadcast_in`) therefore always hits the #63 work-group HARD-REJECT for the default group, committing nothing. The flip guard covers all three competing-state forms (`in_flight() != 0 || !ahead_ranges().is_empty() || subscribers.len() > 1`).

Tests added and confirmed to have teeth: `the_default_group_can_never_be_marked_broadcast` (fails without the guard), `serve_with_an_empty_broadcast_group_is_a_clean_usage_error`, and the over-the-wire `a_disconnect_frees_a_broadcast_groups_group_of_one_slot_over_the_wire` (fails if the disconnect cleanup is removed). Docs (CONTRACTS/CLI/USAGE) state `--broadcast-group` marks a NAMED group only. CI 15/15 green on `a03e011` (incl. windows + musl); clippy pedantic clean.

This PR was held through two full adversarial review rounds: round 1 found two silent-loss BLOCKERs (multi-subscriber drop + unsafe flip), round 2 found a residual default-group bypass, all now closed in code with regression tests. Every reachable silent-loss path is eliminated and the legitimate single-consumer cumulative ack is preserved. Merging on the green gate. Closes #288.

PR #326, opened 2026-06-07
Surface the metric-registry overflow_saturated state as a /metrics gauge (#321)

### LINKED ISSUE #321: [TASK] Surface the metric registry overflow_saturated state as a /metrics gauge
Follow-up from the #320 (#97) re-review. The metric registry's `overflow_saturated` state is tracked internally and documented in METRICS.md as the operator's signal that `ironbus_consumer_lag_records{consumer="__overflow__"}` has become a monotonic lower bound (it triggers only past 1024 distinct over-cap consumers over the broker's lifetime, the defense-in-depth safety valve beyond the 1024-series cap). But it is a Rust accessor only; it is NOT surfaced on `/metrics`, so a Prometheus scraper cannot directly observe when the overflow lag stopped being exact.

Surface it as a gauge, e.g. `ironbus_consumer_overflow_saturated` (0/1), so an operator can alert on it. It is a gauge (no `_total` suffix), so it is excluded from the frozen resilience-counter taxonomy by construction and does not affect `the_resilience_counter_taxonomy_is_frozen`. Small, isolated change in `health.rs` rendering plus a METRICS.md line. Parent: #97 / #16.

ELares:
## What

The metric registry (#97) tracks an internal `overflow_saturated` state: it becomes set only once more than the bounded overflow fold-ledger's capacity of distinct over-cap consumer-lag series have been seen over the broker's lifetime (the defense-in-depth safety valve beyond the 1024-series cap). When set, `ironbus_consumer_lag_records{consumer="__overflow__"}` has become a documented monotonic **lower bound** on the folded lag rather than exact.

Until now this was a Rust accessor only and was **not** surfaced on `/metrics`, so a Prometheus scraper could not directly observe when the overflow lag stopped being exact.

This PR renders it as the gauge **`ironbus_consumer_overflow_saturated`** (value `0` or `1`) on `/metrics`, in the same `health.rs` `consumer_lag_lines` rendering path that emits `ironbus_consumer_lag_records` and the `__overflow__` fold, reading from the registry's existing `overflow_saturated()` accessor. The gauge is `1` once the ledger has saturated, else `0`, so an operator can alert on the transition.

## Out of the frozen taxonomy by construction

It is a **gauge with no `_total` suffix**, so it is excluded from `FROZEN_RESILIENCE_COUNTERS` by that set's `_total` filter and is **not** added to the set. `the_resilience_counter_taxonomy_is_frozen` passes unchanged (this matches how the existing `ironbus_quarantine_bytes` gauge stays out of the taxonomy).

## Changes

- `crates/ironbus-server/src/health.rs`: render the new gauge (HELP + `# TYPE ... gauge` + value) in `consumer_lag_lines`, plus a rendering test (`metrics_renders_the_overflow_saturated_gauge`) asserting the `gauge` TYPE line and value `0` on a fresh registry then `1` after driving the bounded ledger to saturation, in the style of the existing registry-metrics rendering tests.
- `docs/METRICS.md`: document the new gauge (1 = the `__overflow__` series is now a monotonic lower bound because more than the overflow-ledger capacity of distinct over-cap consumers were seen) and add it to the exposition block.
- `CHANGELOG.md`: one terse `Added` bullet.

ironbus-core is untouched (stays IO-free). No new `unwrap`/`expect`/`panic` in lib paths.

## Verification

- `cargo fmt --all -- --check` clean
- `cargo clippy -p ironbus-server --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-server` green (223 passed), including the new gauge test and the unchanged `the_resilience_counter_taxonomy_is_frozen`.

Parent: #97 / #16.

Closes #321


ELares:
## Independent review: SHIP

An independent reviewer confirmed the gauge: `ironbus_consumer_overflow_saturated` renders a well-formed `# TYPE ... gauge` with value `u8::from(overflow_saturated() > 0)` (0 before saturation, 1 after the bounded overflow ledger overflows), sourced from the real registry accessor. The frozen taxonomy is untouched (gauge has no `_total`, excluded by construction; `the_resilience_counter_taxonomy_is_frozen` passes). The test has teeth (fills 1024 series + 1024 ledger slots, asserts 0, drives one more distinct consumer, asserts the flip to 1). ironbus-core untouched, MSRV 1.78, no dashes, diff limited to the gauge + METRICS.md + CHANGELOG. Merging on the green gate. Closes #321.

PR #328, opened 2026-06-07
Quarantine: persisted-footprint gauge + true-FIFO eviction (#315)

### LINKED ISSUE #315: [TASK] Quarantine gauge: surface persisted on-disk footprint + true-FIFO eviction order
Follow-up from the #314 (#134) review. The quarantine store ships sound, but the `ironbus_quarantine_bytes` gauge and the FIFO eviction order have two small gaps worth closing.

## 1. Gauge should reflect the persisted on-disk footprint (SHOULD-FIX)

Today `ironbus_quarantine_bytes` is wired from `Log::quarantined_bytes()`, which is the byte count captured during **this** process's recovery. After a clean reopen with no new corruption, `quarantine_corrupt_span` is never called, so the gauge reads `0` even though the `quarantine/` dir still holds the blobs from previous recoveries (the reopen test at `quarantine_recovery.rs` asserts exactly this).

The HELP text honestly documents the narrow "at the last recovery" scope, so this is a semantic gap rather than a correctness bug (recovery, the live log, and the cap are unaffected). But the intent of a `bytes` gauge is to surface the real disk pressure a bit-rotting node's forensic copies create, which should survive a restart.

The infrastructure already exists: `QuarantineStore::open` calls `scan_bytes` to reconstruct the true on-disk total for cap enforcement. Plumb that persisted total to the gauge (initialize `Log::quarantined_bytes` from a one-time scan on recovery even when there is no corruption skip, or expose it as a second gauge), so the metric reflects what is actually on disk.

## 2. FIFO eviction sorts blob names lexically, not by true offset (NIT)

`make_room_for` evicts oldest-first by sorting blob file names. The segment id is fixed-width hex (sorts numerically), but the `start`/`end` offsets are unpadded decimal, so within one segment two blobs sort lexically (`-100-` before `-9-`) rather than by true offset/recency. This is best-effort forensic ordering only; the cap is strictly enforced regardless of eviction order, so disk can never be exhausted. Zero-pad the offsets (or sort on the parsed numeric tuple) so eviction is true-FIFO.

Both are quality refinements, not safety issues. Parent: #134.

ELares:
Follow-up from the #314/#134 quarantine store review (#315), closing the two gaps the review flagged.

## Part 1: the gauge reflects the PERSISTED on-disk footprint (surviving restart)

`ironbus_quarantine_bytes` was sourced from `Log::quarantined_bytes()`, which was only the bytes captured during THIS process's recovery. After a clean reopen with no new corruption, `quarantine_corrupt_span` is never called, so the gauge read 0 even though `quarantine/` still held the forensic blobs from prior recoveries (real disk pressure). The bytes gauge is meant to surface that footprint, which should survive a restart.

Fix:
- `Log::open`/`recover` now seed `quarantined_bytes` from a one-time **read-only** scan of the persisted quarantine store (`quarantine::persisted_bytes`), reusing the existing `scan_bytes` infra, EVEN WHEN there is no corruption skip. A new corruption-skip capture this recovery re-derives the gauge from the true on-disk total (net of any cap eviction), so the value stays the real persisted footprint.
- **Best-effort and corrupt-never-blocks-recovery:** `persisted_bytes` probes `subdir_exists` FIRST and returns 0 without materializing the subdir when it is absent (so a clean log still never creates `quarantine/`), and a missing or unreadable quarantine dir degrades to 0 and never fails `Log::open`.
- The existing reopen test is updated: it now asserts the gauge reflects the PERSISTED total after a clean reopen (not 0). Added a test that a clean log's scan-on-open never materializes the subdir.

## Part 2: true-FIFO eviction order

`make_room_for` evicted oldest-first by sorting blob file NAMES. The segment id is fixed-width hex (sorts numerically) but the `start`/`end` offsets are unpadded decimal, so within one segment two blobs sorted lexically (`-100-` before `-9-`) rather than by true offset/recency.

Fix:
- Eviction now sorts on the PARSED numeric `(segment_id, start, end)` tuple (`parse_blob_sort_key`) instead of the lexical name, so FIFO is true offset-then-recency order. The blob name format is UNCHANGED, so `scan_bytes` and name parsing still round-trip.
- An OLD-format or foreign blob name that does not parse sorts AFTER every canonical blob (evicted only as a last resort) and is handled best-effort, never a crash.
- Added a test that two blobs in one segment with offsets that sort differently lexically vs numerically are evicted in true numeric (oldest-first) order, and a test that an unparseable old-format blob name is handled best-effort and evicted last.

## Safety properties preserved (#134)

The scan-on-open is strictly **read-only** (it lists and stats blobs, never writes/moves/removes one) and **never fails recovery** (every step degrades to 0). The copy-not-move and never-blocks-recovery properties from #134 are intact. `ironbus_quarantine_bytes` stays a GAUGE (no `_total`), so the frozen resilience-counter taxonomy is unaffected (`the_resilience_counter_taxonomy_is_frozen` passes).

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace ... -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-storage -p ironbus-server` all clean. The quarantine tests, the frozen-taxonomy test, the crash-recovery sweep, and the new/updated tests pass. `docs/METRICS.md` updated to describe the persisted-footprint semantics; terse `CHANGELOG.md` bullet added.

Closes #315


ELares:
## Independent review: SHIP

An independent reviewer verified both parts and the recovery safety. The scan-on-open (`persisted_bytes`) returns `u64` (infallible by signature): `subdir_exists` is consumed by `matches!(.., Ok(true))`, `subdir()` by a `let Ok(..) else return 0`, and `scan_bytes` swallows every list/stat failure, so a missing/unreadable/foreign-blob quarantine dir degrades to 0 and `Log::open` always succeeds (the #134 never-blocks-recovery property). It is strictly read-only and probes `subdir_exists` BEFORE `subdir()`, so a clean log never materializes the subdir (proven by test). The gauge is re-derived (`=`, not `+=`) so no double-count; after a clean reopen it equals the real on-disk total, after recovery+eviction the post-eviction footprint. Part 2 sorts on the parsed numeric `(segment_id, start, end)` tuple (true FIFO), with unparseable/old-format blobs evicted last (never crash) and the cap order-independent so disk can never be exhausted. The #134 copy-not-move bodies are byte-identical; `ironbus_quarantine_bytes` stays a gauge so the frozen taxonomy is unaffected. Tests have teeth. Merging on the green gate. Closes #315.

Issue #20, opened 2026-06-05
[DESIGN] Edge resource constraints: flash wear, RAM ceilings, fsync cost, brownout behavior


ELares:
This issue catalogs the physical realities of edge hardware (SD/eMMC flash endurance, small RAM ceilings, weak ARM CPUs, frequent unclean power loss, missing real-time clocks, and thermal throttling) and turns each into a concrete constraint on the rest of the IronBus design. It is not a feature in itself; it is the gravity well that the storage engine, durability model, retention, and backpressure subsystems all orbit. The deliverable is a documented set of constraints plus a conservative default profile (`tiny`) that is safe to deploy unattended on a battery-less ARM box, and a clear list of which knobs each other subsystem must expose to honor these limits.

## Why this matters for IronBus

This issue is the load-bearing wall under the Edge First and Resilient tenets, and it directly shapes Simple (safe defaults) and Cross Platform (ARM realities differ from the x86 dev box). Edge devices do not behave like datacenter brokers: an SD card has a finite erase-cycle budget, RAM is measured in hundreds of MiB not tens of GiB, the CPU is a slow ARM core that must share cycles between compression, checksums, and the radio, and power can vanish mid-write with no UPS. The systems we borrow from were mostly designed for the opposite environment. Kafka leans on the OS page cache and assumes replicas fail independently (a bad assumption when a whole node shares one battery-less power rail). If IronBus simply copies those defaults, it will burn out flash, OOM under burst, or lose acknowledged data on the first brownout. The safe path must therefore be the default, not an opt-in.

## Prior art

Several systems already encode pieces of the edge story, though none target a single battery-less ARM node directly.

- NATS JetStream writes durable data as append-only `<n>.blk` blocks (default 8 MiB large / 4 MiB medium / 1 MiB small, 2 MiB cap when encrypted) under a per-stream `msgs/` directory with a separate `index.db`. The lesson its own docs draw is explicit: block size must track the device erase block, because large append blocks plus frequent fsync amplify writes and wear SD/eMMC, while too-small blocks add per-block overhead.
[...]

PR #329, opened 2026-06-08
Specify cross-platform segment preallocation; resolve recycling (#40)

### LINKED ISSUE #40: [TASK] Specify cross-platform preallocation and generation-tagged segment recycling
Parent: #4

Specify how the active segment is preallocated and how sealed-then-deleted segments are recycled, with a portable shim so the durability and wear benefits hold on Linux, macOS, and Windows rather than only on Linux. This closes the parent draft's portability hole, where `fallocate` was assumed but the tagline promises Cross Platform.

## Approach

Define a thin platform abstraction with four primitives: preallocate, file-datasync, directory-sync, and sealed-read-map. Preallocation maps to posix `fallocate` on Linux, `F_PREALLOCATE` plus `ftruncate` on macOS, and `SetFilePointer`/`SetEndOfFile` (optionally `SetFileValidData`) on Windows; default ON. Preallocating the active segment to full roll size keeps the file contiguous and avoids an inode-metadata fdatasync on every commit, directly cutting flash wear and commit latency (Edge First). Recycling: keep up to 2 (config knob, default 2) zeroed, renamed former sealed files for reuse as new active segments, each stamped in its header with a monotonically increasing segment generation so stale trailing bytes from a prior life can never be mis-replayed by #7. The generation stamp lives in the record/segment header owned by #5, so coordinate the field there. The fsync semantics align with #6 and the cross-platform build matrix with #17.

## Acceptance criteria

- [ ] The four-primitive platform shim is specified with per-OS implementations and a per-OS durability note.
- [ ] Preallocation of the active segment to roll size (default ON) and its wear/latency rationale are documented.
- [ ] Segment recycling (default 2, zeroed and renamed) with a monotonic generation stamp is specified.
- [ ] The generation field placement is coordinated with #5 and the stale-bytes replay guard with #7.
- [ ] The cross-platform build implications are flagged to #17.

## Failure considerations

Assuming raw `fallocate` everywhere would silently drop preallocation (breaking the power-loss contract) or fail to build on macOS/Windows; the shim prevents that. Recycling a file without a generation stamp risks replaying stale bytes after a crash; the monotonic stamp makes prior-life bytes unambiguously invalid. ENOSPC during preallocation becomes a fail-fast-at-roll event, which #10 and #13 handle cleanly rather than a mid-stream surprise.

ELares:
## What

Adds `docs/PREALLOCATION.md`, a DESIGN spec (M2: Prototype-Ready Design) for portable segment preallocation, and an honest resolution of the segment-recycling question. Docs only, no source changed.

## The preallocation shim

The #135 draft assumed Linux `fallocate`, which neither builds on macOS or Windows nor maps to one syscall across them, leaving a portability hole the Cross Platform tagline does not allow. This closes it with a thin **four-primitive shim** over the existing IO seams (`RandomAccessFile` in `io.rs`, `Filesystem` in `fs.rs`):

| Primitive | Status |
|---|---|
| file-datasync (`sync_data`), directory-sync (`sync_dir`), sealed-read-map (`read_at`) | already IMPLEMENTED in the seam |
| **preallocate** | SPECIFIED here, not yet implemented |

- **Preallocate is mapped per OS** with a per-OS durability note: Linux `fallocate`/`posix_fallocate`; macOS `fcntl(F_PREALLOCATE)` + `ftruncate`; Windows `SetFilePointerEx`/`SetEndOfFile` plus the privileged `SetFileValidData` caveat (`SE_MANAGE_VOLUME_NAME`, exposes stale disk contents). Each has a fallback ladder that bottoms out at today's grow-on-append, so a platform with no preallocation primitive degrades to the shipped v1 behavior, never to an unsafe one.
- **Default ON**, preallocating the active segment to full roll size, with the wear/latency rationale: a contiguous file, and one fewer inode-metadata `fdatasync` per commit (the I2 ack barrier flushes record data without also forcing a length-grow). Edge First.
- **ENOSPC at roll is a fail-fast-at-roll event**, reported as the existing non-fatal `StorageError::AtCapacity` and routed to the #10/#13 disk-full / overflow path (`DropNew` sheds, `DropOldest` force-reaps and retries), not a mid-stream surprise. Records the zero-fill end-of-data (R13) recovery interaction.
- **Maps onto the existing IO seam**: an optional, IO-free-safe default `preallocate` trait method on `RandomAccessFile` that the production `StdFile` overrides per OS; the directory `Filesystem` seam is unchanged. No new plumbing.

## Recycling: resolved, superseded by ADR-0002

The draft's generation-tagged recycling proposal is **SUPERSEDED in v1 by [ADR 0002](https://github.com/ELares/IronBus/blob/main/docs/adr/0002-segments-never-recycled-in-v1.md)**. The at-rest AEAD nonce is `segment_id || record_counter` (#108), so the high 64 bits ARE the `segment_id`; reusing a `segment_id` under a fixed key reuses a nonce, which is catastrophic for GCM and ChaCha20-Poly1305. A generation stamp does not fix this (the nonce is built from `segment_id` alone). So:

- **v1 NEVER recycles** a segment id (fresh monotonic ids forever, pinned by `segment_ids_increase_monotonically_and_are_never_recycled`).
- **No generation stamp** is added to the frozen #5 header: it is unnecessary when ids are never reused, and the reserved header bytes are already spoken for by at-rest encryption.
- The #7 stale-bytes-replay guard the draft wanted is already provided by per-segment framing plus torn-tail truncation (a preallocated zero tail fails the record magic + `header_crc` check).
- Recycling is a **v2 concern gated behind a nonce-safety decision** (a per-life random nonce prefix, or a key rotation on recycle, reusing the new-segments-only rotation machinery). Either would be its own ADR superseding 0002.

The decision was already captured by ADR-0002 and the ADR INDEX (which notes #4's recycle-up-to-2 and #20's generation-tag guard were dropped), so no new ADR is added; the spec cross-references it.

## Build matrix

Flagged to #17: per-OS `#[cfg]` gating, musl-not-glibc verification for the Linux call, the Windows-must-keep-building stub rule, and that the per-OS durability barrier is already in the matrix.

## Files

- `docs/PREALLOCATION.md` (new)
- `docs/WAL.md` (the "Specified but not yet implemented" preallocation bullet now points to the spec and the resolution)
- `docs/README.md` (index entry)
- `CHANGELOG.md` (terse bullet)

No `.rs` touched; no em/en dashes; all internal links resolve.

Closes #40


ELares:
## Independent review: SHIP

An independent reviewer verified every OS primitive claim against the real seam and the frozen docs: the four-primitive shim maps onto the actual `Filesystem`/`RandomAccessFile` traits (the three shipped primitives described accurately); the per-OS preallocate recipes are technically correct (Linux `fallocate`/`posix_fallocate`; macOS `F_PREALLOCATE`+`ftruncate` because F_PREALLOCATE reserves but does not advance EOF; Windows `SetEndOfFile` plus the privileged `SetFileValidData` security caveat), each with an honest durability note and a fallback ladder to grow-on-append. The recycling resolution is sound and consistent with ADR-0002 + AT_REST_ENCRYPTION.md (segment_id is the nonce high bits, so reuse reuses a nonce and a generation stamp cannot fix it; v1 never recycles, no generation field added to the frozen header, recycling is a v2 nonce-safety decision). ENOSPC-at-roll maps to the real `StorageError::AtCapacity` overflow path. No contradictions, honest specified-not-implemented framing, internal links resolve, 16/16 CI green, no dashes. One trivial naming NIT (non-blocking). Merging. Closes #40.

Issue #126, opened 2026-06-05
[TASK] SemVer, MSRV, and the format-version registry policy


ELares:
Parent: #22

Define the three independent compatibility surfaces that protect IronBus downstreams: the Rust API (Cargo SemVer), the Minimum Supported Rust Version, and the two on-disk/wire format-version integers. The central insight is that a storage system's durable bytes must not be forced to re-encode just because the API or the wire codec changed, so the on-disk format (#4, #5) and the wire protocol (#11) are versioned separately from the API and from each other.

## Approach
Follow Cargo SemVer for the Rust API (breaking changes only on a major bump; in 0.x the API may break on any minor). Pin MSRV to Rust 1.78 via `rust-version` in the workspace `Cargo.toml`, tested by a dedicated CI job; bump MSRV only in a minor release, keep the floor at least 6 months old, and add a CHANGELOG note per bump. Introduce two unsigned integer fields, `storage_format_version` (stamped in segment/WAL headers) and `wire_protocol_version` (negotiated in the connection handshake), each tracked in a single `docs/compat/versions.md` registry. Specify the unknown-version behavior: refuse-and-report for storage (never best-effort parse, ties to #7) and negotiate-down-or-reject for wire. State the 0.x promise: API may churn, but both format versions follow their own forward/backward-compat rules from day one.

## Acceptance criteria
- [ ] SemVer policy documented, including the separate format and protocol surfaces and the 0.x promise.
- [ ] `rust-version = "1.78"` set workspace-wide and verified by an MSRV CI job.
- [ ] MSRV bump rule (minor-only, 6-month-old floor, changelog note) is written down.
- [ ] `docs/compat/versions.md` registry exists with a row per format/protocol version.
- [ ] A CI check asserts that header-format diffs also touch the registry.
- [ ] Unknown-version behavior (refuse-and-report vs negotiate-down) is specified.

## Failure considerations
Two PRs could claim the same format version or a PR could change the encoding without bumping at all, mis-parsing durable data; the mandatory registry row turns a collision into a git conflict and the header-diff CI check catches missing bumps, while refuse-and-report prevents an old binary from best-effort-misreading newer data. An MSRV bump sneaking into a patch would break lagging edge toolchains; the minor-only rule plus a patch-series assertion keeps the floor stable for cross-builds (#20).

PR #331, opened 2026-06-08
Frozen v1 conformance corpus + cross-platform byte-identity gate (#45)

### LINKED ISSUE #45: [TASK] Conformance corpus and cross-platform byte-identity test vectors for the v1 format
Parent: #5

Build a frozen, checked-in conformance corpus of byte-exact record and segment fixtures plus a generator, so the v1 format has executable test vectors and cross-platform byte-identity is asserted, not just claimed. Any encoder or decoder (now or future) is validated against this corpus.

## Approach

Provide a deterministic generator that emits a fixed set of fixtures with their expected bytes: a minimal record, a record with key only, a record with key and headers, a compressed record, a maximum-size record at `max_record_bytes`, a multi-record segment with a clean footer, an active segment with no footer, a torn-tail segment (truncated mid-body and truncated mid-trailer), a single-bit-flip mid-log corruption, a zero-window (preallocated) tail, and a newer-version record that v1 must reject. For each, store the exact bytes plus the expected decoder verdict (intact, torn-truncate, skip+report with the loss tuple, fail-closed). Add a cross-platform test: encode the same logical records on x86_64 and aarch64 (CI matrix), assert the output bytes are identical and that each platform's decoder verifies the other's `header_crc` and `body_crc`. Include round-trip property tests (random key/header/payload sizes within caps) asserting decode(encode(x)) == x and that every emitted record passes the frozen read sequence. Wire the corpus into #21 as a regression gate so a format change that is not a deliberate version bump fails CI.

## Acceptance criteria

- [x] A checked-in corpus covers intact, torn-tail (two variants), mid-log corruption, zero-window, max-size, compressed, and reject-newer cases, each with expected bytes and expected verdict.
- [x] A cross-platform CI job asserts byte-identical encoding and cross-verification of CRCs between x86_64 and aarch64.
- [x] Round-trip property tests cover random sizes within the key/header/record caps.
- [x] The corpus is wired as a regression gate that fails on any unintended format change.
- [x] The worked dump from the versioning task is reproduced and verified against a corpus fixture.

## Failure considerations

A silent format drift (someone changes a field width) would break every deployed reader: caught by the byte-exact corpus gate. An endianness or alignment bug that only manifests on aarch64 is caught by the cross-platform matrix rather than in production on an edge box. Non-deterministic fixtures would make the gate flaky, so the generator must be fully deterministic (fixed seeds, fixed timestamps).

ELares:
## What

Builds an executable freeze of the v1 on-disk format (the record frame and the segment header/footer): a deterministic generator that emits byte-exact fixtures with their expected decoder verdicts, a checked-in corpus, a regression gate that fails CI on any byte drift, and a cross-platform byte-identity CI job. The frozen format and the loss taxonomy are unchanged: this PR PINS the format, it does not alter it.

## The corpus and verdicts

The generator (`crates/ironbus-core/tests/conformance/mod.rs`) is fully deterministic (fixed seeds, sequences, timestamps; no `now()`/randomness; encodes only through the IO-free `ironbus_core::{codec, segment}` paths). It emits, each with its expected verdict:

| Fixture | Verdict |
|---------|---------|
| `record_minimal`, `record_key_only`, `record_key_headers_payload`, `record_compressed`, `record_compressed_over_threshold` | intact record |
| `record_max_size` (total exactly `DEFAULT_MAX_RECORD_BYTES`, 16 MiB, carries the xxh3 field) | intact record |
| `segment_sealed_with_footer` | intact segment, clean footer |
| `segment_active_no_footer` | intact segment, no footer |
| `segment_torn_tail_mid_body`, `segment_torn_tail_mid_trailer` | torn-truncate; recovery `TornTail` over the partial frame |
| `segment_mid_log_bit_flip` (a body byte of the 2nd record flipped) | skip+report `CorruptRecordBody`, 1st record survives |
| `segment_zero_window_tail` (preallocated/zero-filled region) | skip+report `CorruptRecordHeader`, 0 records |
| `record_newer_version_reject` (version byte = 2, CRC recomputed) | fail-closed `UnsupportedVersion(2)` |
| `segment_header` | intact segment header |

Each torn/corrupt fixture also stores the `(reason, byte_offset_start, byte_offset_end)` loss tuple recovery emits.

## How byte-identity is gated

The fixtures are checked in under `crates/ironbus-core/tests/corpus/` (small ones as raw `.bin`, compared byte-for-byte; the 16 MiB `record_max_size` as a byte-exact `.digest` of `len + frame_crc + header_crc + body_crc`, to keep git lean while staying a byte-exact gate). The gate test `generator_reproduces_the_checked_in_corpus_bytes_exactly` regenerates every fixture and asserts it equals the committed bytes, so a format change that is not a deliberate `FORMAT_VERSION` bump fails CI as byte drift. `IRONBUS_REGENERATE_CORPUS=1` rewrites the files for a deliberate change (never set in CI).

A new CI job `conformance corpus byte-identity (x86_64-unknown-linux-gnu, aarch64-unknown-linux-gnu)` runs the gate on both production ISAs (aarch64 under qemu via `cross`), and the existing 32-bit `i686` parser job runs `ironbus-core`'s tests too, so the same frozen bytes are regenerated and the same `header_crc`/`body_crc` are cross-verified on x86_64, aarch64, and i686. An endianness or alignment bug that only shows on aarch64 fails CI rather than on a deployed edge box. The in-test `byte_identity_and_cross_crc_verification` re-encodes each record and re-verifies its CRCs explicitly.

## How the regression gate is wired

- `cargo test -p ironbus-core --test conformance_corpus` is a normal test, so it runs in the existing `test (ubuntu/macos/windows)` matrix and the i686 parser job, and fails on any byte/digest drift.
- The storage cross-check `crates/ironbus-storage/tests/conformance_recovery.rs` loads the SAME checked-in segment fixtures and drives them through the real `Log::open` recovery, asserting recovery emits exactly the loss reason and span each fixture declares (so the corpus is a recovery-level gate, not only a decode-level one).
- Round-trip proptests assert `decode(encode(x)) == x` over the full key/header/payload size space within the caps (keys/headers up to a few KiB, payloads from 0 to past the xxh3 threshold) and that every emitted record passes the frozen read sequence.

## Worked format dump (#126)

`reproduces_the_worked_format_dump_from_the_versioning_task` reconstructs the documented `docs/CONTRACTS.md` / `docs/COMPATIBILITY.md` record-frame dump field by field from the documented offsets and verifies it against the `record_minimal` fixture, byte for byte. There is no separate committed hex-dump file, so the dump is reproduced from the doc tables; this is noted in the test.

## Test names

`corpus_is_complete_and_named_uniquely`, `corpus_reason_codes_match_the_frozen_storage_codes`, `generator_reproduces_the_checked_in_corpus_bytes_exactly`, `every_fixture_decodes_to_its_expected_verdict`, `reproduces_the_worked_format_dump_from_the_versioning_task`, `byte_identity_and_cross_crc_verification`, `round_trip_over_the_full_size_space`, `round_trip_straddling_the_xxh3_threshold` (core); `active_no_footer_recovers_all_three_records_cleanly`, `sealed_with_footer_recovers_all_three_records_cleanly`, `torn_tail_mid_body_matches_the_corpus_loss_tuple`, `torn_tail_mid_trailer_matches_the_corpus_loss_tuple`, `mid_log_bit_flip_skips_and_reports_corrupt_body`, `zero_window_tail_recovers_no_records_and_reports_a_corrupt_header` (storage).

## Checks

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, `cargo test -p ironbus-core -p ironbus-storage`, the `io-free-check` AST walk over `ironbus-core/src`, and `actionlint` on the changed workflow all pass locally. No em/en dashes. `ironbus-core` stays IO-free (the generator and fixtures live in `tests/`, not `src/`).

Closes #45


ELares:
## Independent review: SHIP

An independent reviewer confirmed this is a clean, additive freeze of the v1 on-disk format (zero src/ and zero Cargo.toml/lock changes; it PINS, not changes, the format). Verified: the generator is FULLY deterministic (no now()/random/HashMap; a fresh `IRONBUS_REGENERATE_CORPUS=1` regen is byte-identical to the committed fixtures, and the gate fails on any drift); all 13 fixtures match the REAL decoder/`Log::open` verdicts and loss tuples (reason codes pinned to the frozen `ReasonCode` source); the byte-identity gate genuinely runs x86_64 + aarch64 (qemu/cross) + i686, each re-deriving and cross-verifying header_crc/body_crc, so an endianness/alignment bug fails CI not production; the .gitattributes binary-pins the corpus so git never rewrites bytes (the Windows autocrlf trap is fixed) and the trailing-newline tolerance does not weaken the byte-exact format bytes; the round-trip proptests cover the full size space within caps; proptest is dev-only so the crate stays IO-free. 18/18 CI green. Two trivial NITs (CHANGELOG verbose, i686 implicit in the job name). Merging. Closes #45.

Issue #59, opened 2026-06-05
[TASK] Define SkipEvent schema, consumer gap markers, and crash/fuzz test harness


ELares:
Parent: #8

Define the canonical skip-event and metric schema shared with #16, the sparse-offset gap marker delivered to #9, and the crash-injection and fuzz harness that proves the recovery and resync paths under #21. The draft asks for a schema and worked examples but does not pin the struct, the consumer-visible gap contract, or how the paths are tested.

## Approach

Define one SkipEvent struct emitted on every skip: {ts, segment_id, reason_code (RecordCrcMismatch, SegmentHeaderBad, InvariantViolation, TornTailTruncated, ScrubberSuspect), lost_offset_start, lost_offset_end, lost_bytes, resync_offset (nullable), recovered (bool), mode}. Metrics: counters ironbus_corruption_skips_total{reason} and ironbus_corruption_lost_bytes_total{reason} (TornTailTruncated excluded from lost_bytes), gauge ironbus_quarantine_bytes, gauge ironbus_recovery_last_lost_bytes. Logical offsets are SPARSE and stable: a skipped range leaves a permanent hole, the next valid record keeps its original offset, and consumers receive an explicit gap marker (offset range, byte count) on the stream, coordinated with #9. Build a harness under #21 that injects torn tails, mid-log flips, bad segment headers, and planted false magic, asserting the exact skipped range, reported bytes, and resync offset for each.

## Acceptance criteria

- [x] SkipEvent struct defined and agreed with #16.
- [x] Metric set defined; TornTailTruncated excluded from lost_bytes counters.
- [x] Sparse stable offsets + the offset/resync CONTRACT specified (recovery-side, in loss-report.v1). The consumer-visible over-the-wire gap-marker FRAME is the remaining impl residual, tracked in #346.
- [x] Crash-injection cases: torn tail, mid-log flip, bad header, false magic.
- [x] Each case asserts exact skipped range, reported bytes, and resync offset.
- [x] Worked examples documented for the three headline scenarios.

## Failure considerations

Two divergent schemas (here versus #16) would make dashboards and events disagree; a single shared struct prevents it. Counting torn tails as data loss would inflate fleet loss metrics on every brownout; the explicit exclusion fixes it. Silently renumbering offsets to hide a gap would break the reported-loss promise at the consumer boundary; sparse offsets plus gap markers keep loss visible. A weak test harness could let a resync regression ship; the assert-exact-range cases catch it.

ELares:
[...]

PR #332, opened 2026-06-08
Specify optional key-based log compaction, reconciled with the no-manifest model (#83)

### LINKED ISSUE #83: [TASK] Optional key-based log compaction with atomic MANIFEST swap

ELares:
## What

DESIGN spec (specified, NOT implemented) for IronBus's optional, opt-in, Kafka-style KEY-BASED log compaction (#83, parent #13). Adds `docs/COMPACTION.md`, cross-linked from `docs/WAL.md` and `docs/README.md`, with a terse `CHANGELOG.md` bullet. Docs only, no source changed.

## The reconciliation the issue needs: there is NO manifest

The #83 text (and #135's sketch) assume an "atomic MANIFEST swap." IronBus deliberately has **no manifest**: recovery discovers segments by listing the directory and parsing self-describing file names (`naming.rs`: "the directory of self-describing files is the authority, no manifest required"), and the recovery chain (`scan_recover_chain` in `log.rs`) stitches segments by `base_offset`. The spec **adapts** the compaction design to that model:

- **A compacted segment self-describes its coverage in its header.** It carries a COMPACTED marker (one bit of the `flags` u16 at header `[10,12)`) and the COVERED offset range it stands in for: `covered_base_offset` is the existing `base_offset`, and `covered_end_offset` + the highest covered source id are two `u64`s that fit EXACTLY in the reserved 16 header bytes `[44,60)` (CRC-covered, no field grows or moves). This is the information a manifest entry would have held ("offsets [a,b) now live in file X"), placed in the file that owns the range.
- **Honest format consequence: it forces a format-version bump.** A v1 reader treats `flags` as "preserved but not interpreted" and the reserved bytes as zero, so it would misread a compacted segment. A compactor therefore stamps `version = 2` on compacted segments, and a v1 reader refuse-on-unknown-version **fail-closed refuses** the log rather than misreading it. Owned by `COMPATIBILITY.md`/`CONTRACTS.md` at implementation.
- **The single atomic commit point is the durable appearance of the new segment** (the directory fsync after writing + fsyncing it), NOT a manifest entry swap. Before it, the originals are authoritative; after it, the compacted segment is. The originals are then **rename-then-unlinked** (the same discipline the reaper already uses) so an open reader drains its bytes rather than reading freed data.

## Recovery: longest-valid-prefix, NO compaction-specific repair

- **Crash before the commit:** the orphan/torn new compacted segment is redundant (its covered range is fully present as originals), so recovery discards it; the originals win. No special repair.
- **Crash after the commit, mid-retire:** the compacted segment is durable and authoritative; any original it fully covers is unreferenced and recovery unlinks it. No special repair.
- **Overlap resolution is deterministic from the headers alone:** a compacted segment supersedes every original whose offset range it fully covers; in the (crash-only) case of two overlapping compacted segments, the higher segment id wins (ADR 0002 monotonicity is the total tie-break). Recovery sorts by COVERED/ACTUAL offset range, not id, because a compacted segment has a high id but a low covered range. There is no manifest to tear, so there is no torn-manifest case; neither crash window is a reported loss (no durable record is actually lost).

## Full acceptance design covered

At-least-the-last-value-per-key; tombstone retention for `tombstone_ttl` (default 24h, engine-clock-seam, so offline consumers see deletes); never reorder/never rewrite offsets (survivors keep ORIGINAL offsets and sequences, producing sparse ranges, preserving I5/I7); never compact the active segment; keyless records always carried through; sparse-offset reader/index tolerance with a NEW **non-loss `Compacted` gap marker** (honestly flagged: the current poll/`read_from`/offline paths assume dense offsets; coordinated with #9 gap semantics and the #59 SkipEvent, excluded from loss-bytes like `TornTailTruncated`); the cleaner rate-limited + yields to the append actor + OFF by default; and `compact_and_delete` running the age/size/count reaper FIRST so we never compact a segment we are about to delete. The fresh never-recycled id (ADR 0002) keeps the at-rest AEAD nonce unique for rewritten survivors.

## Honesty

Specified, not implemented (no compaction code exists). Opt-in and edge-hostile (CPU + flash cost), for changelog/state-snapshot topics, OFF by default. Verification targets (crash-before/after fault injection, reader-drain, sparse-read, append-priority) are listed in the doc.

## Constraints

Docs only, no `.rs` touched. No em/en dashes. Internal links resolve.

Closes #83


ELares:
## Independent adversarial review: REQUEST-CHANGES (two design-coherence defects)

The no-manifest atomic-swap + recovery design is coherent and crash-safe (single commit = directory fsync of the new segment; rename-then-unlink prevents reading freed bytes; crash-before discards the orphan, crash-after unlinks superseded originals; both resolve from headers with no special repair; the version=2 fail-closed bump is consistent with COMPATIBILITY.md), and every acceptance criterion is present as design. But two SHOULD-FIX defects mean the spec is not yet mutually consistent with the shipped specs nor internally consistent, which is the exact M2 bar.

**SHOULD-FIX 1: header-byte double-claim with #108 (at-rest encryption).** `docs/COMPACTION.md:120-123` claims the covered range needs `covered_end_offset` (u64) + highest-covered source id (u64) = the WHOLE reserved `[44,60)` region. But `docs/AT_REST_ENCRYPTION.md:140` already assigns `[44,60)`: `aead_suite` u8 + a `key_id`, "the rest stay reserved-zero". A segment that is BOTH compacted AND encrypted (a v2 encrypted compacted segment, which the spec itself contemplates at `COMPACTION.md:373-377` since survivors are re-encrypted under the new segment id) cannot fit both in 16 bytes. The flags bits do not conflict (`COMPACTED` vs `SEGMENT_ENCRYPTED` are different bits), but the reserved bytes do. COMPACTION.md pins only the nonce interaction and is silent on the header-byte collision. COMPATIBILITY.md treats `[44,60)` as a single shared future-version space with no registered partition. Reconcile this: state a clear, implementable rule (a per-segment layout that fits both, a dedicated compaction metadata structure outside the cramped reserved bytes, or an explicitly larger v2 header), so the two design docs are mutually consistent and implementable together.

**SHOULD-FIX 2: `covered_base_offset` defined inconsistently.** `COMPACTION.md:110-111` says the covered range is the offset span of the ORIGINAL source segments, but `:121-122`/`:180` equate `covered_base_offset` with `base_offset` (the lowest survivor's offset). These differ when the first source segment's leading records are all superseded: the lowest survivor's offset is then strictly greater than the source set's true start, leaving a gap `[predecessor_end, lowest_survivor_offset)` that the spec's own chain-continuity check (`:250,256-257,276`) would reject. `covered_base_offset` must be the source set's TRUE starting offset, which needs its own field, not an alias of `base_offset`. That adds a third covered-range field (covered_start + covered_end + tie-break), compounding the byte-budget problem in SHOULD-FIX 1 (3 u64s = 24 bytes do not fit the 16-byte reserved region at all), so the two fixes must be resolved together (the covered-range metadata likely cannot live in the reserved header bytes).

NITs (fold in): (3) the real `scan_recover_chain` enforces BOTH `base_offset` AND `base_seq` continuity (`log.rs:395,409,412`); since survivors keep sparse sequences, the `base_seq` continuity must also change, but the spec discusses only the offset half. (4) the loss-bytes analogy at `:316-319` is imprecise: `TornTail` IS reported with non-zero `bytes_skipped` (it is not excluded); the point that a `Compacted` gap is genuinely not loss is sound, but the analogy and the `TornTailTruncated` name are loose. (5) the CHANGELOG bullet (~259 words) is verbose vs the project's terse-bullet preference.

Sending back to reconcile the header layout with #108 and fix the covered-range definition (together, given the shared byte-budget constraint), plus the three nits. `Closes #83` remains the right framing for this M2 design deliverable once it is coherent.

ELares:
## Re-review of the fix: SHIP

The two SHOULD-FIX design-coherence defects are genuinely resolved in the fix (commit ce4231f, rebased to 6614986), and the two docs are now mutually consistent: (1) the covered-range metadata moved OUT of the contested header reserved bytes `[44,60)` into a dedicated 44-byte CRC-protected v2 footer block (5 u64s + block_crc CRC32C over the 40 data bytes), written and fsynced contiguously with the seal so a torn block is indistinguishable from a crash-before-commit (existing longest-valid-prefix recovery, no new repair); `[44,60)` stays owned by at-rest encryption alone (a bidirectional cross-reference was added to AT_REST_ENCRYPTION.md), and the `COMPACTED` flag bit is distinct from `SEGMENT_ENCRYPTED` so a compacted-AND-encrypted segment has room for both. (2) `covered_base_offset` is now its own field = the source set's true start (the predecessor end / lowest covered SOURCE offset), with `covered_base_seq`/`covered_end_seq` handling the parallel base_seq continuity that the real `scan_recover_chain` enforces. The nits are folded (the TornTail loss analogy corrected, CHANGELOG trimmed). Merging. Closes #83.

PR #338, opened 2026-06-08
docs: specify the durability contract and the levels model (#50)

### LINKED ISSUE #50: [TASK] Crash-injection and ack-ordering test harness for the durability contract
Parent: #6

Prove the durability contract holds under power loss and that the producer ack is never sent before its level's condition is met. The contract is only as good as the test that injects failure at the worst moment, so this harness is the gate that lets the levels-spec guarantees be trusted.

## Approach
Build a fault-injection layer that can (a) drop all unsynced bytes at an arbitrary instruction boundary to emulate power loss, and (b) make `fdatasync`/`fsync` return EIO on demand, including the consumed-error case. Drive each level through crash-at-every-step and assert the written worst-case-loss guarantee from the levels-spec sub-issue holds (zero acknowledged loss for `sync`, bounded-and-reported for `batch`/`interval`). Add a property test on the ack path asserting an ack is provably never emitted before the level's condition (the WAL `fdatasync`) returns. Reproduce the fsyncgate scenario and assert the writer freezes and never returns success on a retried failed sync. Reproduce crash-after-segment-create and assert the segment survives via the parent-directory fsync. Run the throughput vs latency vs bytes-at-risk curve on real ARM SD/eMMC hardware, feeding the numbers to #19 and reflecting #20.

## Acceptance criteria
- [ ] Crash injection drops unsynced bytes at arbitrary points; each level's loss guarantee asserted under it.
- [ ] Property test proves ack never precedes the level condition.
- [ ] fsyncgate fault-injection proves no false-success on retried failed sync and a writer freeze.
- [ ] Crash-after-create test proves segments never vanish.
- [ ] Group commit shows syncs-per-record < 1 under concurrent load.
- [ ] Throughput/latency/bytes-at-risk curve across all four levels produced on real ARM SD/eMMC for #19.

## Failure considerations
A harness that only emulates clean truncation would miss torn mid-batch writes; it must also corrupt the tail to exercise the #5 checksum and #7 stop-at-first-bad-frame path. Emulated power loss can diverge from real flash behaviour (FTL reordering), so the real-hardware run on ARM SD/eMMC is required, not optional, to keep the edge guarantees honest.

ELares:
## What

Adds `docs/DURABILITY.md`, the authority for IronBus's DURABILITY CONTRACT and the levels model, plus cross-links from `docs/README.md` and `docs/WAL.md` and a terse `CHANGELOG.md` bullet. This is the M2 DESIGN deliverable for the #50 crash-injection / ack-ordering harness task (under the durability parent #6). Docs only, no source changed.

## The durability-model decision (stated honestly)

#50 assumes a menu of four durability levels (sync / batch / interval / async), each with a per-level loss guarantee. IronBus's FROZEN model is `fdatasync`-BEFORE-ack, so the doc states the honest truth:

- **v1 ships exactly ONE durability level, the safe `sync` level** = ack-implies-durable (invariant I2). A `Pub`/`PubAck` is emitted only after the covering `fdatasync` returns `Ok`; on power loss the unsynced tail is truncated and was never acked, so ACKNOWLEDGED loss is ZERO.
- The #177 append-actor **group commit AMORTIZES** the `fdatasync` (one sync per drained batch) but still acks each record only AFTER that covering sync. So the #50 "batch" level is the SAME safe level observed with the batcher on, not a relaxation, and it collapses into the `sync` row.
- The relaxed `interval` and `async` (the README's `none`) levels are **SPECIFIED, not implemented**: `interval`'s loss is BOUNDED by the flush window (time or bytes since the last completed sync), `async`'s is UNBOUNDED until the next sync. They trade durability for throughput contrary to the Edge-First safe default, so they are off by default, gated behind an explicit data-loss acknowledgement (and a #14 `durability_level` config knob), and cannot be reached from the current `serve` command line at all. The default is and remains the durable level.

## Mapping #50 acceptance to the EXISTING falsification

Every criterion is mapped to a real, named test (all verified present in the tree):

| #50 criterion | Status | Test(s) |
| --- | --- | --- |
| ack never precedes the level condition | PROVEN (`sync`) | `an_ack_is_sent_only_after_the_covering_fsync_completes_i2` (`actor.rs`) |
| crash drops unsynced bytes at arbitrary points; loss guarantee holds | PROVEN (`sync`) | `power_loss_at_every_boundary_no_roll` / `_with_rolling`, `power_cut_with_a_holed_unsynced_tail_loses_no_acked_record`, `power_loss_recovers_the_durable_prefix` (`crash_recovery.rs`) |
| fsyncgate: no false-success on a retried failed sync + writer freeze | PROVEN | `fatal_fsync_freeze_loses_no_acked_record`, `fatal_fsync_freeze_during_a_roll_loses_no_acked_record`, `a_sync_fault_at_any_point_loses_no_acked_record` (`crash_recovery.rs`); `Log::sync` -> `WriterFrozen` (`log.rs`) |
| crash-after-create: segments never vanish | PROVEN | parent-dir fsync in `Log::start_segment` + `recovery_fails_closed_when_a_fault_strikes_the_roll_forward` |
| group commit: syncs-per-record < 1 | PROVEN | `a_batch_of_concurrent_produces_issues_one_fdatasync_not_n` (`actor.rs`) |
| torn-mid-batch + checksum + stop-at-first-bad-frame | PROVEN | the corruption tests in `crash_recovery.rs`, the #45 corpus driven through real recovery (`conformance_recovery.rs`), and `corruption_corpus.rs` |
| per-level loss guarantee asserted (zero/bounded/declared) | PARTIAL | `sync` row proven; `interval`/`async` rows CANNOT be asserted until those levels exist |
| across-levels throughput/latency/bytes-at-risk curve on real ARM SD/eMMC for #19 | RESIDUAL (device) | not host-side reproducible (see below) |

## The residual

Two acceptance items genuinely cannot be satisfied by a design doc or in CI:

1. The **per-level loss assertions for `interval` / `async`** need those levels IMPLEMENTED first (durability #6).
2. The **across-levels throughput/latency/bytes-at-risk curve on real ARM SD/eMMC** is a DEVICE measurement: it needs both the relaxed levels and the reference hardware, must not be faked (emulated power loss cannot reproduce FTL reordering), and feeds #19 via the #111 macro-bench under the SLO ratification process. SLO.md already tables the durability modes side by side with their targets marked not-yet-ratified.

## Closes vs refs

I used **`refs #50`**, not `Closes #50`. The DESIGN deliverable is complete (the contract is specified, the existing falsification is mapped to real tests, the relaxed levels are designed with their loss bounds and opt-in contract, and the device curve is honestly scoped as a residual). But the issue's acceptance list genuinely includes items that require the relaxed levels IMPLEMENTED and the on-ARM-hardware curve RUN, neither of which is host-side or CI-achievable. So the orchestrator can decide whether to close the design and file an implementation (#6) and device (#19) follow-up, rather than this PR claiming the whole harness/levels task is done.

## Constraints honored

- No em/en dashes anywhere (verified).
- Docs only: `git status` shows only `docs/DURABILITY.md`, `docs/README.md`, `docs/WAL.md`, `CHANGELOG.md`. No `.rs` touched, so no SPDX or fmt/clippy/test impact. Did not touch `crash_recovery.rs` / `src/fault.rs` (parallel PR #55) or `ironbus-bench` (parallel PR #284).
- All internal links resolve; every cited test name verified present in the source.

refs #50


ELares:
## Independent review: SHIP (citation nit fixed)

An independent reviewer verified the durability design is complete, honest, and consistent with the frozen fdatasync-before-ack model: v1 ships exactly ONE level (sync = ack-implies-durable, I2; the #177 group-commit amortizes the fsync but still acks only after it), the relaxed interval/async levels are honestly specified-not-implemented (off-by-default, behind an explicit data-loss ack + #14 knob), and every one of the 16 cited tests was grep-confirmed present and verified to prove its property. The one SHOULD-FIX (a mismatched secondary test citation) is fixed in 266e45c: the fsync-histogram sentence now accurately says `produce_records_one_fsync_observation_each` proves one observation per `commit_batch`, with the N-produces-one-sync amortization attributed to the actor test cited just above. The per-level loss assertions (need the relaxed levels) and the on-ARM bytes-at-risk curve (device) are honestly scoped residuals. Merging.

PR #339, opened 2026-06-08
Strengthen the crash harness at every durability write boundary (#55)

### LINKED ISSUE #55: [TASK] Crash-injection and power-cut harness asserting a consistent recovered prefix at every write boundary

ELares:
Closes the remaining #55 gaps so recovery is mechanically falsified at every durability-relevant write boundary, over the deterministic in-memory disk model. The frozen format and the recovery contract are UNCHANGED; this is test/seam-only work (a new test-only `FaultControl` knob, a new dev/test disk model, and new crash-recovery gates). If a new injection had revealed a real recovery bug it would have been reported, not papered over; none did.

## What the harness already covered
The suite (`crates/ironbus-storage/tests/crash_recovery.rs`) already exercised power loss at every op boundary (roll and no-roll), torn tails, tail/mid-log/arbitrary-byte corruption, the fatal-fsync EIO freeze (live and inside a roll), idempotent recovery during recovery, the holed-unsynced-tail class, a sequence-gap reject, and the seeded fault sweeps. This PR adds the three gaps #55 named.

### 1. Rename-boundary injection (the `sync_dir` / directory-publish boundary)
IronBus uses NO `rename`. Its atomic-swap publish point is the `sync_dir` that makes a sealed-or-rolled segment's directory entry durable (segments are created via `create_new` + header `sync_all` + `sync_dir`; the checkpoint is a dual-slot file, no rename). So the rename-boundary analogue is a `sync_dir` crash:
- `FaultControl::set_fail_sync_dir(bool)` + a `sync_dir_count()` (so a test can prove the boundary was actually reached and a stop is never a false pass).
- `FaultFs::sync_dir` counts the publish and, while armed, fails it WITHOUT publishing.
- Tests stop the process immediately BEFORE the publish (the roll's `sync_dir` faults, freezing the writer; the unpublished new segment vanishes on the crash, recovery rolls forward from the sealed predecessor) and immediately AFTER it (a clean roll then a power loss), asserting recovery yields the same consistent acked prefix at both points, plus the fresh-segment create-publish boundary.

### 2. Page-cache reorder/drop power-loss model (#164)
#164 (the block-layer fault umbrella) is closed, having shipped the fsync-EIO/freeze, both-slots-torn, and crash-during-recovery parts; the page-cache reorder/drop model had NOT shipped (the `fault.rs` header still listed it as a follow-up). It lands here to avoid duplication. `InMemoryFile`/`InMemoryFs::simulate_power_loss_reorder(seed)` model the real power-loss guarantee that ONLY fsync'd bytes survive: the unsynced tail persists as a seeded STRICT prefix (the rest dropped), deterministic in the seed (a xorshift step, no ambient randomness). A sweep over 64 seeds asserts recovery still yields a consistent monotonic prefix that loses no acked record, with on-disk byte-state assertions so an injection that did not cross the boundary is caught, plus a two-independent-disk determinism gate.

### 3. `loss_bound == actual discarded suffix`
A shared assertion (`assert_loss_bound_equals_discarded_suffix`) computes the ACTUAL discarded suffix as the active segment's durable byte length at recovery minus the bytes recovery kept (its post-truncation file length, i.e. `valid_end`), and asserts the structured `LossReport`'s claimed byte loss (`total_bytes_skipped`) AND the raw `recovered_truncated_bytes` both equal it, after every injected crash (torn tail swept across chop sizes, body corruption, and the reorder cut). A report that under- or over-claims its loss is falsified; each claimed loss event is also checked to name the right segment with a span matching its byte count.

## Cross-platform (#17)
All new gates use only `InMemoryFs`/`FaultFs`/`InMemoryFile` (no `#[cfg(unix)]`), so they run under `cargo test --workspace` on the ubuntu/macOS/windows matrix, and the storage suite also runs under musl and i686 (the parser/conformance lanes). No boundary test is unix-only, so the prefix/monotonicity/CRC/loss-bound invariants are proven on every platform.

## Verification
- `cargo test -p ironbus-storage`: all pass (213 lib + 27 crash_recovery + the determinism/invariant/conformance gates)
- The new tests are deterministic across repeated runs; no em/en dashes; MSRV 1.78 preserved.

Closes #55


ELares:
## Independent review: SHIP

An independent reviewer verified this is a clean test/seam-only strengthening of the crash harness (no frozen-format/recovery-contract change), deterministic (10/10 loops on the full crash_recovery suite AND the 6 new boundary tests), clippy pedantic clean, no dashes. Criteria delivered for IronBus's actual architecture: the rename-boundary criterion is correctly reinterpreted (the Filesystem trait has no rename and the checkpoint is a dual-slot in-place CRC scheme, so the true atomic-publish point is sync_dir; the PR injects a crash before AND after it with sync_dir_count() proving the boundary was reached); the page-cache power-loss model only ever drops UNSYNCED bytes (durable prefix restored first), with the genuinely-holed-tail case also covered by a pre-existing proptest; and the loss_bound assertion cross-checks total_bytes_skipped against the real (pre-recovery durable length - post-recovery on-disk file length), non-vacuously. 18/18 green. Merging. Closes #55.

PR #340, opened 2026-06-08
Make the injected-stall coordinated-omission self-test a reliable CI gate (#284)

### LINKED ISSUE #284: [TASK] Make the macro-bench injected-stall self-test a reliable CI gate

ELares:
## What

Makes the macro-bench coordinated-omission self-test a RELIABLE, non-flaky CI gate (#284), so it can gate every PR instead of being `#[ignore]`d.

## The problem (#284)

The self-test froze the shipping `ironbus` broker with OS `SIGSTOP` mid-run and asserted the ~200 ms freeze landed in the recorded p99.9 tail. On GitHub's shared runners the kernel did not reliably deschedule the broker for the whole window, so the freeze often delayed no client op at all and vanished from the tail (a representative failing run recorded a 2.6 ms max for a 200 ms freeze). That is an OS-scheduling artifact of the shared host, not a harness-logic bug, so the test was `#[ignore]`d and off the CI critical path.

## The fix: a deterministic in-process freeze (option b)

The new gate `an_injected_stall_shows_up_in_the_recorded_tail` runs the broker **in-process**, using the SAME `ironbus-server` engine + append actor + `serve` the binary ships, over a `FaultFs<InMemoryFs>` (new `crates/ironbus-bench/src/inproc.rs`). It freezes the broker mid-run by **closing the `FaultFs` sync gate** (#177): while the gate is closed, the broker's group-commit `fdatasync` parks on a **condvar**, so EVERY produce that needs that fsync blocks for exactly the held window.

Unlike `SIGSTOP` (which only ASKS the kernel to deschedule the broker and may not actually delay any client op on a shared runner), the gate GUARANTEES the block. The freeze therefore **always** lands in the tail, with no dependence on OS scheduling. That is the whole point: the determinism comes from the gate blocking the fsync, not from any sleep timing or kernel behavior.

## The open-loop measurement stays honest

The harness is unchanged: it still drives the broker through the **real #11 client over a real loopback socket** and records each message's latency from its **intended send time** (the wrk2 anti-coordinated-omission measurement). When the broker's fsync is gated, the sender backs up and the post-thaw backlog drains carrying its **old intended times**, so the freeze lands in p99.9.

The proof is **non-vacuous**: the test calibrates the arrival rate and freeze to the broker's probed op latency, asserts the **healthy baseline p99.9 stays below a separation floor** (so clearing the floor is genuinely the injected freeze, not a saturated baseline), and **FAILS if the freeze is absent from p99.9** (a reintroduced coordinated omission would erase the freeze and leave p99.9 at the healthy floor). The single-worst-message `max` clears the same floor as an independent end-to-end check.

## Determinism result

Verified **20/20** consecutive passes locally on the deterministic gate. The legacy live `SIGSTOP` proof (`an_injected_sigstop_shows_up_in_the_recorded_tail`) is kept behind `--ignored` for a stable host and does **not** gate CI; both tests share one calibration recipe.

## No production change

No freeze hook is added to the shipped `serve` path. The seam is the existing dev-cfg `FaultFs`, wired only in the `publish = false` bench crate. The in-process broker, its deps, and the gate are all `cfg(unix)` (matching the rest of the self-test and the shipped Unix-only broker), so Windows still builds clean. Production timing and the frozen wire protocol / resilience taxonomy are unchanged.

## Verification

- `cargo test -p ironbus-bench`: the de-`#[ignore]`d gate passes; unit tests pass; SIGSTOP proof correctly ignored
- 20x loop on the deterministic gate: **20/20**
- The legacy `--ignored` SIGSTOP proof still passes on a stable host

Docs: `SLO.md` and `RISK_REGISTER.md` (RR-20 resolved, RR-13 note) updated to reflect the now-reliable per-PR gate; terse `CHANGELOG.md` bullet added.

Closes #284


ELares:
## Independent review: SHIP

An independent reviewer verified the deterministic in-process freeze and confirmed it is non-flaky: the freeze is the FaultFs sync gate (#177), which parks the single append-actor thread INSIDE `commit_batch` on a condvar, so the broker block is GUARANTEED for the held window (not OS-scheduling-dependent like SIGSTOP), and the ~200ms freeze always lands in the recorded tail (20/20 determinism). Non-vacuity empirically confirmed: a no-freeze healthy run records p99.9/max ~11ms, far below the 120ms separation floor the assertion requires, so neutering the freeze or reintroducing coordinated omission FAILS the test. The open-loop intended-send-time measurement over the real #11 client + loopback socket is unchanged. No freeze hook was added to the shipped serve path (it is type-parametricity over Filesystem; all FaultFs references in ironbus-server are #[cfg(test)]); Cargo.lock added only path-dep edges (no new external crate, cargo-deny green); the bench crate stays publish=false; inproc.rs is #[cfg(unix)] so Windows builds clean. The #[ignore] is removed from the new gate; the legacy SIGSTOP test is separately ignored and does not gate CI. 18/18 green, no dashes. Merging. Closes #284.

Issue #92, opened 2026-06-05
[TASK] peek/dump/repair/scrub: offline inspection and read-only-by-default repair


ELares:
Parent: #15

Implement the inspection and repair verbs on top of the offline reader: `peek` (seek and show a window of records), `dump` (stream all records as NDJSON), and `scrub`/`repair` (invoke #7 and #8 offline). These are where storage internals become something a human can act on after a brownout, so safety against operator error is the dominant concern.

## Approach

`peek`/`dump` decode straight from `--data-dir` via the shared offline reader, marking (not hiding) records #8 would quarantine, and labeling each with offset, timestamp, bytes, CRC status, codec, and dict id. `dump` of a compressed batch decodes and shows logical messages by default; `--raw` shows the on-disk frame; a missing #12 dictionary degrades to `decoded:false,reason:missing-dict:<id>` and continues, with `--require-dict` to fail strictly (exit 3). `dump` streams NDJSON line-by-line with bounded buffering and honors `--limit`, never collecting a result vector. `scrub`/`repair` are strictly read-only by default: they produce a plan (segments to quarantine, torn-tail truncation point, implied MANIFEST rewrite) and exit. Mutation requires `--apply` plus the exclusive data-dir lock (#4); it fails fast with exit 5 if the broker holds the lock, never edits in place (atomic MANIFEST rename), moves quarantined segments to `quarantine/` rather than deleting, and preserves original uid/gid/mode.

## Acceptance criteria

- [ ] `peek`/`dump` decode with no server running and mark records skipped by #8.
- [ ] `dump` defaults to decoded messages, `--raw` shows the frame, missing dict degrades not aborts, `--require-dict` exits 3.
- [ ] `dump` streams NDJSON within the stated RAM ceiling on a multi-GB segment and honors `--limit`.
- [ ] `scrub`/`repair` default to a read-only plan; `--apply` requires the exclusive lock and exits 5 if the broker is up.
- [ ] `--apply` never edits in place, quarantines (not deletes), and preserves uid/gid/mode.
- [ ] Reviewers from #7 and #8 confirm the plan matches what the broker would do on next start.

## Failure considerations

`--apply` racing a live writer corrupts MANIFEST: mitigated by the exclusive lock, fail-fast exit 5, atomic rename, and quarantine-not-delete. A missing dictionary aborting a whole dump loses forensic value: mitigated by per-record degrade. Concurrent `--apply` by two operators: mitigated by the single exclusive lockfile.

PR #344, opened 2026-06-08
Reconcile SkipEvent to the frozen loss-report.v1 + close its gaps (#59)

### LINKED ISSUE #59: [TASK] Define SkipEvent schema, consumer gap markers, and crash/fuzz test harness

ELares:
## Summary

Defines the canonical SkipEvent schema, the consumer gap marker, and the crash/fuzz coverage for #59 by **reconciling the #59/#137 draft to the shipped, frozen `loss-report.v1`** rather than introducing a divergent second schema. The shipped `LossEvent { segment_id, byte_offset_start, byte_offset_end, bytes_skipped, records_lost_estimate, reason_code }` IS IronBus's canonical per-skip SkipEvent: the single schema shared with the metrics endpoint (#16) and the offline inspector (#15).

The implementation and the frozen contract are canonical; the draft's `lost_offset_start`/`lost_bytes` field names and `RecordCrcMismatch`/`TornTailTruncated` reason names are mapped onto the frozen ones, not adopted.

### 1. Append-only `ScrubberSuspect` (code 6)
Appended `ScrubberSuspect` (numeric code `6`, metric label `scrubber_suspect`, JSON name `"ScrubberSuspect"`) for the future at-rest scrubber (#92). This is **append-only**: codes 1-5 are byte-identical, never reordered or renumbered, and `schema_version` stays **1** (per the frozen versioning policy, a new `ReasonCode` does NOT bump the version). The golden vocabulary test (`golden_reason_code_vocabulary_is_frozen`), `reason_codes_are_stable_and_distinct`, and the labels-frozen test now pin code 6. `loss-report.v1.md` and CONTRACTS.md get the new row.

### 2. TornTail excluded from the data-loss total
A torn tail is a **reported skip** but **not data loss** (a brownout left bytes that were never fully written, never acked data). Counting it as data loss would inflate fleet loss metrics on every clean restart.

- **Finding:** TornTail was already separable in the per-reason `ironbus_recovery_loss_bytes{reason="torn_tail"}` series and already excluded from the forensic quarantine store (`quarantine::is_corruption_skip`), but there was **no explicit data-loss total** that excluded it: `total_bytes_skipped()` (and the `ironbus_bytes_skipped` reconciled gauge) include torn tails.
- **Fix:** added `LossReport::data_loss_bytes()` and `ReasonCode::is_data_loss()` (every reason except `TornTail`, the same boundary quarantine uses), exposed as the new **`ironbus_recovery_data_loss_bytes`** gauge. `total_bytes_skipped()` / `ironbus_recovery_truncated_bytes` still include torn tails. The new metric is a **gauge** (no `_total`), so the frozen resilience-counter taxonomy is unaffected (verified: `the_resilience_counter_taxonomy_is_frozen` stays green).

### 3. SkipEvent reconciliation in the docs
`loss-report.v1.md` now documents that the shipped `LossEvent` IS the canonical per-skip SkipEvent (no divergent second schema), with a full draft->frozen mapping: `lost_offset_start`->`byte_offset_start`, `lost_bytes`->`bytes_skipped`, **the resync point IS `byte_offset_end`** (recovery resumes exactly where the dropped span ends; it never resyncs past a corrupt frame onto a later magic), and the reason-name mapping (`TornTailTruncated`->`TornTail`, `RecordCrcMismatch`->`CorruptRecordHeader`/`CorruptRecordBody`, `SegmentHeaderBad`->`CorruptSegmentHeader`, `InvariantViolation`->`SequenceGap`).

**Schema-bump decision: stay at v1.** The draft's extra fields add no information over the frozen schema and are NOT worth an incompatible v2 bump:
- `resync_offset` is fully captured by the existing `byte_offset_end` (recovery always resumes at the span end).
- `recovered` carries nothing: recovery either accepts a bounded, reported loss (the event lands) or fails closed (no event lands); there is no `recovered=false` event. The bounded-loss caps carry the accept/reject decision.
- `mode` (shutdown kind) is operational state, not per-skip-event state.

So no `schema_version` bump; only the append-only `ScrubberSuspect` reason was added.

### 4. Planted false-magic crash-injection case
Added the missing #59 case to `corruption_corpus.rs`: a **fabricated record magic planted mid-log** must be rejected at the checksum/sequence check, never resynced onto. Two variants:
- `planted_false_magic_mid_log_is_rejected_at_the_checksum`: a real `RECORD_MAGIC` planted inside a middle record's body (where a naive resync scanner could mistake it for a fresh frame) corrupts that record's body checksum; recovery stops at the frame and drops the exact span to EOF (`CorruptRecordBody`), never resyncing onto the deeper planted magic.
- `planted_false_magic_in_a_header_is_rejected_at_the_header_crc`: a valid magic + supported version with a flipped CRC-covered header byte is rejected at the header checksum (`CorruptRecordHeader`).

Both assert the **exact skipped range, reported bytes, and resync offset** via the shared `assert_single_loss` helper. Torn-tail, mid-log-flip, and bad-segment-header are already covered across `corruption_corpus`, `crash_recovery`, and `conformance_recovery` (including #55's boundary sweeps).

### 5. Worked examples
`loss-report.v1.md` gains worked torn-tail / mid-log-corruption / bad-segment-header examples, each with its exact offset range, reason, bytes, and resync point.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test -p ironbus-storage -p ironbus-server` all pass (golden loss-report, frozen vocabulary, the new false-magic cases, `data_loss_bytes_excludes_torn_tail`, the frozen resilience-counter taxonomy, and the crash gates). No em/en dashes.

Closes #59


ELares:
Independent adversarial review: **SHIP**. All five frozen-contract invariants verified with teeth (renumbering TornTail->99 fails both `reason_codes_are_stable_and_distinct` and `golden_reason_code_vocabulary_is_frozen`; a wrong expected-prefix fails the planted-magic body test):
1. Codes 1-5 byte-identical (values, JSON names, labels unchanged).
2. `ScrubberSuspect` append-only at code 6; `schema_version` stays 1 (golden serialization still pins `"schema_version": 1`).
3. Golden/freeze tests pin all 6 (code, label, JSON-name) triples and assert 1-5 by value.
4. `is_data_loss()` / `data_loss_bytes()` exclude ONLY TornTail, byte-for-byte the same predicate as `quarantine::is_corruption_skip`; `total_bytes_skipped()` still counts torn tails; the new `ironbus_recovery_data_loss_bytes` is a GAUGE (no `_total`), so the frozen resilience-counter taxonomy is untouched.
5. The planted-false-magic corpus tests have real teeth (recovery stops at the corrupt frame and never resyncs onto a deeper planted magic).
Tests: ironbus-storage loss 34/34, corruption_corpus planted_false_magic 2/2, ironbus-server 223/223, clippy clean, zero em/en dashes.

CI: 18/18 green on head `2a0ecb2` (fail=0 pending=0).

Merging with `refs #59` (not Closes): #344 lands the canonical SkipEvent schema (the shipped frozen `loss-report.v1` LossEvent), the metric set with TornTail excluded from the data-loss total, all four crash-injection cases with exact-range asserts, and worked examples. The one remaining #59 criterion is the consumer-visible sparse-offset GAP-MARKER frame delivered on the DELIVER stream (coordinated with #9); no `GapMarker` FrameType exists yet. Tracking that as a follow-up.

PR #345, opened 2026-06-08
Make the injected-stall gate robust to a noisy CI baseline (#284)

ELares:
## The flake

The coordinated-omission gate `an_injected_stall_shows_up_in_the_recorded_tail` (`crates/ironbus-bench/tests/injected_stall.rs`) flaked on the shared `test (macos-latest)` runner. The failing assertion was the non-vacuity guard in `assert_stall_is_measured`:

```rust
assert!(healthy.percentiles.p999_us < cal.floor_us, ...);
```

The freeze itself is already deterministic (#284 closed the OS-scheduling flakiness by parking the in-process broker on the `FaultFs` sync gate). The remaining problem is the SEPARATION FLOOR: `cal.floor_us` was a FIXED fraction of the freeze (`~0.6 * stall`, i.e. ~120 ms for a ~200 ms stall), an ABSOLUTE threshold. On a loaded macOS CI runner the in-process broker contends with the harness for CPU, so the HEALTHY baseline's own p99.9 spikes ABOVE that fixed floor, and the guard fails even though nothing is wrong with the freeze. This is a noisy-baseline-vs-fixed-floor problem, not a coordinated-omission regression.

## The fix: an ADAPTIVE separation tied to the measured baseline

Replace the fixed floor with a separation RELATIVE to the MEASURED healthy baseline. The stalled tail must now exceed the baseline by a freeze-tied margin:

- `stalled.p999_us >= healthy.p999_us + SEPARATION`
- `stalled.max_us >= healthy.p999_us + SEPARATION`

where `SEPARATION = 0.5 * cal.stall_us`. Both sides scale with the runner, so a loaded host that lifts the baseline lifts the stalled tail by the same amount and the relative check still clears. It cannot be defeated by a slow/noisy runner, and it still FAILS on a coordinated-omission regression (which would leave `stalled.p999 ~= healthy.p999`, far below `baseline + 0.5*stall`).

The flaky `healthy.p999 < fixed_floor` assertion is replaced with a GENEROUS sanity backstop (`healthy.p999 < 0.8 * cal.stall_us`) that only catches a pathologically slow host, not a merely-loaded one. The minimum freeze is raised from 200 ms to 300 ms for extra absolute headroom. The shared `assert_stall_is_measured` helper is used by BOTH the deterministic in-process gate and the legacy `--ignored` SIGSTOP path, so the hardening applies to both.

## Verification (run under load to reproduce the flaky condition)

- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: clean.
- Gate test run 20x UNLOADED: 20/20 pass.
- Gate test run 8x under HEAVY CPU LOAD (20 `yes > /dev/null` generators on a 10-core host, so the baseline gets noisy): 8/8 pass. The relative `baseline + 0.5*stall` threshold holds across baselines that varied with the load.
- Non-vacuity confirmed: temporarily injecting a ZERO freeze makes the stalled p99.9 sit at the baseline (`11215 us` vs baseline `11002 us`, required `>= 161002 us`) and the test FAILS with the coordinated-omission message. So the gate still catches a coordinated-omission regression; it was not weakened into vacuity.

Bench crate stays `publish = false` / `cfg(unix)`; the freeze stays deterministic; no shipped-broker hook added.

refs #284


ELares:
Self-review resolution: CI fully green (18/18, fail=0 pend=0), including the `test (macos-latest)` job that was previously flaking on `an_injected_stall_shows_up_in_the_recorded_tail`.

The fix replaces the fixed non-vacuity floor (0.6*stall) with an adaptive baseline-relative threshold: `threshold_us = healthy.p999_us + cal.separation_us`, with a SANITY backstop (`healthy.p999 < 0.8*stall`) checked first. Both sides of the separation scale with the runner, so a slow/loaded host can no longer defeat the assertion, yet coordinated omission (which leaves stalled.p999 ~= healthy.p999) still FAILS the test. Empirically: 20/20 unloaded, 8/8 under 20-yes-generator CPU load, and a deliberate zero-freeze injection FAILS as required (stalled p99.9 11215us vs baseline 11002us, required >= threshold). Non-vacuous and robust. Merging.

PR #349, opened 2026-06-08
bench: load generator with production-safety and flash-endurance guards (#94)

### LINKED ISSUE #94: [TASK] bench: load generator with production-safety and flash-endurance guards
Parent: #15

Implement `ironbus bench`, the publish/subscribe/round-trip load generator that reports throughput, p50/p99/p999 latency, fsync cost, and bytes/op, honoring the #12 codec so the ratio is measured on real payloads. Beyond performance, the dominant requirement is that bench cannot damage production data or burn out edge flash, which the first draft underspecified.

## Approach

bench publishes over the #11 wire format and measures end-to-end and fsync latency using the same path as the broker. By default it targets an isolated synthetic topic namespace (for example `ironbus-bench-<random>`) and a fresh ephemeral cursor; it refuses to publish to an existing non-bench topic or join an existing named consumer group unless the operator passes `--i-understand-this-is-live`. It auto-creates and auto-deletes the synthetic topic and reports if cleanup fails. To protect edge flash write endurance, bench supports a `--no-fsync` dry-run mode and requires a bounded `--duration` or `--count` (no unbounded default run). Round-trip mode measures producer-to-consumer latency through the real #6 path so the fsync-cost number is honest. Output honors the global `--json` contract (versioned schema) with the latency histogram fields named explicitly.

## Acceptance criteria

- [x] bench reports throughput, p50/p99/p999 latency, fsync cost, and bytes/op in human and versioned `--json` output.
- [x] Latency and ratio are measured on #12-encoded real payloads, not synthetic incompressible noise by default.
- [x] Default target is an isolated synthetic topic and fresh ephemeral cursor; touching an existing topic or named group requires `--i-understand-this-is-live`.
- [x] The synthetic topic is auto-created and auto-deleted; cleanup failure is reported with a non-zero exit.
- [x] `--no-fsync` dry-run exists and a bounded `--duration`/`--count` is required (no unbounded default).
- [x] fsync cost is measured through the real #6 path in round-trip mode.

## Failure considerations

bench pointed at a live topic injects junk or steals messages in queue-group mode: mitigated by the synthetic-namespace default and the explicit live-topic guard. Unbounded bench burns edge flash endurance: mitigated by the required duration/count bound and `--no-fsync` dry-run. A leftover synthetic topic after a crash wastes space: mitigated by auto-delete and a reported cleanup-failure exit code.

ELares:
Implements `ironbus bench` (#94, parent #15): a publish / subscribe / round-trip load generator that reports throughput, p50/p99/p999 latency, fsync cost, and bytes/op over the real #11 wire format and the real #6 produce path, with the safety guards that are the issue's dominant requirement.

## Acceptance criteria (#94)

| # | Criterion | Status | Where |
| - | --- | --- | --- |
| 1 | throughput, p50/p99/p999 latency, fsync cost, bytes/op in human and versioned `--json` | DONE | `bench.rs` `bench_json`/`write_human`; `bench_run.rs` `finish_report` |
| 2 | latency/ratio on real (codec-friendly) payloads by default, not noise | DONE | `bench.rs` `PayloadShape::Realistic` default + `fill_payload`; `--payload-shape random` opt-in |
| 3 | default isolated synthetic target + fresh ephemeral cursor; existing topic/named group needs `--i-understand-this-is-live` | DONE | `bench.rs::parse_bench` guards; `bench_run.rs::run_isolated` (own broker + fresh `ironbus-bench-<random>` dir + group) |
| 4 | synthetic target auto-created and auto-deleted; cleanup failure reported with non-zero exit | DONE | `bench_run.rs::run_isolated` (spawn over fresh dir, `remove_dir_all` always); `bench.rs::cleanup_failed_error` (exit 70) |
| 5 | `--no-fsync` dry run + a bounded `--duration`/`--count` required (no unbounded default) | DONE | `bench.rs::parse_bench` bound guard; `--no-fsync` batches the bench broker's cursor checkpoints |
| 6 | fsync cost measured through the real #6 path in round-trip mode | DONE | `bench_run.rs::produce_one` times the produce call (PubAck only after the covering `fdatasync`, I2); median is the reported fsync cost, isolated from queue-wait |

## Key decisions

**Where it ships / dependency graph (kept clean, ZERO new deps).** The shipped `ironbus` binary must not pull the `publish = false` `ironbus-bench` crate (which brings `hdrhistogram`, `rand`, `rand_distr`, `serde_json`). So `ironbus bench` is a thin, self-contained subcommand inside `ironbus-cli` that reuses only crates the binary already ships (`ironbus-client`, `ironbus-proto`, and `serve` itself for the isolated broker). It computes percentiles with a tiny in-module nearest-rank quantile and emits JSON by hand like every other CLI verb. `Cargo.toml`/`Cargo.lock` are unchanged, so `deny.toml` needs no new allowance and the shipped graph is byte-for-byte the same. The heavyweight open-loop, coordinated-omission-free SLO instrument stays in `ironbus-bench` (its injected-stall self-test is untouched and still passes).

**Topic isolation under the single-queue model.** IronBus is one durable ordered queue today; there is no multi-topic abstraction, so I did not fake one. The real isolation unit is (a) the data directory (the "topic" namespace) and (b) the named consumer group (the cursor). The safety INTENT holds for real:
- Default ISOLATED mode spawns its OWN broker over a fresh, randomly-named `ironbus-bench-<random>` data directory under the temp dir, read through a fresh random `ironbus-bench-<random>` group. The synthetic dir is physically separate from any real broker's data dir, so bench cannot corrupt real data, and the random group means it cannot steal a real group's messages. The dir is auto-created and always auto-deleted; a failed delete is reported and exits 70.
- LIVE mode (`--addr`) targets an existing broker and is REFUSED without `--i-understand-this-is-live`. A `--group` that is not a `ironbus-bench-*` name is likewise refused without the flag (it could be a real consumer group whose messages bench would steal).

**Honest fsync cost.** Round-trip mode measures producer-to-consumer latency through the real durable path, but attributing the round-trip p50 as "fsync cost" would over-claim (it includes queue-wait under the per-ack durable path). Instead the reported fsync cost is the MEDIAN produce-call latency: a `Pub` returns its `PubAck` only after the covering group-commit `fdatasync` (invariant I2), so the produce-call median isolates the durable-write cost. An integration test pins that the reported cost is strictly below the round-trip max latency, so a regression that conflates them fails.

**`--no-fsync`.** `serve` always fsyncs (the safe `sync` durability level, per `docs/DURABILITY.md`); there is no broker flag to disable the produce fsync. So `--no-fsync` is an honest dry-run that batches the bench-spawned broker's cursor checkpoints (a large checkpoint interval instead of one durable cursor write per ack), cutting the bench's own flash writes for a quick capacity probe, and the report flags `fsync_measured: false` / `fsync_cost_us: null` so no one mistakes a dry-run number for an honest one.

## Tests with teeth

- Unit (`src/bench.rs`, run on every platform): the required-bound guard, the live-target guard, the non-bench-group guard, the zero-bound and payload-below-token rejections, the versioned-JSON shape with explicitly-named latency fields, publish-mode null latency, `--no-fsync` flagging, the realistic-vs-random payload distinctness, and the cleanup-failure mapping (asserts a 70+ non-clean exit). Each guard test FAILS if its guard is removed.
- Integration (`tests/bench.rs`, real binary subprocess, Unix): a 500-message round-trip run exits 0, emits versioned JSON, measures the fsync cost through the real path, the fsync cost is strictly below the round-trip max (isolation teeth), and the synthetic dir is auto-deleted (checked in a per-test private `TMPDIR`). Plus publish-mode null latency + cleanup, `--no-fsync` flagging, and the three guard refusals exiting 1.

## Local gate (all green)
- `cargo test --workspace`: 886 passed, 0 failed, 1 ignored (the pre-existing live SIGSTOP proof). The deterministic injected-stall self-test still passes.
- No em/en dash anywhere in the diff. No new dependencies (`Cargo.toml`/`Cargo.lock` unchanged).

## Engineering invariants
edition 2021 / MSRV 1.78, no `unwrap`/`expect`/`panic`/`unreachable`/`todo` in non-test paths (typed `CliError` at boundaries), no `unsafe`, `ironbus-core` untouched. The run path is `#[cfg(unix)]` like `serve`; the cross-platform parsing/guards/JSON compile and unit-test on every target, and the non-Unix `cmd_bench` stub consumes its args so the Windows `-D warnings` build stays clean. No `serve`/`ServeConfig` field was added, so the `cfg(not(unix))` `cmd_serve` stub footgun does not apply.

## Follow-up gaps (honest)
- Real end-to-end cleanup-FAILURE injection (a genuinely undeletable synthetic dir) is environment-fragile in CI (needs root / another user / a read-only mount), so it is covered by the mapping unit test plus the real auto-delete-success integration test rather than a real undeletable-dir run. A device/CI fault-injection harness for it could be a small follow-up.
- The `--rate` open-loop pacing here is a simple closed-form sleep, not the Poisson, coordinated-omission-free schedule the `ironbus-bench` crate uses; for honest SLO tail numbers use that crate. This `bench` is the operator-facing, safety-guarded tool the issue specifies.

Closes #94


ELares:
Independent adversarial review was SHIP-WITH-NITS: all five production-safety checks passed under break-and-confirm (default isolation is a separate data dir + random group so it physically cannot touch a real broker or steal a real group; the live-target guard refuses `--addr`/non-bench `--group` without `--i-understand-this-is-live` and the guard has teeth; a run with no `--duration`/`--count` is a usage error; `--no-fsync` batches the durable cursor; fsync cost is the median real produce-call latency through the I2 path). Zero new dependencies (Cargo.lock untouched), correct exit-code mapping, no dashes.

The one real defect (NIT-1) is now FIXED on this branch: the synthetic-dir cleanup was an explicit `remove_dir_all` that a `?` on a broker spawn failure or a panic in the load loop would skip, leaking the directory (proven by injection, and contradicting the unconditional-auto-delete safety claim). It is now an RAII `DataDirGuard` armed right after the dir is created and disarmed only after the normal-path explicit cleanup, so it removes the dir on every early-return/panic path while the explicit cleanup stays authoritative for the exit-70 failure report; a cleanup failure is also surfaced on stderr when the run itself failed. Added guard tests with teeth (armed -> removed, disarmed -> kept).

CI: 18/18 green on head `f7fe309` (fail=0 pending=0). Merging.

Issue #78, opened 2026-06-05
[TASK] Trained per-type dictionary lifecycle (train, content-hash dict_id, self-contained distribution)


ELares:
Parent: #12

Add trained per-message-type zstd dictionaries, the highest-ratio lever for IronBus's tiny similar payloads, with a lifecycle that never strands an offline reader. Deferred from v1 on purpose: a referenced-but-absent dictionary is permanently undecodable data, so this lands only once the distribution story below is real. The v1 frame already reserves dict_id, so this is additive with no wire-format break.

## Approach
Ship a #15 subcommand that collects sample records per message type and runs ZDICT training to produce a dictionary. dict_id is a u32 truncation of the content hash of the dictionary bytes, collision-checked at train time, so dictionaries are immutable and never reused (no central registry). Make dictionaries self-contained on disk: the producing node writes each referenced dictionary as a sidecar (dicts/<dict_id>.zstd) alongside the segments, AND the active set is embedded in the static binary per #17. A reader resolves dict_id from the on-disk sidecar first, then the embedded set; an unresolved dict_id is bounded poison via #8 (per the decoder-resilience sub-issue), never a silent drop or crash. Dictionaries are append-only; rotation means writing a new dict_id, never mutating an existing one.

## Acceptance criteria
- [ ] CLI trains a dictionary from collected samples and emits a content-hash dict_id.
- [ ] A frame written with a dictionary is decodable using only the on-disk sidecar (binary-independent).
- [ ] The same frame is decodable from the embedded set when the sidecar is absent.
- [ ] An unresolved dict_id is quarantined via #8, not crashed or silently dropped.
- [ ] Dictionaries are immutable; a re-train yields a new dict_id and leaves old data readable.
- [ ] A documented before/after ratio on real telemetry shows the dictionary win over no-dictionary per-batch.

## Failure considerations
The defining risk is undecodable data from a missing dictionary; self-contained per-segment sidecars make a segment decodable by whoever holds it, even after a binary downgrade, and the #8 route bounds the damage when even that fails. dict_id reuse would silently reinterpret old data; the content-hash id makes reuse structurally impossible. Embedding stale dicts in a downgraded binary is covered because the on-disk sidecar is resolved first.

ELares:
[...]

PR #353, opened 2026-06-08
docs(#78): specify the trained per-type dictionary lifecycle

ELares:
Specifies the design of #78 (parent #12) so it can close for the **M2: Prototype-Ready Design** milestone. Adds `docs/DICTIONARY_LIFECYCLE.md`, the normative DESIGN spec for trained per-message-type compression dictionaries, with every #78 decision made concrete and testable.

**Docs only.** No source changed, no dependency added. The spec is gated behind the existing ADR-0003 opt-in zstd feature; the `ZDICT` training call itself is an explicit impl follow-up, not done here (IronBus keeps a deliberately minimal dep graph: lz4_flex default, zstd opt-in behind a feature, no vendored C on the default path).

## What the spec pins (acceptance-criterion map)

| #78 acceptance criterion | Status |
|--------------------------|--------|
| CLI trains a dictionary from collected samples and emits a content-hash dict_id | SPECIFIED (S1, S2); the `ZDICT` call is IMPL-RESIDUAL |
| A frame written with a dictionary is decodable using only the on-disk sidecar (binary-independent) | SPECIFIED (S3a, S4) |
| The same frame is decodable from the embedded set when the sidecar is absent | SPECIFIED (S3b, S4) |
| An unresolved dict_id is quarantined via #8, not crashed or silently dropped | SPECIFIED (S5); emitting `UnresolvedDictId` is IMPL-RESIDUAL |
| Dictionaries are immutable; a re-train yields a new dict_id and leaves old data readable | SPECIFIED (S2, S6) |
| A documented before/after ratio on real telemetry shows the dictionary win | METHOD SPECIFIED (S7); the measured number is IMPL-RESIDUAL |

## Two load-bearing decisions

- **dict_id = `u32` truncation (first 4 bytes, little-endian) of BLAKE3-256 of the dictionary bytes**, collision-checked and REFUSED fail-closed at train time. Content-addressed, so reuse is structurally impossible with NO central registry. BLAKE3 is pure-Rust (no ADR-0003 violation); the truncation matches the reserved u32 dict_id width, which is exactly why the train-time collision guard is mandatory.
- **Unresolved dict_id maps to a NEW append-only loss reason `UnresolvedDictId` (code 7, label `unresolved_dict_id`, counts as data loss).** It is a distinct failure class from the corruption codes: framing and CRCs are INTACT, only the decode input (the dictionary) is absent (the at-rest "unknown key-id" shape), so reusing `CorruptRecordBody` would misreport bit-rot. Per the frozen `loss-report.v1` rule it is APPEND-ONLY: codes 1-6 stay byte-identical and `schema_version` stays `1` (exactly the move code 6 `ScrubberSuspect` already made). A v1 reader still reads a code-7 event's numeric span.

## Additive, no wire break

The v1 record frame already RESERVES a `u32 dict_id` in the compressed payload's self-describing descriptor behind `RecordFlags::COMPRESSED` (bit 0). In v1 `COMPRESSED` is never set and `dict_id=0` is the no-dictionary sentinel (verified in COMPATIBILITY.md: the codec/dict id spaces are not present yet, stored codec always "none"). A future dictionary-aware writer sets the bit and a non-zero dict_id; no existing byte moves, so the change is provably additive.

## Distribution and resolution

Self-contained two ways: a per-segment on-disk `dicts/<dict_id>.zstd` SIDECAR (content-named, durable before the referencing segment is acked, parallel to `quarantine/`, invisible to the recovery walk) PLUS the active set embedded in the static binary via `include_bytes!` (#17). Readers resolve the SIDECAR FIRST then the EMBEDDED set, the order that survives a binary downgrade (the sidecar travels with the data; the embedded set is fixed at the reader's build time and can be stale). Append-only rotation = a new dict_id, never mutate an existing blob, so old data is readable forever.

Consistency: ADR-0003 (lz4_flex default / zstd opt-in), the frozen `ironbus.loss-report.v1` schema, and the frozen frame/flag set. Keep-a-Changelog entry under `## [Unreleased]`; linked from `docs/README.md`. No em/en dashes (verified). DCO-signed.

refs #78


ELares:
Independent adversarial review: **SHIP-WITH-NITS** (now fixed). The four high-risk design commitments were each verified against the real code, not the doc's self-description:
- FROZEN loss-report.v1 schema: PASS. UnresolvedDictId at code 7 is genuinely the next free code; codes 1-6 stay byte-identical (golden tests pin them); schema_version stays 1 (the same append-only move code 6 ScrubberSuspect made); is_data_loss=true is correct (undecodable record, same boundary as quarantine::is_corruption_skip).
- Wire additivity: PASS. The COMPRESSED flag (bit 0) and the dict_id=0 sentinel are real and unused in v1; dict_id lives in the compressed-payload descriptor that v1 never writes, so it is provably additive (no byte moves, no tag renumber).
- dict_id collision safety and sidecar-first-then-embedded resolution (survives a binary downgrade, content-hash-revalidated, never strands an offline reader with the unresolved case bounded via #8): SOUND.

Fixed the nits on this branch: Section 2 now refuses a candidate deriving dict_id 0 (consistent with the Section 8 sentinel reservation); the BLAKE3 choice is disclosed as a new crate needing a deny.toml justification; and the reason-class wording is precise (SequenceGap has valid CRCs; the dict_id is reserved by flag interpretation, not a physical field).

CI: 18/18 green on 3f104c9. Merging with refs #78; the ZDICT training + sidecar IO + embed + the measured ratio are the tracked impl residual.

PR #354, opened 2026-06-08
docs: edge resource constraints design synthesis for #20 (constraint-to-knob table, tiny profile, RTC-less clock model)

ELares:
Completes the design of #20 (edge resource constraints) for the M2 (Prototype-Ready Design) milestone. Most of #20 was already specified by merged docs; this PR writes the three genuinely-missing DESIGN deliverables in a new normative doc, `docs/EDGE_CONSTRAINTS.md`, and cross-links the rest so a reader sees the whole #20 picture in one place. Docs only, no source changed.

## What this adds

1. **Hardware-constraint-to-knob mapping table** (#20 criterion 1, #117). One row per limit (flash endurance, RAM, CPU, power loss, clock/RTC, thermal): the failure if unhandled, the governing IronBus knob, its `tiny` default, and the owning doc/issue. Every knob name is cross-checked against the merged config (`serve` flags / `EngineConfigSnapshot` / RAM_BUDGET.md); conceptual knobs (the RSS boot guard, `mmap_max_bytes`, the monotone-wall-clock wrapper) are labelled **SPECIFIED-NOT-YET-A-FIELD** against their owning issue so the table never invents a field.
2. **The `tiny` profile, fully specified** (#20 criteria 2 and 8, #115). Concrete default values (8 MiB segment, 8 / 256 KiB consumer credit, 32 connections, 64 groups, 256 in-flight, drop-new, 1024 checkpoint, `sync` durability, lz4_flex) shown summing ~9 MiB under the 64 MiB RAM ceiling using RAM_BUDGET.md's budget, documented as THE default for unattended battery-less ARM, and reachable in the single static binary with no external dependency. The runtime auto-selection flag and the refuse-to-boot RSS guard are explicitly the #115 / #17 / #87 implementation residual, NOT claimed as enforced.
3. **Clock-skew / missing-RTC model** (#20 criterion 6, #117). The broker-assigned monotonic `seq` (record-header sequence / log offset, durable, recovered from disk alone) is the sole, clock-independent ordering authority; the wall-clock `timestamp` is producer-supplied advisory metadata with a Redis-Streams never-emit-a-smaller broker-timestamp rule for broker-stamped times; age retention is best-effort and fail-safe under a backward jump (a bad clock only makes the reaper more conservative, never prematurely destructive; the byte/count bounds are clock-free). Tied to the `Clock` trait seam.

Plus a section reconciling the four already-satisfied #20 criteria to their merged owners (bounded buffers -> RAM_BUDGET.md; power-loss-safe default -> DURABILITY.md; large/sequential/batched + measurable write-amp -> DURABILITY.md + EDGE_RUN_DISCIPLINE.md; codec justified, cold-only zstd -> ADR-0003).

## Consistency

- **No contradictions** found with the merged docs (RAM_BUDGET.md, DURABILITY.md, EDGE_RUN_DISCIPLINE.md, EDGE_TUNING.md, SLO.md, ADR-0003) or the source (`format.rs`, `clock.rs`, `segment.rs`, `log.rs`, `engine.rs`). The original #20 draft's illustrative `compression = "lz4"` / `clock_source = "monotonic_plus_logical"` TOML is superseded by the resolved decisions (lz4_flex per ADR-0003; monotonic-`seq` ordering), noted as superseded, not contradicted.
- Complements (does not duplicate) EDGE_TUNING.md: that doc is the operator runbook for shipped flags; this is the design-level map including the conceptual knobs and the profile/clock design #20 owns.
- No em/en dashes (verified with `grep -rnP '[\x{2013}\x{2014}]'` over the diff). `cargo fmt --all -- --check` clean. All markdown links resolve. Linked from `docs/README.md`; terse CHANGELOG entry under `### Documentation`.

refs #20 #115 #117


ELares:
Independent adversarial review: **SHIP-WITH-NITS** (now fixed). The reviewer independently re-verified every load-bearing claim against the real code: the knob table invents no field (each is a real flag or honestly labeled SPECIFIED-NOT-YET-A-FIELD with its owning issue, defaults cross-checked); the tiny-profile budget re-adds to ~9 MiB, well under the 64 MiB ceiling, with enforcement honestly marked an impl residual; the clock-skew/missing-RTC model is backed by the real code (monotonic seq/offset is the ordering authority, the reaper reads the clock seam and only ever gets MORE conservative under a backward jump, byte/count bounds are clock-free). No contradictions with merged docs, no em/en dashes.

Fixed the two precision nits on this branch: PubBody.timestamp_ms (not Append.timestamp_ms) is the producer timestamp, and seq is its own monotonic counter distinct from the offset (not derived from it).

CI: 18/18 green on cf055f8. Merging with refs #20.

Issue #29, opened 2026-06-05
[TASK] Document the open prior-art gaps for IronBus-novel features


ELares:
Parent: #2

#2 thoroughly catalogs what to take and leave from existing systems but never names the places where NO surveyed system provides a usable answer. IronBus's riskiest promises (bounded-and-REPORTED data loss, automatic poison-record and poison-segment skipping with a machine-readable loss report, and structural producer non-blocking via disk spill and load shedding) have no clean prior-art. This task gives those features grounded starting points instead of a blank page.

## Approach
Add an 'open prior-art gaps' subsection. For each IronBus-novel feature, name the closest partial analog and state precisely what the analog does NOT cover. Poison-record and poison-segment skipping: closest analog is RocksDB `kPointInTimeRecovery` (longest-valid-prefix replay), which stops at the first bad record rather than skipping a bad segment and resuming, and produces no operator-facing loss report; SQS DLQ routes poison messages but only after delivery, not during recovery. Bounded-and-reported loss accounting: no surveyed system emits a machine-readable record of exactly which offsets were dropped and why; the closest is Kafka's recovery logs, which are unstructured. Structural producer non-blocking via spill: closest analog is Kafka's producer-side buffer that blocks or fails when full, which is the opposite of spill-to-disk-and-shed; SQS's in-flight cap sheds on the consumer side only. For each, state the concrete behavior IronBus must invent and which downstream issue (#7, #8, #10) owns it.

## Acceptance criteria
- [ ] Each IronBus-novel feature has a named closest analog and an explicit statement of what the analog does not cover.
- [ ] Poison-segment skip, loss accounting, and producer spill are each addressed.
- [ ] Each gap names the downstream issue that owns the invention.
- [ ] The subsection makes clear these are inventions, not borrowings.
- [ ] #7, #8, and #10 can cite this as their starting point.

## Failure considerations
The risk this addresses is designing the hardest parts from scratch with no lessons-learned. A residual risk is overclaiming novelty when a real analog exists; each entry must cite the analog and state only the specific delta IronBus adds. Another risk is that 'reported loss' stays vague, so the subsection must specify that the loss report is machine-readable and lists offsets and reasons, matching the product's bounded-and-reported tenet.

Issue #28, opened 2026-06-05
[TASK] Reconcile page-cache vs direct-DMA and fsync-floor vs async into one edge stance


ELares:
Parent: #2

#2 contains two unreconciled contradictions: it says to copy Kafka's page-cache reliance while praising Redpanda's no-page-cache direct-DMA design, and it recommends both a group-committed fsync floor and a page-cache-async path without stating which wins on the SAME battery-less edge box. A prior-art issue exists to settle exactly this, so leaving both halves standing lets #4, #19, and #6 each pick a different half and diverge.

## Approach
Write a short reconciliation subsection that resolves both tensions for the edge target. State the stance plainly: IronBus relies on the OS page cache for read warmth and for coalescing sequential writes (the Kafka model), and forces a group-committed fsync at the durability boundary (the RocksDB model), so the cache and the fsync are complementary, not alternatives. Reject Redpanda's no-page-cache direct-DMA stance for the edge default with the cited reasons the draft already gives: it loses warm-cache-on-restart and requires core-pinning with no work-stealing, both wrong for a single-core-constrained ARM node. State that page-cache-async is an opt-in only, never the default, and cross-reference #6 as the owner of the exact group-commit sizing. Add a one-line decision table: read path (page cache), write coalescing (page cache), durability boundary (forced fsync), so no downstream issue can read the survey two ways.

## Acceptance criteria
- [ ] The reconciliation subsection states a single coherent edge stance for page cache plus fsync.
- [ ] Redpanda's direct-DMA stance is explicitly rejected for the edge default with a cited reason.
- [ ] Page-cache-async is named as opt-in only, not the default.
- [ ] #4, #19, and #6 can each cite one unambiguous stance.
- [ ] No remaining sentence in #2 can be read as recommending no-fsync-by-default.

## Failure considerations
The risk this task removes is incoherent inheritance: without it, one downstream issue builds async durability while another builds a fsync floor on the same box. A residual risk is over-pinning a perf number here that belongs to #19; the subsection deliberately states the stance, not the tuning constants, and defers sizing to #6. Another risk is that the reconciliation reads as taste, so each claim cites the specific edge condition (single core, restart warmth, battery-less power) that decides it.

PR #362, opened 2026-06-08
docs: reconcile the edge IO stance and the open prior-art gaps (#28, #29)

ELares:
Adds `docs/PRIOR_ART_AND_IO_STANCE.md`, two short consolidating spec sections so the M1 architecture-spec tasks #28 and #29 can close. Both are cases where the DECISION was already made and scattered across merged docs and code, but the reconciling prose was never written. Docs only, no source changed.

## #28: the single edge IO stance
- States the stance: IronBus uses the OS page cache for reads (warm-cache-on-restart) and for write coalescing, AND a forced group-committed `fdatasync` as the durability boundary (ADR-0001, DURABILITY.md I2). These are complementary (the Kafka cache model plus the RocksDB forced-barrier model), not alternatives.
- Page-cache-async is OPT-IN ONLY, off by default, cannot be reached from the `serve` command line today, requires an explicit data-loss acknowledgement, and is labeled not power-loss safe (SLO.md async row, EDGE_CONSTRAINTS.md).
- EXPLICITLY REJECTS Redpanda-style no-page-cache direct-DMA / thread-per-core-pinned IO for the edge default, with the cited reasons: single-core ARM edge targets, lost warm-cache-on-restart, and core-pinning without work-stealing wasting a scarce core.
- One-line decision table: read path = page cache; write coalescing = page cache; durability boundary = forced group-committed `fdatasync`; async = opt-in only, not power-loss safe.

## #29: the open prior-art gaps
For each IronBus-novel feature, the closest partial analog, the precise delta, and the owning issue:
- Poison-record / poison-segment SKIP with bounded-and-reported loss. Analogs: RocksDB `kPointInTimeRecovery`, ZFS scrub, Kafka log truncation. Delta: per-event typed loss accounting with hard caps that fail closed (the `64 MiB` per-event cap and the `1%` global cap, freeze read-only on exceed). Owner #8.
- Bounded-and-reported recovery loss accounting. Analog: Kafka's unstructured recovery logs. Delta: a frozen, machine-readable `ironbus.loss-report.v1` of exactly which offsets were dropped and why, vs silent truncation. Owner #7.
- Producer SPILL / SHED backpressure. Analogs: Kafka's blocking producer buffer, SQS consumer-side in-flight cap. Delta: a durable spill (the log itself) plus a deterministic reported shed (drop-new / drop-oldest) plus a rate-independent CoDel sojourn signal. Owner #10.
- Honest novelty boundary: the novelty is in the COMBINATION and the bounded-and-reported loss CONTRACT, not in any single primitive.

## Consistency
Cross-checked against and cites the merged ADR-0001, DURABILITY.md, SLO.md, EDGE_CONSTRAINTS.md, and BACKPRESSURE.md (plus RAM_BUDGET.md, BASELINE_RIG.md, the loss-report schema, INVARIANTS.md, METRICS.md). No contradictions found. Linked from `docs/README.md` with a terse Keep-a-Changelog entry. No em-dash or en-dash anywhere. DCO signed.

refs #28 #29


ELares:
Independent adversarial review: **SHIP-WITH-NITS**. Every load-bearing claim verified against merged docs/source with no contradiction: the page-cache-for-reads + write-coalescing + forced-group-committed-fdatasync stance (ADR-0001, DURABILITY.md I2), the explicit Redpanda direct-DMA rejection with all three cited reasons, and critically the claim that page-cache-async is opt-in/off/not-CLI-reachable was confirmed by enumerating every serve flag (no --durability/--async flag exists). #29 per-feature analog/delta/owner is fair and accurate (poison-skip->#8, recovery-loss-accounting->#7, spill/shed->#10), honest about the novel-combination boundary. No em/en dashes. Merging with refs; closing #28/#29.

PR #368, opened 2026-06-08
test(server): scoped loom concurrency models for commit index, WAL handoff, and refcounts (#122)

### LINKED ISSUE #122: [TASK] loom scoped concurrency tests for commit index, ring-buffer handoff, and refcounts

ELares:
Closes #122

## What

Bounded, exhaustive `loom` permutation tests for the three genuinely cross-thread hot paths whose corruption would violate I1 (lost ack) or the commit watermark. loom was entirely absent from the workspace before this.

The tests live in `crates/ironbus-server/tests/loom_concurrency.rs` (gated `#![cfg(loom)]`) and model:

1. **COMMIT-INDEX publish/observe** (`commit_index_*`) -> mirrors the `AckCursor` committed watermark in `ironbus-core::cursor`. A committer publishes a monotonic index with `Release` after writing the covered data; a reader observes it with `Acquire`. Invariant: an observed index implies its covered data is visible, and the watermark never tears or regresses.
2. **RING-BUFFER / WAL HANDOFF** (`wal_handoff_*`) -> mirrors the bounded `std::sync::mpsc::sync_channel` handoff between connection handlers and the committer in `ironbus-server::actor`. A bounded ring with ATOMIC slot reservation (`fetch_add`/`compare_exchange` on `tail`). Invariants: no lost or duplicated item, FIFO per producer, backpressure (block, never overwrite) when full, no deadlock.
3. **REFCOUNT** (`refcount_*`) -> mirrors `ironbus-server::server::ConnectionSlot`'s `active.fetch_sub(1, AcqRel)` release and the shared `Arc`. Invariants: the resource is dropped exactly once, only after the last reference releases, never while a reader pins it (no use-after-free analog); every slot acquire is matched by exactly one release.

## Approach: faithful models, not cfg-switched production

The three production structures are not shaped to be cfg-switched cheaply (the actor's bounded `sync_channel` has no loom replacement; `AckCursor` is a pure IO-free value type with no atomics; the refcount is an `AtomicUsize` slot plus `Arc`). cfg-switching them would touch hot code and risk perturbing the std path. Per the accepted scoped-loom pattern, each model is a small bounded structure built from loom's replacement primitives that replicates the EXACT atomic ordering of the real code, with a comment cross-referencing the real symbol it stands in for. The normal (std) build and all existing tests are unchanged.

## Constraints honored

- Each model: `<= 3` threads, `<= 4` shared ops per thread; `LOOM_MAX_PREEMPTIONS=3` in CI.
- Correctness rests only on `AcqRel`; no SeqCst-only reasoning (per loom's documented modeling limits). Scope limits and loom false-positive/false-negative caveats are documented in the module header.
- Each test asserts it explored `> 1` interleaving (guards against a model that silently collapses to one path).

## Model counts (proof of exhaustive exploration)

```
commit_index_publish_observe:     19 interleavings
commit_index_monotonic_advance:   59 interleavings
wal_handoff_spsc_bounded:        182 interleavings
wal_handoff_mpsc_backpressure: 61184 interleavings
refcount_pin_vs_recycle:       10626 interleavings
refcount_connection_slot_balance: 90 interleavings
```

All 6 pass in ~15s total.

## Teeth

Temporarily weakening the commit-index `observe` load from `Acquire` to `Relaxed` makes loom find the failing interleaving (reader sees index=1 with stale data=0), proving the test genuinely depends on the real ordering. Reverted.

## CI and supply chain

- New non-blocking nightly `loom` job runs the tests under `RUSTFLAGS="--cfg loom"` with `LOOM_MAX_PREEMPTIONS=3`. actionlint-clean; the new action `uses` are SHA-pinned (reuses the existing pinned actions).
- `loom` is a `[target.'cfg(loom)'.dev-dependencies]` (MIT, already on the deny.toml allowlist), so it never enters the shipped graph (`cargo tree -i loom` on the default graph is empty) and `cargo deny check` stays green.

## Local gate

- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: clean (also clean under `--cfg loom`)
- `cargo test --workspace`: all pass (loom target runs 0 tests under normal cfg, correctly gated)
- `RUSTFLAGS="--cfg loom" cargo test -p ironbus-server --test loom_concurrency`: 6 pass
- `cargo deny check`: green
- No em/en dashes anywhere.


ELares:
Independent adversarial review: **APPROVE**. The separate-workspace fix genuinely keeps loom and its env-filter tree out of the shipped graph (verified: root Cargo.lock byte-identical to main, no loom in `cargo tree -p ironbus-cli`, shipped tracing-subscriber has no env-filter). The three models are bounded (<=3 threads, <=4 ops) and have real teeth (weakening Acquire->Relaxed in models 1 and 2 made loom find the failing interleaving). Faithfulness: model 3 (refcount) is a literal match to ConnectionSlot AcqRel; models 1 and 2 are pattern-only (the real cursor watermark is single-threaded, made visible via the actor channel; the real handoff is std sync_channel), which is the standard use of scoped loom models. The overclaim wording the reviewer flagged is fixed on this branch (CHANGELOG and nightly comment now say "concurrency patterns" not "mirrors AckCursor"; the vestigial deny.toml-allowlist note dropped). Normal build unchanged + green. Merging; Closes #122.

Issue #95, opened 2026-06-05
[TASK] Embedded HTTP server with /metrics, /healthz, /readyz and secure bind defaults


ELares:
Parent: #16

Stand up the single embedded HTTP server that exposes the observability surface on one port, with secure-by-default binding and the liveness/readiness semantics that an orchestrator and #7 depend on. This is the front door for metrics-style scraping and the health probes; it must do no background work until scraped so an unscraped edge node pays essentially zero.

## Approach

Serve four routes on default port 9095 bound to 127.0.0.1: `GET /metrics` (OpenMetrics text), `GET /healthz` (liveness), `GET /readyz` (readiness), and `GET /admin/...` (read-only JSON, detailed in the sibling task). Binding to any non-loopback interface requires an explicit config key AND a configured bearer token or TLS client auth, validated at startup with #18; otherwise the process refuses to start and logs why. Liveness returns 503 only after a hysteresis window (default 10 s, configurable) of no event-loop progress reported by a monotonic-clock watchdog; a slow-but-progressing fsync never fails liveness, and a read-only frozen node still returns liveness 200. Readiness returns 503 while #7 replay is in progress or the node is frozen, and 200 once it accepts writes. All routes use a monotonic clock for any duration.

## Acceptance criteria

- [ ] Default bind is 127.0.0.1:9095; a wider bind without auth refuses to start and emits a clear config error.
- [ ] `/healthz` flaps only after the configured hysteresis window of no progress; a slow progressing fsync keeps it 200.
- [ ] A frozen node returns `/healthz` 200 and `/readyz` 503.
- [ ] `/readyz` returns 503 throughout #7 replay and 200 only when writes are accepted.
- [ ] An unscraped node spawns no mandatory background thread and shows no measurable steady-state CPU from the server.

## Failure considerations

An unauthenticated non-loopback bind would leak topology and offsets and invite scrape-amplification DoS, so the startup refusal is the primary guard. A liveness probe with no hysteresis would restart the node mid-fsync and crash-loop it, so progress is measured on a monotonic clock with a window. Wall-clock skew must not drive any health timing.

ELares:
PR #369 implements the two open gaps (the four-route loopback server already shipped in `crates/ironbus-server/src/health.rs`):

[...]

PR #369, opened 2026-06-08
feat(health): secure-bind guard and liveness hysteresis watchdog (#95)

ELares:
Implements the two open gaps in the embedded health server (#95): the fail-closed SECURE-BIND guard and the `/healthz` liveness HYSTERESIS watchdog. The four-route loopback server already shipped; this PR adds the security-critical bind refusal and the monotonic-clock liveness watchdog.

### Secure-bind guard (fail-closed)
- `serve` REFUSES to start (exit 1, before any listener opens) when `--health-addr` resolves to a non-loopback address, naming the address and explaining that the health surface is unauthenticated and unencrypted (TLS #107 and an auth identity #106 are specified but not wired).
- New `--health-allow-public` flag (`IRONBUS_HEALTH_ALLOW_PUBLIC`) acknowledges the risk and binds the public surface with a loud startup WARN on every start. Loopback binds ignore it and never warn.
- Classification is on the RESOLVED address via `ToSocketAddrs`, so a hostname or the wildcards `0.0.0.0`/`::` that map to a routable IP are caught; the resolved addresses are exactly what gets bound. No override exists for the wire `--addr` bind.

### Liveness hysteresis watchdog
- The wire accept loop ticks a monotonic-clock progress beacon (`LivenessBeacon`, new `ironbus-server/src/liveness.rs`) on EVERY iteration, including the idle would-block poll, so idle is progress.
- `/healthz` returns 503 only after `--health-liveness-window-ms` (default 10 s, `0` = disabled) with no tick, so a slow-but-progressing fsync stays 200, a healthy idle node stays 200, and only a stuck loop sheds. The watchdog reads the clock directly (not through the append actor), so a frozen writer still returns `/healthz` 200 (liveness != readiness).
- `/readyz` is unchanged (writer-frozen / shutdown gate).

## Per-criterion status
- [x] Default bind is loopback; a wider bind without auth refuses to start with a clear config error. `crates/ironbus-cli/src/main.rs` `health_bind_decision` / `health_non_loopback_refusal` / the top-of-`cmd_serve` guard.
- [x] `/healthz` flaps only after the configured hysteresis window; a slow progressing fsync keeps it 200. `crates/ironbus-server/src/health.rs` `/healthz` branch + `crates/ironbus-server/src/liveness.rs`; `crates/ironbus-server/src/server.rs` `serve` ticks the beacon.
- [x] A frozen node returns `/healthz` 200 and `/readyz` 503. The watchdog measures the accept loop, not writer health, so a frozen writer is still live; `/readyz` already gates on `is_healthy()`.
- [ ] PARTIAL: `/readyz` 503 throughout #7 replay. Recovery completes before the listener opens today (a started broker has already replayed), so there is no serving-during-replay window to gate; left as a noted follow-up rather than adding a replay-in-progress flag. `crates/ironbus-server/src/health.rs` `/readyz` unchanged; documented in METRICS.md.
- [x] An unscraped node spawns no mandatory background thread. The health server runs only with `--health-addr`; the watchdog is a single relaxed atomic store on the existing accept loop, no new thread.

## Tests
- Watchdog: 200 while progress is fresh, 503 after the window with no progress, no flip on normal idle, and a zero window disables it (`crates/ironbus-server/src/health.rs`, `crates/ironbus-server/src/liveness.rs`).
- Secure bind: loopback / `::1` classify loopback; `0.0.0.0` / `::` / a routable IP classify non-loopback; unresolvable address errors; a non-loopback bind WITHOUT the ack is refused; loopback binds silently; a non-loopback bind WITH the ack warns; end-to-end through `run`, the refusal is exit 1 naming the address (`crates/ironbus-cli/src/main.rs`).

## Invariants
- `ironbus-core` unchanged (IO-free preserved; the watchdog lives in `ironbus-server`).
- The `#[cfg(not(unix))]` `cmd_serve` stub consumes both new `ServeConfig` fields (`health_liveness_window_ms`, `health_allow_public`).
- No em/en dashes (`grep -rnP '[\x{2013}\x{2014}]'` clean). Docs updated: CLI.md, CONFIG.md, METRICS.md, CHANGELOG.
- Local gate green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, `cargo test --workspace` (974 passed, 0 failed).

refs #95


ELares:
Independent adversarial review: **APPROVE**. The bind guard is genuinely fail-closed and no bypass was found: classification is on the RESOLVED SocketAddr set (all-IPs-loopback), so a hostname resolving to a public IP, a wildcard 0.0.0.0/::, an IPv4-mapped IPv6, a mixed-resolution hostname, and an empty/malformed addr are all refused before any side effect (no dir lock, no engine, no socket); break-and-confirm proved the refusal test has teeth. The hysteresis watchdog reads the monotonic clock seam directly (a wedged writer stays /healthz-200 but /readyz-503), ticks on every accept-loop iteration incl idle (no flap), and flips 503 only past the window. Windows stub consumes both new ServeConfig fields. The /readyz-during-replay residual is moot today (recovery completes synchronously before the listener binds, verified). No blockers. Merging with refs; closing #95 (the secure-bind + hysteresis deliverables are complete; the readyz-replay gate is a latent defense only relevant if recovery ever becomes lazy).

PR #370, opened 2026-06-08
feat(cli): scrub and repair offline inspection + read-only-by-default repair (#92)

### LINKED ISSUE #92: [TASK] peek/dump/repair/scrub: offline inspection and read-only-by-default repair

ELares:
Closes #92

Implements the `scrub` and `repair` offline verbs on top of the shared offline reader (`peek`/`dump` already ship).

## What landed

- **`scrub --data-dir <D>`**: strictly READ-ONLY full integrity scan sharing the recovery decode path (`OfflineReader`). Reports every corruption/torn-tail/checksum span (the plan), marks-not-hides what recovery would quarantine, never writes.
- **`repair --data-dir <D>`**: defaults to the SAME read-only plan as scrub (print what it would do, change nothing). `--apply` acquires the EXCLUSIVE data-dir lock first (exit 5 if a broker holds it, reusing `dirlock.rs`), then runs recovery via `Log::open`: quarantine-copies each corrupt span to `quarantine/` (never deletes), truncates the active segment to the longest valid prefix with the atomic fsync+rename + dir-fsync discipline, and preserves the data dir's uid/gid/mode. It is recovery made explicit and offline; it never makes the data less recoverable than recovery already would.
- **Exit-code-3 gate** (per `docs/CLI_CONTRACT.md`): exit 3 on real data loss (a corruption skip), exit 0 on clean OR torn-tail-only (the `ReasonCode::is_data_loss` boundary), exit 2 on a missing dir, exit 4 on a structurally-unreadable chain (BLOCKED, distinct from 3 = finished-and-reported), exit 5 when a broker holds the lock on `--apply`.
- **`--json`**: emits the versioned `ironbus.cli.scrub.v1` / `ironbus.cli.repair.v1` result object on every exit path, carrying `ok` and `exit_code`; `events[]` field names match `ironbus.loss-report.v1`. Registered in `docs/compat/versions.md`.
- **`dump --raw` / `--require-dict`**: accepted as the frozen compression-inspection surface but inert until on-disk compression (#12) lands (codec is always `none` today). Documented as a residual gated on #12, not faked.

## Read-only-by-default + quarantine-not-delete under the exclusive lock

`repair` without `--apply` only ever calls `OfflineReader::open` (which the storage crate documents and a test here proves never writes). `--apply` is the only write path: it takes `DirLock::acquire` (the same `flock(LOCK_EX|LOCK_NB)` `serve` holds) BEFORE touching anything, so a live broker blocks it with a fail-fast exit 5 and nothing is mutated; then `Log::open` quarantine-COPIES the corrupt span, truncates to the valid prefix, and preserves the dir in place. `AlreadyLocked` is mapped to exit 5 here (vs `serve`'s exit-70 double-open guard).

## Tests with teeth

Real on-disk dirs, planted corruption, real recovery: scrub clean exits 0; scrub planted corruption exits 3 and reports the exact span; scrub torn-tail-only exits 0; scrub is byte-for-byte read-only (no quarantine created); repair plan changes nothing; repair `--apply` quarantines + truncates + preserves mode (and scrubs clean afterward, proving recovery agreement); repair `--apply` exits 5 under a held lock and mutates nothing.

## Local gate

- `cargo test --workspace`: 985 passed, 0 failed (11 new scrub/repair tests)
- No em-dash/en-dash (`grep -rnP '[\x{2013}\x{2014}]'`): clean

## Residual

`dump --raw` / `--require-dict` / missing-dict-degrade is only meaningful once on-disk compression (#12) lands; the flag surface is implemented but inert (codec always `none`), noted as gated on #12.

Engineering invariants held: no unwrap/expect/panic/unreachable/todo in non-test paths, typed errors mapped to the frozen exit codes, no unsafe added, ironbus-core unchanged, rustdoc on the public-ish surface, non-unix stubs consume every param. DCO-signed.


ELares:
Independent adversarial review: **APPROVE**. repair is provably safe: read-only by default (only OfflineReader read path, no write reachable), the exclusive flock is acquired BEFORE any mutation with exit 5 if the broker holds it (break-and-confirm: moving the lock after Log::open made the held-lock test fail on leaked quarantine+truncation), it quarantine-copies before truncating to the longest valid prefix exactly as Log::open recovery does (so never less recoverable than a normal restart), and uid/gid/mode are preserved. scrub is strictly read-only (a byte-for-byte before/after dir snapshot test). Exit codes correct (3 handled-data-loss with TornTail-only staying 0 via is_data_loss, 4 blocked, 5 lock-held, 2 missing, 70 IO). 185/185 cli tests, no dashes. The LogConfig::default() segment-cap divergence is marginal (same on-disk truncation; repair never loses more data). Merging; Closes #92.

Issue #118, opened 2026-06-05
[TASK] Edge metrics and power-loss fault-injection test contract


ELares:
Parent: #20

Specify the edge-specific metrics (flash write amplification, RAM headroom, throughput-collapse thermal signal, reported loss counters) and the power-loss fault-injection test contract that verifies the safety claims this issue makes. The draft asserts a power-loss-safe default, torn-tail truncation, and bounded RAM but never says how any of it is verified, so the guarantees are untested. This sub-issue gives concrete signals to #16 and a fault-injection contract to #21.

## Approach

Export, at minimum: `logical_bytes_written`, `physical_bytes_written`, a derived `write_amp_ratio` gauge, an estimated flash-erase-cycles-per-day counter, a RAM-headroom gauge (ceiling minus RSS), per-topic `dropped_records` and `recovered_loss_records` counters, and a portable throughput-collapse saturation signal (with a chip-temperature gauge only where a platform source exists). Add an opt-in `daily_physical_write_budget_bytes` that, when set and exceeded, sheds via #10 rather than weakening durability; default unset (metric-only). For testing, define a fault-injection harness with #21: a write interceptor that can drop the tail of the last write, flip a record checksum, and make fsync return an error, plus a RAM-pressure burst test asserting RSS stays under `ram_ceiling_bytes`. These tests gate the `tiny` profile.

## Acceptance criteria

- [ ] Write-amplification, RAM-headroom, drop, and recovered-loss metrics are exported and documented.
- [ ] The thermal signal is throughput-derived and portable; temperature is a best-effort add-on where available.
- [ ] `daily_physical_write_budget_bytes` is opt-in (default unset) and sheds via #10 when exceeded.
- [ ] A fault-injection harness can drop a write tail, flip a checksum, and force an fsync error.
- [ ] Power-loss and RAM-burst tests gate the `tiny` profile in CI.

## Failure considerations

The failures this work guards against are flash wear-out (caught by `write_amp_ratio` and the erase-cycles counter), silent data loss on brownout (caught by the drop-the-tail and fsync-error injectors plus `recovered_loss_records`), and RAM overrun under burst (caught by the RSS assertion). A risk specific to this sub-issue is the thermal signal binding to Linux-only sysfs and breaking the cross-platform binary; keep the trigger on the portable throughput signal and treat temperature as an optional gauge.

ELares:
PR #371 implements the edge metrics and the arming checksum-flip fault injector.

Acceptance criteria:
[...]

PR #371, opened 2026-06-08
feat(metrics): edge resource metrics and an arming checksum-flip injector (#118)

ELares:
Implements the edge metrics and the arming checksum-flip fault injector from #118. The fsync histogram and recovery-loss counters already shipped; this fills the EDGE-metrics gap and promotes the checksum flip from a static corpus to an arming injector.

## A) Edge metrics on `/metrics` (additive)

All new gauges carry no `_total` suffix, so the FROZEN resilience-counter taxonomy stays intact; the one new `_total` (a shed counter) is added to the frozen taxonomy deliberately. Both golden tests (`the_resilience_counter_taxonomy_is_frozen`, `the_metric_name_and_type_contract_is_frozen`) stay green.

- **Write amplification**: `ironbus_logical_bytes_written` / `ironbus_physical_bytes_written` counters and the derived `ironbus_write_amp_ratio` gauge. Physical = record frames + segment headers/footers (the real flash-wear write volume); logical = user payload. Tracked in the append/segment path in `ironbus-storage` (`Log::charge_physical`).
- **RAM headroom**: `ironbus_ram_headroom_bytes` = `ram_ceiling_bytes - RSS`, with a `-1` unavailable sentinel when no ceiling is set or RSS is unreadable. RSS is read best-effort with NO `unsafe`: `VmRSS` from `/proc/self/status` on Linux, `ps -o rss=` on macOS, unavailable elsewhere, behind `crates/ironbus-server/src/rss.rs`.
- **Throughput-collapse signal**: `ironbus_produce_saturated` gauge (1 once the broker has shed a produce). Portable and throughput-derived, NOT a thermal sensor; chip temperature is left as an optional device-only add-on.
- **Opt-in daily write budget**: `daily_physical_write_budget_bytes` log knob (off by default). When set and today's physical writes reach it, a produce sheds via the existing #10 drop-new path (counted in `ironbus_produce_rejected_total`), surfaces `ironbus_daily_write_budget_sheds_total`, flips `ironbus_daily_write_budget_over`, and never weakens durability. The UTC-day meter resets on the clock seam. Also exposes `ironbus_daily_physical_write_budget_bytes` and `ironbus_physical_bytes_written_today`.

## B) Fault injection

`FaultControl::arm_checksum_flip(offset)` promotes the checksum flip to an arming one-shot injector: a covering `write_all_at` persists a bit-flipped byte and SUCCEEDS, so a complete-but-corrupt record lands on disk. A power-loss test asserts recovery detects and reports the `CorruptRecordBody` loss.

## Tests with teeth

Write-amp counters increment to the exact expected logical/physical/ratio; the ram-headroom gauge reflects `ceiling - measured RSS`; the daily-budget over-budget signal fires when exceeded (and resets at the UTC day boundary); the checksum-flip injector corrupts and recovery reports it; both golden tests stay green.

## Local gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all pass (987 tests, 0 failures). No em/en dashes.

## Residuals

- The "gate the `tiny` profile in CI" acceptance criterion depends on #87 (the tiny profile does not exist yet); the metrics and the RAM-burst RSS-ceiling assertion are covered.
- The chip-temperature gauge is intentionally a device-only add-on (no portable source).
- A dedicated `serve` flag for the two knobs is a small follow-up; both are reachable via the library config today and echoed on `/admin`.

refs #118


ELares:
Re-review of the fix: **SHIP**. The BLOCKER is fixed: the daily-write-budget shed now returns a distinct `StorageError::DailyWriteBudgetExceeded`, and the DropOldest reap-gate (`append_with_policy`) keys on the StorageError-level `is_at_capacity()` which is byte-cap-only, so a budget shed gets the producer-facing rejected-produce reply but NEVER enters the force-reap loop. The regression test `drop_oldest_treats_a_daily_budget_shed_as_final_and_never_force_reaps` asserts force-reaps == 0 and is teeth-proven (fails under the pre-fix routing). Both NITs fixed (produce_saturated now reflects produce_rejected || budget sheds matching its HELP; the duplicated comment removed). Frozen taxonomy + metric-name golden tests stay green; I2/group-commit/crash-recovery green. Merging with refs #118; the tiny-profile CI gate is the residual owned by #87.

PR #372, opened 2026-06-08
feat(cli): compiled-in versioned profiles and materialized-config logging (#87)

### LINKED ISSUE #87: [TASK] Compiled-in versioned profiles and materialized-config logging

ELares:
Implements #87: three compiled-in, versioned `serve` tuning profiles and a materialized-config startup log.

## What

- **`--profile <edge-tiny|balanced|throughput>`** serve flag (and `IRONBUS_PROFILE` env var; default `balanced`). The selected preset is applied FIRST, then any explicit env var or flag overrides an individual knob, so the precedence is **profile < env < flag** (consistent with `docs/CONFIG.md` section 2). An unknown name is a usage error (exit 1).
- **Three presets** with the exact `docs/CONFIG.md` section 6 values, `edge-tiny` cross-checked byte-for-byte against `docs/EDGE_CONSTRAINTS.md`:
  - `edge-tiny`: 8 MiB segments, consumer_credit 8, consumer_credit_bytes 256 KiB, max_connections 32, max_groups 64, max_in_flight 256, checkpoint_interval 1024, disk_full_policy drop-new.
  - `balanced` (default): IS the compiled `DEFAULT_*` set (64 MiB / 64 / 8 MiB / 256 / 1024 / 1024 / drop-new / 1024 / 30 s / 5), written against those constants so zero-config is byte-identical to before.
  - `throughput`: 256 MiB segments, consumer_credit 512, consumer_credit_bytes 64 MiB, max_connections 1024, max_groups 4096, max_in_flight 8192, checkpoint_interval 4096, disk_full_policy drop-oldest. Every throughput value is fixed by CONFIG.md (none had to be chosen freely).
- **`PROFILE_SCHEMA_VERSION`** (= 1) carried in the materialized config. `serve` logs one structured `materialized-config` line at startup with the active profile, the schema version, and every resolved knob.
- A profile **never sets `data_dir` or any network/TLS key**.
- **Breaking-change convention** noted: a profile content change bumps `PROFILE_SCHEMA_VERSION` and is a documented breaking-change CHANGELOG entry (a `balanced` change is also a default change).

## Precedence proof (profile < env < flag)

One knob set at three layers: `--profile edge-tiny` sets `max_connections` 32; `IRONBUS_MAX_CONNECTIONS=50` overrides to 50; `--max-connections 64` overrides to 64. Tests assert the flag wins (64), and that with the flag removed the env (50) beats the profile (32).

## Windows footgun

The new `--profile` field and the materialized-config line are read only on the unix serve path, so they are consumed in the `#[cfg(not(unix))]` cmd_serve stub. Verified with `cargo check --target x86_64-pc-windows-gnu` under `-D warnings` (the cross check actually caught `Profile::name`/`materialized_config_line` being unused before the fix).

## #118 cross-reference

This unblocks the #118 tiny-profile CI gate. The gate (boot `serve --profile edge-tiny`, assert edge metrics / RAM headroom within the tiny budget) is left a small follow-up for #118.

## Tests (cli, platform-independent)

Each profile resolves to its exact values; flag-over-profile, env-over-profile, flag-over-env; profile-by-env with flag override; unknown profile is exit 1; default is byte-identical to today's defaults; the materialized-config line carries the resolved values + profile + schema version; presets are distinct and `balanced` equals the default set. The acceptance / golden-path serve-output parsers were widened to tolerate the new startup line.

## Local gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all green. No em/en dashes (`grep -rnP '[\x{2013}\x{2014}]'` clean).

Closes #87


ELares:
Independent adversarial review: **APPROVE**. Every profile value matches docs/CONFIG.md section 6 exactly, and `balanced` is structurally guaranteed to equal the compiled-in DEFAULT_* set (so no --profile is byte-identical to the prior behavior). Precedence is profile<env<flag with teeth (--profile edge-tiny --max-connections 64 yields 64; env beats profile; unknown profile is exit 1). The ubuntu SIGPIPE regression fix is sound and proven: the materialized-config line is now on stderr (config logging belongs on the log stream), so a consumer that closes stdout after the listen line can no longer SIGPIPE serve, with the migrate seed test (which does exactly that close) as the teeth. The Windows -D warnings stub consumes the new --profile field. No blockers. Merging; Closes #87.

Issue #33, opened 2026-06-05
[TASK] Specify the effectively-once dedup window: dual bounds, dedup-hit return, and optional persistent producer identity


ELares:
Parent: #3

Specify the opt-in effectively-once mechanism the parent #3 sketches: a producer-supplied `msg_id` deduplicated within a bounded window. This task resolves the three sub-problems the draft conflated: window sizing, the dedup-hit response (the most error-prone corner), and whether dedup survives a restart. Keying is `msg_id` only, never the body, matching JetStream and SQS FIFO.

## Approach

Dedup is OFF by default (`dedup_window = 0`); it activates per-producer when a `msg_id` is present, with a default dual bound of max 100,000 ids OR 2 minutes, evicting on whichever is hit first. The structure is a per-producer ring of `(msg_id, offset)`. A dedup hit returns `rc=0` success with the ORIGINAL offset and `duplicate=true`, never an error, so an idempotent retry over a lossy edge link does not loop. A republish outside the window creates a genuine new offset and is delivered again; consumers stay idempotent regardless. Dedup is session-scoped (in-memory) by default and lost on broker restart or reconnect. An optional stable `producer_id` + monotonic `epoch` persists the per-producer last-seen `msg_id` high-water to the WAL so dedup survives broker restart; a higher epoch fences an older zombie session. The time bound uses a monotonic clock source so an NTP step cannot mis-expire the window. Expose duplicate-hit and out-of-window-republish counters to #16; document the exact worst case (offline past the window republishes duplicates) in #14.

## Acceptance criteria

- [ ] Dedup default OFF; opt-in by presence of `msg_id`; default dual bound (100k ids AND 2 min) stated with the eviction rule.
- [ ] Dedup-hit return is specified: original offset, `duplicate=true`, `rc=0`, never an error.
- [ ] Out-of-window republish behavior is documented as a new offset; consumer idempotency requirement restated.
- [ ] Session-scoped default and optional persistent `producer_id` + `epoch` (with fencing) are both specified.
- [ ] Window time bound uses a monotonic clock; duplicate and out-of-window counters are observable.

## Failure considerations

[...]

PR #373, opened 2026-06-08
feat(dedup): opt-in effectively-once dedup window (#33)

ELares:
Implements #33: the opt-in effectively-once dedup window the parent #3 sketches. A producer opts in by attaching a `msg_id` to a publish; the broker deduplicates it within a bounded per-producer window. Default OFF (no `msg_id` = today's behavior, byte-for-byte unchanged).

## What it does

- **Dedup engine** (`ironbus-core`, pure + IO-free): `DedupRegistry` holds a per-producer ring of `(msg_id -> offset)` bounded by BOTH a count (default 100k) AND a monotonic time window (default 2 min), evicting on whichever is hit first. The monotonic clock comes through the seam, so an NTP step cannot mis-expire the window (I6).
- **Dedup-hit return**: a duplicate within the window returns the ORIGINAL offset with `duplicate = true` and `rc = 0` (a benign hit, NEVER an error), and appends NO second copy. A republish outside the window is a fresh offset.
- **Epoch fencing**: an optional stable `producer_id` + monotonic `epoch` fences a zombie session (a higher epoch supersedes an older one; a stale-epoch produce is rejected).
- **Counters**: `ironbus_dedup_hits_total` and `ironbus_dedup_out_of_window_total`, pinned in the frozen resilience taxonomy and the metric-name golden contract.

## Wire approach (frozen-safe, additive)

The FROZEN `PubAck` (tag 14, 8-byte offset) body is left untouched. A dedup hit is signaled by a NEW append-only frame `PubAckDuplicate` (tag 20, body = the 8-byte original offset), pinned in the `type_tags_have_their_exact_frozen_wire_values` test. The `msg_id`/`producer_id`/`epoch` ride an opt-in block on the `Pub` body behind a wire-only flag bit (`PUB_FLAG_HAS_DEDUP`, bit 7), masked off before the byte becomes a record flag, so a no-dedup produce is byte-for-byte the historical layout. An old client that never sends a `msg_id` never receives tag 20. The client surfaces `duplicate = true` via `produce_dedup`.

## I2 / group-commit

The actor parks BOTH fresh appends and dedup hits behind the covering group-commit fsync, so a hit on an id recorded earlier in the same uncommitted batch never replies before that offset is durable.

## Config

`--dedup-max-ids` / `--dedup-window-ms` serve flags (the non-unix `cmd_serve` stub consumes them); `DedupConfig` added to `EngineConfig`.

## Tests with teeth

A duplicate within the window returns the original offset and appends no second record (log length unchanged); an id evicted by the count OR time bound is treated as fresh; no `msg_id` = no dedup; epoch fencing rejects a stale-epoch produce; the dedup counters increment; the frozen frame-tag + metric golden tests stay green. Coverage spans core, engine, actor, session-wire, and client-end-to-end.

## Local gate

- `cargo test --workspace`: all pass (0 failures)
- No em/en dashes (`grep -rnP '[\x{2013}\x{2014}]'`): clean

## Residual

The optional persistent `producer_id` high-water in the WAL (survives restart) is deferred; the in-memory window is session-scoped (lost on restart by default), as the spec documents.


ELares:
Independent adversarial review: **APPROVE** (one RAM caveat, now FIXED). The review confirmed the frozen contracts are preserved (PubAck body unchanged, PubAckDuplicate tag 20 append-only + pinned, the opt-in Pub dedup block truly additive with the wire flag bit 7 stripped before storage, proven), dedup correct with teeth at every layer, I2/ordering/group-commit intact (Duplicate replies parked behind the covering fsync), and the taxonomy preserved. The one real caveat (the per-producer map was keyed by attacker-chosen producer_id and never reaped, an unbounded-memory DoS, with a wrong RAM_BUDGET claim) is now fixed: a max_producers cap (default 4096) with approximate-LRU eviction + empty-window reaping, a 256-byte producer_id/msg_id length cap enforced at the wire boundary, and the corrected total-memory bound documented in RAM_BUDGET.md + THREAT_MODEL T7 + RISK_REGISTER RR-24. Teeth-proven regression tests (a flood of distinct producer_ids stays <= cap; an evicted producer falls back to fresh; an active producer is not evicted). 1033 tests pass; frozen wire + taxonomy golden tests green. Merging; Closes #33 (the persistent WAL high-water remains a tracked residual under #3).

PR #377, opened 2026-06-08
test(cli): tiny-profile edge-budget CI gate (#118)

### LINKED ISSUE #118: [TASK] Edge metrics and power-loss fault-injection test contract

ELares:
Adds the last remaining residual of #118: the tiny-profile edge-budget CI gate (the edge metrics and the arming checksum-flip injector already shipped in #371; the `--profile edge-tiny` switch shipped in #87). With this gate, #118 fully closes.

## What it is

A Rust integration test (`crates/ironbus-cli/tests/edge_tiny_budget.rs`) in the existing `test` CI matrix, so no new workflow is needed. It boots the REAL `ironbus serve --profile edge-tiny --data-dir <tmp> --addr 127.0.0.1:0 --health-addr 127.0.0.1:0` (parsing the bound wire and health addresses from the startup stdout lines, the same pattern as `acceptance.rs`; the #87 materialized-config line is on stderr now and is captured separately), runs a small bounded workload (a few pubs), scrapes `GET /metrics`, and asserts the documented tiny-profile edge budget.

## What the gate proves (all CI-runnable, structural/budget properties)

1. **The tiny-profile knobs are in effect.** The #87 materialized-config startup line reports `profile=edge-tiny` plus the resolved knobs; the test asserts the `edge-tiny` preset from `docs/EDGE_CONSTRAINTS.md` section 2 (8 MiB segments, 8 / 256 KiB consumer credits, 32 connections, 64 groups, 256 in-flight, `drop-new`, 1024 checkpoint). A profile-content drift or a dropped knob fails here.
2. **The write-amplification budget holds.** `docs/EDGE_CONSTRAINTS.md` fixes the flash-endurance edge gate at `>= 4x fails the run`. After the workload, `ironbus_write_amp_ratio` is finite, non-zero, and strictly under 4.0 (measured ~1.2x, a large margin); the two raw counters advanced and `physical >= logical`.
3. **The RAM-headroom gauge is honest.** The `edge-tiny` serve path wires no runtime RAM ceiling yet (`ram_ceiling_bytes = 0`; a `--ram-ceiling-bytes` flag is a documented #115/#118 follow-up), so `ironbus_ram_headroom_bytes` reports the documented `-1` unavailable sentinel, not a misleading positive number. The test asserts exactly that.
4. **The edge series are present and bounded.** `ironbus_produce_saturated` and `ironbus_daily_write_budget_over` read `0` (no shed, governor unset by default).

## What it deliberately leaves device-only

A precise RSS-under-the-64-MiB-ceiling assertion. The CLI does not configure a runtime ceiling yet (the `-1` sentinel above is the honest consequence), and a shared CI runner's RSS is not a meaningful or stable edge signal. The RAM-burst-RSS-ceiling enforcement lands with the `--ram-ceiling-bytes` serve flag (#115/#118 follow-up) and is exercised on the reference device. The gate asserts the honest sentinel and the write-amp + knob budget that ARE meaningful in CI, so it is neither flaky nor vacuous.

## Robustness

Deterministic (fixed payloads, single producer; write-amp is exactly 1.212x every run). All timing is bounded (recv timeouts, read timeouts, bounded retries); the bound wire/health addresses are parsed, not pre-picked; `pub` retries only a connection-time race and fails fast if the broker died. Verified non-flaky over a 10x local loop (10/10 pass).

## Local gate

- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: clean (0 warnings)
- `cargo test --workspace`: green (all suites pass, including the new test)
- No `unwrap`/`expect`/`panic` in non-test paths; no `unsafe`; no em/en dashes.

Docs: CHANGELOG bullet, `docs/METRICS.md` edge-series gate note, `docs/EDGE_CONSTRAINTS.md` residual update.

Closes #118


ELares:
Reviewed (test + docs only, no production code changed): the gate is meaningful and non-vacuous. It boots the real serve --profile edge-tiny, asserts the tiny knobs are in effect (via the materialized-config line), that ironbus_write_amp_ratio is finite/non-zero and strictly under the documented >= 4x flash-endurance gate (milli-unit integer-exact), that physical >= logical and both advanced, and that ram_headroom reports the honest -1 sentinel (no runtime ceiling wired), while honestly leaving the precise on-device RSS-under-ceiling assertion as a documented device residual (the --ram-ceiling-bytes flag is a #115 follow-up). The Booted guard reaps the broker on drop. 23/23 CI green; 10/10 non-flaky locally. Merging; Closes #118.

PR #390, opened 2026-06-08
fix(cli): harden upgrade rollback against the two-rename re-entry window (#348)

### LINKED ISSUE #348: [TASK] Harden the upgrade rollback against the two-rename re-entry window (#104 follow-up)
Low-priority robustness follow-up noted during the #104 (#335) re-review. Not a blocker; the shipped contract (no double-count, a healthy binary never rolls back, N genuine failed starts roll back) holds.

## The window
`rollback_to_prev` (`crates/ironbus-cli/src/upgrade.rs`) does a two-rename atomic swap (dest -> `.prev`, then tmp/`.prev`-bytes -> dest) and then clears the failed-start counter. If power is lost in the sub-microsecond window BETWEEN the two `rename(2)` calls AND before the counter clear, the next boot's `record-start --check` can re-fire the rollback while `.prev` now holds the bad bytes, which could promote the bad binary.

This is intrinsic to any retain-prev-then-swap rollback (`scripts/install.sh` and `upgrade` share the property, documented as "leaves either the old or the new binary"). The window is two adjacent same-directory metadata ops with no I/O between them, so it is astronomically narrow, but it is a real edge in the rollback path.

## Possible hardening (pick during impl)
- Make rollback idempotent under a crash: a small durably-fsynced intent/journal record ("rollback in progress, generation G") written before the first rename and cleared after the counter reset, so a re-entered `--check` resumes deterministically rather than re-applying.
- Or a content check: refuse to promote `.prev` if its checksum matches the binary that just failed N starts (record the failing binary's hash alongside the counter).
- Add a power-cut test in the upgrade/rollback harness that interrupts between the two renames and asserts the next boot converges to a good binary.

Refs #104.

ELares:
Closes #348.

## The window (a #104 follow-up)

`rollback_to_prev` reused the generic two-rename atomic swap (retain `dest` as `ironbus.prev`, then rename the new bytes over `dest`) and then cleared the failed-start counter. A power cut in the sub-microsecond window BETWEEN the two `rename(2)`s, or after the swap but before the counter clear, could let a re-entered `record-start --check` re-fire the rollback while `ironbus.prev` momentarily held the just-failed bytes, which could promote the known-bad binary. The shipped contract (no double-count, a healthy binary never rolls back, N genuine failed starts roll back) held; this closes the narrow re-entry window.

## Mechanism (chose BOTH options the issue lists, justified)

The two options are complementary and together converge a re-entry to the GOOD binary:

1. **A rollback swap that never destroys the good `.prev`.** `rollback_to_prev` now uses `restore_prev_over_dest`, which stages `.prev`'s bytes to a sibling temp, fsyncs, then renames the temp OVER `dest` directly WITHOUT first moving `dest` onto `.prev`. The last known-good bytes survive the whole rollback, so a crash at any point leaves `dest` bad-or-good with `.prev` always good, and a re-entry repeats safely. (The destructive retain-then-swap was the actual root cause: it overwrote the good `.prev` with the bad `dest` bytes in the window.)
2. **A durable known-bad content guard.** At the fall-back cap, `record_failed_start` records the failing binary's content fingerprint (crc32c + byte length) as a known-bad guard, written with the atomic write-temp + fsync + rename + dir-fsync discipline. `rollback_to_prev` REFUSES to promote `.prev` if its fingerprint matches (typed `UpgradeError::PrevIsKnownBad`, exit 1), so even a pathological `.prev` holding the just-failed bytes can never be promoted. The guard is cleared only AFTER a completed rollback (bytes restored + counter reset) or a healthy `record-start --ok`, so a crash before the clear keeps the re-entry deterministic.

`crc32c` is already in the tree (ironbus-core record/segment checksums) and on the `deny.toml` allow-list; it is a non-cryptographic content identity, not a security boundary. `ironbus-core` stays IO-free.

## Tests + teeth

- New `a_power_cut_between_the_renames_never_promotes_the_bad_binary` (library) and `..._over_the_real_binary` (the `upgrade_migrate` real-binary harness): drive the counter to the cap (recording the guard), reproduce the EXACT mid-rollback state a power cut leaves (`dest` absent, `.prev` = bad bytes), then re-enter the rollback and assert it REFUSES to promote the known-bad bytes.
- New `a_crash_during_rollback_then_re_entry_converges_to_the_good_binary` (library + real-binary): a crash before the counter clear, re-entry completes the rollback to the GOOD binary and clears the counter + guard.
- Plus `the_hardened_rollback_never_destroys_the_good_prev`, `a_healthy_start_clears_the_known_bad_guard`, `the_guard_is_only_recorded_at_the_cap_not_below`, and a fingerprint identity test.
- **Teeth proven:** the power-cut-between-renames test FAILS on the pre-fix shape (temporarily reverting `rollback_to_prev` to the generic destructive swap, no guard, makes it return `Ok` and promote the bad bytes) and PASSES with the fix.

## Contract intact

The existing no-double-count / healthy-never-rolls-back / N-failed-starts-roll-back tests (`installer_verify.rs`, the `unit_wiring_*` and `the_fall_back_counter_*` tests in `upgrade_migrate.rs`) are unchanged and green.

## Docs

`docs/DISTRIBUTION.md` (rollback protocol), the `packaging/systemd/ironbus.service` comment, and a terse CHANGELOG entry updated. No unit change is required; the hardening lives entirely in the verbs.

## Local gate

- `cargo test --workspace`: 1075 passed, 0 failed, 1 ignored
- No em-dash/en-dash anywhere (`grep -rnP '[\x{2013}\x{2014}]'` = 0 hits)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ to follow Semantic Versioning once it reaches a tagged release.
 
 ## [Unreleased]
 
+### Fixed
+- Hardened the upgrade rollback against the two-rename re-entry window (Closes #348).
+  - `ironbus rollback` now restores `ironbus.prev` over the destination WITHOUT first moving the destination onto `.prev`, so a power cut between the renames preserves the last known-good bytes and a re-entry converges to the good binary.
+  - At the fall-back cap, `record-start --failed` records the failing binary's content fingerprint (crc32c + length, fsynced); `rollback` refuses to promote `ironbus.prev` if it matches, so a re-entry can never promote the known-bad bytes. The guard clears after a completed rollback or a healthy `--ok`.
+  - New power-cut-between-renames and crash-during-rollback tests (library + the real-binary `upgrade_migrate` harness); the existing no-double-count / healthy-never-rolls-back / N-failed-starts contract is unchanged.
+
 ### Added
 - Docs link-check CI and a slug-token issue index (Closes #30, #25).
   - Per-PR `docs-links` job runs `scripts/ci/relative-link-check.sh` (in-repo relative links + section anchors, zero network, never flaky) and the offline `scripts/ci/issue-index-audit.sh` structural pass.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -474,6 +474,7 @@ dependencies = [
 name = "ironbus-cli"
 version = "0.0.0"
 dependencies = [
+ "crc32c",
  "ctrlc",
  "ironbus-client",
  "ironbus-core",
--- a/crates/ironbus-cli/Cargo.toml
+++ b/crates/ironbus-cli/Cargo.toml
@@ -31,6 +31,13 @@ ironbus-client = { path = "../ironbus-client", version = "0.0.0" }
 ironbus-server = { path = "../ironbus-server", version = "0.0.0" }
 ironbus-storage = { path = "../ironbus-storage", version = "0.0.0" }
 
+# A content fingerprint for the upgrade rollback's known-bad guard (#348): the failed-start counter
+# records the fingerprint of the binary that failed N starts, so a rollback re-entered after a power
+# cut in the two-rename window can never PROMOTE the known-bad bytes. crc32c is already in the tree
+# (ironbus-core's record/segment checksums) and on the deny.toml allow-list (MIT OR Apache-2.0); it
+# is a non-cryptographic content identity (paired with the byte length), never a security boundary.
+crc32c = "0.6"
+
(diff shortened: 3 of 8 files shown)


ELares:
Independent adversarial review: APPROVE. The re-entry window is genuinely closed: restore_prev_over_dest never renames dest over .prev (the destructive root cause is gone), so the last-known-good bytes survive; plus a durable crc32c+length known-bad fingerprint makes rollback refuse to promote a matching .prev. Every traced power-cut interleaving converges to a good binary and never bricks the broker. Teeth proven (reverting to the destructive swap makes the power-cut test fail). The no-double-count / healthy-never-rolls-back / N-failed-starts contract stays green; no new crate. Merging; Closes #348.

PR #391, opened 2026-06-08
feat(storage): cross-platform segment preallocation primitive on the IO seam (#330)

### LINKED ISSUE #330: [TASK] Implement the cross-platform preallocate primitive on the IO seam
Follow-up from #40 (PR #329), which SPECIFIED the cross-platform preallocation shim in `docs/PREALLOCATION.md` but deliberately deferred the implementation to a separate task (M2 was a design milestone).

Implement the `preallocate` primitive on the `Filesystem`/`RandomAccessFile` IO seam per the spec:
- Add an IO-free-safe default trait method (`set_len` + `sync_all` semantics for `InMemoryFile`) and a real `StdFile` implementation with the per-OS recipes: Linux `fallocate`/`posix_fallocate`; macOS `fcntl(F_PREALLOCATE)` (F_ALLOCATECONTIG then F_ALLOCATEALL) + `ftruncate`; Windows `SetEndOfFile` (and document/guard the privileged `SetFileValidData` path). Each with the fallback ladder bottoming out at today's grow-on-append.
- Default-ON: preallocate the active segment to full roll size on creation; ENOSPC-at-preallocate is a fail-fast-at-roll event routed to the existing `StorageError::AtCapacity` / disk-full overflow path (do not surprise mid-stream).
- A `--preallocate`/config knob (per #14) to disable it.
- Tests: preallocation reserves the file to roll size; the grow-on-append fallback path; ENOSPC-at-roll degrades cleanly; the crash-recovery sweeps still pass (a preallocated zero-window tail recovers as no-records, consistent with the #45 zero-window fixture).

The design, per-OS durability notes, and rationale are in `docs/PREALLOCATION.md`. refs #40, #4, #20.

ELares:
Implements the cross-platform preallocate primitive on the IO seam, per `docs/PREALLOCATION.md` and the never-recycle decision (ADR-0002).

## What

- New `RandomAccessFile::preallocate(len)` on the IO seam, with a **no-op default** body (non-breaking, IO-free-safe). `ironbus-core` stays IO-free; the primitive lives in `ironbus-storage`.
- Production `StdFile` reserves real blocks **per OS, keeping the logical size** (so the append cursor stays the end of data):
  - **Linux**: `fallocate(fd, FALLOC_FL_KEEP_SIZE, 0, len)`
  - **macOS**: `fcntl(fd, F_PREALLOCATE)` (`F_ALLOCATECONTIG` then `F_ALLOCATEALL`), no `ftruncate`
  - **Other Unix / Windows**: degrade to the no-op default (grow-on-append)
- `InMemoryFile` tracks the reservation (`preallocated_to`) **without changing bytes or length**, so the deterministic disk image stays byte-identical; `FaultFile` delegates.
- Wired into `start_segment`: each new active segment is preallocated to `max_segment_bytes` **best-effort**. A preallocate failure is swallowed and falls back to grow-on-append, so the broker still starts and appends on any filesystem. `ENOSPC` from the keep-size call is surfaced at the seam (the route-to-`AtCapacity` refinement is a documented forward improvement).

## Why it keeps recovery correct over a zero region

The keep-size reservation does **not** advance the logical length, so recovery and the offline `dump`/`scrub` tools find the end of data from `file.len()` exactly as before. A preallocated tail is zeros, and a zero word is never a valid record magic, so the torn-tail scan truncates it (the frozen #45 zero-window rule): a freshly preallocated empty segment recovers as **no records**, a partially-written one to its **longest valid prefix**. The determinism, crash-recovery, and invariant-gate sweeps stay green.

## Unsafe / deps

The one `unsafe` is a thin per-OS libc FFI wrapper (`fallocate`/`fcntl`), each call carrying a `// SAFETY:` argument, under the workspace `unsafe_code = "warn"` + `#[allow(unsafe_code)]` convention the existing `clock.rs`/`broker.rs` libc calls use. `libc` is **reused** (already in the tree, pure-Rust binding, on the `deny.toml` pure-Rust-syscall allowlist), no new C dependency.

## Tests (with teeth)

- `preallocate_reserves_real_blocks_on_a_supporting_fs`: asserts `st_blocks` covers the request (real reservation, not a no-op).
- `preallocate_keeps_the_logical_length_and_appends_round_trip`: logical length unchanged, no zero tail past the end, appends round-trip.
- `start_segment_preallocates_the_active_segment_to_the_roll_size`: first segment and a rolled-to segment each preallocated to the roll size.
- `a_preallocated_then_crashed_empty_segment_recovers_as_no_records` and `..._recovers_its_longest_valid_prefix`: recovery correctness over the zero region.
- `a_preallocate_failure_is_non_fatal_and_the_broker_still_starts_and_appends`: a fail-prealloc fs (counts failures, proving the path is taken) still opens, rolls, and appends.

## Local gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all green (1076 passed, 0 failed). No em/en dashes. DCO-signed.

Closes #330


ELares:
Independent adversarial review: APPROVE. The keep-size deviation (no logical-length grow) is the right call and is the crux of recovery correctness: an empty preallocated segment recovers as zero records, a partial one recovers its longest valid prefix, and the offline tools see end-of-data unchanged (the frozen zero-window classification is untouched). The 3 unsafe FFI wrappers are thin, cfg-gated, and soundly justified over an owned fd with correct errno handling (EOPNOTSUPP/ENOSYS/ENOTSUP non-fatal, ENOSPC surfaced); libc was already a dep. A preallocate failure is non-fatal; I2 is unaffected. 228 storage tests + offline tests green. Merging; Closes #330.

PR #393, opened 2026-06-08
ci(nightly): pin fuzz to the host gnu target + stop tracking leaked agent worktrees

ELares:
Fixes the failing nightly run (https://github.com/ELares/IronBus/actions/workflows/nightly.yml).

## 1. fuzz soak (parsers) build failure
The job built the fuzz targets for `x86_64-unknown-linux-musl` and failed with `sanitizer is incompatible with statically linked libc` and a missing musl `std`. Root cause: `cargo-fuzz` infers its default `--target` from its OWN binary triple, and `taiki-e/install-action` now binstalls the musl-static `cargo-fuzz` build, so the fuzz build flipped to musl even though the rustc host is gnu (`host: x86_64-unknown-linux-gnu` in the log). AddressSanitizer cannot link a static libc. Fix: pin `cargo +nightly fuzz run --target x86_64-unknown-linux-gnu`, the always-installed host target, independent of how cargo-fuzz is packaged.

## 2. Leaked agent-worktree gitlinks
37 `.claude/worktrees/agent-*` directories had been committed to main as gitlinks (`160000 commit`) from a `git add -A` inside an agent worktree, causing the CI cleanup warning `No url found for submodule path '.claude/...'`. This untracks all 37 and adds `.claude/` to `.gitignore` so agent worktrees and session state can never be committed again. No source or release artifact is affected.

Note: the nightly fuzz job only runs on the nightly cron, so PR CI does not exercise it; I will trigger a manual nightly run after merge to confirm the fuzz job is green.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -66,7 +66,14 @@ jobs:
           # toolchain unambiguous and matches the documented local command in fuzz/README.md.
           for target in record_codec frame_decode cursor_snapshot segment_scan pub_body ack_body deliver_body dead_letter_body; do
             echo "::group::fuzz ${target}"
-            cargo +nightly fuzz run "${target}" -- -max_total_time="${MAX_TOTAL_TIME}"
+            # Pin the fuzz build to the rustc HOST target (gnu), not cargo-fuzz's self-detected
+            # default. cargo-fuzz infers its default --target from its OWN binary triple, and
+            # taiki-e/install-action now binstalls the musl-static cargo-fuzz build, so without
+            # this flag it builds the fuzz targets for x86_64-unknown-linux-musl. AddressSanitizer
+            # is incompatible with a statically linked libc (and the musl std for the sanitizer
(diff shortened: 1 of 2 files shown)


ELares:
Self-verified CI/infra fix (the nightly fuzz job runs only on cron, so PR CI does not exercise it, but the build/test/actionlint jobs are green confirming the gitlink removal + gitignore are safe, and actionlint validates the nightly.yml edit). Root cause is empirically confirmed in the failing log: rustc host is gnu but cargo-fuzz was binstalled as a musl-static binary, so cargo-fuzz defaulted the fuzz build --target to musl, which AddressSanitizer cannot link (static libc). Pinning --target x86_64-unknown-linux-gnu (the always-installed host) fixes it independent of how cargo-fuzz is packaged. Also untracks 37 leaked .claude/worktrees/agent-* gitlinks and gitignores .claude/ (the source of the No-url-for-submodule cleanup warnings). I will dispatch a manual nightly run after merge to confirm the fuzz job builds + runs green. Merging.

Issue #51, opened 2026-06-05
[TASK] Define the checkpoint file format, double-buffered slots, and lower-bound write protocol


ELares:
Parent: #7

Specify and implement the on-disk CHECKPOINT artifact that names the recovery point (last segment, durable byte offset, durable logical sequence) and the protocol that keeps it a conservative lower bound on durable data. This is the anchor the whole recovery path reads first, so its format, atomicity, and ordering contract must be pinned before the recovery state machine is built. Coordinates with #4 for the segment id space and with #6 for the fsync ordering the lower-bound invariant depends on.

## Approach
A dedicated fixed-size CHECKPOINT file in the log directory with two alternating slots (A/B). Each slot holds magic, format_version, last_segment_id, durable_offset, durable_seq, a monotonically increasing write_seq, and a CRC32c over the slot. Readers pick the highest write_seq whose CRC validates; a torn write to one slot always leaves the other intact. The file is created once via temp + fdatasync + rename + parent-dir fsync. Steady-state slot writes are lazy (no hot-path fsync) because a lost update only enlarges the next rescan. The load-bearing invariant: a slot naming (durable_seq, durable_offset) is written only AFTER the fsync covering those bytes has returned, per #6. Remove any clean-shutdown trust field. Cadence (segment roll, default 1000 ms timer, 4 MiB byte trigger) and the format_version are exposed as knobs in #14.

## Acceptance criteria
- [ ] Byte-exact slot layout documented with field widths, endianness, and CRC coverage.
- [ ] A/B slot selection algorithm specified, including both-slots-torn fallback to oldest live segment.
- [ ] Written statement that the checkpoint is a lower bound, persisted only after the covering fsync, with the ordering referenced to #6.
- [ ] Cadence defaults (1000 ms / 4 MiB / on roll) defined and wired to #14.
- [ ] format_version present and a forward-compat rule for unknown versions.

## Failure considerations
A power cut mid slot-write must never destroy both slots: the A/B scheme plus per-slot CRC guarantees a valid prior slot. An optimistic slot (naming un-fsynced data) is the one fatal bug, prevented by the after-fsync write rule and asserted in crash tests. Both slots torn degrades to oldest-live-segment fallback, not failure to open.

ELares:
[...]

PR #395, opened 2026-06-08
feat(server): durable per-message attempt counter so MaxDeliver survives an unclean restart (#358)

### LINKED ISSUE #358: [TASK] Durable per-message attempt counter so MaxDeliver survives an unclean restart (#60 residual)
Carved out of #60 (closed for M1: the durable-cursor architecture is the cursor-checkpoint snapshot model, ratified in CONTRACTS.md/WAL.md). One genuine correctness residual remains.

Today durable consumer state is the AckCursor checkpoint snapshot; the per-message delivery-attempt counter is in-memory and resets on restart (CONTRACTS.md: the separate consumers/<name>/ state-event log is Not implemented). Consequence: a poison message MaxDeliver count resets across an unclean restart, so on an edge device that reboots often a poison record could be redelivered past its MaxDeliver cap and never reach the DLQ.

Scope: persist the per-message attempt counter (and optionally the pending-entries list) durably so MaxDeliver/DLQ routing is a hard guarantee across an unclean restart rather than best-effort. Decide between (a) a compact attempt-count map alongside the cursor checkpoint, or (b) the drafted append-only consumers/<name>/ state-event log with snapshot+replay. Must stay within the RAM budget and the single-writer model.

Refs #60, #63, #9.

ELares:
Closes #358

## Problem

Durable consumer state was the `AckCursor` checkpoint snapshot only; the per-message delivery-attempt counter lived IN-MEMORY in the `LeaseTable` and reset on restart. So a poison record's `MaxDeliver` count reset across an unclean restart, and on an edge device that reboots often a poison record could redeliver past its `MaxDeliver` cap and NEVER reach the DLQ.

## Mechanism (option a, in its own checkpoint)

Persist each group's in-flight `{offset -> attempt}` map durably, reusing the #51 crash-safe dual-slot CRC checkpoint discipline. The 64-byte cursor slot is far too small for a `max_in_flight`-bounded map, so the attempt map gets its OWN two-slot checkpoint (`attempts.ckpt` for the default group, `attempts-<hex>.ckpt` for a named group) with a 32 KiB per-slot cap, written on the same cursor-checkpoint cadence and the clean-shutdown flush. This is option (a)'s spirit (a compact attempt-count map alongside the cursor checkpoint) without overflowing the tiny cursor slot, and lighter than the drafted append-only state-event log (b).

- **ironbus-core** (IO-free): a versioned, CRC32C attempt-snapshot codec (`attempt.rs`), plus `LeaseTable::resume_attempts` / `attempt_counts`. On the first post-restart claim of an offset, the lease resumes its delivery count at `carried + 1` instead of 1.
- **ironbus-storage**: `AttemptsCheckpoint`, a dual-slot checkpoint alias (like `CountersCheckpoint`).
- **ironbus-server**: persist/resume the map (clamped to the watermark and the durable head); named-group recovery discovers groups from BOTH cursor and attempts files, so a poison in flight but uncommitted is still rediscovered.

## The guarantee

A message delivered N-1 times that survives a crash redelivers as attempt N, so the poison reaches the DLQ after exactly `MaxDeliver` attempts TOTAL across restarts (not `2*MaxDeliver`). A clean ack clears the count. The DLQ move stays crash-atomic and idempotent (the no-duplicate-DLQ tests are unchanged).

## Additive, bounded, safe

- ADDITIVE format: a pre-#358 data directory has no attempts file and decodes as "resume at attempt 1" (the historical behavior); a torn snapshot degrades the same way and never blocks open.
- BOUNDED by `max_in_flight` per group (documented in `RAM_BUDGET.md`); the carried map is empty in steady state.
- Engineering invariants honored: no unwrap/expect/panic in non-test paths, typed errors, no unsafe, ironbus-core IO-free, frozen loss-report/taxonomy intact.

## Teeth tests

- N-1-then-crash+restart reaches the DLQ at `MaxDeliver` TOTAL (DLQ entry records attempt `MaxDeliver + 1`).
- A clean ack clears the count across a restart.
- An old snapshot (no attempt counts) decodes as attempts = 0.
- A torn attempts snapshot falls back to no carried counts.
- The named-group path survives a restart and reaches its own DLQ.
- The on-disk codec round-trips and detects every single-bit flip (proptest).

## Local gate

- `cargo test --workspace`: 1107 passed, 0 failed, 1 ignored

No em-dash/en-dash anywhere.


ELares:
Independent review APPROVE (guarantee wording made precise). The durable counter resumes each redelivery near its true attempt across an unclean restart; teeth proven (no-op the resume and the poison reaches the DLQ at 2x or never). A crash between checkpoints replays only the un-checkpointed tail, never regressing below the durable floor, so the honest guarantee is at-least MaxDeliver (no longer unbounded across reboots), now stated precisely in the docs. attempts.ckpt is dual-slot CRC with a safe torn-fallback, the snapshot is additive, union-discovery is sound (no orphan resurrection/double-count), memory is bounded by max_in_flight, and no-duplicate-DLQ stays crash-atomic. Merging.

PR #396, opened 2026-06-08
bench: WAL-append, CRC32C hw-vs-portable, and zero-copy handoff micro-benches (#386)

### LINKED ISSUE #386: [TASK] WAL-append, CRC32C hw-vs-portable, and ARM-core micro-benches (#112 residual)
Carved out of #112 (closed for M0: the criterion micro-bench suite shipped for the IO-free hot paths, PR #273: record encode/decode/decoded_len at 64B/4KiB and the AckCursor acks + snapshot, criterion dev-dep, harness=false targets). Four bench additions remain.

Scope: a WAL-append bench at representative batch sizes (1B/16KiB/1MiB) in the storage crate (which has no bench today); CRC32C benched with the hardware path and the portable fallback measured separately; the zero-copy bytes::BytesMut freeze-to-Bytes O(1) slice-handoff bench; and running the suite on at least one ARM reference core with the micro results archived in the shared versioned format.

Refs #112, #111, #114.

ELares:
Closes #386

Finishes the #112 micro-bench suite with four criterion micro-benches (the four #112 residuals). All are dev-only, `harness = false`, deterministic, architecture-neutral, and run on demand; they are NOT a per-PR CI gate (the regression gate is #114).

### The four benches
1. **WAL-append** (`crates/ironbus-storage/benches/wal_append.rs`, the storage crate's first bench): `Log::append` at 1 B / 16 KiB / 1 MiB over the in-memory filesystem (`InMemoryFs`). It measures the IO-light framing + CRC32C + buffer cost and deliberately does NOT call `sync()`, so it never charges a device-dependent `fdatasync`/`fsync` (disk-flush latency is not a stable micro-benchmark; the macro rig #111/#114 owns end-to-end durable throughput on a reference device). Adds a `criterion` dev-dependency and the `[[bench]]` table to `ironbus-storage`.
2. **CRC32C hardware-vs-portable** (`crates/ironbus-core/benches/crc32c_hw_vs_portable.rs`): the realized `crc32c::crc32c` path (hardware-accelerated when the host CPU has SSE4.2 / ARMv8-CRC, the crate's own software fallback otherwise) vs an in-bench portable table-driven reference (the reflected Castagnoli polynomial). The `crc32c` crate auto-selects at runtime and exposes NO public portable-only entry point (its `sw` module is private), so the bench is honest that "realized" means hardware-when-available, and asserts the portable reference matches the crate before timing (apples-to-apples). Sizes 64 B / 4 KiB / 64 KiB.
3. **Zero-copy handoff** (`crates/ironbus-core/benches/zero_copy_handoff.rs`): the issue scopes this as the `bytes::BytesMut` freeze-to-`Bytes` handoff, but `bytes` is NOT a dependency and is not added just for a bench. So it benches the equivalent handoff IronBus actually uses: an `Arc<[u8]>` freeze (the `BytesMut::freeze` analogue) plus the O(1) refcounted slice handoff (the `Bytes::slice` analogue), contrasted against an O(n) `to_vec` copy.
4. **ARM reference core**: all three benches are architecture-neutral source (no x86-only intrinsics), so `cargo bench -p ironbus-storage` / `-p ironbus-core` runs on an aarch64 device. The committed numbers are x86; the on-device ARM run is a documented device residual (it requires ARM hardware and is not faked here).

### Invariants
- `criterion` stays dev-only: it is not in the shipped (normal) dependency graph; `Cargo.lock` adds only the `ironbus-storage -> criterion` dev-edge (package count unchanged, 155 -> 155). The release binary is byte-for-byte unaffected.
- `ironbus-core` stays IO-free (the WAL bench lives in `ironbus-storage`); no `unsafe`; no `unwrap`/`expect`/`panic` in production code (only in bench code).

### Local gate (all green)
- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: clean (benches are an `--all-targets` target).
- `cargo bench --workspace --no-run`: all four new benches compile.
- `cargo test --workspace`: all green.
- Sample numbers (x86, short run): `crc32c/realized/4096` ~17.4 GiB/s vs `crc32c/portable/4096` ~620 MiB/s (~28x hardware speedup); `wal_append/1048576` ~3.6 GiB/s framing throughput in-memory.
- No em-dash / en-dash anywhere (`grep -rnP '[\x{2013}\x{2014}]'` clean).


ELares:
Independent review APPROVE. criterion stays dev-only (cargo tree --edges normal has no criterion; the lock package count is 155 before and after, zero new shipped deps); all four benches compile and run; the CRC32C portable reference is honest (re-derived the reflected Castagnoli 0x82F63B78, asserts equal to the crate before timing); the WAL bench never calls sync; bytes was not added; the benches are arch-neutral with the ARM run honestly a device residual. Merging.

PR #397, opened 2026-06-08
test(storage): seeded fault scheduler + same-seed determinism gate + seed sweep (#384)

### LINKED ISSUE #384: [TASK] Seeded fault scheduler and same-seed determinism gate (#119 residual)
Carved out of #119 (closed for M0: the deterministic-simulation SUBSTRATE shipped, the Clock seam ManualClock, the IO seam Filesystem/RandomAccessFile + InMemoryFs/FaultFs, the fault model FaultControl, and the byte-identical-disk-image determinism meta-tests in tests/determinism.rs + the recovery-is-pure-function tests). The seeded-scheduler half is the residual (overlaps the closed #151).

Scope: a single seeded ChaCha8Rng (or equivalent pure-Rust PRNG) driving every fault decision with a replayable printed seed; a Scheduler seam for ordering/interleaving; a same-seed-twice determinism gate comparing the event trace + the recovered-log hash (not just the disk image); a cfg-guarded sim-mode lint deny-list (reject now()/real-thread-spawn/unordered-map iteration in sim mode); and a per-PR fixed-256-seed fault sweep under a 5-minute cap as a hard gate (today the deep sweep is nightly + proptest-case-count).

Refs #119, #151, #21.

ELares:
Closes #384

Carves the SEEDED half out of the #119 deterministic-simulation residual. The existing arming model (`fault.rs`) and the byte-identical-disk-image determinism gates stay; this adds the seed-driven companion. Test/sim infrastructure only, production code unaffected.

## What

- **In-tree PRNG, no new dependency.** `ironbus_storage::sim::SplitMix64`: a ~15-line, fully in-tree, deterministic, cross-platform-reproducible PRNG (pure wrapping `u64` arithmetic, pinned against the published reference vector). No `rand`/`rand_chacha`.
- **Seeded fault scheduler.** `FaultSchedule` drives every fault decision (which op class faults, and which fault) from one `u64` seed, recording a replayable fault/op event trace. `FaultControl::seeded_schedule(seed, probs)` is the additive seeded constructor; the arming model is unchanged and the boundary-pinning tests keep using it.
- **Same-seed determinism gate** (`tests/seeded_faults.rs`): one workload under one seed run twice yields an identical event TRACE and an identical recovered-log HASH (stronger than the disk-image gate in `determinism.rs`), and a DIFFERENT seed varies the schedule (asserted, so the gate is not vacuous).
- **Per-PR fixed-256-seed recovery sweep**: drives faults through the crash + recovery path asserting I1 to I4 hold, under a few-second budget, printing the seed of any failing case for replay. Deterministic (a second meta-test reruns the sweep and asserts seed-for-seed agreement), not a flaky cron.
- **Recovery-side arm**: a fault injected DURING recovery must hold the invariants on success or fail closed with a typed error, never panic.

## Scope

In: the seeded FAULT schedule, the same-seed gate, the seed sweep (the highest-value parts).
Deferred, still open (noted in `RECOVERY.md` / `INVARIANTS.md`): the broader async-scheduler seam that also seeds ORDERING/interleaving, and the cfg-guarded sim-mode lint deny-list (#119, #151).

## Invariants / gate

- Existing determinism + crash-recovery tests stay green (production behavior unchanged).
- `ironbus-core` stays IO-free; no `unsafe`; no `unwrap`/`expect`/`panic` in production paths (the seeded PRNG + scheduler live in the storage test/sim seam).
- Local gate green: `cargo fmt --all -- --check`; `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`; `cargo test --workspace`.


ELares:
Independent review APPROVE. The SplitMix64 stream was independently recomputed and matches the pinned seed-0 reference vector; the same-seed gate has teeth (a perturbed second run fails) and the fault-schedule-depends-on-seed property is guarded; the 256-seed sweep genuinely asserts I1 to I4 and prints the seed on failure; 5 runs, no flakes, ~0.02s. No new dependency, production paths untouched, the arming model additive-only. Merging.

PR #400, opened 2026-06-08
feat(server): relaxed durability levels behind a strict opt-in (#341, #379)

### LINKED ISSUE #341: [TASK] Implement the relaxed durability levels (interval / async) behind an explicit opt-in
Follow-up from #50 (PR #338, `docs/DURABILITY.md`). The durability CONTRACT and levels MODEL are specified; the relaxed levels are designed but NOT implemented (M2 was a design milestone). v1 ships only the safe `sync` level (ack-implies-durable, I2).

Implement the relaxed durability levels per `docs/DURABILITY.md` (owned by #6):
- `interval`: ack before the periodic fsync, bounded loss = `min(flush_interval_ms, flush_max_bytes)`; a #14 config knob, off by default.
- `async`/`none`: ack immediately, unbounded loss until the next fsync; behind an EXPLICIT data-loss acknowledgement, off by default, flagged as contrary to the safe default.
- Per-level worst-case-loss ASSERTIONS in the crash harness (the #55-extended harness can drive them once the levels exist).
- Wire the per-level `durability_level` config (#14) and the relevant metrics (#16).

The safe `sync` level stays the default. refs #50, #6, #14, #16.

### LINKED ISSUE #379: [TASK] Relaxed durability levels (none/interval) plus their observability and the none safety gate (#49 impl)
Carved out of #49 (closed for M1: the durability-observability DESIGN is complete in docs/DURABILITY.md, and the fsync-latency histogram ships). The implementation is blocked because v1 ships exactly ONE durable level (sync, ack-implies-durable) by deliberate design; none/interval are SPECIFIED-not-implemented (DURABILITY.md section 3), so the level-labeled metrics and the none-gate have nothing to gate yet.

Implement the relaxed durability levels (interval flush, none/async) behind an explicit opt-in, and with them the #49 observability: the durability_level metric label, the sticky power_loss_unsafe gauge, syncs_per_record / batch_fill / commit_queue_depth / oldest_unsynced_age, and the i_accept_power_loss_data_loss (async_loss_ack) fail-closed startup gate that refuses to relax durability without the explicit acknowledgement. Monotonic-clock sourced.

Refs #49, #6, #14.

ELares:
Closes #341
Closes #379

Implements the relaxed durability levels (`interval`/`async`/`none`) as a STRICTLY OPT-IN relaxation of I2. The default stays `sync`, so an operator who changes nothing keeps ZERO acknowledged loss on a power cut (the historical durable broker, byte-for-byte).

## Per-level ack semantics and the exact worst-case-loss bound
- `sync` (default): ack ONLY after the covering `fdatasync` returns. I2 holds. Worst-case acked loss on a power cut: ZERO.
- `interval`: ack after the page-cache `write()`; a background window forces an `fdatasync` every `flush_interval_ms` of MONOTONIC time (clock seam, so an NTP step never mis-fires it) OR after `flush_max_bytes` of unsynced record bytes, whichever first. Worst-case loss: BOUNDED by the SMALLER of the two triggers (at most one flush window of acked records).
- `async`: ack after the page-cache `write()`; fsync only opportunistically (a segment roll's seal or a clean shutdown). Worst-case loss: UNBOUNDED until the next such barrier (only the OS dirty-writeback window flushes the page cache).
- `none`: like `async` with NO periodic fsync; the only barriers are a segment roll and a clean shutdown. The LARGEST loss window.

## The safe default (the teeth test)
`commit_batch` is level-aware: under `sync` it issues the covering `fdatasync` before the actor releases the ack (I2). The engine durability suite proves a default broker loses NOTHING on a power cut (`the_default_level_is_sync_and_acks_only_post_fsync_zero_acked_loss`, `async_loses_the_unsynced_tail_on_a_power_cut_but_sync_loses_nothing_same_harness` runs sync and async over the SAME harness: sync recovers all, async loses the unsynced tail).

## The none/async safety gate
`async`/`none` REFUSE to boot without `--async-loss-ack` (fail-closed usage error, exit 1, naming the level, the waived I2 invariant, the worst-case loss, and the flag). `interval` with both flush triggers at 0 is also refused (it would silently degrade to `async`). When any relaxed level is active the broker logs a LOUD startup WARN that I2 is WAIVED, that it is NOT power-loss safe, and the worst-case loss for the active level.

## Observability (additive, frozen taxonomy re-pinned)
New GAUGES `ironbus_durability_level_info{level=...}`, `ironbus_durability_power_loss_unsafe`, `ironbus_durability_unsynced_bytes`. `FROZEN_METRIC_TYPES` extended and re-pinned; the resilience-counter taxonomy (`FROZEN_RESILIENCE_COUNTERS`) is untouched (a relaxed level is an opt-in, not a shed). The level + `power_loss_safe` are also in the materialized-config startup line. Both frozen golden tests stay green.

## I2 conditioned on sync
`INVARIANTS.md` now states I2 (ack-implies-durable) holds ONLY under `durability_level=sync`; the relaxed levels are the explicitly-labeled, off-by-default exception, each carrying its own weaker documented bound. The crash tests assert the documented bound per level, not zero loss.

## Engineering invariants
No new `unsafe`; no `unwrap`/`expect`/`panic` on the non-test paths; the new `ServeConfig` fields are consumed in the `#[cfg(not(unix))]` `cmd_serve` stub (the Windows `-D warnings` footgun). Local gate green: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, `cargo test --workspace` (1146 passed, 0 failed). No en/em dashes.


ELares:
Independent adversarial review APPROVE, no blockers, on the most safety-critical change of the set. The cardinal rule holds: the default is sync and acks only after fdatasync (zero acked loss, teeth-proven by reverting to ack-before-sync and watching the zero-loss + DLQ + recovery tests fail). Each level matches its documented bound (sync zero; interval <= the time/byte window, proven by a same-harness contrast; async/none = the last barrier), the storage layer separates the visible head from the durable synced head so the loss is real and bounded, async/none refuse to boot without --async-loss-ack and loudly warn (teeth-proven), I2 is conditioned on sync in INVARIANTS.md, the non-unix cmd_serve stub consumes all four new fields, and the three new durability gauges are additive + re-pinned in the frozen taxonomy. Merging.

Issue #336, opened 2026-06-08
[TASK] Implement the backpressure controls (CoDel shedding + retry budget + egress AIMD)


ELares:
Follow-up from #68/#69 (PR #333), which SPECIFIED the backpressure controls in `docs/BACKPRESSURE.md`. Implementation is future work (M2 was a design milestone; today IronBus ships only drop-new/drop-oldest overflow + per-consumer credit).

Implement, per `docs/BACKPRESSURE.md`:
- CoDel time-in-queue (sojourn) shedding (#68): monotonic-clock sojourn at enqueue, TARGET 5ms / INTERVAL 100ms (clamped #14 keys), INTERVAL/sqrt(count) drop spacing, suspend-reset, and the depth/byte backstop independent of sojourn; shed-count + sojourn-estimate metrics (#16).
- The per-client retry budget (10% / 60s), the fire-and-forget token bucket, and the egress AIMD limiter (#69); shed/retry-ratio/egress-limit metrics (#16).
- BLOCKED on a #11 frozen-protocol extension for the `retry_after_ms` (u32, sentinel 0xFFFFFFFF = do-not-retry) + `shed` nak fields. The wire encoding is owned by #11; the retry-budget broker-side re-check and the do-not-retry signaling cannot fully land until that wire extension does.

Design, config keys, metrics, and the proofs-of-intent are in `docs/BACKPRESSURE.md`. refs #68, #69, #10, #11, #16.

ELares:
Implemented in PR #401 (`feat/backpressure-controls-336`).

- **CoDel sojourn shedding** (#68): pure RFC 8289 controller in `ironbus-core::backpressure` (sojourn min-tracking, dropping state, INTERVAL/sqrt(count) spacing, suspend-reset), wired into the produce-admission path. Sheds a NEW produce before the append (no data loss, I2 holds) with a typed `Err` reply; depth/byte backstop is sojourn-independent. Clamped TARGET [1ms,1s] / INTERVAL [20ms,10s]; OFF by default.
- **Retry budget** (#69): Google-SRE accept-based throttle (10%/60s), broker-side re-check.
- **Egress AIMD + fire-and-forget token bucket** (#69): +1/x0.5 bounded [4,128] behind an estimator seam; the bucket caps the un-credited tier.
[...]

PR #401, opened 2026-06-08
feat(backpressure): CoDel admission shedding, retry budget, egress AIMD (#336)

ELares:
Implements the backpressure framework for #336. The CoDel produce-admission shedding is wired live (sheds a new produce before the append, never losing an accepted record, typed self-announcing Err). The retry budget, fire-and-forget bucket, and egress AIMD are implemented + unit-tested with their seams in place; their live enforcement call sites land with the redelivery/egress wiring, tracked in #402.

refs #336
refs #402


ELares:
Independent adversarial review APPROVE on data-safety (shed-before-append proven with teeth, inert defaults, Clock seam not real-clock, existing credit-flow + durability intact, frozen taxonomy re-pinned, all 8 ServeConfig fields in the non-unix stub). The review flagged a scope/honesty point: only CoDel is wired live; the retry-throttle, fire-and-forget bucket, and egress AIMD are tested seams without production call sites. Addressed by correcting the BACKPRESSURE.md enforcement wording, filing #402 for the live wiring, and merging as refs #336 (not Closes) so #336 stays open until the controls act live. Merging the live CoDel framework.

Issue #357, opened 2026-06-08
[TASK] Implement the trained dictionary lifecycle (ZDICT training, sidecar IO, embed, UnresolvedDictId, measured ratio)


ELares:
Parent: #12. Implements the design specified in docs/DICTIONARY_LIFECYCLE.md (#78, #353). All decisions are frozen in that doc; this is the build, behind the opt-in zstd feature (ADR-0003: no vendored C on the default path).

Scope:
1. ZDICT training. Add the opt-in zstd dependency and wire `ironbus dict train` to call ZDICT_trainFromBuffer over the per-type sample corpus, enforcing the MIN_SAMPLES / TARGET_SAMPLES / MIN_DISTINCT_BYTES floors. Emit dicts/<dict_id>.zstd and the --json summary. Derive dict_id as the u32 (first 4 bytes, LE) truncation of BLAKE3-256 of the dictionary bytes, with the train-time collision check (byte-identical = no-op, differing = fail-closed refuse) and the dict_id=0 refusal (the no-dictionary sentinel). BLAKE3 is a new crate: add a deny.toml justification (CC0-1.0 OR Apache-2.0 via the Apache-2.0 arm).
2. Sidecar IO. On write, persist each referenced dictionary as dicts/<dict_id>.zstd (content-named; fsync the blob then the dicts/ dir before the referencing segment is acked). On read, resolve sidecar-first then embedded, re-deriving the content hash to validate the blob; a failed hash is treated as absent.
3. Embed. build.rs + include_bytes! to compile the active dictionary set (dicts/active/<dict_id>.zstd) into the binary keyed by dict_id (#17); retired dicts not embedded.
4. UnresolvedDictId. Append ReasonCode::UnresolvedDictId (code 7, label unresolved_dict_id, is_data_loss() == true) to crates/ironbus-storage/src/loss.rs, extend golden_reason_code_vocabulary_is_frozen and reason_codes_are_stable_and_distinct, and route the decompress-time unresolved-dict path through the #8 quarantine flow. Codes 1-6 unchanged; schema_version stays 1.
5. Measured ratio. Run the two-arm ironbus bench (no-dict dict_id=0 vs trained dict_id=D, same per-batch unit, same level, same real per-type corpus) on the reference device under the #113 run discipline; record the realized before/after ratio and wire it into the #114 matched-workload rig as a tracked number.

[...]

Issue #76, opened 2026-06-05
[TASK] Decoder resilience: bounded decompression, unknown-codec and corrupt-unit quarantine


ELares:
Parent: #12

Define and implement what a reader does when it cannot honor a self-describing frame: a corrupt unit, an unknown codec id, or a frame that would decompress to an unsafe size. The contract is uniform: never crash, never silently drop, always hand the unit to #8 as bounded, reported loss. This closes the offline-edge data-loss gap the first draft left open and protects the low-RAM device from decompression bombs (#18).

## Approach
Before decompressing, validate the descriptor. If codec is not a compiled-in id (not 0/1/2), the unit is poison: route to #8, report, advance. Validate uncompressed_len against a configurable per-unit decompressed cap (default 8 MiB, must be >= the batch byte cap) BEFORE allocating, and stream-decompress into a fixed bounded output buffer rather than trusting the header length; abort and quarantine the instant output would exceed the cap. Any zstd decode error is likewise a quarantined poison unit, not a panic. All three paths (unknown codec, over-cap length, decode error) converge on the same #8 entry point so there is one quarantine story. This sub-issue depends on the frame/descriptor sub-issue and the #8 interface.

## Acceptance criteria
- [ ] A unit with an unknown codec byte is quarantined via #8, reported, and the reader advances without crashing.
- [ ] A frame declaring uncompressed_len > the cap is rejected before allocation.
- [ ] A frame that expands past the cap mid-decode aborts into quarantine with a fixed-size buffer (no header-driven allocation).
- [ ] An arbitrary corrupt zstd frame never panics the reader; it is quarantined.
- [ ] The decompressed-size cap is configurable via #14 with an 8 MiB default and a floor of the batch byte cap.
- [ ] A fuzz target over random/corrupted frames produces zero panics and zero unbounded allocations.

## Failure considerations
The core risk is an OOM-kill on edge from a trusted length; the before-allocation validation plus fixed buffer is the mitigation, asserted by a test that feeds a tiny frame declaring a huge length. A second risk is a panic path turning one bad unit into a consumer crash loop; the fuzz target is the gate. A third is divergent handling across the three error classes; routing all to one #8 entry point keeps behavior uniform and testable.

ELares:
[...]

PR #403, opened 2026-06-08
feat(compress): lz4_flex codec runtime with decoder resilience (#387)

### LINKED ISSUE #387: [TASK] Compression codec runtime: lz4_flex default + opt-in zstd, decoder resilience, musl verification (#75/#76/#77 impl)
Carved out of #75, #76, #77 (closed for M0: the codec STRATEGY is decided and frozen in ADR-0003, lz4_flex pure-Rust default + zstd opt-in behind a feature, never on the default path; the descriptor + the unknown-codec POISON classification are specified in CONTRACTS.md / compat/versions.md; docs/DICTIONARY_LIFECYCLE.md). The runtime is deliberately deferred per ADR-0003 and tracked here.

Scope (per ADR-0003, the title's zstd-default is historical): #75 the per-batch compress/decompress runtime, the self-describing codec descriptor (codec/level/dict_id=0/uncompressed_len), the 256 KiB + linger batch bound, the 64-byte / never-expand raw-store guard, CRC32C over compressed-as-stored bytes, lz4_flex default + zstd behind an opt-in feature (a deny.toml C-FFI allowlist decision #102). #76 decoder resilience: the per-unit decompressed cap (8 MiB default, cap-before-alloc, stream into a bounded buffer, mid-decode abort), corrupt-zstd-never-panics, and converge unknown-codec / over-cap / decode-error on the #8 quarantine entry point, with a corrupted-compressed-frame fuzz target. #77 static-musl verification: vendored zstd static-links on aarch64/armv7 musl, zstd + lz4_flex round-trip from the static binary, and a gating fail-loudly-if-zstd-fails-to-link CI job. The musl matrix #100 already exists; this is the codec-specific verification.

Refs #75, #76, #77, #12, #8, #102, #17.

ELares:
Implements #387: the on-disk compression codec runtime per ADR-0003. `lz4_flex` (pure-Rust) is the default codec; `zstd` stays a deferred opt-in. Touches the FROZEN on-disk record format additively, so `FORMAT_VERSION` stays 1 and the format-registry digest is unchanged.

## The codec descriptor (additive, byte-identical for uncompressed records)

When the already-RESERVED `COMPRESSED` flag bit is set, the record's payload is a self-describing compressed object: a fixed 9-byte descriptor (`codec_id` u8, `dict_id` u32 LE, `uncompressed_len` u32 LE) then the codec stream. The descriptor lives INSIDE the payload, inside the CRC-covered body, so it consumes no new header bytes and shifts no field. No `pub const` layout declaration in `format.rs` changes, so `scripts/check-format-registry.sh` stays green (digest `59773de...` unchanged) at `FORMAT_VERSION = 1`.

A record stored RAW (sub-64-byte, never-expand fallback, or `--compression none`) keeps the `COMPRESSED` bit CLEAR and writes no descriptor, so it is BYTE-FOR-BYTE the pre-compression frame. An integration test proves a raw-stored record's bytes equal the no-compression encoder's bytes, and the conformance corpus passes unchanged: every existing record and vector reads identically (backward compat).

## Write guards

- Raw-store threshold: a payload below ~64 bytes is stored raw (descriptor overhead + near-zero ratio on a tiny buffer makes compression a net loss).
- Never-expand: if descriptor + stream is not STRICTLY smaller than the raw payload, store raw, so compression never expands a record.

## CRC-before-decompress ordering

`body_crc` (and the optional xxh3-64) cover the STORED, post-compression bytes, so the existing verify-body-crc-before-anything path is unchanged: `codec::decode` verifies the CRC over the stored bytes FIRST, and only those verified bytes are handed to `decompress_payload`. A test flips a compressed-body byte and asserts `decode` returns `BadBodyCrc` before any decompress runs.

## Decoder resilience (#76)

- A per-unit DECOMPRESSED-size cap (8 MiB default, derived from the 16 MiB max-record bound) is checked against the descriptor's claimed `uncompressed_len` BEFORE allocation, so a decompression bomb cannot drive an unbounded allocation. lz4 then decompresses into a buffer sized to exactly the claimed length.
- A corrupt lz4 stream returns a typed `DecompressError`, never a panic (proptest + a `compressed_unit` fuzz target sweep arbitrary bytes).
- An UNKNOWN codec id (a future `zstd` = 2, or garbage) or an unresolved `dict_id` is classified POISON, routed to the #8 quarantine via `ReasonCode::for_decompress_error`, never a crash.

Teeth tests: compressed round-trip; uncompressed byte-identity; below-threshold / would-expand stays raw; corrupt unit -> typed error (no panic); decompression bomb rejected by the cap before allocation; unknown codec -> POISON; CRC verified before decompress.

## #357 assessment (partial: lifecycle seam lands, ZDICT deferred)

The lifecycle SEAM is implemented: the descriptor's `dict_id` field, the `DictResolver` sidecar seam (`NoDictionaries` default resolver), and `ReasonCode::UnresolvedDictId` (code 7, `unresolved_dict_id`, data-loss) appended append-only with no `schema_version` bump. An unresolved `dict_id` is POISON, not a crash. The ZDICT TRAINING, sidecar IO, embed, and measured ratio are zstd-specific and stay DEFERRED behind the opt-in zstd feature (ADR-0003). So #357 does NOT fully close: this PR uses `refs #357`, not `Closes`.

## Deps / supply chain

`lz4_flex` 0.10 (MIT, already on the license allow-list; pinned to 0.10 for the 1.78 MSRV since 0.11+ require 1.81) with the `safe-encode` / `safe-decode` features: no `unsafe`, no `links` key, no C-builder. `deny.toml` notes it; `scripts/ci/cffi-ban.sh` stays clean (42 shipped crates, none C-FFI); the embedded SBOM and `Cargo.lock` both reflect it so the SBOM-diff gate stays consistent.

## serve

New `--compression <none|lz4>` (and `IRONBUS_COMPRESSION`), default `lz4` per ADR-0003, rendered in the materialized-config line; `zstd` is rejected on the default build. The non-unix `cmd_serve` stub consumes the new `ServeConfig` field.

## Local gate (all green)

- `cargo test --workspace`: all pass (incl. the new compress unit/integration/proptest legs and the unchanged conformance corpus)
- `scripts/check-format-registry.sh`: digest unchanged (FORMAT_VERSION stays 1)
- `scripts/ci/cffi-ban.sh`: clean
- no em/en dashes (`grep -rnP '[\x{2013}\x{2014}]'`)

## Docs

CONTRACTS.md (descriptor implemented), compat/versions.md (codec id `lz4` + dict_id POISON implemented), COMPATIBILITY.md, METRICS.md + loss-report.v1.md (code 7), ADR-0003 (status), DICTIONARY_LIFECYCLE.md (seam), CHANGELOG (terse bullet).

Closes #387
refs #357


ELares:
Independent adversarial review APPROVE after one fix. The review caught a CRITICAL latent remote-DoS: lz4_flex 0.10 with safe-decode but without checked-decode panics (out-of-bounds index) on a stream decoding past the claim-sized buffer, the claim and stream being independently attacker-controlled. Fixed by enabling checked-decode (MSRV-safe) + a regression test pairing a valid 4096-decoding block with a claimed len of 16 (the gap the single-blob proptest/fuzz could not reach), verified to return CorruptStream not panic. The rest is sound: an uncompressed record is byte-identical (FORMAT_VERSION 1, format-registry digest unchanged, conformance green), CRC is verified before decompress, the bomb cap rejects before alloc, unknown-codec/unresolved-dict are POISON, ReasonCode 7 is additive, and the RUSTSEC-2026-0041 ignore is verified sound and scoped (safe-decode on, fresh zeroed buffer per call). Merging.

Issue #380, opened 2026-06-08
[TASK] Config-reload engine and runtime CONFIG admin verbs (#88 impl)


ELares:
Carved out of #88 (closed for M1: the config-system DESIGN is complete in docs/CONFIG.md, the 5-layer precedence + the hot/cold/coupled knob table + coupled-set validation). The implementation is blocked on two deferred dependencies:
1. The config-FILE parser + precedence resolver (#85, M0): there is no config file to reload until the TOML parser + the effective-config resolver land (a deny.toml dependency decision).
2. The mutating-admin surface: CONFIG SET/RELOAD/SAVE mutate runtime state, and the shipped /admin is deliberately READ-ONLY because exposing unauthenticated config mutation is unsafe until the connection-scoped AUTH impl lands (the #106 spec is done; the impl is deferred).

Implement, once those land: the hot/cold/coupled reload engine (validate the whole effective config, then one atomic swap; reject a cold-key change atomically), file reload, and the CONFIG GET/SET/LIST/DIFF/RELOAD/SAVE verbs (read-only GET/LIST may surface on /admin; mutating SET/SAVE behind auth). SAVE is atomic temp+fsync+rename+dir-fsync, comment-preserving.

Refs #88, #85, #106, #14.

ELares:
PR #404 (Closes #382) delivers the AUTH-FREE part of this issue: the immutable `Arc<EffectiveConfig>` read via a single atomic pointer load, the cold/hot key classification, and the atomic re-read RELOAD (validate the whole config, reject a cold-key change atomically, swap only on full success; a broken reload keeps the old config). The read-only materialized-config surface is the read half.

Still OPEN here (deferred behind auth): the MUTATING wire `CONFIG SET`/`SAVE`/`DIFF` admin verbs, which change runtime state and need the #106 connection-scoped auth (no unauthenticated remote config mutation surface). The crash-safe comment-preserving SAVE (temp + fsync + rename + dir-fsync) ties to that surface. Also a follow-up: wiring the reload to a live SIGHUP without regressing the #195 graceful-stop binding (ctrlc's `termination` feature currently catches SIGHUP for shutdown), so the engine ships with a re-read trigger here.

ELares:
## Assessment of the residual, with one finding that changes the SIGHUP call

Re-reviewing this with fresh eyes against the merged tree. Status of the named parts:

[...]

PR #404, opened 2026-06-08
feat(config): TOML config-file layer, literal grammar, strict validation, atomic reload (#382)

### LINKED ISSUE #382: [TASK] Config-file (TOML) layer, parser, coupled-set validator, and atomic reload (#85/#86 residual)
Carved out of #85 and #86 (closed for M0: the flag>env>default precedence RESOLVER and the typed knob table with per-flag range validation shipped, #308; the five-layer model + the literal grammar + the coupled-set rules are specified in docs/CONFIG.md). The file-backed half is blocked on a dependency decision and tracked here.

Scope: add the TOML config-FILE layer (a toml parser dependency is a deny.toml decision), the immutable Arc<Config> effective-config with a single atomic pointer swap, the shared literal-grammar parser (duration {ms,s,m,h,d} / binary byte-size {B,KiB..TiB}, unit-required, decimal-SI rejected, overflow-safe), the unknown-key reject-with-did-you-mean + --allow-unknown-config, the whole-config coupled-set validator (each subsystem's pure validator), and CONFIG RELOAD that validates the whole effective config then swaps only on full success (cold-key change rejected atomically). Reload trigger + the mutating CONFIG verbs tie to #380 (and need auth).

Refs #85, #86, #380, #14.

ELares:
Closes #382
refs #380

Implements the config-FILE layer and the auth-free atomic reload.

## The file layer (flag > env > FILE > default)
- `serve --config <path>` loads a TOML file. The file slots BETWEEN env and default, so the precedence becomes **flag > env > FILE > default**. A flag beats env beats the file beats the compiled default. With NO `--config`, resolution is byte-for-byte the historical flag > env > default (the no-config default path is unchanged). The FILE layer composes the env seam (env-then-file), so the existing resolvers gain the file with no change to their relative order.
- Pure-Rust `toml` dependency (maintainer-approved). `deny.toml` documents it; its licenses are already on the allow-list; `cffi-ban.sh` and the SBOM round-trip stay clean (toml is pure Rust, no `links`/C-builder).

## The literal-grammar parser (IO-free core)
- `ironbus-core::config`: durations `{ms,s,m,h,d}` and binary byte sizes `{B,KiB,MiB,GiB,TiB}`, unit-REQUIRED, decimal-SI REJECTED, overflow-checked against u64 (32/64-bit safe). Shared by the file and (available to) the flags.

## Strict validation
- An unknown key is REJECTED with a Levenshtein did-you-mean; `--allow-unknown-config` downgrades it to a warning. The typed key table checks type/range. The coupled-set validator runs the cross-key rules as a UNIT (segment-fit when a max-record cap is configured, the durability none/async gate + interval-trigger, retention-requested-but-all-off, the drop-oldest no-cap warning). A broken file fails fatally with the path + line/column. A bad config is a clean usage exit (1), never a panic.

## Immutable Arc + atomic reload (#380, no-auth part)
- The resolved config is an immutable `Arc<EffectiveConfig>` read via a single atomic pointer load (`config_reload`, safe-Rust `RwLock<Arc<...>>`, no per-message re-parse). The re-read RELOAD validates the whole candidate, rejects a COLD-key change atomically, and swaps in ONE store ONLY on full success; a broken reload KEEPS the old config. Wired as a startup re-read self-check.

## #380 split
- DELIVERED: the SIGHUP/re-read reload engine + the read-only materialized-config surface.
- DEFERRED to #106: the MUTATING wire `CONFIG SET`/`SAVE` admin verbs (they need connection-scoped auth; this PR adds NO unauthenticated remote config mutation). SIGHUP is currently bound to graceful-stop via ctrlc's `termination` feature, so re-binding it to reload is a follow-up that must not regress #195; the engine and the safe re-read trigger ship here.

## Tests (teeth)
- file > default, env > file, flag > env > file; the no-config path is unchanged; the literal grammar accepts durations/sizes and rejects a unitless or decimal-SI value; an unknown key is rejected with a suggestion (and accepted under `--allow-unknown-config`); a coupled-set violation is rejected; a broken file fails with path + line; the reload swaps on success and keeps the old config on a broken reload and rejects a cold-key change.

## Local gate
- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all green. No unsafe; ironbus-core stays IO-free; no em/en dashes; deny/cffi-ban/io-free checks pass. The non-unix `cmd_serve` stub consumes the new fields. DCO-signed.


ELares:
Independent adversarial review APPROVE, no blockers. Security crux clean: zero proto/wire/admin changes, no unauthenticated remote config-mutation surface, the reload is triggered only by a local re-read (SIGHUP stays bound to graceful-stop), the mutating CONFIG verbs are deferred to the open #380. The reload is atomic (validate-whole, cold-key-reject, swap-on-success-only, keep-old-on-failure, all teeth-proven), the no-config default path is byte-identical, a malformed file is a clean exit 1 not a panic, and toml 0.5 adds no new shipped crate (serde-only, MSRV-favorable). The Windows stub consumes the new fields. Merging; #380 stays open for the authed verbs.

ELares:
Independent adversarial review APPROVE, no blockers (security crux clean: no unauthenticated remote config mutation, atomic reload teeth-proven, no-config default byte-identical, toml 0.5 adds no new shipped crate, malformed file = clean exit). Rebased onto main after the #403 codec merge; resolved the lib.rs (both modules) and main.rs (the #387 compression parsing + the #382 mutable ParsedServe, compression field present and consumed in the non-unix stub) conflicts by hand, fmt + clippy + CI green. Merging.

Issue #71, opened 2026-06-05
[TASK] Connection lifecycle, INFO/CONNECT negotiation, and the auth carrier


ELares:
Parent: #11

Specify the full connection lifecycle from TCP/TLS open through INFO/CONNECT capability and version negotiation, keepalive, and graceful and ungraceful close, plus the generic auth carrier that #18 fills without touching framing. Document it as a state diagram. No code; this is the handshake contract every client implements.

## Approach

On accept, the server sends INFO advertising a capability bitset and limits: max_frame_size (default 1 MiB, hard ceiling 16 MiB), default credit window (finite, not 65535), supported compression codecs and dictionary ids (for #12), header support, auth methods, and a protocol version. The client replies with CONNECT echoing the subset it will use and carrying `[u8 auth_method][varint auth_len][auth_blob]`. Auth methods: 0 none (allowed only on loopback or unix socket), 1 bearer-token inline, 2 SCRAM-style challenge-response, 3 mTLS-only (blob empty, identity from cert). For method 2 the server may reply ERR(AUTH_CONTINUE 0x0100) with a nonce, prompting a second CONNECT. Define PING/PONG for liveness and RTT with a default interval and a missed-PONG close. Credit is always reset and re-advertised per (re)connect, never carried across connections. Specify graceful close (drain then FIN) versus ungraceful (fatal ERR then close). Version negotiation is Kafka-style: unknown verbs above 0x1F and unknown user header keys are ignored-or-error per a capability flag, so new frames do not break old peers.

## Acceptance criteria

- [ ] State diagram covering open, INFO/CONNECT, optional AUTH_CONTINUE, established, PING/PONG, graceful and ungraceful close.
- [ ] INFO field set and CONNECT echo rules specified.
- [ ] Auth carrier shape frozen with the four method ids and the loopback-only rule for method 0; handoff to #18 documented.
- [ ] Version and capability negotiation rules for unknown verbs and header keys defined.
- [ ] PING interval, missed-PONG threshold, and credit-reset-per-connect rules stated.

## Failure considerations

Reconnect storms arise when a client cannot tell fatal from transient; mitigate with the fatal-vs-transient ERR split and a clear close protocol. A desynced capability set (client uses a feature the server did not advertise) is a fatal ERR, not undefined behavior. Auth-method downgrade is prevented by the server refusing method 0 off-loopback regardless of client request, coordinated with #18.

ELares:
[...]

PR #406, opened 2026-06-09
feat: negotiated per-consumer credit in the Connect/Info handshake (#292)

### LINKED ISSUE #292: [TASK] Wire-advertised / negotiated per-consumer credit in the Connect/Info handshake (follow-up to #275)

ELares:
Closes #292

Wire-advertised / negotiated per-consumer credit in the Connect/Info handshake, the deferred WIRE half of #275/#65. The handshake bodies were EMPTY; they now carry the credit negotiation in a versioned, length-prefixed, forward-compatible body. This is an ADDITIVE wire change: the frozen `Connect`/`Info` FrameType tags (1, 2) are unchanged, only the previously-empty body gained a format.

## Body format (versioned, length-prefixed, forward-compatible)

Both bodies share: `body_version: u8` (`HANDSHAKE_BODY_VERSION` = 1), `field_len: u16` (the known-field block length), then the v1 block.
- `ConnectBody` v1 block: `flags: u8` (presence bits), `requested_credit: u32 LE`, `requested_credit_bytes: u64 LE`.
- `InfoBody` v1 block: `flags: u8`, `credit {negotiated, cap}: u32 LE`, `credit_bytes {negotiated, cap}: u64 LE`.

Trailing bytes past the block (a future version's appended fields, e.g. the #71/#11 `wire_protocol_version` + capability bitset) are tolerated and ignored, so this body is the seam those fields slot into without a re-break.

## Negotiation

The effective per-consumer credit is `min(client request, server cap)`, or the server default when the client requests nothing. A request can only TIGHTEN the server cap, never raise it; no unbounded/`request(MAX)` value is representable. The negotiated credit governs the existing `consumer_credit` / `consumer_credit_bytes` consumer pull.

## Backward-compat (both directions)

- An OLD client's EMPTY `Connect` decodes to no-request, so the server uses its defaults.
- A new client tolerates an OLD server's EMPTY `Info` (decodes to no-advertisement) and keeps its own local credit.

## Decode safety

Cap-before-alloc (the declared `field_len` is bounds-checked against the body before any read); an unknown `body_version` is a typed `BadHandshakeVersion`; a malformed body is a typed error / clean close (server replies `Err`, client surfaces `ClientError::Body`), never a panic or over-allocation. New `connect_body`/`info_body` fuzz targets + regression seeds; proto proptests; the server reads its caps LOCALLY off the engine handle (no actor round-trip), preserving the #177 head-of-line property.

## Tests with teeth

A client requesting a credit gets `min(request, cap)`; an empty Connect gets the server default; an empty Info leaves the client on its local credit (both directions); a malformed handshake body is a typed error not a panic; the negotiated credit governs the pull; the FrameType freeze + wire round-trip + fuzz replay stay green.

## Local gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all pass; no em/en dashes; FrameType golden test intact.

Docs: CONTRACTS.md (body format), COMPATIBILITY.md + compat/versions.md (additive handshake-body change + empty-body backward-compat), FLOW_CONTROL.md (negotiated credit), CHANGELOG.


ELares:
Independent adversarial review APPROVE after one fix. The review caught a CRITICAL flow-control bypass: a client requesting Some(0) byte-credit hit 0, which aliases UNLIMITED in the pull path, disabling its own in-flight-byte ceiling (memory-exhaustion bypass regressing #275). Fixed: a 0 (unlimited) request now clamps DOWN to a finite cap, with regression tests on both dimensions. Everything else teeth-proven clean: backward-compat both directions, the message-cap clamp, cap-before-alloc decode, the FrameType freeze, and the #177 head-of-line invariant (caps read locally). Rebased onto main (resolved the CHANGELOG keep-both); CI green. Merging.

PR #407, opened 2026-06-09
feat(cli): offline mutating admin verbs (consumer-reset, dlq-redrive) (#299)

ELares:
## What

The AUTH-FREE, broker-stopped subset of the mutating admin surface (#299): two OFFLINE CLI verbs that operate on a STOPPED broker's data directory. A mutating surface OVER THE WIRE needs connection-scoped auth (an unauthenticated remote data-destruction endpoint is an RCE/data-loss-class hole), so this PR ships only the parts that are safe without auth: the offline verbs an operator who already owns the filesystem runs while the broker is stopped. Both take the same exclusive data-dir lock `serve` holds, so they refuse a running broker (exit 5) and can never race a live writer (the broker-stopped contract).

### Delivered (offline, auth-free)

- **`admin consumer-reset --data-dir <dir> --group <name> --to <offset|earliest|latest>`** rewrites a work-group's durable cursor checkpoint to a chosen offset, clamped to the durable range `[earliest_retained, head]`. An out-of-range explicit offset (or an invalid group name the broker's discovery would skip) is rejected (exit 1). It reuses the broker's **exact codecs** (the dual-slot CRC `Checkpoint` + the `AckCursor` snapshot), so the file is byte-for-byte one the broker's recovery resumes from, not a new variant.
- **`admin dlq-redrive --data-dir <dir>`** re-injects the dead-lettered records from the durable DLQ sink (`dlq/`) back onto the main log via the log's own append path: **append + fsync the records FIRST, then advance a durable `dlq-redrive.ckpt` watermark** (the same dual-slot CRC checkpoint). The watermark makes a completed redrive **idempotent** (a re-run re-injects nothing); a crash before the watermark advance re-redrives the suffix (at-least-once), **never corrupting the log**. The DLQ sink is preserved.

### Deferred (and why)

- **The mutating WIRE forms** (the same actions on a LIVE broker) and **`force-reap`** (reaping stuck leases on a running broker, which has no offline meaning) need connection-scoped auth, so they are deferred to the authed admin surface (#380 / #106). `admin force-reap` is a clean usage error here, naming the deferral.

## Crash-safety + correctness

- Reuses the atomic discipline: the dual-slot CRC checkpoint (a torn write reverts to the prior durable slot) and the segmented log's append+fsync. A crash mid-operation leaves a recoverable data directory, never a corrupt log.
- The canonical cursor-checkpoint naming now lives in `ironbus-storage::naming` and the **engine delegates to it** (single source of truth, so the offline rewrite and the broker's writes target byte-for-byte the same file).

## Tests with teeth

- The **real broker** (engine) resumes a named group from the reset offset.
- Reset clamps to the durable range; an out-of-range offset and an invalid group name are rejected and write nothing.
- DLQ redrive re-injects every poison record and is **idempotent on a re-run** (no duplicates); a crash before the watermark advance leaves a recoverable log and **re-redrives at-least-once**.
- Both verbs **refuse a running broker** (the dirlock, exit 5).
- The versioned `--json` schemas (`ironbus.cli.admin-consumer-reset.v1` / `ironbus.cli.admin-dlq-redrive.v1`) and the frozen exit codes are pinned.

## Engineering invariants

No `unwrap`/`expect`/`panic` in non-test paths, typed errors to the frozen exit codes, no `unsafe`, `ironbus-core` stays IO-free, the on-disk formats are written EXACTLY as the broker writes them (same codecs, no new variant). No em/en dashes.

## Local gate (green)

- `cargo fmt --all -- --check`: pass
- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: pass
- `cargo test --workspace`: pass (1301 tests)

## Docs

`CLI.md`, `CLI_CONTRACT.md` (the new verbs, their `--json` schema, exit codes, the broker-stopped contract), `CONTRACTS.md` (the `dlq-redrive.ckpt` watermark, the shared cursor-checkpoint naming), `compat/versions.md` (schema rows), and a terse CHANGELOG bullet.

refs #299 (stays open to track the deferred authed/online wire admin verbs and force-reap)


ELares:
Independent adversarial review APPROVE after one fix. The review found a cross-feature bug: redrive dropped the COMPRESSED flag so a redriven compressed record (default lz4 codec) decoded as garbage. Fixed: DlqEntry carries the stored flags and redrive re-applies COMPRESSED (append re-derives HAS_KEY/HAS_XXH3), with a teeth test. Everything else teeth-proven clean: offline-only (no wire surface), crash-safe with the broker exact codecs (append+fsync then watermark), idempotent, clamp+reject, dirlock before any write. Rebased onto main (CHANGELOG + versions.md keep-both); CI green. #299 stays open for the authed verbs. Merging.

Issue #337, opened 2026-06-08
[TASK] Implement optional key-based log compaction (v2 segment, atomic swap, sparse offsets)


ELares:
Follow-up from #83 (PR #332), which SPECIFIED optional key-based log compaction in `docs/COMPACTION.md`. Implementation is future work (M2 was a design milestone; the spec is explicit that compaction is opt-in, off-by-default, and edge-hostile).

Implement per `docs/COMPACTION.md`:
- The cleaner: read N adjacent dirty sealed segments, build a key-to-latest-offset map, write survivors keeping ORIGINAL offsets (sparse) into a fresh `version=2` compacted segment, fsync, then the single atomic commit (durable appearance), then rename-then-unlink the originals.
- The v2 compaction-metadata footer block (covered_base_offset/covered_end_offset/covered_base_seq/covered_end_seq/highest_covered_source_id + block_crc) and the `COMPACTED` flag bit; the `version=2` fail-closed bump (a v1 reader refuses a compacted log).
- Recovery resolves overlapping ranges from headers/footer-block via longest-valid-prefix with no compaction-specific repair; tombstone TTL (24h); rate-limited + append-yielding + off-by-default; `compact_and_delete` runs the reaper first.
- DEPENDS ON: sparse-offset reader/index tolerance + the non-loss `Compacted` gap marker (coordinate with #9 reader gap semantics and the #59 SkipEvent), which the current dense-offset read paths (`Engine::poll`, `Log::read_from`, the offline reader) do not yet have. Those are prerequisites.

Design, the no-manifest atomic-swap reconciliation, and the crash-safety argument are in `docs/COMPACTION.md`. refs #83, #13, #9, #59.

ELares:
Implemented in #409 (off by default, opt-in): the cleaner + the additive v2 compacted segment format (COMPACTED flag bit, 44-byte covered-range footer block, fail-closed version=2 bump) + the no-manifest atomic swap + recovery's overlapping-range resolution from the footer metadata + sparse-offset reads (the #346 Compacted gap reason). Crash-safety proven with the fault-fs + ManualClock (crash before/after the commit point, before-vs-after-identical recovery, no record lost or doubled); I1 to I5 hold across compaction + crash; v1 segments byte-identical; format-registry re-pinned. Deferred as safe follow-ups: byte-based min_dirty_ratio accounting, a standalone CLI compact verb, an explicit tombstone flag bit.

PR #408, opened 2026-06-09
feat(proto): consumer-visible sparse-offset gap marker on DELIVER (#346)

### LINKED ISSUE #346: [TASK] Deliver the consumer-visible sparse-offset gap marker on the DELIVER stream (#59 residual)
Follow-up carved out of #59 (closed for M2: the SkipEvent schema, the offset contract, the metric set, the crash-injection harness, and worked examples all landed in #344). One #59 acceptance criterion is an IMPL task that remains open and is tracked here.

The recovery-side half is done: skipped ranges leave a permanent hole, the next valid record keeps its original offset, and `loss-report.v1` records each skip with `byte_offset_start` / `byte_offset_end` (the resync point). The **consumer-facing half is not implemented**: there is no over-the-wire gap marker. The proto has only `Deliver` (tag 13); no `GapMarker` FrameType exists, so a consumer reading the DELIVER stream across a recovered skip sees offsets jump with no explicit signal that a bounded, reported loss occurred in the hole.

## Scope
- Add a consumer-visible gap marker on the DELIVER stream (a new FrameType, frozen-tag-pinned, append-only after the current max tag) carrying the gap's offset range and byte count, sourced from the already-frozen `loss-report.v1` skip record.
- Coordinate the delivery semantics with the #9 consumer model (cursor advancement across a hole, at-least-once interaction, whether the marker is at-least-once or best-effort, group/broadcast behavior).
- Client-side: surface the gap to the consumer API so a reader can observe "a reported loss of N bytes spanned offsets [a, b)" rather than inferring it from an offset jump.
- Tests: a consumer reading across a recovered torn-tail / mid-log-corruption skip receives the exact gap marker (range + bytes) matching the recovery-side `loss-report.v1` event.

## Why it was deferred
#59's design, schema, metric, and recovery-side crash/fuzz harness are complete and frozen; the over-the-wire delivery contract depends on the #9 consumer model and is a wire-protocol addition (new frame tag) better done as its own reviewed change than folded into the schema-reconciliation PR.

Refs #59, #9, #8.

ELares:
Closes #346

## What

Implements the consumer-facing half of the #59 sparse-offset contract: a consumer-visible **gap marker** on the DELIVER stream that tells a reader the half-open offset span `[from, to)` is PERMANENTLY ABSENT (skipped), so a consumer that tracks contiguity learns an offset jump is a bounded, REPORTED gap rather than message loss.

## Mechanism (additive frame, opt-in)

- **New `GapMarker` FrameType (tag 21)**, appended after the current max tag (`PubAckDuplicate` = 20). `GapMarkerBody` = `from: u64`, `to: u64` (exclusive), `bytes_skipped: u64`, `reason: u8` (`1` = trimmed, `2` = compacted/reserved for #337), a fixed 25 bytes, sourced from the frozen `loss-report.v1` skip record. The **`Deliver` (tag 13) body is UNCHANGED** (dedicated frame, not a flag/field on DELIVER), so the DELIVER codec stays frozen.
- **Per-consumer OPT-IN** via an AND-negotiated handshake capability bit (`Connect` bit 2 `CONNECT_FLAG_WANTS_GAP_MARKER` / `Info` bit 2 `INFO_FLAG_GAP_MARKER`), a pure flag carried in the existing #292 handshake block (no new field bytes).

## Where it is wired

- **Engine unchanged.** `Poll::Truncated { earliest_retained, skipped }` is the existing skip outcome. The **session** maps it: for a gap-marker consumer it emits `GapMarker` with `to == earliest_retained`, `from == earliest_retained - skipped`, `reason = TRIMMED`, `bytes_skipped = 0` (a force-reap trim is byte-untracked); for a legacy consumer it emits the unchanged `Truncated` (tag 18). The two NEVER double-signal.
- **Client** surfaces a skipped span as a typed `Fetch::gaps` event (distinct from a delivery and end-of-stream), opt-in via `ClientConfig::request_gap_marker` / `Client::gap_marker_enabled`.

## Backward-compat

The bundled client REFUSES an unknown frame tag (`ClientError::UnknownFrameType`), so the new frame is gated behind the handshake capability: the server only sends tag 21 to a consumer that advertised it. An old consumer keeps `Truncated` and is **never sent the new tag**; a new client against an old server (empty `Info`) leaves the capability off and keeps `Truncated`. `OFFSET_TRIMMED` is subsumed cleanly (one signal per gap, richer for a capable consumer). Offset monotonic / never-reuse invariants are untouched; the marker only makes the existing sparse-offset gap explicit.

## Tests with teeth

- Server: a gap-marker consumer reading across a force-reaped span gets EXACTLY ONE `GapMarker` with the exact `[from, to)` + `TRIMMED` reason and ZERO `Truncated` (no double-signal); a contiguous stream yields no spurious marker; `Info` confirms the capability only when the client advertised it.
- Client: a `GapMarker` frame surfaces as the typed `Gap`; it counts against credit; an old server leaves the capability off and the legacy `Truncation` still surfaces.
- Proto: round-trip proptest, cap-before-alloc short/overlong rejection, unknown-reason tolerance, re-pinned FrameType golden test + `ALL_TYPES`.
- Fuzz: `gap_marker_body` target + content-addressed regression seeds, wired into the replay test, `TARGETS`, and the nightly soak.

## Local gate

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, and `cargo test --workspace` all green. No em/en dash. No unwrap/expect/panic in non-test paths; no unsafe; `ironbus-core` untouched (IO-free).


ELares:
Independent adversarial review APPROVE, no blockers. The FrameType freeze is extended additively (tag 21 appended, 1..20 byte-for-byte unchanged, golden test re-pinned), backward-compat is opt-in AND-negotiated (an old client is never sent tag 21, teeth-proven by breaking the gate), there is exactly one marker with the correct range+reason and zero double-signal on a contiguous stream, the fixed-25-byte body decodes with no attacker-controlled length (over-alloc structurally impossible, fuzz target + seeds), and the offset invariants are untouched (engine unchanged). Merging.

PR #409, opened 2026-06-09
feat(compaction): optional key-based log compaction (v2 segment, atomic swap, sparse offsets) (#337)

### LINKED ISSUE #337: [TASK] Implement optional key-based log compaction (v2 segment, atomic swap, sparse offsets)

ELares:
Closes #337

Implements the optional, opt-in, off-by-default key-based log compaction designed in `docs/COMPACTION.md`: the cleaner, the additive v2 compacted segment format, the no-manifest atomic swap, the fail-closed `version=2` bump, recovery's overlapping-range resolution, and the sparse-offset read path. Off by default; a broker that does not opt in is byte-for-byte unchanged.

## The cleaner (`crates/ironbus-storage/src/compaction.rs`)
read N adjacent dirty SEALED segments -> build a key-to-latest-offset map -> write the survivors (latest record per key + a tombstone within its 24h TTL + every keyless record) into a fresh `version=2` compacted segment, KEEPING their ORIGINAL offsets/sequences (sparse, never renumbered, so I5 holds) -> fsync the file -> dir-fsync (THE single atomic commit point) -> unlink-then-dir-fsync the originals. It runs OFF the hot path after the produce-path reaper (`compact_and_delete`: reaper first, then compactor); it only ever reads sealed segments and writes a NEW file, never the active segment, so it never races or blocks an append.

## Crash-safety
A crash at ANY step recovers deterministically and is proven with the fault-fs + ManualClock (`tests/compaction_crash.rs`):
- crash BEFORE the commit (the commit dir-fsync fails) -> the originals win, the orphan compacted segment is discarded;
- crash AFTER the commit during retire (the compacted segment is durable, some/all originals still present) -> the compacted set wins, the superseded originals are dropped, the overlapping range resolved from the v2 footer metadata;
- before-vs-after-commit recover IDENTICALLY (the swap is crash-atomic); no record lost or doubled; neither reconciliation emits a `LossReport` event.

## The v2 format (additive)
A `COMPACTED` segment-header flag bit (`0x0001`), a 44-byte covered-range footer block (`covered_base/end_offset`, `covered_base/end_seq`, `highest_covered_source_id` + `block_crc`), and the `version=2` FAIL-CLOSED bump (a v1-only reader REFUSES a compacted log with a typed `UnsupportedVersion`). A v1 segment is byte-identical; the format-registry digest is re-pinned (v1 lines unchanged, v2 consts added).

## Recovery + sparse offsets
`Log::recover_with_compaction` resolves the overlap via longest-valid-prefix from the self-describing v2 metadata (no manifest, no compaction-specific repair), advancing the offset/seq expectation by the COVERED span across a compacted segment. The storage read path and the engine poll SKIP a compacted hole (the #346 `Compacted` gap-marker reason, not loss) and the cursor advances past it. A compacted segment takes a fresh never-recycled id (ADR 0002); the active roll skips past it (`next_fresh_segment_id`).

## Scoped in vs deferred
IN: the cleaner, the v2 format, the atomic swap, the recovery overlap resolution, the fail-closed bump, sparse-offset reads, the off-by-default `serve --compact` knob + the off-hot-path engine pass. DEFERRED as safe follow-ups (the core is complete): the byte-based `min_dirty_ratio` accounting (count is shipped), a standalone `ironbus compact` CLI verb, and an explicit TOMBSTONE flag bit (the empty-payload convention is shipped). None affect the crash-safety or the format.

## Gate
`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`, `cargo test --workspace`, `scripts/check-format-registry.sh`, the conformance vectors + corpus, and the recovery/crash/determinism/seeded-fault sweeps all pass. No em/en dashes (`grep -rnP '[\x{2013}\x{2014}]'` clean). Also resolves the pre-existing committed merge-conflict markers in `CHANGELOG.md` and `compat/versions.md`.

Docs: `COMPACTION.md` flipped from specified to implemented, plus `WAL.md`, `CONTRACTS.md` (the v2 footer + flag byte layout), `INVARIANTS.md` (compaction upholds I5), `RECOVERY.md` (overlapping-range resolution).


ELares:
Independent adversarial review APPROVE, no critical blockers, on the largest/most data-critical change in the project. Crash-safety of the atomic swap is verified with break-and-confirm teeth (removing the data fsync-before-commit makes the crash tests fail, so the ordering is load-bearing and tested), a v1 log is byte-identical (only additive v2 consts, registry re-pinned), the v2 version bump is fail-closed (a v1 reader REFUSES a compacted segment, proven on-disk), recovery resolves an overlapping range deterministically via the covered-range footer, the offset invariants hold (survivors keep original sparse offsets, the fresh never-recycled compacted id cannot collide either direction, I1-I4 pass across a compaction+crash), sparse reads skip the hole, and the feature is off-by-default + off-hot-path. The one substantive nit (the engine advances SILENTLY over a compacted hole rather than emitting GapMarker(reason=COMPACTED)) is corrected in the docs and filed as #411; it is not a loss/correctness gap. Merging.

Issue #67, opened 2026-06-05
[TASK] fsync-derived producer admission window and WAL backpressure seam


ELares:
Parent: #10

Make the durable log's fsync throughput a first-class backpressure source so a fast producer cannot grow an unbounded pre-fsync write buffer on edge flash. The draft assumes the log absorbs spikes; this task closes the seam where fsync IOPS, not the consumer, is the real ceiling, in coordination with #6 and #4.

## Approach
The WAL/segmented-log writer publishes an admission credit derived from group-commit headroom. A producer must acquire one admission slot before appending; slots release on fsync completion. Default `wal_admission_window` = 4 group-commit batches in flight. When slots are exhausted, the topic's overflow policy fires: durable topics block up to `ack_timeout` (default 5 s) then error; telemetry topics drop-oldest. This bounds dirty in-flight writes to the admission window rather than to RAM. Expose `wal_admission_slots_in_use` and fsync p99 to #16 so the operator can see WAL backpressure distinctly from consumer-side backpressure. The spill tier's writes share the same fsync and wear budget (#6), so spill and main-log admission draw from a coordinated budget, not two independent ones.

## Acceptance criteria
- [ ] Admission slot acquired before append, released on fsync completion; default window = 4 batches.
- [ ] Exhaustion fires the topic overflow policy (durable block-then-error at 5 s, telemetry drop-oldest).
- [ ] Dirty pre-fsync in-flight writes are provably bounded by the admission window.
- [ ] `wal_admission_slots_in_use` and fsync p99 exposed to #16.
- [ ] Spill and main-log admission share one coordinated fsync/wear budget.
- [ ] No path allows an unbounded pre-fsync buffer (audited).

## Failure considerations
A producer outrunning fsync would otherwise OOM via an unbounded write buffer: the admission window blocks it and converts the spike into bounded latency or an explicit error. Spill and main log competing for fsync could double-spend the wear budget: a coordinated budget prevents it. A wedged fsync (stuck device) is surfaced by the slots-in-use gauge pinned at capacity rather than hidden as a latency spike.

ELares:
[...]

PR #413, opened 2026-06-09
feat(server): fsync-headroom admission credits bounding the un-fsynced write frontier (#378)

### LINKED ISSUE #378: [TASK] fsync-headroom admission credits (refinement of the #67 WAL backpressure seam)
Carved out of #67 (closed for M1: the WAL backpressure SEAM is satisfied by the #177 bounded append-channel, which backpressures producers when the disk stalls while preserving I2). This is the deferred OPTIMIZATION half.

Replace the coarse bounded sync_channel bound with an fsync-HEADROOM-derived admission window: acquire an admission slot before append, release it on fsync completion, default wal_admission_window ~4 batches, exhaustion fires the topic overflow policy, and expose ironbus_wal_admission_slots_in_use + an fsync-p99 admission metric. The current bounded channel already bounds dirty-pre-fsync writes; this derives the credit from group-commit headroom for tighter, observable control. Must preserve I2 / group-commit / no-deadlock.

Refs #67, #177, #10.

ELares:
Closes #378

Refines the #67 / #177 WAL backpressure seam with an **fsync-headroom admission credit**: it bounds how far the BUFFERED (appended-but-not-yet-`fdatasync`'d) write frontier may run ahead of the DURABLE (synced) frontier, so a producer outrunning fsync cannot grow an unbounded un-fsynced backlog. This is a memory guard under any level and a bounded-loss-window guard under a relaxed durability level, distinct from CoDel's queue-latency shed.

## The admission credit (reusing `unsynced_bytes()`)

The storage log already tracks the exact un-fsynced exposure as `unsynced_bytes()` (the #341 relaxed-durability accounting, reset to `0` on a real `sync` or a segment roll's seal). The credit does ONLY the threshold math against that frontier, so the pure decision (`FsyncHeadroom::would_admit`) lives in the IO-free `ironbus-core`; the engine feeds it the live frontier through the storage seam.

The append actor consults it BEFORE the append (so it rejects NEW work only). On exhaustion it forces a group-commit DRAIN first, then re-checks:
- Under the default `sync` level (and a due `interval` window) the drain is a real `fdatasync`, so the frontier resets to `0` and the record is admitted by the no-wedge floor: the headroom **THROTTLES** (drain-then-admit), never sheds, never loses.
- Under a relaxed `async`/`none` level a commit DEFERS the fsync, so the frontier does not drain; the re-check still fails and the new produce is SHED with the typed `ProduceOutcome::WalHeadroomShed` (the bare `Err` message `wal fsync headroom exhausted`, distinct from CoDel's `shed under load` and the byte cap's `at capacity`).

## How it composes with #336 (CoDel) and #341 (durability)

- **CoDel (#336):** orthogonal. CoDel sheds on standing admission SOJOURN (latency); the headroom sheds on the un-fsynced BACKLOG SIZE. Each is consulted before the append and admits-or-sheds NEW work only, so neither weakens the other.
- **Durability (#341):** under `sync` the headroom bounds the GROUP-COMMIT backlog (a RAM guard); under a relaxed level it bounds the LOSS WINDOW in bytes by shedding new produces once the backlog fills (the `interval` byte trigger applied as an admission gate).

## No data loss + safe default

The shed rejects NEW work only and is decided before the append, so no accepted (acked or to-be-acked) record is ever dropped or left non-durable: the durability/ack path and I2 are untouched (the #336 shed contract). The knob defaults to `0` (OFF), so an unchanged broker behaves exactly as today; a small headroom is the opt-in for a tight RAM/loss-window bound. An empty backlog always admits, so an oversized produce never wedges.

## Observability (re-pinned taxonomy)

- `ironbus_wal_fsync_headroom_shed_total` (counter): pinned in `FROZEN_RESILIENCE_COUNTERS` (a shed is never silent) and `FROZEN_METRIC_TYPES`.
- `ironbus_wal_fsync_headroom_bytes` (gauge): the configured window, pinned only in `FROZEN_METRIC_TYPES`.

Both frozen golden tests re-pinned and green.

## Knob + non-unix stub

New `serve --wal-fsync-headroom-bytes` / `IRONBUS_WAL_FSYNC_HEADROOM_BYTES` (default `0` = off), wired through `ServeConfig`, the `EngineConfig`, the materialized-config startup line, and the `#[cfg(not(unix))]` `cmd_serve` stub's `let _ = (...)` (the Windows `-D warnings` footgun).

## Tests with teeth (fault-fs + ManualClock)

- A tight headroom under a relaxed level sheds the over-backlog produce with the typed signal and NO accepted record lost (the durable head still reflects every accepted record); the shed counter increments exactly once.
- A sync drains the backlog and re-admits.
- The default (off) headroom is unchanged behavior; under `sync` the headroom throttles via the group-commit drain and never sheds.
- The headroom admits within the window and sheds past it against the live `unsynced_bytes()` frontier (the loss-window cap).
- The typed wire mapping (`WalHeadroomShed` -> `wal fsync headroom exhausted`, distinct from the other sheds) and the flag/env/default resolution + materialized-config line.

## Local gate

- `cargo test --workspace`: **1347 passed, 0 failed** (the frozen taxonomy + durability/crash/determinism tests stay green)
- No em/en dashes anywhere (`grep -rnP '[\x{2013}\x{2014}]'` clean)

Engineering invariants held: no unwrap/expect/panic in non-test paths, typed errors, no unsafe, `ironbus-core` stays IO-free (pure headroom math), the FROZEN taxonomy extended and re-pinned.


ELares:
Independent adversarial review APPROVE, no blockers. No data loss (the headroom check sheds a NEW produce before the append and drains the parked batch durably first, teeth-proven by removing the drain), no deadlock (an empty backlog always admits an oversized record, teeth-proven by disabling the floor), safe default (off=byte-for-byte unchanged, teeth-proven), composes orthogonally with CoDel #336 and correctly with the #341 durability levels (sync throttles, relaxed sheds, I2 intact), and the two new metrics are additive + re-pinned in the frozen taxonomy. IO-free core, non-unix stub consumes the field. Merging.

PR #414, opened 2026-06-09
docs(readme): edge-focused quick-start (install, producers, consumers)

ELares:
Adds an easy-to-follow **Quick start** section to the main README so someone can get IronBus running on an edge device and wire up producers and consumers.

It covers, in order:
1. **Install** (build from source today, or the static musl binary for the device; the released `curl|sh` / `.deb` / container channels are noted as going live at the first tag).
2. **Start the broker on the edge** (`serve --data-dir --profile edge-tiny`, the loopback default + `fdatasync` zero-loss note, binding `0.0.0.0` for remote, health/metrics, graceful stop, systemd), with a security caveat that the wire protocol is not yet authenticated or encrypted.
3. **Producers, one or many** (`ironbus pub`, key, stdin, a burst loop).
4. **Consumers, one or many** (`ironbus sub` with `--ack`/`--nack`/`--term`/peek, then the three many-consumer patterns: competing, `--key-shared-group`, and `--broadcast-group` + `cumulative-ack`).
5. **Inspect the data dir offline** (`peek`/`dump`/`scrub`).

Every command was cross-checked against `crates/ironbus-cli/src/main.rs` and `docs/CLI.md`, so it uses only verbs the binary actually ships. Also fixes two stale CLI mentions that named unimplemented commands (`tap`/`wire` in the Simple tenet, `info`/`consumer ls`/`lag` in the CLI section).

Docs-only: README + a terse CHANGELOG bullet. No em/en dashes; the relative-link check (anchor + doc links) passes.


ELares:
Independent review APPROVE, no blockers. Every command in the new Quick start was cross-checked against crates/ironbus-cli/src/main.rs + docs/CLI.md (a per-command table with line citations): serve/pub/sub/cumulative-ack/peek/dump/scrub all match real arms + flags + defaults, only shipped verbs are used, the two stale mentions (tap/wire, info/consumer-ls/lag) are correctly fixed, the not-yet-encrypted/authenticated security caveat is accurate (no TLS/auth on the wire path), the edge-tiny 64 MiB refuse-to-boot ceiling + the sync-default zero-acked-loss + the competing/key-shared/broadcast semantics all match the engine, the anchor + doc links resolve (relative-link-check 0 problems), and there are no em/en dashes. Merging.

PR #415, opened 2026-06-09
ci(release): automatic calendar-versioned rolling release on every main push

ELares:
## Why

Installing IronBus on an edge device should be "grab the binary for your arch from the latest release" (or run the existing `curl | sh` installer, which already auto-detects arch and pulls the latest release). Today no release exists because the only release workflow fires on a `v*` tag (reserved for the maintainer's `v0.1.0`), so the installer has nothing to pull. This adds a continuous channel that publishes a release on every push to main, making install seamless.

## What

**A new rolling, calendar-versioned release channel.** `.github/workflows/rolling-release.yml` publishes a NORMAL GitHub Release on every push to main (plus `workflow_dispatch`), so `releases/latest` and `scripts/install.sh --version latest` always resolve to the newest build.

- Trigger `on: push: branches: [main]` + `workflow_dispatch`; `concurrency: rolling-release` (cancel-in-progress false) so close pushes serialize and never collide on the build number; `permissions: contents/id-token/attestations: write`.
- Version `YYYY.MMDD.N`: UTC `date -u +%Y.%m%d` plus a per-day build number (count of existing same-day release tags + 1), so the first build of a day is `.1`, the next `.2`, and the next day resets. The leading-zero month/day is intentional CalVer (never semver-parsed).
- `[skip release]` in the head commit message skips a build (the `workflow_dispatch` path always runs).
- Build matrix reuses `release.yml`'s 3-triple static-musl setup VERBATIM (same `cross`, same reproducible `RUSTFLAGS` + `SOURCE_DATE_EPOCH`, the static-link `readelf` assertion, the `dist/ironbus-<triple>` + `.sha256` staging), plus `export IRONBUS_BUILD_VERSION="$VERSION"` so the calendar version is compiled in.
- Publish job downloads the three binaries, builds a consolidated `SHA256SUMS` (self-checked with `sha256sum -c` before publish), attests Sigstore build-provenance over the binaries + `SHA256SUMS`, and `gh release create`s a NORMAL release (no `--prerelease`, no `--draft`) whose notes carry the one-line install and a per-arch CPU -> asset table plus the commit SHA. No changelog gate (rolling builds are continuous).
- Every `uses:` is SHA-pinned to the SAME commit as `release.yml` (the #142 convention). `release.yml` is untouched: it stays the formal `v*` channel with the `.deb`, container, CycloneDX SBOM, and changelog gate.

**Build-time version override.** `ironbus version` now prints `option_env!("IRONBUS_BUILD_VERSION").unwrap_or(env!("CARGO_PKG_VERSION"))`, so the published calendar version is stamped in at compile time without bumping `Cargo.toml` (a version bump would break `cargo build --locked`, since `Cargo.lock` pins the workspace crates at `0.0.0`; `option_env!` leaves the lockfile untouched). `Cross.toml` forwards `IRONBUS_BUILD_VERSION` into the cross container. The version unit test asserts the same fallback expression, so it passes with the env unset (dev/CI/test) and when set.

**Docs.** README quick-start leads with the seamless path (`curl | sh`, then the per-arch latest-release table), keeping build-from-source as the developer path and stating every push to main publishes a fresh `YYYY.MMDD.N` build. `docs/DISTRIBUTION.md` and `RELEASING.md` document the rolling vs formal-tagged channels. Terse `CHANGELOG` `### Added` bullet.

## Local gate (all green)

- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic`: clean.
- `cargo test --workspace`: all pass, including the version test (passes with `IRONBUS_BUILD_VERSION` unset).
- `actionlint .github/workflows/*.yml`: clean (shellcheck on `run:` blocks included).
- `sh scripts/ci/relative-link-check.sh`: 0 problems.
- `cargo build --locked` with `IRONBUS_BUILD_VERSION` set: succeeds, `Cargo.lock` untouched.
- Proof both ways: `IRONBUS_BUILD_VERSION=2026.0609.1 cargo run -q -p ironbus-cli -- version` -> `ironbus 2026.0609.1`; plain `cargo run -q -p ironbus-cli -- version` -> `ironbus 0.0.0` (the fallback).
- No em/en dashes anywhere in the diff.

## Note

Merging this PR will itself trigger the FIRST rolling release (the merge commit lands on main), so the merge is the end-to-end test: it should publish `<today>.1` with the three binaries + `SHA256SUMS` + provenance, and `releases/latest` will then resolve so the installer works.


ELares:
Independent review APPROVE, no blockers; the one robustness nit is addressed. The CalVer YYYY.MMDD.N computation is now max(same-day build numbers)+1 (gap-robust, proven on a .1/.2/.5 gap -> .6 and the empty -> .1 cases), using startswith + split so the date dots need no escaping; the release is a NORMAL release (no prerelease/draft) so releases/latest + the installer default resolve; the option_env! version override is proven both ways and leaves --locked + Cargo.lock untouched (Cross.toml forwards IRONBUS_BUILD_VERSION); the build reuses release.yml exact reproducible steps with the env exported before cross build; every uses: is SHA-pinned to release.yml SHAs, actionlint clean, write perms now scoped to the publish job, release.yml untouched; the [skip release] guard cascades via needs; never creates a v* tag; and the README/install per-arch table matches install.sh detect_target. Merging triggers the first rolling release (the end-to-end test).

PR #419, opened 2026-06-10
feat(backpressure): wire retry-throttle, egress AIMD, and the #11 QoS-0 fire-and-forget tier

### LINKED ISSUE #336: [TASK] Implement the backpressure controls (CoDel shedding + retry budget + egress AIMD)

### LINKED ISSUE #402: [TASK] Wire the backpressure controls to live call sites: retry-throttle enforcement, fire-and-forget tier, egress AIMD (#336 residual)
Carved out of #336. The CoDel produce-admission shedding is WIRED LIVE and shipped (it sheds a new produce before the append, never losing an accepted record). The other three controls are IMPLEMENTED + UNIT-TESTED with their enforcement seams in place, but have no production call site yet:

- Retry-throttle ENFORCEMENT: the retry-budget accounting (record_accept/record_shed) is wired in the append actor, but `retry_budget_should_throttle` is never called in production, so a redelivery storm is not actually throttled today. Wire the throttle into the redelivery path (with #9 lease/redelivery).
- Fire-and-forget bucket: built + tested, but there is no fire-and-forget wire tier to gate yet (lands with the wire QoS-0 / uncredited tier, depends on #11).
- Egress AIMD: the +1 / x0.5 bounded [4,128] limiter + estimator seam are built + tested, but the call sites at a live downstream sink (egress_on_success/failure/record_shed, egress_limit) land with that downstream egress feature.

All three are observable already (the counters/gauges are pinned in the frozen taxonomy). This issue tracks attaching them to live call sites so they actually act, not just account. The CoDel mechanism, the pure RFC-8289/AIMD/SRE-retry math (crates/ironbus-core/src/backpressure.rs), and the metrics shipped under #336.

Refs #336, #9, #11.

ELares:
Attaches the three built-but-unwired backpressure controls (#336) to live call sites and adds the wire QoS-0 fire-and-forget tier they need (#11).

- **QoS-0 fire-and-forget tier**: an additive PUB flag (`PUB_FLAG_FIRE_AND_FORGET`, bit 6, wire-only, masked off before the stored RecordFlags, distinct from the dedup bit 7). The client opt-in `produce_fire_and_forget` does not wait; the broker DROPS the produce under the fire-and-forget token bucket without acking (counted in `ironbus_fire_and_forget_shed_total`) or appends it durably with NO PubAck. The default `produce` is unchanged at-least-once; the FrameType tag vocabulary is unchanged.
- **Retry-throttle enforcement**: the retry budget is now consulted in the redelivery path. An exhausted budget DEFERS a redelivery (re-arms the lease deadline by the attempt backoff via a new pure `LeaseTable::defer_redelivery`) rather than dropping it, so a storm is rate-limited but every at-least-once message still redelivers until MaxDeliver routes it to the DLQ. `ironbus_retry_shed_total{side}` counts it.
- **Egress AIMD**: adjusts the effective per-consumer egress credit WITHIN the negotiated #292 ceiling (never exceeds it): additive increase on a prompt ack, multiplicative decrease on would-block / slow-ack / nack. `ironbus_egress_limit` / `ironbus_egress_shed_total` move. OFF by default.

All three default to inert (a zero-config broker is byte-for-byte unchanged). No data loss for at-least-once: only the explicitly-flagged QoS-0 tier may drop, by contract; the retry throttle defers, never drops; AIMD only adjusts within caps. I2 untouched.

Gate: fmt + clippy (all-features, pedantic, -D warnings) + the full workspace test suite green locally; no em/en dashes.

Note: the implementing subagent died mid-flight after committing (a platform outage); the session re-ran the full local gate on its commit and is opening this PR on its behalf.

Closes #402
Closes #336


ELares:
Independent adversarial review: REQUEST CHANGES with two proven blockers, both now fixed and re-gated. (1) The CLI --egress-limit defaulted to 16, silently enabling the egress AIMD on every zero-config broker against the inert-default contract; now 0 (off) with a pinning test. (2) A QoS-0 produce hitting any non-appended disposition (CoDel shed, byte-cap, headroom shed, fence, failure, fatal, too-long dedup id) still received an Err frame, desyncing the fire-and-forget reply stream; the session now sends no frame on every disposition, with a per-disposition mock test. Everything else was verified clean with break-and-confirm teeth: the QoS-0 drop is flag-gated only (an unflagged produce can never be dropped), the wire-only mask holds, the retry throttle defers without touching deliveries/generation and the poison still DLQs at MaxDeliver, the AIMD never exceeds the negotiated cap and never starves to zero, I2 untouched, frozen taxonomies intact, no dashes. Merging.

PR #426, opened 2026-06-10
fix(installer): enforce HTTPS on the wget fallback, fail closed otherwise

### LINKED ISSUE #423: [BUG] install.sh: wget fallback lacks the HTTPS/TLS enforcement of the curl path (and -q hides all error detail)
## Summary

`scripts/install.sh`'s download helper enforces TLS on the curl path but not on the wget fallback:

- curl path: `--proto '=https' --tlsv1.2` — an `http://` redirect or downgrade is refused. Good.
- wget path: no `--https-only` / `--secure-protocol=TLSv1_2` equivalent — a plain-HTTP redirect **is followed**, silently weakening the installer's fail-closed design exactly on the hosts most likely to take this path (minimal/embedded images where wget — often BusyBox — is the only tool present).

`-q` on the wget invocation also suppresses all error detail, so when a download does fail on such a host the operator gets no indication of why.

The checksum verification downstream does still gate what gets installed — but SHA256SUMS itself is fetched over the same downgradable channel, so a network-position attacker who can redirect both fetches to HTTP defeats the verification entirely. The TLS pin is the part of the chain that makes the checksums meaningful.

## Suggested fix

```sh
wget --https-only --secure-protocol=TLSv1_2 -nv -O "$out" "$url"
```

with a fallback note for BusyBox wget (which lacks `--secure-protocol`; it does honor `--https-only` in most builds — if neither is available the script should say so and fail closed rather than proceed downgradable). Replacing `-q` with `-nv` keeps output terse but preserves error lines.

Found during code review while deploying release 2026.0610.1 to an armv7 edge device.

ELares:
## Problem

`scripts/install.sh`'s download helper enforces TLS on the curl path (`--proto '=https' --tlsv1.2`) but the wget fallback had no equivalent: a plain-HTTP redirect was silently followed, exactly on the hosts most likely to lack curl (minimal/BusyBox images). `SHA256SUMS` is fetched over the same channel as the binary, so a network-position attacker who redirects both fetches to HTTP defeats the checksum gate entirely. The `-q` flag also suppressed all error detail, so a failed download on a wget-only host gave the operator no indication of why.

## Fix

The helper now probes `wget --help 2>&1` for HTTPS enforcement support (new `wget_tls_flags` function), keeping the helper's interface and the curl-preferred ordering unchanged:

- GNU wget: invoked with `--https-only --secure-protocol=TLSv1_2`, mirroring the curl path's scheme pin plus TLS version floor.
- A limited (BusyBox-style) build that advertises `--https-only` but not `--secure-protocol`: accepted with `--https-only` alone, still HTTPS-pinned (a plain-HTTP URL or redirect is refused); only the TLS version floor is unavailable on such a build, and the comment in the script notes that.
- A wget with no `--https-only` at all (real BusyBox): FAIL CLOSED. The helper dies with a clear error saying the available wget cannot enforce HTTPS and suggesting curl or GNU wget. The install never proceeds downgradable and no download is attempted.
- `-q` is replaced with `-nv`, so output stays terse on success but the error line survives on failure.

A CHANGELOG bullet was added under `## [Unreleased]` / `### Security`.

## How it was tested

- Four new tests in `crates/ironbus-cli/tests/installer_verify.rs`, using the existing `IRONBUS_INSTALL_SH_SOURCED=1` sourcing harness plus a stub-tool PATH (PATH is restricted to a directory of stub `curl`/`wget` scripts, so no network and the host's real tools are invisible):
  - a stub wget advertising `--https-only` and `--secure-protocol` is invoked with `--https-only --secure-protocol=TLSv1_2 -nv` (and never `-q`);
  - a stub wget WITHOUT `--https-only` in its help makes `download` fail closed: nonzero exit, the "cannot enforce HTTPS" error naming curl or GNU wget, and no download invocation and no destination file;
  - a stub wget advertising only `--https-only` is accepted and is not passed the unsupported `--secure-protocol`;
  - with both stubs present, curl is preferred (with its existing TLS pin) and wget never runs.
- `cargo test -p ironbus-cli --test installer_verify`: 12 passed (8 existing + 4 new).
- `shellcheck scripts/install.sh`: clean.

Closes #423


ELares (commented):
## Self-review record (adversarial deep review, security lens)

The panel returned **needs-changes with a blocker, and it was right**: the original approach (`--https-only --secure-protocol=TLSv1_2` on GNU wget) was empirically FALSE. On GNU Wget 1.25.0 the exact invocation downloaded a plain http:// URL (exit 0) and followed an HTTPS-to-plain-HTTP 302 (in wget 1.x, opt.https_only is consulted only by recursive mode's download_child). The four original tests pinned argv, not behavior, which is exactly how the false invariant passed CI.

**Resolved in 139066b (full rework):**
- wget2 was then tested rather than assumed, and DISPROVEN too (wget2 2.2.1: --https-only skips the command-line URL; its redirect refusal exits 0 with no failure signal; --https-enforce=hard silently falls back to plaintext when the TLS connect fails; combining the flags disables the upgrade entirely).
- Decision table now: curl (unchanged, preferred, `--proto '=https' --tlsv1.2`) or nothing. EVERY wget flavor is refused fail-closed BEFORE any fetch, classified from `LC_ALL=C wget --version`, with the empirical evidence in the code comment and a flavor-specific error naming the remedy.
- Tests now pin BEHAVIOR: three refusals (wget 1.x, wget2, BusyBox/unrecognized: nonzero exit, no fetch attempted, clear error) plus curl-preferred ordering. CHANGELOG rewritten to the honest claim.
- Panel nits landed: LC_ALL=C probe; download() header comment aligned with the die-based fail-closed path (no call site is a subshell, verified by the panel).

**Hardware record:** on the target hive (GNU wget 1.20.1, curl hidden from PATH) the reworked script refuses with the full evidence-bearing error, exit 1, nothing fetched. The BusyBox-stub refusal was also verified on-device.

**Process note:** the panel also caught that a shared CARGO_TARGET_DIR across worktrees can serve stale cross-worktree test binaries that cargo reports as fresh; all verification since uses per-worktree target dirs.

Verdict: blocker resolved with evidence; this comment is the review of record.

PR #427, opened 2026-06-10
fix(installer): failure-safe single-rename swap and idempotent same-version re-run

### LINKED ISSUE #421: [BUG] install.sh: install_binary swap is not failure-safe — a failed final mv strands the host with no binary, and the two-rename window breaks supervised restarts
## Summary

`install_binary()` in `scripts/install.sh` implements the binary swap as **two separate renames with no failure recovery**, which breaks the installer's own fail-closed contract in two ways:

1. **A failed final `mv` strands the host with NO binary.** The existing binary is first moved to `<dest>.prev`; if the final `mv "$tmp_dest" "$dest"` then fails (ENOSPC, I/O error, target dir becomes read-only, etc.), the function returns non-zero **without restoring `.prev`** — `$dest` no longer exists even though known-good bytes are sitting right next to it. An "installer that can only leave you with the old or the new binary" should never be able to leave you with neither.

2. **The two-rename window is a real race on supervised hosts.** Between `mv $dest $dest.prev` and `mv $tmp_dest $dest` there is a moment when `$dest` does not exist. On a host where the broker runs under systemd with `Restart=`, a crash/restart landing in that window fails to exec. The docs for the `ironbus upgrade` verb describe the correct ordering (stage sibling temp → fsync → retain `.prev` → **single atomic rename**); install.sh should match it: `cp` (or hardlink) the current binary to `.prev` first, then do exactly one `rename()` over `$dest`.

## Suggested fix

```sh
# retain rollback copy WITHOUT unlinking the live binary
cp -p "$dest" "$dest.prev" 2>/dev/null || true   # or ln + mv for atomicity of .prev itself
# single atomic swap; on failure the live binary is untouched
mv -f "$tmp_dest" "$dest" || { echo "install failed; existing binary untouched" >&2; return 1; }
```

This removes both the stranded-host failure mode and the no-binary window in one move. (Alternatively, have install.sh delegate the swap to the already-correct `ironbus upgrade --new-binary "$tmp_dest" --dest "$dest"` when a prior binary exists.)

Found during code review while deploying release 2026.0610.1 to an armv7 edge device.

### LINKED ISSUE #422: [BUG] install.sh: re-running the installer for the same version clobbers <dest>.prev, destroying the only rollback copy
## Summary

Re-running `scripts/install.sh` for the **same version** destroys the only rollback copy. `install_binary()` unconditionally retains the current binary as `<dest>.prev` before installing the new one, so:

1. Upgrade v1 → v2: `.prev` = v1 (correct, rollback available).
2. Re-run the installer for v2 (idempotent re-provision, config-management convergence run, a retry after a partial failure elsewhere in the script, etc.): `.prev` is overwritten with v2 — **bytes identical to the live binary** — and the v1 rollback copy is gone.

The fall-back-after-N mechanism in the systemd unit (and `ironbus rollback`) then "recovers" a failing v2 by reinstalling v2.

This matters most on exactly the fleet this project targets: edge nodes managed by configuration management that re-runs installers idempotently.

## Suggested fix

Compare checksums before the swap: if the staged binary's SHA256 equals the live binary's, skip both the `.prev` retention and the install (report "already at <version>, nothing to do"). The script already has SHA256 plumbing for asset verification, so the comparison is nearly free. (The same guard belongs in `ironbus upgrade` if it doesn't have one.)

Found during code review while deploying release 2026.0610.1 to an armv7 edge device.

ELares:
## Problem

`install_binary()` in `scripts/install.sh` broke the installer's own fail-closed contract in two ways, both in the same swap sequence:

1. **#421: the swap was two renames with no failure recovery.** The live binary was first moved to `<dest>.prev`; if the final `mv "$tmp_dest" "$dest"` then failed (ENOSPC, IO error, directory turned read-only), the function returned non-zero without restoring `.prev`, leaving the host with NO binary at `$dest` even though known-good bytes sat right next to it. Between the two renames there was also a window where `$dest` did not exist, so a systemd `Restart=` respawn landing in that window failed to exec.
2. **#422: a same-version re-run destroyed the only rollback copy.** Re-running the installer for the version already live unconditionally overwrote `<dest>.prev` with bytes identical to the live binary (upgrade v1 to v2, re-run v2: `.prev` becomes v2, v1 is gone), so the fall-back-after-N mechanism would "recover" a failing v2 by reinstalling v2.

## Fix

One coherent rework of `install_binary` (plus a small `main` hook), keeping the POSIX sh, fail-closed, heavy-comment conventions:

- **Same-version guard first.** If a binary already exists at `$dest` and its SHA256 equals the staged binary's (reusing the existing `sha256_of` tool detection, `sha256sum` or `shasum`), the run is an idempotent re-provision: skip both the `.prev` retention and the swap, print "ironbus \<version\> already installed at \<dest\>, nothing to do", and exit 0. If no sha256 tool exists the guard falls through to a normal install, never skipping on an ambiguous answer.
- **Copy-based retention that never unlinks the live binary.** `cp -p "$dest"` to a sibling temp, then one atomic `mv` onto `<dest>.prev`, so `.prev` itself is replaced atomically and the live binary is never moved. This mirrors how `ironbus rollback` was hardened in #348 to never move the live binary onto `.prev`.
- **Exactly one atomic swap.** A single `mv -f "$tmp_dest" "$dest"` is now the only operation that changes `$dest`. If it fails, the live binary is untouched (the retention copied it instead of moving it), so the failure path just reports and returns non-zero with the host still healthy. There is no longer any moment where `$dest` does not exist.

## How it was tested

- `cargo test -p ironbus-cli --test installer_verify`: 11 passed (8 existing plus 3 new). The new tests source the real script via the existing `IRONBUS_INSTALL_SH_SOURCED=1` harness:
  1. a same-version re-run preserves `ironbus.prev` byte-for-byte and exits 0;
  2. the rollback copy still holds the OLD version's bytes after an upgrade followed by a same-version re-run;
  3. a failed swap leaves the original binary present at `$dest`: first with the destination's parent directory made read-only after the original binary is in place (skipped under root), then by failing exactly the final rename via an `mv` override, the ENOSPC case from #421.
- Teeth check: with `scripts/install.sh` reverted to main, all 3 new tests fail; with the fix they pass.
- `shellcheck scripts/install.sh`: clean.

Closes #421
Closes #422


ELares (commented):
## Self-review record (adversarial deep review, installer correctness lens)

Panel verdict was **needs-changes**; all findings resolved in 813652a and 46a9959. Disposition:

**Resolved:**
- Acceptance-gate failure (4 CI jobs): the golden-path test "upgraded" with byte-identical bytes, which the new #422 guard correctly no-ops. Step 10 now stages a byte-distinct, never-executed intermediate (macOS code-signing safe), pins the genuine upgrade's `.prev`, adds an acceptance-level same-version no-op pin, then upgrades back to the pristine binary for the post-upgrade execution assertions.
- Reproduced should-fix: the `.prev` commit happened BEFORE the final swap, so a failed swap (ENOSPC) destroyed a pre-existing good rollback copy (dest=v2, prev=v1, failed v3 left prev=v2). The commit is now AFTER the swap lands; a failed swap leaves dest AND the old `.prev` untouched. The deliberate trade (a crash between swap and commit leaves `.prev` one version stale, an older known-good) is documented in the contract comment and pinned by a new test with a pre-existing distinct-bytes `.prev`.
- Rust twin parity: `atomic_swap_with_prev` had the exact #421 pattern (rename-based retention, stranded-host on failed final rename, dest-missing window) and no #422 guard, while module docs and DISTRIBUTION.md claimed parity. The verb now has copy-staged retention, fsync, single atomic dest rename, then `.prev` commit, plus a byte-compare SkippedSameVersion no-op; failure paths never remove dest nor clobber the old `.prev`. The #348 rollback hardening contract is untouched (its power-cut re-entry tests still pass). Parity docs are true again, and DISTRIBUTION.md states honestly that the shell flow cannot portably fsync; the verb is the durably-synced path.
- Temp hygiene: the staging-cp failure branch removes its partial temp; main's EXIT trap also clears install-dir temps.
- Nits: unreadable-dest fail-closed behavior change documented; symlink-dest semantics documented; same-version message reads "ironbus (latest)" when unpinned; failed-swap test comment now states exactly which phase has #421 teeth.

**Hardware record (RPi4, real ARM cp/mv):** fresh install, genuine upgrade (.prev holds the prior bytes), and same-version no-op (.prev fingerprint preserved) verified on-device via the sourced harness, plus the real-network same-version no-op against the live GitHub release. The failed-swap path is pinned by the test suites on Linux and macOS.

**Spawned issues:** #433 (installer dest /usr/local/bin vs unit's /usr/bin split, found during hardware testing).

Verdict: findings resolved and pinned; this comment is the review of record.

PR #428, opened 2026-06-10
fix(engine): retention protect floor skips a never-touched default group

### LINKED ISSUE #424: [BUG] Retention is silently a no-op for named-group-only deployments: the always-present default group pins the protect floor at offset 0
## Summary

For any deployment that consumes only via **named** groups, every retention bound (`--max-retained-bytes`, `--max-age-ms`, `--max-messages`) is silently a **no-op**, and the disk fills to `--max-total-bytes` and stays there shedding every produce (`drop-new`) — the slow-motion version of exactly the outage retention was configured to prevent.

Cause: the retention protect floor is `min_committed_offset()` (`crates/ironbus-server/src/engine.rs`), the minimum committed offset across **every** work-group — and the default group `""` always exists. The code comment is explicit:

> The default group (`""`) always exists, so the iterator is never empty; a fresh group sits at offset 0, which keeps the floor at 0 (reaping nothing) until it has consumed something, exactly the safe behavior.

"Safe" per-group, but combined with "the default group always exists and is never evicted," a deployment whose consumers all use `--group <name>` has a permanent phantom consumer pinned at offset 0. Observed live on an edge node (2026.0610.1, armv7): after publishing and fully draining a named group, metrics still showed `ironbus_consumer_lag_records{consumer=""} 6` — the floor was pinned at 0 by a group nothing would ever consume.

There is also no online way out: `cumulative-ack` is broadcast-only (and `""` cannot be broadcast), and `admin consumer-reset` is offline-only, so the only mitigations are "stop the broker and reset the default group's cursor" or "remember to drain `""` periodically."

## Asks (any subset)

1. **Document it loudly** in USAGE.md "Bounding disk use": today nothing warns that named-group-only usage disables retention.
2. **Exempt the default group from the protect floor while it is virgin** (committed == 0 **and** never subscribed/leased): a group no consumer has ever touched arguably shouldn't pin the floor; the moment anything subscribes to `""` it gets full protection. This keeps the conservative behavior for real default-group consumers.
3. Alternatively an explicit opt-out (`--retention-exempt-group ""` or similar) so the operator states the intent.
4. A startup (or periodic) WARN when retention is enabled but the floor has been pinned at 0 by a group with no activity for the whole window — the observability hook that would have surfaced this immediately.

Happy to discuss; (2) + (1) seems like the right default for an edge box where "disk quietly stops reclaiming" is a field-visit-grade failure.

ELares:
## Problem

The retention protect floor is the minimum committed offset across every work group, and the boot-created default group `""` always exists. A deployment that only consumes through NAMED groups therefore has a phantom consumer pinned at offset 0 forever:

- every retention bound (`--max-retained-bytes`, `--max-age-ms`, `--max-messages`) is silently a no-op,
- the log grows to `--max-total-bytes`,
- then `drop-new` sheds every produce: the slow-motion version of exactly the outage retention was configured to prevent.

Observed live on an armv7 edge node running 2026.0610.1: after fully draining a named group, `ironbus_consumer_lag_records{consumer=""}` still showed the whole log, and the floor sat at 0.

## Fix

A `WorkGroup` now carries a `touched` flag.

- `touched` starts `true` for every group except the boot-created default group with NO durable consumer state (no `cursor.ckpt`, no `attempts.ckpt`). A named group only ever comes to exist through a consumer op or a durable-state resume, so it is always touched.
- Any poll, ack, nack, term, progress, cumulative ack, or explicit subscribe marks the group touched (the touch sites are exactly the existing `last_activity` stamp sites plus `subscribe_in`).
- `min_committed_offset` skips untouched groups. With no touched group at all the floor is the durable head (no consumer has ever shown intent, so every sealed record is reapable once a bound trips).
- A touched group keeps the unchanged slow-consumer protection, including across restarts via its durable checkpoint.

Documented trade (no new durable state added): a default-group consumer whose commits were never yet checkpointed resumes untouched after an UNCLEAN restart, until it interacts again. A graceful stop checkpoints, and the cadence (`checkpoint_interval`) bounds the window. This is called out in the field doc comment and in USAGE.md.

This implements asks (1) and (2) of #424. Ask (4), a pinned-floor WARN, is observability follow-up and is not in this PR; ask (3), an explicit opt-out flag, is unnecessary once (2) exists.

## Tests

Four new engine tests pin the contract:
- `an_untouched_default_group_does_not_pin_retention`: named-groups-only consumption reaps past the bound.
- `a_polled_default_group_pins_retention_like_any_consumer`: ONE default-group poll restores the pre-change pinning.
- `an_untouched_default_group_stays_untouched_across_reopen`: a restart does not resurrect the phantom consumer.
- `a_checkpointed_default_group_resumes_touched_across_reopen`: a real default-group consumer's unconsumed record survives retention across a restart.

`cargo test --workspace --all-features --locked` green locally (350 ironbus-server lib tests), `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` clean, `cargo fmt --all --check` clean.

## Docs

USAGE.md (Bounding disk use) documents the touched rule and the restart nuance, and corrects a stale claim in the same paragraph: it said no disk-full drop-oldest policy exists, but `--disk-full-policy drop-oldest` shipped in #82.

Closes #424


ELares (commented):
## Self-review record (adversarial panel, two lenses: engine invariants, test quality)

Both lenses returned needs-changes; all findings are now resolved in 6790235 or tracked. Full disposition:

**Resolved in this PR:**
- Flaky test (invariants lens, reproduced 2 in ~30 full-suite runs): the worker drain helper silently swallowed `Poll::Truncated`, ending the drain early. The drain is now exhaustive (Truncated continues, unknown outcomes panic). 10 consecutive isolated runs and a full-suite run pass; the strictness also surfaced WHY test 3 needs it (the re-created worker takes a one-time #84 truncation reset after reopen, by design).
- Unpinned head fallback (test lens, mutation-verified: `unwrap_or(0)` passed the whole suite): new test `a_producer_only_deployment_reaps_with_no_consumer_at_all` pins it.
- Unpinned subscribe touch (test lens, mutation-verified): new test `a_subscribed_default_group_pins_retention_before_its_first_poll` pins the only touch path that can mark the default group before its first poll.
- Overstated checkpoint-cadence claim: the field doc, USAGE.md, and CHANGELOG now state the real mechanics (offset-gated live write via `maybe_checkpoint`; the shutdown write skips a cursor with nothing to record, so a polled-but-never-committed consumer resumes unprotected even after a graceful stop).
- Factually wrong field-doc claim about named-group creation: serve-time declared broadcast/key-shared groups exist before any consumer and deliberately keep the conservative pinning; the doc and the constructor comment now say so.
- "Cumulative ack" dropped from the default-group touch lists (the default group can never be broadcast, so the op is impossible there; kept for named broadcast groups).

**Accepted as-is (panel rated nit, rationale recorded):**
- The other four touch sites (cumulative_ack_in, nack_in, progress_in, poll_in_member) are unpinned by tests; for the default group none is reachable as a FIRST interaction (a lease token requires a prior poll, which already touches; cumulative ack is impossible there).

**Tracked separately, not in this PR:**
- Idle-eviction interaction (invariants lens): an evicted group's durable checkpoint does not pin the floor, so a returning consumer can truncate. Pre-existing for deployments with an active default group; #428 extends reach to named-only deployments. The ghost-checkpoint semantics are a design decision: filed as #432 with three options.

Hardware note: the underlying footgun was observed live on an armv7 edge node (see #424); post-merge validation on the same hardware is planned with tiny retention bounds on a scratch data dir before #424 is closed.

Verdict: the panel's blocking findings are resolved and mutation-verified locally; CI is 28/28. Approving formally is not possible on one's own PR; this comment is the review of record.

PR #429, opened 2026-06-10
docs(staleness): fix docs and help text that lag shipped features

ELares:
## Problem

The project releases several times a day and several docs (and one block of CLI help text) still describe shipped features as nonexistent, or describe the pre-implementation design instead of what landed:

- `docs/EDGE_TUNING.md` said there is no `--profile` flag, no TOML config file, no reload, that the `interval`/`none` durability modes are not exposed as `serve` flags, and that the #115 RAM refuse-to-boot guard is unwired. All of those shipped (#87, #382, #341/#379, #115).
- `docs/WAL.md` said preallocation does not exist (#330 shipped it), that the fsync-headroom admission credits are absent (#378 shipped them), and that "edge profile" is not a profile (#87 shipped `--profile edge-tiny`).
- `docs/CONFIG.md` quoted CONTRACTS.md as saying "There is NO TOML config document in the implementation today" (CONTRACTS.md was already updated to IMPLEMENTED #382) and said on-disk compression "is not landed" without mentioning the shipped `--compression` knob.
- The `docs/README.md` index still labeled PREALLOCATION, COMPACTION, DICTIONARY_LIFECYCLE, and BACKPRESSURE as "specified, not yet implemented" (all implemented: #330, #337, #357, #336/#402), and `docs/CLI_CONTRACT.md` claimed the binary "maps only the six exit codes 0/1/2/4/5/70" (exit 3 ships for scrub/repair, with versioned `--json` schemas).
- The root README compression row said "per-batch" and "dictionaries deferred" (shipped design is a per-record descriptor; dictionaries shipped opt-in behind the `zstd` feature, #357).
- The CLI help text near `dump --raw` said "On-disk compression (#12) is not yet implemented (the codec is always none)". Verification showed this is SUBTLE: the codec runtime (#387) and the `serve --compression <none|lz4>` knob (default `lz4`) DID ship, and the materialized-config line echoes `compression=lz4`, but the serve write path never invokes the runtime (`compress_payload` has no callers in `ironbus-server`/`ironbus-storage`; `EngineConfig` has no compression field), so records are in fact still stored raw. The help text was factually right about bytes on disk but silent about the shipped-but-unwired knob, which is confusing next to a broker logging `compression=lz4`.

## Fix

[...]

PR #434, opened 2026-06-10
docs(signals): state that SIGHUP is graceful stop and reload is restart-only today

### LINKED ISSUE #431: [BUG] Docs claim SIGHUP reloads config, but SIGHUP is graceful stop: an operator following CONFIG.md takes a production broker down
## Summary

CONFIG.md and CONTRACTS.md state the SIGHUP config re-read reload path is wired, but in code SIGHUP is bound to graceful stop (the #195 disentanglement residual, per the `cmd_serve` comment), and the re-read reload exists as a startup self-check plus a callable engine that no signal invokes. The docs overclaim in the shipped direction.

Found during the #429 docs staleness sweep and deliberately left out of that PR because it is an overclaim (docs ahead of code) rather than a stale claim (docs behind code), and the right fix is a decision, not a wording tweak:

1. Bind the reload engine to a signal or admin verb (the #88/#380 config-reload surface) and keep the docs, or
2. correct CONFIG.md and CONTRACTS.md to say SIGHUP is graceful stop today and the reload engine is not yet reachable at runtime.

Either way, today an operator following CONFIG.md will SIGHUP a production broker expecting a config reload and get a graceful shutdown instead, which on an edge node behind `Restart=on-failure` (a CLEAN stop) means the broker stays DOWN until manually started.

refs #88, refs #380, refs #195

ELares:
## Problem

CONFIG.md and CONTRACTS.md stated that the SIGHUP config re-read reload path is wired. In code (`crates/ironbus-cli/src/main.rs`), SIGHUP is bound to graceful stop (the #195 disentanglement residual: ctrlc's `termination` feature covers SIGINT/SIGTERM/SIGHUP with one shutdown-flag handler), and the validate-whole-then-swap re-read engine (`config_reload.rs`) runs exactly once, as a startup self-check; no signal or runtime trigger invokes it. The docs overclaimed in the shipped direction: an operator following them would `kill -HUP` a production broker expecting a reload and get a clean stop instead, and under the packaged unit's `Restart=on-failure` a clean stop (exit 0) is not restarted, so the broker stays down until started again.

This implements resolution option 2 of #431: the docs state the truth everywhere. PR #429 already corrected `docs/CONFIG.md` and `docs/EDGE_TUNING.md`; this PR fixes the remaining spots. Docs and comments only, no behavior change. The runtime reload trigger stays tracked in #380 (refs #88, refs #195).

## Fix

Corrected the remaining overclaims:

- `docs/CONTRACTS.md` (config-models intro): "the hot/cold/coupled reload classes are wired for the SIGHUP/re-read path" now says the classes are enforced by the re-read engine, which runs once as a startup self-check, with no runtime trigger wired (SIGHUP is graceful stop, #431; trigger is #380).
- `docs/CONTRACTS.md` (implementation-status Config note): "the safe SIGHUP/re-read reload ship" now lists the runtime reload trigger as a deferred residual alongside the #106 mutating verbs.
- `docs/CONTRACTS.md` (#137-draft discrepancy entry): "the only deferred half is the authed mutating wire verbs" now names both deferred halves (the #380 runtime trigger and the #106 verbs).
- `docs/CLI.md` (env-mapping residual note): the re-read RELOAD engine is implemented as a startup self-check; applying a config change requires restarting the broker.
- `crates/ironbus-cli/src/main.rs` (`mod config_reload` doc comment): "this is the safe SIGHUP/re-read reload path only" now says it is the safe local re-read path, run once as a startup self-check, with no signal invoking it at runtime (comment only).

Made the SIGHUP=stop behavior explicit where signal handling is described, with the `Restart=on-failure` warning:

- `docs/CLI.md` (`serve`): SIGHUP is NOT a config reload; under a supervisor that restarts only on failure, a SIGHUP'd broker exits 0 and stays down.
- `docs/USAGE.md` (Stopping and restarting): all three signals mean stop, none means reload; config changes are edit-and-restart; under the packaged unit a clean stop is not restarted.
- `docs/DISTRIBUTION.md`: new short "Signals and config changes (the unit)" note next to the unit wiring (the section that documents `Restart=on-failure`).
- `README.md` quick start: SIGHUP also stops, it does not reload config; a clean stop stays down under the unit.
- `CHANGELOG.md`: Fixed bullet under [Unreleased].

## Verification

- `sh scripts/ci/relative-link-check.sh`: inspected ~755 in-repo links across 44 files, 0 problems.
- `sh scripts/ci/changelog-unreleased.sh`: ok, [Unreleased] non-empty.
- `cargo fmt -p ironbus-cli --check`: clean (the main.rs change is a doc comment only).
- Re-grepped `SIGHUP`/`reload` across docs/, README.md, and main.rs: every remaining mention is either the corrected truth or future-work framing (#380/#88/#106).
- No em or en dashes introduced (diff scanned).

Closes #431


ELares (commented):
## Self-review record (adversarial fact-check panel)

Panel verdict was needs-changes; resolved in 2dddf3d. Disposition:

- **Blocker (panel caught real damage):** the CHANGELOG edit had REPLACED the existing #421/#422 installer parent bullet instead of inserting a sibling, erasing "Closes #421, Closes #422" from the changelog and misattributing the three installer sub-bullets to a docs-only change. Restored.
- **Should-fix:** the two rustdoc overclaims in `config_reload.rs` (module doc and `reload_from`, both labeling it "the SIGHUP/re-read path") survived because the verification grep skipped `crates/` beyond main.rs. Both now state that no signal invokes the reload and the startup self-check is the only caller.
- **Nit:** "runs exactly once" tightened to "at most once" in three places; the self-check is gated on `--config`.

The panel verified every corrected passage against code (signal handler, the single gated `reload_effective_config` call site, the unit's `Restart=on-failure`) and confirmed the stays-down warning is present at all four operator surfaces. Link check 0 problems, changelog gate ok. This comment is the review of record.

ELares (commented):
## Self-review record (adversarial fact-check panel)

Panel verdict was needs-changes; resolved in 2dddf3d. Disposition:

- **Blocker (panel caught real damage):** the CHANGELOG edit had REPLACED the existing #421/#422 installer parent bullet instead of inserting a sibling, erasing "Closes #421, Closes #422" from the changelog and misattributing the three installer sub-bullets to a docs-only change. Restored.
- **Should-fix:** the two rustdoc overclaims in `config_reload.rs` (module doc and `reload_from`, both labeling it "the SIGHUP/re-read path") survived because the verification grep skipped `crates/` beyond main.rs. Both now state that no signal invokes the reload and the startup self-check is the only caller.
- **Nit:** "runs exactly once" tightened to "at most once" in three places; the self-check is gated on `--config`.

The panel verified every corrected passage against code (signal handler, the single gated `reload_effective_config` call site, the unit's `Restart=on-failure`) and confirmed the stays-down warning is present at all four operator surfaces. Link check 0 problems, changelog gate ok. This comment is the review of record.

PR #435, opened 2026-06-10
feat(installer): IRONBUS_INSTALL_DEST override and a unit-path mismatch warning

### LINKED ISSUE #433: [BUG] install.sh installs to /usr/local/bin while the systemd unit and its rollback machinery operate on /usr/bin: silent version split
## Summary

The fail-closed installer and the shipped systemd unit disagree about where the binary lives, and nothing warns about it:

- `scripts/install.sh` installs to `/usr/local/bin/ironbus` when run as root (first writable choice), `~/.local/bin` otherwise.
- `packaging/systemd/ironbus.service` hardcodes `ExecStart=/usr/bin/ironbus serve` and `Environment=IRONBUS_BIN=/usr/bin/ironbus` (the `.deb` path), and the whole fall-back-after-N mechanism (`record-start`, `rollback`, `ironbus.prev`) operates on that path.

Observed end-to-end on a Raspberry Pi 4 gateway: the script cheerfully installed to `/usr/local/bin/ironbus` while the running unit kept executing `/usr/bin/ironbus`. Consequences for an operator following the README one-liner and then adopting the repo unit:

1. The unit either fails to start (no `/usr/bin/ironbus`) or silently keeps running a DIFFERENT binary than the one just installed or upgraded.
2. `/usr/local/bin` precedes `/usr/bin` on a default PATH, so interactive `ironbus` and the service's broker can be two different versions indefinitely; `ironbus upgrade` runs against whichever the operator happens to name, and the unit's rollback machinery never sees script-side upgrades.

## Asks (any of)

1. Teach install.sh to detect the unit (`systemctl cat ironbus` or the file's presence) and either install to `IRONBUS_BIN`'s path or warn loudly that the service uses a different binary.
2. Honor an `IRONBUS_INSTALL_DEST` env override and document pairing it with the unit (`IRONBUS_INSTALL_DEST=/usr/bin/ironbus`).
3. At minimum, document the split in docs/DISTRIBUTION.md (script = /usr/local/bin standalone; .deb + unit = /usr/bin) and make the unit's docs say script users must override IRONBUS_BIN and ExecStart together.

refs #421, refs #420

ELares:
## Problem

The fail-closed installer and the shipped systemd unit disagree about where the binary lives, and nothing warns about it (#433). `scripts/install.sh` installs to `/usr/local/bin/ironbus` (first writable choice) or `~/.local/bin`, while `packaging/systemd/ironbus.service` hardcodes `ExecStart=/usr/bin/ironbus serve` and `Environment=IRONBUS_BIN=/usr/bin/ironbus`, where the whole fall-back-after-N machinery (`record-start`, `rollback`, `ironbus.prev`) operates. An operator following the README one-liner next to the packaged unit upgrades a binary the service never executes; `/usr/local/bin` usually precedes `/usr/bin` on PATH, so the interactive CLI and the running broker can be two different versions indefinitely.

## Fix

All three asks of the issue:

1. **`IRONBUS_INSTALL_DEST` env override** (new `resolve_install_dest` in `scripts/install.sh`): when set it is the FULL destination path for the binary (e.g. `/usr/bin/ironbus`), bypassing the default `/usr/local/bin` vs `~/.local/bin` selection and `--bin-dir`. Validated fail-closed, each failure dying with an error naming the problem: must be an absolute path, must not be (or syntactically name) a directory, and its parent directory must exist and be writable. The override flows through `install_binary` unchanged, so the same-version no-op guard and the `.prev` rollback retention operate on the override path exactly as on a default one. It is also documented in the script header, so `--help` prints it.
2. **Unit-awareness warning** (`unit_exec_binary` + `warn_if_unit_binary_differs`): after dest selection (override or default), the installer detects the unit's ExecStart binary cheaply and POSIX-safely: `systemctl cat ironbus` when `command -v systemctl` succeeds (mirroring systemd's later-wins/empty-resets override semantics), falling back to `/etc/systemd/system/ironbus.service` then `/lib/systemd/system/ironbus.service` if readable. The extracted path is the first word after `ExecStart=` with a leading `+` stripped, parsed with shell builtins only (no grep/awk/sed), so the detection survives the test harness's restricted PATH. A found unit binary that differs from the chosen dest prints a LOUD multi-line warning to stderr (never an error; exit 0 continues) naming both paths, the version-split consequence, and the remedy: `IRONBUS_INSTALL_DEST=<unit path>`. No systemctl and no unit file = no warning, so non-systemd hosts are unaffected.
3. **`docs/DISTRIBUTION.md`**: a new subsection in the installer section documents the script-vs-`.deb` destination split explicitly and the `IRONBUS_INSTALL_DEST` pairing for unit users, including the sudo one-liner.

`CHANGELOG.md` gets the bullet under `[Unreleased]` / `### Added`.

## Tests

New in `crates/ironbus-cli/tests/installer_verify.rs` (all 21 tests in the harness pass):

- `an_ironbus_install_dest_override_is_honored_end_to_end`: the binary lands exactly at the override through the same resolve-then-`install_binary` sequence `main` runs; a same-version re-run no-ops there (exit 0, "nothing to do", no fabricated `.prev`).
- `a_relative_ironbus_install_dest_is_refused_with_nothing_installed` and `a_directory_ironbus_install_dest_is_refused_with_nothing_installed`: each dies with the clear error naming the variable and the problem; nothing is installed.
- `a_unit_running_a_different_binary_triggers_the_loud_mismatch_warning`: a stub `systemctl` (the only tool on a restricted PATH) whose `cat ironbus` prints `ExecStart=+/usr/bin/ironbus serve` (exercising the `+` strip, the first-word cut, and the ExecStartPre non-match) triggers the stderr warning naming both paths and the `IRONBUS_INSTALL_DEST=/usr/bin/ironbus` remedy, with exit 0.
- `no_systemctl_and_no_unit_file_means_no_mismatch_warning`: restricted PATH with no systemctl produces no output at all.

Gates: `shellcheck scripts/install.sh` clean, `sh -n` clean, `cargo test -p ironbus-cli --test installer_verify --locked` 21/21, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` clean, `cargo fmt --all --check` clean.

Closes #433


ELares (commented):

Panel verdict was needs-changes with three should-fixes, all empirically verified; all resolved in 3cb1739:

- **Prefix stripping:** the ExecStart parser handled only "+", so any other systemd Exec prefix ("@", "-", ":", "!", "!!", stackable) produced a spurious LOUD warning whose printed remedy was itself rejected by the validator (reproduced end to end by the reviewer). The parser now strips all stacked prefixes via a POSIX-safe bracket class, verified identical in dash, bash, and sh. The inaccurate "only prefix this repo's unit uses" comment is corrected.
- **Hermetic detection:** the file fallback hardcoded the real /etc and /lib unit paths, so the no-warning test would fail on any host with the .deb installed (exactly the hives this project targets) and the fallback branch had zero coverage. IRONBUS_UNIT_FILES now overrides the candidate list (documented as test/advanced surface); every warning test pins it into its own tempdir; new tests cover the fallback branch (Environment lines, an ExecStart= reset, stacked prefixes) and the inert-when-absent case.
- **Toothless assertions (mutation-proven):** the warning test's substring assertions survived deleting the first-word cut. Assertions are now exact-line matches plus a no-space check plus ExecStartPre teeth (/bin/false sentinels). Mutation re-check: deleting the cut now fails 2 tests on the exact-remedy-line assertion.
- **Nits landed:** whitespace tolerated around the unit key's "=", leading value whitespace trimmed (was a silent reset), IRONBUS_INSTALL_DEST validated eagerly before any download, the systemctl-may-block trade accepted in a comment, --help sed range updated with a drift-risk note.

Hardware validation of the override and warning runs on the hive before merge. This comment is the review of record.

PR #436, opened 2026-06-10
fix(engine): an idle-evicted group's ghost checkpoint keeps pinning the retention floor

### LINKED ISSUE #432: [DESIGN] Idle-evicted groups do not pin the retention protect floor: returning consumers can truncate; decide ghost-checkpoint semantics
## Summary

`--group-idle-evict-ms` and the retention protect floor interact so that an idle-evicted group's durable position does not pin the floor, even though its `cursor-<hex>.ckpt` sits on disk declaring it. `evict_group` removes both the in-memory group and its `group_last_checkpointed` entry, so `min_committed_offset` no longer sees it; every segment sealed after the eviction becomes reapable once a bound trips, and a briefly-offline consumer (idle longer than the evict window) can return to a one-time truncation and miss the records produced during its absence.

This is not new with #424/#428: pre-#424, any deployment whose default group was actively consumed and caught up had the same window (floor = head once the idle group was evicted). #428 extends the reachable cases to named-groups-only deployments (where the pinned-at-0 default group previously masked it). Surfaced by the adversarial review of #428; filing separately because the right semantics are a design decision, not a bug fix:

- If an evicted group's ghost checkpoint pins the floor, an evicted-forever group pins retention at its eviction point indefinitely, which is the same disease #424 cured (a phantom consumer holding every retention bound).
- If it does not pin (today's behavior), `NEVER-LOSE-COMMITTED-POSITION` holds (the group resumes exactly at its checkpoint and redelivers nothing it acked) but records produced during the eviction window are not retained for it, which softens what "committed position" buys an intermittent consumer under retention.

## Options

1. Document today's semantics loudly next to `--group-idle-evict-ms` (an evicted group does not hold retention; pair the flag with bounds sized for the longest expected consumer absence).
2. Keep the `group_last_checkpointed` entry on eviction and have the floor take the min over live touched groups plus evicted ghosts, with an explicit unsub (which already reclaims the slot immediately) also releasing the floor pin: pinning becomes opt-out by explicit unsubscribe rather than implicit by absence.
3. A `--retention-protect-evicted-ms` window: an evicted ghost pins the floor for at most this long, bounding both the truncation risk and the phantom-pin risk.

Option 2 plus the existing explicit-unsub release seems closest to the project's never-lose-position posture, but all three are defensible.

refs #424, refs #277, refs #240, refs #13

ELares:
## Problem

`--group-idle-evict-ms` and the retention protect floor interacted so that an idle-evicted group's durable position stopped pinning the floor: `evict_group` removed both the in-memory group and its `group_last_checkpointed` entry. Every segment sealed after the eviction became reapable once a bound tripped, so a briefly-offline consumer could return to a one-time truncation and miss the records produced during its absence. A restart already re-pinned the durable cursor (recovery resumes every checkpoint as a live touched group), so runtime and restart semantics disagreed. Design options were laid out in #432; this implements option 2 plus the explicit-unsub release.

## Fix

- A new private `GhostPolicy` decides the `group_last_checkpointed` entry's fate at eviction. The idle sweep KEEPS it: the entry becomes a ghost that pins the floor at the eviction-point head, so eviction reclaims memory, never retention protection.
- `min_committed_offset` now chains live touched groups with ghost entries (entries whose group is not live). Ghosts are touched by construction (they carried durable state). The producer-only fallback (no touched groups, no ghosts: floor = durable head) is unchanged.
- A returning group supersedes its ghost: `ensure_group` resumes the cursor at exactly the ghost's value and inserts the live group, so the floor follows the consumer's live progress immediately.
- An explicit Unsub RELEASES the pin via `evict_group_if_idle`: the live-group path evicts with `Release` (byte-identical to before), and the not-live path now removes a leftover ghost (reachable when the sweep evicted the group while the connection stayed subscribed). Pinning is opt-out by unsubscribe, never implicit by absence. The release is in-memory only; the checkpoint file is never deleted, so a restart conservatively re-pins.
- Documented limits: the pin binds only the consumer-safe reaper (`drop-oldest` force-reaps still ignore the floor), and an evicted-forever group pins at its eviction head until an unsub or return (the deliberate option-2 trade, called out in the CLI help and USAGE.md).

## Tests

Four new engine tests, all driving the sweep through a drained named driver group so the default group stays untouched (#424 interplay): the ghost pins and an explicit unsub releases (durable bytes cross the bound exactly at the release), a returning group supersedes its ghost and the floor follows it back to the bound, a reopen resumes the evicted group live and touched at its checkpoint, and the producer-only head fallback still reaps with no ghosts. Plus a no-ghost assertion added to the existing explicit-unsub test. `cargo test -p ironbus-server --lib` 348 green; workspace clippy pedantic clean; fmt clean.

Closes #432


ELares (commented):
## Self-review record (adversarial engine-invariants review)

Panel verdict: **approve**, with four follow-ups, all resolved in c6285e6 or accepted with rationale:

- **Should-fix (resolved):** the resume-at-ghost supersede claim overreached. It holds for the consumer paths (`poll_in` -> `ensure_group` resumes the cursor at exactly the ghost's value), but the serve-flag declared-group paths (`set_key_ordering_in`, `set_broadcast_in`) create an absent group fresh at offset 0 without resuming, superseding the ghost with a LOWER (more conservative, safe-direction) pin. The `ensure_group` comment and the floor doc now scope the claim precisely. No code change: the floor can only move down on that path, never up, and eviction's caught-up-at-head requirement means no regressing checkpoint can be written.
- **Question (accepted, documented):** any wire connection can release any ghost by name (SUB then UNSUB on an absent group). That matches the pre-#106 trust model, where an unauthenticated peer can already produce, consume, and ack on any group; per-group rights land with the authed surface. Stated in the unsub arm's comment.
- **Nit (accepted, documented):** the ghost set is unbounded by `max_groups`; its bound is the distinct group names ever checkpointed, the same class of bound as the on-disk checkpoint files that open already loads unboundedly by design. Stated in the floor doc.
- **Nit (resolved):** USAGE.md and the CLI help now state that an unsub release is in-memory only, so a restart re-pins the renounced group at its durable checkpoint.

The panel mutation-tested the new tests (reverting the Keep policy and the unsub release each fail their pinning tests) and verified the drained-driver pattern does not mask the claims. Formal approval is impossible on one's own PR; this comment is the review of record.

PR #437, opened 2026-06-10
feat(engine): wire serve --compression into the write path

### LINKED ISSUE #430: [BUG] serve --compression is inert: the codec runtime has no write-path callers, yet materialized-config echoes compression=lz4
## Summary

`serve --compression` (default `lz4`) is accepted, validated, and echoed in the startup materialized-config line (`compression=lz4`), but the knob is inert: the codec runtime that #387 shipped has no callers on the write path. `compress_payload` has zero call sites in `ironbus-server` and `ironbus-storage`, and `EngineConfig` carries no compression field, so every record is stored raw (codec id none) regardless of the flag.

Observed on a deployed 2026.0610.1 broker (armv7 edge node): the materialized-config line reports `compression=lz4` while `dump --raw` decodes every record as uncompressed.

This is worse than a missing feature: the broker tells the operator compression is on. An operator sizing flash or bandwidth budgets from that line is misled, and a support engineer reading the log will chase a codec bug that does not exist.

## Asks (either resolves the issue)

1. Wire the #387 codec runtime into the produce/append path so `--compression` does what it says (the decode/resilience half already ships: an unknown-codec record is poison, never a crash), or
2. until then, stop echoing the inert knob: refuse `--compression` values other than `none` with a clear "not yet wired" error, and drop `compression=` from the materialized-config line, so the config echo never claims something the bytes on disk do not do.

PR #429 (docs staleness sweep) already corrects the surrounding docs to state precisely that the knob shipped but the write path is unwired; this issue tracks closing the gap in code.

refs #387, refs #12, refs #75

ELares:
## Problem

`serve --compression` (default `lz4`) parsed, validated, and echoed `compression=lz4` in the materialized-config line, but the knob was inert: `compress_payload` had zero call sites on the write path and `EngineConfig` carried no compression field, so every record was stored raw regardless of the flag. The broker actively told the operator something the bytes on disk did not do (issue #430, observed on a deployed 2026.0610.1 armv7 broker). PR #429 corrected the docs to say "shipped but unwired"; this PR closes the gap in code, taking the issue's option 1.

## Design

Per ADR-0003 / what #387 actually shipped: PER-RECORD compression (the 9-byte descriptor + codec stream inside the CRC-covered payload, behind `RecordFlags::COMPRESSED`), not the historical per-batch text in #12/#75.

- **One seam.** `EngineConfig::compression` (`ironbus_core::compress::Codec`) is applied in `Engine::append_no_sync`, BEFORE `append_with_policy`. Every produce entry funnels through it (single produce, the actor group-commit drain, the dedup path, the bench broker), and everything downstream is automatically correct: the CRC (and the >= 64 KiB xxh3) covers the STORED bytes, the byte cap / segment roll / `durable_record_bytes` / #118 write-amp meters account post-compression stored bytes (the on-flash truth). `produced_bytes` deliberately keeps counting the ORIGINAL logical bytes (documented on the field).
- **Pass-through guard (load-bearing).** A message whose flags already carry `COMPRESSED` is appended UNTOUCHED: the wire legally delivers bit 0 (`PUB_WIRE_ONLY_FLAGS` masks only bits 6/7) and the DLQ redrive re-appends stored records; compressing either again would wrap a descriptor in a descriptor.
- **Write guards unchanged.** The 64-byte raw-store threshold and the never-expand guard mean a small or incompressible payload is stored byte-for-byte the historical layout; a compress error falls back to a raw store, so the produce error taxonomy is unchanged.
- **Client transparency.** The broker stays store-and-forward (delivers stored flags + bytes verbatim, the Kafka model #12 cites); `ironbus-client` now decompresses on deliver (bit cleared on `Message::flags`, bounded by the 8 MiB bomb cap, `NoDictionaries`), so the end-to-end pub/sub payload contract is byte-identical under `none` and `lz4`. An undecodable delivery (unknown codec, missing dict, bomb, corrupt stream) is the new typed `ClientError::Decompress`, never a panic. The CLI `sub` printer inherits this. `ironbus-core` moves from a dev-dependency to a runtime dependency of `ironbus-client` (pure Rust on the default build; the C-FFI ban is unaffected).
- **dump/peek render the truth.** The real stored codec from the descriptor; the decoded form decompresses (`decoded:true`, original length), degrades structurally (`decoded:false` + `reason`, the frozen #136 `missing-dict:<id>` wording); `--raw` shows the stored frame; `--require-dict` fails strictly (exit 3); the previously inert #92 plumbing in `run_dump` is live. Resolver: sidecar-first `CachingDictResolver` on a `zstd` build (opened READ-ONLY: only when `dicts/` already exists), `NoDictionaries` on the default build (correct by construction: lz4 never references a dictionary; a zstd record is unknown-codec poison there first).
- **Off switch byte-identity.** Every existing `EngineConfig` literal (~68 sites) passes `Codec::None`, so every existing determinism / conformance image is unchanged; the frozen corpus (`record_compressed.bin` and all fixtures), `FORMAT_VERSION`, and `scripts/check-format-registry.sh` are untouched. `zstd` stays rejected at parse on the default build (pinned by the existing test); no new server cargo feature.

## Tests

- Engine (5 new): lz4 produce/poll round trip (flag set, stored bytes strictly smaller, decode recovers the original, `produced_bytes` logical); sub-64-byte raw store; never-expand raw store (high-entropy payload), both byte-identical vs a `Codec::None` engine; already-`COMPRESSED` pass-through (verbatim store, one decode recovers); the byte cap metering stored bytes (raw trips a 2 KiB cap at the 3rd 1 KiB produce, lz4 stays under it for 8).
- Determinism: new lz4 run-vs-run byte-identity case (with a non-vacuity check that the lz4 image undercuts the raw image); deliberately NO frozen compressed bytes, so the format gate stays decoupled from the `lz4_flex` version.
- CLI (3 new/updated): an lz4 data dir written through the real `open_disk_engine` dumps `codec=lz4 decoded=true` with the original length; `--raw` shows the strictly smaller stored frame with no decode; NDJSON carries the same fields; `--require-dict` does not trip on lz4; a `--compression none` dir renders the historical form; the renamed raw/require-dict test pins that raw-stored records render historically under both flags.
- Client (3 new): lz4-broker end-to-end fetch returns the original payload with the bit cleared; a scripted bit-0 deliver decodes to the original; a hostile unknown-codec deliver is the typed error, never a panic.
- Live e2e: the acceptance competing-group step now round-trips broker-compressed >= 64 B payloads through the client transparently (it failed before the client commit, proof the loop was genuinely open).
- Three CLI tests whose deterministic sizing arithmetic is defined over RAW bytes on disk (golden_path drop-oldest, acceptance step 8, the edge-tiny write-amp gate) pin `--compression none` with per-site rationale (the behavior under test is codec-independent; lz4 would shrink the repeated-letter payloads below the caps).

## Gates (all run locally)

- `cargo test --workspace --all-features --locked`: PASS
- `cargo test --workspace --locked` (default features): PASS
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: PASS
- `cargo fmt --all --check`: PASS
- `cargo build --workspace --features zstd --locked` + `cargo test --workspace --features zstd --locked` (the CI zstd-feature job's invocation): PASS
- `cargo test -p ironbus-core --test conformance_corpus --locked` (frozen corpus byte-identical, no regeneration): PASS
- `cargo test -p ironbus-server --test fuzz_regression_replay --locked` + `sh fuzz/seed-regression-corpus.sh --check`: PASS
- `sh scripts/check-format-registry.sh`: PASS (registry digest unchanged)

Note: `ironbus-storage`'s `preallocate_reserves_real_blocks_on_a_supporting_fs` is environment-flaky on this APFS host (it failed identically on an untouched `origin/main` checkout and passes intermittently); unrelated to this change.

## Docs

CHANGELOG bullet under [Unreleased] / Added (Closes #430, refs #12 #387 #75). The #429 docs that deliberately said the knob was unwired are flipped to the wired truth: `docs/CONFIG.md` (coupled set 6), `docs/EDGE_TUNING.md` (knob row + the #12 deferred bullet), `docs/EDGE_CONSTRAINTS.md` (codec row), `docs/CLI.md` (`--raw`/`--require-dict` rows + the #136 paragraph + the record codec note), `docs/CLI_CONTRACT.md` (dump-record schema note), `docs/USAGE.md` (record-line codec note), and the CLI help/doc comments (`DEFAULT_COMPRESSION`, `CompressionArg`, `ServeConfig.compression`, the materialized-config builder, the non-Unix consume, the dump help).

Accounting decisions, date-stamped 2026-06-10: `produced_bytes` = pre-compression logical bytes (producer-facing throughput; stored truth is `durable_record_bytes` + the #118 meters); the bench in-process self-test pins `Codec::None` so its recorded baselines stay comparable (the open-loop harness proves latency methodology, not the codec).

Closes #430. Refs #12, #387, #75, #429, #92, #136.


ELares (commented):
## Self-review record (adversarial two-lens panel: format/write-path invariants, tests/contracts)

Both lenses returned needs-changes; all findings resolved across 6bf6748, a29502c, 4c7726e, b91da18, 1d7de64. Disposition:

**Blocker (panel reproduced empirically):** the seam had no upper guard against the readers' decompressed-size cap, so a compressible payload in (8 MiB, 16 MiB] was compressed, durably ACKED, and then refused by every shipped reader (DecompressedTooLarge): consumer stall, then pseudo-poison dead-lettering of an honest record. The seam now stores raw above DEFAULT_MAX_DECOMPRESSED_BYTES (the same ironbus-core constant the readers enforce, no duplicated number). Boundary tests pin one-over (raw) and at-cap (compressed); mutation check: deleting the guard fails the test.

**Resolved should-fixes:**
- Pass-through guard test had no teeth (the panel proved deleting the guard survived the whole suite because the fixture could not be re-shrunk and the never-expand guard masked the deletion). The fixture is now a recompressible stored object (lz4 of 1 MiB of zeros) with a self-check that it genuinely recompresses; the mutation now fails the test.
- A mid-batch decompress failure desynced the client connection (batch-mates dropped, FlowEnd unread, no offset to skip). Client::fetch now drains the window before surfacing the error; ClientError::Decompress carries source, offset, and generation so callers can skip the poison record; the variant documents the post-error batch state; a test proves the next fetch on the same client is clean. Broker-side produce-time descriptor validation is a tracked follow-up.
- Rollout flag-day: COMPATIBILITY.md gains a mixed-version subsection and the CHANGELOG states the rule (upgrade consumers to a #430+ client first, or run --compression none during the transition; a client downgrade after compressed data exists regresses consumers to raw stored bytes).
- Metrics semantics: METRICS.md, the /metrics HELP text, the Log::append comment, and EDGE_CONSTRAINTS.md now state that ironbus_logical_bytes_written meters STORED bytes post-#430 and what that does to the write-amp ratio; produced_bytes stays producer-logical. Docs only, no accounting changed.

**Verified sound by the panel (no change needed):** the seam is genuinely single (produce, actor group-commit, dedup, fire-and-forget all funnel through it; the offline DLQ redrive appends verbatim via Log::append and preserves the flag); CRC covers stored bytes and recovery/scrub treat torn compressed records exactly like torn raw records; Codec::None disk images are byte-identical (image-comparison proven); no golden/corpus file regenerated; check-format-registry digest unchanged; no format bump required (the descriptor-inside-payload design was pre-ratified in #387 and the compat registry rows pre-exist); codec id 2 on a default build stays poison-not-crash.

**Follow-ups filed:** #438 (broker-side descriptor-shape validation when a wire PUB carries the COMPRESSED bit, or a deliver-side capability bit per the GapMarker precedent) and #439 (a default-codec lz4 write-amp gate and an entropy-honest bench mode, since every byte-budget CI gate now pins --compression none).

Hardware validation (compression on-disk via dump --raw, decode round trip, mixed-version smoke) runs on the hive before #430 closes. This comment is the review of record.

PR #441, opened 2026-06-10
test/bench: lz4 companion write-amp gate + entropy-honest bench harness (#439)

### LINKED ISSUE #439: [TASK] CI gates and bench baselines cover --compression none only; add a default-codec (lz4) write-amp gate and an entropy-honest bench mode
## Summary

After #430 (PR #437) the shipped default codec is a live lz4 write path, but every end-to-end byte-budget gate now pins `--compression none` for sound individual reasons (raw-byte sizing arithmetic in golden_path drop-oldest and acceptance step 8; the historical framing/checkpoint regression contract in the edge-tiny write-amp gate; baseline continuity in the in-proc bench). Net effect: no CI gate exercises the write-amplification, segment-roll cadence, or reap behavior of the configuration that actually ships, and the wire bench's all-zeros payload makes its default-codec numbers best-case rather than representative.

## Asks

1. An lz4 companion to the edge-tiny write-amp gate with a contract defined for compressed brokers (the `>= 4x fails` threshold in EDGE_CONSTRAINTS.md is defined over stored bytes; for small compressible payloads the ratio inflates while real flash wear per user byte falls, so the compressed-gate threshold needs its own derivation).
2. A payload-entropy knob on `ironbus bench` (or a documented incompressible mode) so recorded baselines are not silently best-case-compressible.

refs #430, refs #118, refs #19

ELares:
Closes #439.

Both asks of the issue: the shipped default lz4 configuration now has a write-amp regression gate, and the regression-gate bench harness no longer measures a pathological all-zeros best case against an unstated codec.

## 1. lz4 companion to the edge-tiny write-amp gate

New test `edge_tiny_write_amp_is_bounded_under_the_shipped_default_lz4_codec` in `crates/ironbus-cli/tests/edge_tiny_budget.rs`:

- Boots the same `serve --profile edge-tiny` with NO `--compression` flag, so it exercises the codec the shipped binary actually defaults to, and pins `compression=lz4` via the materialized-config line (a default-codec flip fails loudly and forces a re-derivation instead of silently invalidating the bound's comment).
- Runs the identical payload mix as the raw gate (factored into a shared `publish_gate_workload` so the two cannot drift), asserts the #118 meters advance with `physical >= logical`, and bounds `ironbus_write_amp_ratio` at a derived `20.000x`.
- Derivation (in the `WRITE_AMP_GATE_LZ4_MILLI` doc comment): observed `logical=54` stored bytes, `physical=382`, ratio `7.074x`, byte-identical across 30 consecutive runs (deterministic workload: 6 sequential pubs, no timing-dependent checkpoint, deterministic `lz4_flex`). Bound `20x` gives ~2.8x headroom, the same proportional philosophy as the raw gate's typical `1.2x` vs `4x`; an unexplained 10x regression (`70.7x`) fails by 3.5x.
- Why the bound differs from the raw gate's 4x: post-#430 the ratio's denominator is STORED (post-compression) bytes (`docs/EDGE_CONSTRAINTS.md`). lz4 collapses this workload's 512/1024-byte repeated-letter payloads to ~23 stored bytes each while the physical framing/page cost stays fixed, so the metric inflates to 7x even though real flash wear per USER byte improves (`0.25x` here vs `1.212x` raw). Inheriting 4x would fail a configuration whose real wear got better.
- The none-pinned gate is untouched and remains the historical framing/checkpoint regression contract; its pin rationale moved verbatim to the call site of the now-parameterized boot helper.
- Drive-by fix the second test exposed: `Scratch` paths collided (same pid plus the same wall-clock nanosecond for two concurrently starting tests on a coarse clock), sending the second broker into the #89 single-broker data-dir lock while the colliding `remove_dir_all` raced the first broker's live data dir. Reproduced at ~20% per run; a process-wide atomic counter now disambiguates, 30/30 green after.

`docs/EDGE_CONSTRAINTS.md` documents the companion gate next to the existing gate description.

## 2. Entropy-honest bench harness

- `crates/ironbus-bench/src/harness.rs`: new `PayloadEntropy` threaded through `RunConfig`. The default `CompressibleRealistic` body mirrors the CLI bench's `--payload-shape realistic` telemetry shape generator byte for byte (not importable: `ironbus-cli` ships only a binary target; mirroring is documented on both the function and its tests). `Incompressible` re-fills per message (LCG seeded by seq) so the codec can never amortize repeated bytes. Both fills are deterministic; the token region is never touched (tested).
- `crates/ironbus-bench/src/broker.rs`: `Broker::spawn` now pins `--compression lz4` explicitly, documented as the shipped ADR-0003 default, so a recorded baseline self-describes what it measured and a future default-codec change cannot silently shift the #114 rolling-median comparison. `extra` args are appended after and the serve parser takes the last occurrence, so an explicit `--compression none` comparison run still works.
- `ironbus-bench` binary grows `--payload-entropy <realistic|random>` (default realistic), recorded in the provenance JSON (`config.payload_entropy`, an additive field, schema version unchanged) and in the reproduce command.
- Baseline continuity (`src/regression.rs` consumers): pre-#439 all-zeros runs are not comparable with realistic runs; per the prior review NO baseline has been archived yet, so no recorded history is invalidated and the first archived baseline starts honest. Noted in the send_loop comment, the provenance field doc, and the CHANGELOG.
- The deterministic in-proc stall broker keeps `Codec::None` (unchanged): it proves the open-loop latency harness, not byte budgets, and its comment already records that decision.

End-to-end proof the knob and the pinned codec both bite (same 2 s, 500 msg/s run):

| entropy | write amp |
|---|---|
| `realistic` (default) | 0.78x |
| `random` | 1.43x |

## Gates

- `cargo test --workspace --all-features --locked`: green (includes the injected-stall gating proof and the 5 `regression_gate` CLI tests, unchanged).
- The new lz4 gate: 30 consecutive runs green with a byte-identical measurement, plus 3 post-derivation runs.
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: clean.
- `scripts/ci/slo-table-check.sh`, `scripts/ci/fmea-coverage.sh`, `scripts/ci/changelog-unreleased.sh`, `scripts/ci/relative-link-check.sh`: all pass.
- CHANGELOG under `[Unreleased]` / `### Added` (Closes #439).


ELares (commented):

Panel verdict: **approve**. Evidence-checked: the 20x lz4 bound reproduced byte-identical across 21 reviewer runs (logical=54, physical=382, write_amp=7.074x), every derivation step re-verified, and the assertion mutation-tested (a 7.000x bound fails the gate). The none-pinned raw gate is behaviorally identical pre/post refactor, the Scratch atomic-counter fix closes the observed collision, and the entropy modes provably bite (0.78x vs 1.43x on the same run).

**Resolved should-fix:** the mirrored payload generator was an unpinned fork (the harness cannot import the CLI fill; the tie was comment-only). Both crates now pin their generator's output for both entropy modes against shared 48-byte golden vectors, so either side drifting fails its own test against the same constants.

**Accepted nits, rationale on record:** the #114 RunPoint schema does not carry payload_entropy (history builders must check provenance; acceptable until baselines exist); the 20x headroom lets a clean 2x regression pass, exactly matching the raw gate's proportional-headroom philosophy (1.212x observed vs 4x contract), and the measurement's determinism makes tightening easy later if wanted.

Issue #35, opened 2026-06-05
[TASK] Author the conformance test vectors for the semantics contract


ELares:
Parent: #3

Produce a language-agnostic, golden conformance suite that pins every observable behavior the parent #3 contract promises, so any IronBus implementation or client can be checked against the same vectors. This is the executable spec for the contract; it depends on the four sibling tasks (offset lifecycle, delivery/ack, dedup, key routing) and feeds #21.

## Approach

Define vectors as input sequences plus expected observable outputs, independent of internal layout. Cover: (1) ordering, a producer sends N records, a broadcast consumer reads offsets 0..N-1 in exact order with no gaps; (2) at-least-once, an un-acked offset past its lease is redelivered, an acked one is not; (3) dedup, the same `msg_id` within the window returns the ORIGINAL offset with `duplicate=true` and `rc=0`, while the same `msg_id` after window eviction yields a NEW offset; (4) ack rejection, a cumulative ack in competing mode returns `ERR_CUMULATIVE_ACK_NOT_ALLOWED` and a foreign-offset ack returns `ERR_ACK_NOT_OWNED`; (5) key routing, records sharing a key all reach one competing member and preserve per-key order, and a member leave reshuffles only that member's keys; (6) broadcast cursors, first attach defaults to from-start, `cursor=latest` sees only new, `cursor=resume` for a named consumer survives a simulated restart; (7) trim, a read of a trimmed offset returns `OFFSET_TRIMMED`. Encode vectors as a checked-in data file (JSON or similar) plus a harness contract, so the same suite runs in #21 and against external clients.

## Acceptance criteria

- [x] Vectors exist for ordering, at-least-once redelivery, dedup hit/eviction, ack rejection, key routing, broadcast cursor positions, and trim.
- [x] Each vector states inputs and exact expected outputs (offsets, flags, error codes) with no implementation dependence.
- [x] Dedup vectors assert original-offset-with-`duplicate=true` in window and a new offset after eviction.
- [x] Ack-rejection vectors assert the exact named error codes.
- [x] The suite is a checked-in data file with a documented harness contract reusable by #21 and external clients.

## Failure considerations

[...]

PR #442, opened 2026-06-10
feat(server): produce-time COMPRESSED descriptor shape validation (#438)

### LINKED ISSUE #438: [TASK] Validate the descriptor shape when a wire PUB carries the COMPRESSED bit, or negotiate deliver-side compression capability
## Summary

Any wire producer can publish `RecordFlags::COMPRESSED` (bit 0) over arbitrary non-descriptor bytes: `PUB_WIRE_ONLY_FLAGS` masks only bits 6 and 7, the broker is store-and-forward, and nothing validates the descriptor shape at produce time. Post-#430 (PR #437), every consumer of that offset attempts the decode and gets a typed `ClientError::Decompress`, cycling fetch failures until max-deliver dead-letters the record: a per-record consumer stall costing max-deliver visibility-timeout cycles, injectable unauthenticated. Pre-#430 the same bytes were delivered verbatim and the pipeline kept moving.

PR #437's scoped mitigation makes the client survive it (the window is drained, the error carries the offset so a caller can skip, and a subsequent fetch is not desynced), but the record still burns delivery attempts for every consumer group that encounters it.

## Asks

1. Broker-side shape validation when a PUB carries the COMPRESSED bit: the payload must parse as a well-formed descriptor (codec id known-or-poison, lengths sane: stored length consistent with the frame, `uncompressed_len` within the read-side cap). Reject the PUB with the existing error taxonomy rather than acking a record no reader can decode. This is cheap (a 9-byte header parse) and runs once at produce rather than per-consumer per-attempt.
2. Alternatively or additionally: a deliver-side capability bit in CONNECT (the GapMarker precedent in `docs/compat/versions.md`) so the broker knows whether a consumer can decode compressed deliveries at all; that is the bigger #106-adjacent design and the GapMarker row is the template.

refs #430, refs #12, refs #106

ELares:
Implements ask 1 of #438: when a wire PUB carries `RecordFlags::COMPRESSED` (bit 0), the broker validates the payload's descriptor SHAPE at produce time and rejects a malformed one instead of acking a record no reader can decode. One undecodable record previously cost EVERY consumer group max-deliver visibility-timeout cycles of typed `ClientError::Decompress` failures before it dead-lettered, injectable unauthenticated; the gate is a 9-byte header parse that runs once at the source instead of per-consumer per-attempt.

Ask 2 of #438 (a deliver-side capability bit in CONNECT, the GapMarker precedent) is deliberately NOT in this PR: it is the bigger #106-adjacent capability-negotiation design and stays a follow-up tracked on #438/#106.

Closes #438

## What changed

- `ironbus-core::compress::validate_descriptor_shape` (new, pure, IO-free): header-only shape rules, each with a why comment, reusing the existing `DecompressError` taxonomy. NO decompression at produce (shape only, no codec CPU on the hot path).
- `Session::handle_pub`: applies the gate at the WIRE boundary, beside the existing dedup-id-cap rejections. Rejection replies the existing `Err` (tag 12) frame with `malformed compressed descriptor: <detail>`; nothing is appended and no offset is consumed; the connection stays open. Fire-and-forget violations are dropped with NO frame (the QoS-0 no-frame contract).
- Docs: CHANGELOG (`[Unreleased]/Added`, Closes #438), `docs/CONTRACTS.md` PubBody (what a bit-0 PUB must carry and what the broker validates, including the normative non-empty-stream rule and the codec-id rollout coupling), the `docs/compat/versions.md` codec-id row (rollout coupling), and the `ClientError::Decompress` doc in `ironbus-client` (broker-side produce-time validation is no longer "a follow-up, tracked separately"; it ships here).

## Validation rules (mirroring the read side, with ONE documented exception)

1. The payload parses as a well-formed descriptor (`read_descriptor` succeeds; at least the 9-byte `DESCRIPTOR_LEN`).
2. The codec id is one of the REGISTERED ids: `none`(0) / `lz4`(1) / `zstd`(2) per the append-only registry in `docs/compat/versions.md`. An unregistered id (a typo'd future producer) fails fast at the source instead of poisoning every consumer. Rollout coupling (now documented in CONTRACTS.md and the registry row): registering a future codec id requires brokers to learn it BEFORE producers emit it.
3. The claimed `uncompressed_len` is at most `DEFAULT_MAX_DECOMPRESSED_BYTES`, the SAME `ironbus-core` constant every shipped reader's #76 bomb guard and the #437 write seam use (no duplicated number).
4. Per-codec stream shape: a `none` stream's length equals the claim EXACTLY (the read side's `BadRawLength` rule, covered in both directions: shorter AND longer than the claim); an `lz4`/`zstd` stream is non-empty. For lz4 the non-empty rule mirrors the read side exactly; for zstd it is deliberately STRICTER on exactly one degenerate input, see decision 4.

`dict_id` and stream CONTENT are deliberately NOT judged at produce: dictionary resolution is a reader capability (sidecar + embedded set, unknowable at the broker), and checking content would mean decompressing on the hot path. A structured-generator proptest pins the safety direction honestly: the gate NEVER rejects a payload `decompress_payload` accepts, EXCEPT the single enumerated zstd-empty-stream-claim-0 degenerate descriptor (decision 4), which deterministic tests pin on both feature sets.

## Decisions, with code evidence

1. Rejection surface: REUSE the existing `Err` (tag 12) prose frame; no new wire error code, no new conformance `ErrorCode`.
   - The #35 `ErrorCode` taxonomy (`crates/ironbus-server/src/codes.rs`) maps ENGINE outcomes (`ErrorCode::of_engine_error`); the existing session wire-boundary rejections ("malformed pub body", "producer_id too long", "msg_id too long" in `handle_pub`) have no `ErrorCode` and ride bare `Err` prose. This rejection never reaches the engine, so it follows that precedent exactly.
   - `codes.rs` states the wire `Err` bodies are NOT a pinned surface (the frozen things are the `ErrorCode` spellings, the FrameType tags, the record format); `docs/compat/versions.md` registers no `Err`-string vocabulary, and CONTRACTS.md documents the existing pattern of distinct self-announcing `Err` texts ("shed under load" vs "at capacity") riding tag 12 with no new tag. So a distinct message is additive and compat-safe.
   - The message carries the typed detail (`malformed compressed descriptor: unknown compression codec id 9 (poison)`) like `handle_connect`'s dynamic `reply_err(out, &e.to_string())` precedent, so a producer can fix the input.
2. `COMPRESSED` + codec `none` IS wire-legal. The read side (`decompress_payload`, `Codec::None` arm in `crates/ironbus-core/src/compress.rs`) returns the stream verbatim when `stream.len() == uncompressed_len` and rejects a mismatch as `BadRawLength`. The gate mirrors both halves; pinned by `shape_validation_pins_the_none_codec_exact_length_rule` plus `a_none_codec_stream_longer_than_its_claim_is_rejected_by_both_sides` (which assert the read side agrees on the same bytes) and the session test `wire_legal_descriptor_variants_are_still_acked`.
3. Registered ids, not build-implemented ids. The gate checks `CODEC_ID_NONE | CODEC_ID_LZ4 | CODEC_ID_ZSTD` (consts always defined) rather than `Codec::from_id` (which returns `None` for id 2 on a default build). The broker is store-and-forward: a `zstd` record produced through a default-build broker is decodable by a `zstd`-capable consumer, so narrowing the registry to the broker's own feature set would reject legal traffic. Pinned by `shape_validation_accepts_the_registered_zstd_id_on_every_build` (runs on both the default and the `--features zstd` gate).
4. Empty `lz4`/`zstd` stream is rejected for ANY claimed length, including 0; for zstd this is deliberately STRICTER than the read side on exactly one input. Verified empirically, both halves:
   - lz4: `lz4_flex::block::decompress_into(&[], ...)` returns `Err(ExpectedAnotherByte)` even into a 0-length buffer (the canonical lz4 encoding of empty input is the single token byte `0x00`), so for lz4 the rule rejects nothing the read side accepts. Pinned by `shape_validation_rejects_an_empty_stream_under_a_compressing_codec`, which asserts the read side fails the same bytes.
   - zstd: the locked `zstd` 0.13.3 `Decompressor::decompress_to_buffer(&[], &mut [])` returns `Ok` with 0 bytes written, so the 9-byte descriptor (codec `zstd`, dict 0, claim 0) over an EMPTY stream is decoder-ACCEPTED yet gate-rejected. The gate stays strict because the wire contract in CONTRACTS.md normatively requires a non-empty stream under a compressing codec, and a genuine zstd encoder never emits an empty frame (compressing the empty payload still emits a frame header, asserted in-test), so only hand-crafted bytes hit the gap. The rustdoc, the gate comment, CONTRACTS.md, and the CHANGELOG all state this exception explicitly instead of claiming a perfect read-side mirror. Pinned by `zstd_empty_stream_claim_zero_is_the_documented_strictness_exception` (zstd build: read side accepts, gate rejects) and `zstd_empty_stream_claim_zero_is_rejected_by_both_sides_on_a_default_build` (default build: both reject, the read side as unknown-codec poison).
5. Fire-and-forget with a bad descriptor: silent drop, NO frame, NO counter. The QoS-0 no-frame contract (#11, `a_fire_and_forget_pub_sends_no_frame_on_every_non_appended_disposition`) forbids any frame on any disposition. The existing wire-boundary rejections (the dedup-id caps in `handle_pub`) drop fire-and-forget violations silently with no counter; `ironbus_fire_and_forget_shed_total` increments only inside `Engine::fire_and_forget_admit` (the engine-decided token-bucket shed), and `Session` has no metrics surface, so counting here would either conflate load sheds with malformed input or require a new actor API. Matching the established precedent; pinned by `a_fire_and_forget_compressed_pub_over_garbage_sends_no_frame_and_appends_nothing`.
6. Engine-internal paths are untouched (verified before writing the gate). The DLQ redrive re-injects via `Log::append` directly (`ironbus_storage::admin::redrive_dlq`) and never passes through a session; the #437 seam's own compressed output is produced below the wire boundary. Both documented in the gate's comment, and the round-trip test runs through an lz4-compression engine so the #437 pass-through guard (not a disabled codec) is what preserves the bytes.

## Review fixes (adversarial panel)

1. SHOULD-FIX, false rejection: the panel proved the zstd-empty-stream-claim-0 descriptor is decoder-accepted on the locked zstd 0.13.3 while the gate rejects it. Resolution: keep the gate strict (the contract is normative; no genuine encoder emits the shape), make every claim honest, and pin the exception deterministically on both feature sets. See decision 4.
2. SHOULD-FIX, vacuous proptest: the old property drew raw random bytes, which decode with probability ~0 (panel: 0 hits in 1e6), so the safety premise never fired. Measurement during the fix showed even structured streams stay vacuous (0 decodable in 10000) while the claim is drawn INDEPENDENTLY, because decode requires claim == decoded length exactly. The property now draws from a three-arm generator (CHAOS: independent fields over empty/arbitrary/genuine-codec streams; HONEST: genuine encodes with claim == payload length, decodable by construction; CLAIM-0 BOUNDARY: compressing codecs over empty streams and the genuine streams of the empty payload). Measured with a deterministic 10000-case probe: the read side accepts ~42% of cases on the default build and ~50% on a zstd build, and the degenerate exception branch itself fired 555 times. A deterministic premise case set (`the_shape_gate_accepts_every_genuinely_decodable_premise_case`, zstd cases cfg-gated) guarantees the premise regardless of generator luck.
3. SHOULD-FIX, mutation survivor: weakening the none-codec exactness from `!=` to `<` survived the whole suite (no test had a stream LONGER than its claim). Killed by `a_none_codec_stream_longer_than_its_claim_is_rejected_by_both_sides`; see the mutation check.
4. QUESTION resolved as docs: the codec-id rollout coupling (brokers must learn a new registered id BEFORE producers emit it, because this gate rejects unregistered ids) is now stated in the CONTRACTS.md produce-gate paragraph and the versions.md codec-id row. The format-registry digest gate hashes only the `format.rs` pub-const layout, so the prose edit does not trip it (verified: `scripts/check-format-registry.sh` ok).

## Mutation check

With the gate disabled (`if false && ...` on the `COMPRESSED` check in `handle_pub`):
- FAILED (as required): `a_compressed_pub_over_garbage_is_rejected_and_appends_nothing`, `a_compressed_pub_with_an_over_cap_claim_is_rejected`, `a_compressed_pub_with_an_unregistered_codec_id_is_rejected`, `a_fire_and_forget_compressed_pub_over_garbage_sends_no_frame_and_appends_nothing` (the garbage was durably acked).
- Still passing (as required): `a_well_formed_producer_compressed_pub_still_round_trips`, `wire_legal_descriptor_variants_are_still_acked`.
Gate restored; all 82 session tests pass.

With the none-codec exactness weakened from `!=` to `<` (review finding 3), one site at a time:
- mutant A (`validate_descriptor_shape`): `a_none_codec_stream_longer_than_its_claim_is_rejected_by_both_sides` FAILED; the other 29 compress tests (including the pre-existing exact-length pin) still passed, confirming this mutant previously survived.
- mutant B (`decompress_payload`, none arm): the same single test FAILED, 29 passed.
Both sites restored; full suite green on both feature sets.

## Gates

- `cargo test --workspace --all-features --locked`: 40 suites, 1458 passed, 0 failures
- `cargo test --workspace --locked` (default build, exercises the `cfg(not(feature = "zstd"))` tests): 40 suites, 1429 passed, 0 failures
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (workspace lints include `clippy::pedantic`): clean
- `cargo build --workspace --features zstd --locked` + `cargo test --workspace --features zstd --locked` (the CI zstd gate): 40 suites, 1450 passed, 0 failures
- `sh scripts/check-format-registry.sh`: digest unchanged (1ba45310...)
- `sh scripts/ci/relative-link-check.sh`: 757 links, 0 problems


ELares (commented):

Panel verdict was needs-changes; all findings resolved in 7d28359, 54047ab, 99fa1bf, 4ecaf74. Disposition:

- **False-rejection counterexample (panel proved it empirically):** the zstd empty-stream/claim-0 descriptor is decoder-accepted on the locked zstd 0.13.3 but gate-rejected. Resolution: the gate stays strict as the NORMATIVE wire contract (a genuine zstd encoder never emits an empty frame, asserted in-test), and every claim that said "rejects nothing the read side accepts" is now scoped honestly (true for lz4, documented strictness exception for zstd), with deterministic pins on both feature sets.
- **Vacuous proptest:** the panel's measurement (0 decode-Ok in 1e6 random cases) stands, and the fix went further: the panel's prescribed generator was ALSO measured vacuous (an independent claim almost never equals the decoded length), so the property now draws from a three-arm generator (chaos, honest-by-construction, claim-0 boundary) with a measured 42-50% decoder-accept rate and the exception branch firing 555/10,000; a deterministic premise set guarantees coverage regardless of generator luck.
- **Mutation survivor:** the none-codec longer-than-claim case now pins BOTH halves (validator and decoder); each weakened-to-< mutant fails exactly this test, independently verified.
- **Rollout coupling (question):** one sentence each in CONTRACTS.md and the versions.md codec-id row: brokers must learn a future codec id before producers emit it, because this gate rejects unregistered ids. Format-registry digest unchanged.

Verified sound by the panel, unchanged: seam placement (a rejected PUB mutates nothing: no dedup entry, no counter, no offset; fire-and-forget drops with no partial state), wire compat (Err tag 12 prose only, no frozen ErrorCode touched, old clients treat prose as opaque), no PUB-flag path bypasses handle_pub, the #437 engine guard remains needed for internal appends, header-parse-only cost on the hot path.

Hardware note: normal-path compressed publish/consume regression-validated on the deployed hive; the rejection paths are wire-level and pinned by the session test suite. Ask 2 of #438 (CONNECT capability negotiation) remains deferred to the #106-adjacent surface as the issue allows. This comment is the review of record.

Issue #445, opened 2026-06-10
[TASK] Memory-mode validation: acceptance variant, OOM-bound gate, bench (child of #443)


ELares:
## Summary

Validation for `serve --storage memory` (child of the core task): the mode ships only with the same gate discipline the disk path has.

## Work

1. **Acceptance**: a golden-path variant over `--storage memory` (install to first message, produce/consume/ack, group cursor redelivery semantics within one process lifetime, clean SIGTERM exit 0); an explicit restart test pinning the CONTRACT that the queue is empty after a broker restart (the inverse of the disk path's recovery test).
2. **OOM protection is the headline risk**: a test that produces past `--max-total-bytes` and asserts drop-new sheds with the existing `at capacity` error while broker RSS stays bounded (the #115 ram-ceiling guard composing with the byte cap); mutation-style check that removing the required-bound boot validation fails a test.
3. **Bench**: `ironbus bench --storage memory` (or bench inheriting the flag) for honest RAM-path numbers, reported next to the disk numbers; the #439 entropy modes apply unchanged.
4. **CI**: the new tests run in the existing workspace gates; no new jobs unless the acceptance variant needs one; conformance suites already cover the in-memory filesystem, note that explicitly in the PR.

Parent: #443, refs #439, refs #115

ELares:
Scope additions from the #446 review panel:

1. The #115 RAM-ceiling boot guard models connections/credits/groups/in-flight only; in memory mode the STORE itself is RAM (up to `--max-total-bytes` of stored bytes, briefly doubled by the in-memory durable-image clone at sync points), so `--ram-ceiling-bytes 64MiB` plus `--max-total-bytes 1GiB` boots today. The OOM-bound work here should fold the store (and the clone headroom) into the memory-backend footprint proof, or document the exclusion loudly.
2. PR #446 already shipped a real-wire golden-path memory test (produce/consume/ack, no-redelivery, clean SIGTERM, no-files-created assertions over a scratch cwd and isolated TMPDIR). Trim item 1 here to what remains: the restart-revives-EMPTY contract test and install-to-first-message if applicable.

ELares:
Hardware validation record (shipped in 2026.0611.4, armv7 edge node):

[...]

Issue #444, opened 2026-06-10
[TASK] Memory-mode operational surface: offline verbs, boot validation, and docs (child of #443)


ELares:
## Summary

Operational surface and documentation for `serve --storage memory` (child of the core task): every place that assumes a data DIRECTORY exists must decide what memory mode means and say it out loud.

## Work

1. **Offline verbs**: `peek`, `dump`, `scrub`, `repair`, `admin consumer-reset`, `admin dlq-redrive`, and `migrate` operate on a STOPPED broker's `--data-dir`. For a memory broker there is no directory and nothing to inspect after exit. Decide per verb and pin with tests: a clear "memory-mode broker leaves no data dir to inspect" usage error beats a confusing `data dir not found` exit 2. The LIVE surfaces (`/healthz`, `/readyz`, `/metrics`, `/admin`, `ironbus top --addr`) work unchanged and become the ONLY introspection story; say so.
2. **Flag interplay validation at boot**: `--storage memory` plus `--data-dir` is a usage error (or a documented ignore, pick one); retention bounds, `--checkpoint-interval`, and cursor checkpoints become in-memory no-ops or in-memory-only (decide: group cursors still matter for redelivery semantics WITHIN the process lifetime, so checkpoint machinery should keep working against the in-memory fs rather than being special-cased).
3. **Docs**: CLI.md and USAGE.md rows for the flag and the loss contract; CONFIG.md (env var `IRONBUS_STORAGE`); EDGE_TUNING.md (when memory mode is the right call: flash-wear-bound deployments that can tolerate reboot loss; cite the measured SD vs tmpfs numbers); COMPATIBILITY.md (no on-disk artifacts means no downgrade story to document, say it); DISTRIBUTION.md unit section (Restart= revives an EMPTY broker; pair memory mode with consumers that tolerate replay-from-live); CLI_CONTRACT.md exit-code rows for the new usage errors.
4. **The tmpfs baseline** gets one honest paragraph in EDGE_TUNING.md regardless, since it works on every shipped version today.

Parent: #443

ELares:
Hardware note (2026.0611.4 on the armv7 edge node): the documented operational surface checks out live; the offline-verb rejection (`peek --data-dir x --storage memory` = usage exit 1) and the live-only introspection story were exercised during the #443/#445 validation. The DISTRIBUTION.md memory-unit recipe (UnsetEnvironment drop-in) is documentation-validated against the shipped unit's env fallback, which this hive runs.

PR #446, opened 2026-06-11
feat(cli): serve --storage memory, a first-class ephemeral in-memory engine

### LINKED ISSUE #443: [TASK] serve --storage memory: a first-class ephemeral in-memory engine (explicit loss contract, config-bounded RAM, no flash wear)
## Summary

Add a first-class ephemeral in-memory engine: `serve --storage <disk|memory>` (default `disk`, byte-for-byte unchanged). In `memory` mode the broker runs the existing engine over the existing in-memory filesystem with NO files, NO fsync, and explicitly NO power-loss or restart durability: a queue for hot-path fan-out, spill-to-RAM buffering, and test rigs, on devices where flash wear or fsync latency is the binding constraint.

## Why a first-class mode when tmpfs already works

Both workarounds were measured on real edge hardware (RPi4, SD card, 256 B realistic payloads, 2026.0610.10):

| | fsync cost | throughput | physical writes |
| --- | --- | --- | --- |
| SD card data dir (default `sync`) | ~9.0 ms | 50 msg/s | ~70 KB per 256 B message |
| tmpfs data dir (`/dev/shm`) | ~0.5 ms | 92 msg/s (at the CPU ceiling) | RAM only |

tmpfs works today and stays the documented zero-code baseline. A first-class mode is still worth having because: (1) the semantics become EXPLICIT and machine-checkable (the materialized-config line says `storage=memory`; an operator cannot mistake a tmpfs mount for durable storage); (2) no dependency on a tmpfs mount existing, being writable, or being big enough (the queue is bounded by broker config, not by mount size: a full `/dev/shm` shared with other tenants fails as ENOSPC mid-write, while `--storage memory` sheds via the existing `at capacity` produce error); (3) the fsync syscall round trip disappears entirely instead of being cheap; (4) `ironbus-storage` already ships the deterministic in-memory `Filesystem` used by every engine test and the conformance suites, so the stored-format invariants (CRC-over-stored-bytes, torn-tail recovery within the process lifetime, retention, compression #430) hold identically.

## Design constraints (the tenets bind)

- **Resilient stays honest.** An ack in memory mode survives a connection drop and an engine hiccup, never a process exit. Mirroring the `--durability-level async/none` precedent, the broker REFUSES to boot with `--storage memory` unless the operator passes the existing explicit consent flag (or a dedicated `--ephemeral-loss-ack`); the startup banner and materialized-config line state the loss contract.
- **RAM bounds become load-bearing.** On disk, an unbounded log fills the SD card; in memory it OOMs the device. `--storage memory` REQUIRES an explicit `--max-total-bytes` (refuse to boot at 0/unlimited) and composes with the #115 `--ram-ceiling-bytes` refuse-to-boot guard; document that the byte cap meters stored (post-#430-compression) bytes.
- **Same engine, same format.** No new storage code: `Engine<InMemoryFs>` (already exercised by the test and conformance suites) behind the flag; the CLI's `open_disk_engine` grows an enum seam. The wire surface, group semantics, retention, ghosts (#432), and the produce gate (#438) are unchanged.
- **Restart semantics.** A clean stop or crash loses the queue by contract; under the packaged unit `Restart=on-failure` revives an EMPTY broker. The unit docs must say so (see the operational-surface task).

## Non-goals (v1 of this mode)

- No hybrid spill-to-disk; no snapshot/restore of the in-memory state; no Windows enablement ride-along (serve stays Unix-only even though the on-disk store is the documented reason).

refs #115, refs #430, refs #432, refs #438

ELares:
## Problem

Closes #443. refs #444, refs #445.

Edge deployments where flash wear or fsync latency is the binding constraint (queue for hot-path fan-out, spill-to-RAM buffering, test rigs) had only the tmpfs workaround, which is implicit (a tmpfs data dir is indistinguishable from durable storage on the startup line), mount-dependent (fails as ENOSPC mid-write when a shared /dev/shm fills), and still pays the fsync syscall round trip. #443 asks for a first-class ephemeral mode with an explicit, machine-checkable loss contract.

## What this does

`serve --storage <disk|memory>` (flag > `IRONBUS_STORAGE` env > default `disk`):

- `disk` (default): byte-for-byte the historical durable broker. No behavior change; the existing suites prove it.
- `memory`: the SAME engine and the SAME resolved `EngineConfig` over `ironbus_storage::fs::InMemoryFs` (the deterministic in-memory filesystem every engine test and conformance suite already exercises). NO files, NO fsync, NO power-loss or restart durability. The stored-format invariants (CRC-over-stored-bytes, retention, #430 compression, the #438 produce gate, group cursors and redelivery within the process lifetime) hold identically. `serve` stays Unix-only.

## Design decisions

**Dedicated consent flag.** Memory mode refuses to boot without `--ephemeral-loss-ack`. It is deliberately NOT the reused `--async-loss-ack`: that flag consents to a relaxed fsync schedule on a durable store (bounded or opportunistic sync, data survives a clean stop), while memory mode consents to losing EVERY acked message on any process exit. Two distinct loss contracts must never be conflated by one flag, so the memory gate's refusal names --async-loss-ack to separate the two contracts (the durability gate, untouched here, does not mention the ephemeral flag: an async/none operator is not at risk of confusing the two) to make the boundary explicit.

**Four fail-closed boot refusals** (all `CliError::Usage`, exit 1, before any listener opens):
1. `memory` without `--ephemeral-loss-ack`: states the loss contract (a clean stop or crash loses every acked message; an ack still survives a connection drop and an engine hiccup) and that `--async-loss-ack` does not cover it.
2. `memory` with `--max-total-bytes` 0/unset (unlimited): on disk an unbounded log fills the SD card, in RAM it OOMs the device. The message states the cap meters STORED (post-compression) bytes, the same accounting the disk store uses, and that it composes with the #115 `--ram-ceiling-bytes` guard unchanged.
3. `memory` with an explicit `--data-dir`: the broker keeps no on-disk state, so a given path would only LOOK durable; the refusal echoes the conflicting directory. (This is the one flag interplay that cannot wait for the #444 sweep, because `--data-dir` is otherwise required.)
4. An unknown `--storage` value: usage error naming its source (the flag, or `IRONBUS_STORAGE` when env-sourced).

**Dispatch shape.** `cmd_serve` is a static two-armed match; each arm opens its engine (`open_disk_engine` over `StdFs`, `open_memory_engine` over `InMemoryFs::new()`) and calls the new `run_broker<F: Filesystem>` carrying the whole serve body (actor spawn, bind, startup logging, config handle + reload self-check, signals, health server, accept loop, graceful drain). One monomorphized broker per backend, no dyn dispatch on the hot path, and one shared `open_engine_with` so the two backends can never drift in engine behavior. Memory mode takes no data-dir lock (each process owns a private filesystem; there is no path to lock) and `data_dir` becomes `Option<&Path>` through the serve path with a stable `none` sentinel for the reload cold-key comparison.

**Observability.** The stdout listening line names the backend (`storage memory (ephemeral)`) instead of a path that does not exist; the second startup line states the ephemeral contract on every memory boot (mirroring the loud I2-waived durability warning); the stderr materialized-config line gains an ADDITIVE trailing `storage=` field with `data_dir=none` in memory mode, so a script can machine-check the backend and an operator can never mistake a tmpfs mount for durable storage, nor an ephemeral broker for a durable one.

## Tests

- Unit (CLI): each of the four refusals asserts its message content and the usage exit code; flag > env > default precedence for the backend; the `storage=` materialized-config echo (disk and memory, with the `data_dir=none` sentinel); USAGE lists both flags; memory with consent + cap validates.
- Real wire (golden-path harness): boots the real binary with `--storage memory --ephemeral-loss-ack --max-total-bytes 16777216` (no `--data-dir`), asserts the listening line and the banner, pubs three, subs with `--ack`, re-fetch sees nothing within the process lifetime, clean SIGTERM exits 0, and the stderr materialized-config line contains `storage=memory`.
- The default disk path is untouched; every other scenario in the existing suites proves it.

## Gates (all run locally on this branch)

- `cargo test --workspace --all-features --locked`: 1473 passed, 0 failed
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: clean
- zstd job as CI runs it (`cargo build --workspace --features zstd --locked` + `cargo test --workspace --features zstd --locked`): green
- `scripts/check-format-registry.sh`: ok (no on-disk format change)
- `sh scripts/ci/relative-link-check.sh`: 757 links, 0 problems
- `sh scripts/ci/changelog-unreleased.sh`: ok
- Windows path: `cargo check --workspace --all-targets --locked --target x86_64-pc-windows-msvc`: 0 errors/warnings (the cfg(not(unix)) serve stub consumes the new `storage`/`ephemeral_loss_ack` fields, keeping the windows-latest `-D warnings` surface dead-code clean)

## Out of scope (tracked)

- #444: memory-mode operational surface (offline verbs, full boot-interplay sweep, unit docs). CLI.md and USAGE.md get one minimal row each here.
- #445: acceptance variant, OOM-bound gate, bench.
- Non-goals per the issue: no spill-to-disk hybrid, no snapshot/restore, no Windows serve enablement.


ELares (commented):
## Self-review record (adversarial two-lens panel: correctness, tests/contract)

Correctness lens: **approve**, all seven lanes verified live on the branch (disk-path operation order traced line-by-line as byte-identical with only the additive storage= field; two memory brokers on different ports fully independent and a same-port second fails on bind; all four refusals fire before any side effect and the data-dir refusal covers the flag, the env var, AND the config-file key; SIGTERM drains over the Filesystem trait with no path ops; the #432 ghost machinery is fully generic; no in-repo parser breaks on the data_dir=none sentinel; a restarted memory broker comes back empty and healthy).

Tests lens was **needs-changes**; all resolved in c9aef8c:
- A conflated consent gate (accepting --async-loss-ack for memory mode) survived the entire suite because the consent test only asserted message text. A new case boots with ONLY the async consent and pins the refusal itself; the env-ignoring consent mutant is likewise killed by a new IRONBUS_EPHEMERAL_LOSS_ACK test.
- The golden-path memory test never asserted the no-files contract. It now runs the broker with a scratch cwd and an isolated TMPDIR and asserts both stay empty across boot, produce, consume, ack, and the graceful drain (the panel verified the implementation is honest today via lsof; the test makes regressions fail loudly).
- CLI.md gains the dedicated --ephemeral-loss-ack row with the env form and a scripting note (key on storage=, never data_dir=none alone); the serve synopsis no longer shows --data-dir as unconditionally required; the PR body's bidirectional-cross-reference overclaim is corrected.

Panel questions resolved: the #115 RAM-ceiling/store-footprint gap is folded into #445's scope (commented there), as is trimming #445's overlap with the golden-path test shipped here; the reload cold-key behavior is pre-existing and unchanged.

PR #447, opened 2026-06-11
Memory-mode operational surface: offline verbs, boot validation, and docs (#444)

### LINKED ISSUE #444: [TASK] Memory-mode operational surface: offline verbs, boot validation, and docs (child of #443)

ELares:
Closes #444. Child of #443, which shipped as #446 (`serve --storage memory` with the consent gate, the byte-cap requirement, the data-dir refusal, the `storage=` echo, and the ephemeral banner).

## Decisions, with rationale

**1. Offline verbs vs memory mode: docs plus a pinning test, NO new code guard.**
Verified in the parsers: `--storage` (and `--ephemeral-loss-ack`) are collected only by `collect_serve_flags`. Each of the seven data-dir verbs (`peek`, `dump`, `scrub`, `repair`, `admin consumer-reset`, `admin dlq-redrive`, `migrate`) has its own strict parser that rejects ANY unknown `--flag` with a usage error (exit 1) naming the flag and the verb, before touching the filesystem. So `peek --storage memory` already fails clearly at usage level, and the issue's feared confusing `data dir not found` exit 2 is reachable only by pointing `--data-dir` at a path that does not exist, which is the same honest error for any wrong path (a memory broker leaves no artifact by which a verb could detect "this used to be a memory broker", so a tailored error there is impossible, not just unbuilt). The offline verbs read no `IRONBUS_*` env vars, so `IRONBUS_STORAGE=memory` in a unit env file cannot leak into them. The minimal honest surface change is therefore documentation (the live surfaces `/healthz` `/readyz` `/metrics` `/admin` and `top --addr` are the ONLY introspection story) plus one test pinning the unknown-flag rejection for all seven verbs and `top --storage memory` (the zstd-only `dict install`/`dict ls` reject it identically but stay out of the pin so it passes on a default build), so a future shared-parser refactor cannot silently start accepting and ignoring the flag.

**2. Boot-validation interplay sweep (beyond what #446 shipped), each decision pinned and documented per knob in `docs/CLI.md`:**
- `storage.data_dir` config-FILE key: REFUSED under memory mode, exactly like the flag and env forms. This was already true mechanically (the #382 env mapping carries the file key into the #446 refusal); it is now pinned by a test at the parse + `finish_serve` seam (re-seamed from `run` so a mapping regression fails as an immediate panic instead of booting a real broker that hangs the test).
- Segment size, the three retention bounds, `--disk-full-policy`: stay LEGAL with identical semantics. They all operate on the in-memory filesystem exactly as on disk, and the reclaim knobs are MORE load-bearing on a RAM-backed store, where the mandatory byte cap is the OOM guard.
- `--checkpoint-interval`: stays legal and MEANINGFUL. Group cursors, attempts counts, and redelivery semantics hold within the process lifetime; the checkpoint machinery runs against the in-memory fs, deliberately not special-cased.
- `--compact`: ALLOW and document (the issue's preferred option). Compaction rewrites segments in RAM exactly as on disk, and for a keyed changelog topic it genuinely RECLAIMS RAM under the byte cap, so refusing it would remove a real tool; for a non-keyed workload it only spends CPU, which is equally true on disk and is said out loud in the docs.
- Durability knobs (`--durability-level`, flush triggers, `--async-loss-ack`): legal but VACUOUS (no fsync exists on the in-memory fs). The gates are still enforced uniformly (`async`/`none` still demand `--async-loss-ack`; `validate_durability` runs before `validate_storage`, backend-independent), so a script that flips `--storage` never silently changes what it consented to.
- `--ram-ceiling-bytes`: composes unchanged; the docs note stored records are resident RAM ON TOP of the metered buffers, so the two bounds are sized together.

**3. Docs sweep (seven documents):**
- `docs/CLI.md`: the per-knob memory-mode interplay table, the offline-verb story, `IRONBUS_STORAGE` / `IRONBUS_EPHEMERAL_LOSS_ACK` env-mapping rows, the memory-mode startup lines, and the `storage=` echo / `data_dir=none` sentinel note.
- `docs/USAGE.md`: a "Memory mode" subsection (when to use it, the loss contract and the three fail-closed gates, live-only introspection, restart-revives-EMPTY under a supervisor, pair with consumers that tolerate replay-from-live); "Restart and resume" and "Offline inspection" now state they describe the disk backend.
- `docs/CONFIG.md`: the two flag/env-only rows in the `[storage]` section and why they deliberately have no config-file key; the `data_dir` inversion under memory mode.
- `docs/EDGE_TUNING.md`: "When memory mode is the right call" with the #443 measured baseline (SD 9.0 ms fsync / ~50 msg/s / ~70 KB physical flash per 256 B message vs tmpfs 0.5 ms / 92 msg/s, CPU ceiling), a constraint-table row, and the honest tmpfs-baseline paragraph (tmpfs works on every shipped version, keeps broker-restart survival within an uptime and the offline verbs; memory mode is the same trade made fail-closed and self-describing).
- `docs/COMPATIBILITY.md`: memory mode writes no on-disk artifacts, so there is NO upgrade/downgrade/migration story and `migrate` has nothing to gate, said explicitly.
- `docs/DISTRIBUTION.md`: under the unit's `Restart=`, a crashed memory broker is revived EMPTY; no `StateDirectory=`, no writable container volume.
- `docs/CLI_CONTRACT.md`: NO new exit codes; every memory-mode refusal is an ordinary exit 1 usage error under the existing scheme, and the offline-verb rejection is the ordinary unknown-flag exit 1.

**4. Tests:** three new unit tests in `crates/ironbus-cli/src/main.rs` (the seven-verb `--storage` rejection pin, the config-file `storage.data_dir` refusal pin, the shared-engine-knobs-stay-legal pin including `--compact`). No production code changed.

## Gates

- `cargo test --workspace --all-features --locked`: PASS (1478 tests)
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`: PASS
- `cargo fmt --check`: PASS
- `sh scripts/ci/relative-link-check.sh`: PASS (~763 in-repo links, 0 problems)
- `sh scripts/ci/changelog-unreleased.sh`: PASS (`[Unreleased]` / `### Added` entry, Closes #444)


ELares (commented):
## Self-review record (adversarial fact-check panel)

Panel verdict was needs-changes, narrowly; all findings resolved through 816c379. Disposition:

- **Should-fix (the panel's one genuine catch, empirically proven):** the memory-mode unit recipe would have crashlooped verbatim: the shipped unit's Environment=IRONBUS_DATA_DIR fallback cannot be unset from an EnvironmentFile (an empty value still arrives set and is refused), so every start exits 1, the rate-limiter-disabled Restart= loops it, and after N=3 the fall-back machinery performs a spurious BINARY rollback for a config mistake. The recipe now requires the drop-in with UnsetEnvironment=IRONBUS_DATA_DIR, cited against systemd.exec(5) primary wording ("may undo assignments from any configuration source", added in 235, matching the doc's existing floor), and names the consequence of omitting it.
- **Nits resolved:** the env claim is scoped to "no IRONBUS_* environment variables" (top reads NO_COLOR; the load-bearing half holds) in the doc, CHANGELOG, test comment, and PR body; the rejection pin now covers eight verbs including top (with the zstd-only dict equivalents named and excluded for default-build reasons); the config-key refusal test is re-seamed off run() so a mapping regression fails as an immediate panic instead of a hung broker, with a Drop cleanup guard and a TEST-NET-3 tripwire.

Verified true by the panel, unchanged: the offline-verb rejections are usage-level before any filesystem access (mutation-checked), the interplay decisions match live behavior (a memory broker with --compact and every shared knob booted and served), the quoted hardware numbers match issue #443's record exactly, and the link check passes. Zero production code in this PR.

PR #448, opened 2026-06-11
test(cli): memory-mode validation: restart contract, OOM-bound gate, bench --storage memory (#445)

### LINKED ISSUE #445: [TASK] Memory-mode validation: acceptance variant, OOM-bound gate, bench (child of #443)

ELares:
Closes #445 (parent #443 shipped in #446; refs #439, refs #115).

Per the issue comment thread, item 1 is trimmed to what remained after #446 (which already shipped the real-wire golden-path memory round trip: produce/consume/ack, no redelivery, clean SIGTERM, no-files assertions), and item 2 folds the store into the memory-backend RAM-footprint proof.

## 1. The restart contract (`crates/ironbus-cli/tests/golden_path.rs::memory_mode_restart_revives_an_empty_broker`)

Real binary, two lives on the SAME flags: life 1 produces 3, acks all 3, SIGTERMs (exit 0); life 2 boots fresh and is asserted EMPTY in both machine-checkable senses: a consumer fetches 0 messages AND the durable-offset counter restarts at 0 (the first produce returns offset 0, not 3). The exact inverse of the disk recovery tests, with the contract pinned loudly in comments: if data ever survives the restart, the ephemeral backend has silently grown persistence. The memory-broker scratch-dir harness also gained a process-wide counter so parallel memory-mode tests cannot remove each other's no-files scratch dirs.

## 2. The OOM-bound gate

**(a) Real-wire shed + bounded RSS** (`golden_path.rs::memory_mode_sheds_past_the_byte_cap_and_rss_stays_bounded`): offers 192 MiB of INCOMPRESSIBLE (per-message LCG noise, so the default lz4 write path cannot shrink stored bytes under the cap) payloads against a 2 MiB `--max-total-bytes`. Asserts: the at-capacity drop-new shed reaches the producer as the existing typed `at capacity` error on a connection that STAYS OPEN; the broker process survives the whole flood and still sheds promptly afterwards; RSS (read externally via `ps -o rss=`, sampled every 256 produces DURING the flood and again after) stays under a generous fixed bound of `2 * cap + 96 MiB`. Measured on this run: accepted 32, shed 3040, final RSS ~16.6 MiB against the 100 MiB bound while 192 MiB was offered, i.e. RSS does not grow with offered load.

Boundedness scope (documented design exclusion, panel finding 3): the bound holds for ack-progressing (or, as in this test, consumer-free) workloads. The dead-letter sink is DELIBERATELY byte-uncapped (a poison record is the durable evidence of a dropped message and must never itself be shed), and in memory mode it lives on the same in-memory filesystem, OUTSIDE `--max-total-bytes` and the #115 modeled floor, so a poison-heavy workload (consumers that never ack, dead-lettering at `--max-deliver`, default 5) grows RSS past the modeled floor. Capping the DLQ would shed poison evidence, a different design decision this PR does not make unilaterally; the exclusion is now stated in the `worst_case_buffer_bytes` model rustdoc, the `validate_ram_ceiling` rustdoc, the refusal message's store note, `docs/RAM_BUDGET.md` term4mem, and the `--storage` rows of `docs/CLI.md` / `docs/USAGE.md` (pair memory mode with consumers that ack, monitor `ironbus_dlq_records_total`, tune `--max-deliver`).

**(b) The #115 interplay, the store fold**: `validate_ram_ceiling` modeled connections/credits/groups/in-flight only, so a memory-mode config whose BUFFER terms fit the ceiling booted no matter how large its in-RAM store cap was: the edge-tiny knobs (~15 MiB of buffers) under their 64 MiB ceiling booted with `--max-total-bytes 1GiB --storage memory`. (Under the server-sized balanced defaults that ceiling was already refused on term 1 alone, 256 connections x 8 MiB of credit bytes, fold or no fold; an earlier revision of this PR claimed the default-knob pair "booted before this fold", which was false as stated and is corrected everywhere it appeared.) The model now is:

```
disk:   worst_case = term1(conns x credits) + term3(groups x in-flight) + term5(fixed + stacks)        (bit-for-bit unchanged)
memory: worst_case = term1 + term3 + term5 + IN_MEMORY_STORE_IMAGES (2) * max_total_bytes
```

The 2x is not a guess: `InMemoryFile` (read from `crates/ironbus-storage/src/{fs,io}.rs`) keeps TWO byte images per file, the live bytes plus the durable image each `sync_data` clones them into (the power-loss-simulation contract); the directory-level `sync_dir` clone is only a map of `Arc` pointers, no third byte image. The 2 is the steady-state RETAINED set, not an instantaneous bound (panel finding 4): the durable image's `clone_from` realloc transient can briefly exceed it mid-sync; amortized, exactly two images per file are retained, which is what a steady-state footprint proof may honestly charge. So the worst-case in-RAM store is `2 * max_total_bytes`, charged as a new additive `in_memory_store_bytes` term in `ironbus_server::rss::RamFootprintConfig` (0 in disk mode). A memory-mode ceiling below the modeled floor REFUSES TO BOOT with a message naming the store term, the 2x clone headroom, and `--max-total-bytes` as the knob to lower.

Pinned, with the multiplier itself mutation-verified (panel findings 1 and 2):

- the rss model test asserts the memory worst case grows over disk by a HARD LITERAL `2 * cap`, never `IN_MEMORY_STORE_IMAGES` (an expectation computed from the constant is self-referential and follows a drifted constant); the refusal-direction model test's lower bound is a literal 2 GiB for the same reason
- a validate-level test places the 64 MiB edge-tiny ceiling BETWEEN the one-image floor (~47 MiB, boots) and the two-image floor (~79 MiB, refuses) for a 32 MiB store cap, so a 1x mutant boots that config and fails
- the real-binary refusal test boots `--profile edge-tiny` with a 1 GiB in-RAM store, so the ceiling covers every buffer term and ONLY the store term can force the refusal (an invocation that genuinely booted before the fold); it polls with a 10s deadline, so a fold-less mutant that boots fails loudly instead of hanging the suite
- mutation evidence, recorded in the step commits: `IN_MEMORY_STORE_IMAGES = 1` fails `rss::tests::the_in_memory_store_is_charged_at_two_byte_images`, `rss::tests::a_memory_store_past_the_ceiling_is_refused_where_the_disk_config_fits`, and `tests::the_store_fold_charges_two_images_a_one_image_mutant_boots_here`; removing the fold (`StorageArg::Memory => 0`) boots the real binary and fails `memory_mode_refuses_a_ram_ceiling_below_the_store_floor` at its deadline ("the broker BOOTED"). Both mutants were applied, observed failing, and reverted; the suite is green on the restored code

`docs/RAM_BUDGET.md` (which the refusal message cites) states `term4mem`, the steady-state caveat on the 2x, and the DLQ exclusion.

## 3. `ironbus bench --storage memory`

Threaded exactly as the comment shaped it: the bench broker is the real serve path, so bench threads `--storage memory` plus the consent plus a default cap through its spawn. `IsolatedBroker` is now generic over the `Filesystem` (mirroring `run_broker`'s static two-armed dispatch); the memory arm spawns via the shared `open_memory_engine`, with bench supplying `ephemeral_loss_ack` (its synthetic broker's data is disposable by design, the same way it owns the disk dir auto-delete) and a default 256 MiB in-RAM byte cap; the spawned config is run through the REAL `validate_serve_config` boot gates so bench can never spawn a config serve itself would refuse. No synthetic data dir exists on the memory arm (the test asserts none is ever created).

Honest reporting: `fsync_measured` is `false` in memory mode and `fsync_cost_us` is null (the in-memory engine issues NO fsync, so there is no durable-write cost to attribute; the human view names the in-memory target and states that reason, distinct from the `--no-fsync` dry-run wording). The JSON gains an ADDITIVE `storage` field (schema version unchanged, the #439 `payload_entropy` precedent). `--storage` with `--addr` is refused (the flag shapes only the spawned broker). The #439 entropy modes apply unchanged and are pinned composing with the memory backend at both the parse and binary level; the #439 golden-vector fill tests are untouched and green.

## 4. CI

No new job: every new test is an ordinary `ironbus-cli` unit/integration test and runs inside the existing workspace `test` matrix (the golden-path additions live in `tests/golden_path.rs`, which the workspace test job already executes; the named acceptance CI job is a separate single scripted scenario and is untouched). The storage conformance suites already exercise `InMemoryFs`, the same filesystem the memory backend serves over.

## Panel review follow-ups (4 findings, all resolved on this branch)

1. **2x multiplier unpinned**: literal-2 expectations in the model tests plus a validate-level test whose ceiling sits between the 1x and 2x floors; `IN_MEMORY_STORE_IMAGES = 1` now fails three tests (was: zero).
2. **Refusal test did not pin the fold**: the real-binary test now uses edge-tiny connection terms so only the store term can refuse, and a bounded poll turns a booting mutant into a failure; fold removal now fails the test (was: passed). The false "booted before this fold" wording is corrected in the rustdoc, test comments, CHANGELOG, and this body.
3. **In-RAM DLQ exclusion**: documented as a deliberate design exclusion (no behavior change; capping the DLQ would shed poison evidence) at the model, refusal message, RAM_BUDGET, and operational docs; "RSS stays bounded" claims are scoped to ack-progressing workloads.
4. **"Exactly 2 images" nit**: documented as the steady-state retained set with the `clone_from` realloc transient caveat everywhere the 2x appears.

## Gates (all re-run locally on the review-fix commits, all clean)

- `cargo test --workspace --all-features --locked`: exit 0 (40 suites ok, 0 failed)
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` with `RUSTFLAGS="-D warnings"` (the workspace lints set clippy `all` + `pedantic` to warn, promoted to deny): exit 0
- zstd job as CI runs it: `cargo build --workspace --features zstd --locked && cargo test --workspace --features zstd --locked`: exit 0 (40 suites ok)
- `sh scripts/check-format-registry.sh`: ok (digest unchanged; no stored-format change in this PR)
- Environment note for the record: on the dev macOS host (APFS at 99% capacity), `ironbus-storage`'s pre-existing `io::std_file_tests::preallocate_reserves_real_blocks_on_a_supporting_fs` (a real-filesystem block-reservation tooth, byte-identical to `main`, untouched by this PR) flaked under full-suite disk pressure and passed standalone and on rerun; the green exits above are full honest runs with that test passing. CI runs this suite on ubuntu ext4 where the reservation is deterministic.

CHANGELOG: `[Unreleased]/### Added` entry (Closes #445).


ELares (commented):

Panel verdict was needs-changes; all four findings resolved through e322b6c. Disposition:

- **2x store multiplier was unpinned (panel mutation-proved: a 1x mutant survived the whole suite via a self-referential expectation):** the model test now asserts against the hard literal with the derivation comment (InMemoryFile retains exactly live + durable; sync_dir clones Arc pointers only), and a new validate-level test places the ceiling BETWEEN the one-image and two-image floors. The 1x mutant now fails three tests; restored, green.
- **The refusal test did not pin the fold (balanced defaults' 2 GiB connection term tripped the ceiling regardless; "booted before this fold" was false as stated):** the real-binary test now boots edge-tiny (~15 MiB of buffer terms under the 64 MiB ceiling) with a 1 GiB in-RAM store so ONLY the store term can force the refusal, with a bounded poll so a booting mutant fails instead of hanging. Fold-removal mutation now fails at the deadline; the false wording is corrected everywhere it appeared.
- **The in-RAM DLQ exclusion (a real hole in the boundedness story):** documented, not capped (shedding poison evidence is a design decision this PR must not make unilaterally): the exclusion is stated in the footprint model's NOT-counted list, the refusal message's store note, RAM_BUDGET.md, and both flag docs; every "RSS stays bounded" claim is scoped to ack-progressing workloads with the monitoring guidance (ironbus_dlq metrics, --max-deliver).
- **"Exactly 2 images" scoped as the steady-state retained set** with the clone_from realloc-transient caveat at every site.

Unchanged and verified: the restart-revives-empty contract test (asserted two ways), the OOM real-wire gate (192 MiB offered against a 2 MiB cap: 32 accepted / 3040 shed on a live connection, RSS ~16.6 MiB), disk ceiling verdicts bit-identical, bench --storage memory with honest fsync_measured=false and the #439 entropy pins intact.

PR #449, opened 2026-06-11
fix(bench): allow the unix-only fsync_is_measured on non-unix builds

ELares:
## Problem

The #448 squash broke the `test (windows-latest)` lane on main: `BenchConfig::fsync_is_measured` is called only from `bench_run.rs`, whose callers sit behind `cfg(unix)` (the bench broker is the real `serve` path, Unix-only in v1). On a non-unix build the method is dead code, and the CI wall (`RUSTFLAGS=-D warnings`) refuses the compile:

```
error: method `fsync_is_measured` is never used
```

The branch CI showed the failure but the merge was gated on checks settling rather than all-green; that process gap is mine, recorded here for honesty.

## Fix

One attribute with the explanation: `#[cfg_attr(not(unix), allow(dead_code))]` on the method. The unit tests that call it compile on every platform (tests are not cfg-gated), the unix builds keep the full warning wall, and the method stays available for a future Windows-enabled serve.

Verified: unix clippy pedantic clean, ironbus-cli unit tests green. The Windows lane on this PR is the authoritative check (the local cross-target stdlib is not installed).

### DIFF
--- a/crates/ironbus-cli/src/bench.rs
+++ b/crates/ironbus-cli/src/bench.rs
@@ -200,6 +200,10 @@ impl BenchConfig {
     /// (the in-memory engine issues NO fsync at all, so there is no durable-write cost to
     /// attribute; reporting one would be dishonest) and only outside the `--no-fsync` dry run
     /// (which batches the cursor checkpoints, so the per-ack durable path is not exercised).
+    /// The callers live in `bench_run.rs` behind `cfg(unix)` (the bench broker is the real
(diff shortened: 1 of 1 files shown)


ELares (commented):
Review of record: one-attribute hotfix for the Windows dead-code break that #448's merge landed (my gating ran on checks-settled rather than all-green; this PR merged only after the windows lane was explicitly verified green, 28/28). The method stays available on every platform; unix warning wall intact.

Issue #450, opened 2026-06-11
[PERF] Pipelined publish: amortize the group commit across an in-flight window


ELares:
# [PERF] Pipelined publish: amortize the group commit across an in-flight window

## Hypothesis

The engine ALREADY group-commits: the append actor (#177, #49) drains every queued `Command::Produce` into one pass, appends each with `append_no_sync`, and issues ONE covering `commit_batch` (`fdatasync`) before releasing any reply. But a single connection can never form a group, because both ends serialize on one awaited ack per publish. If ONE connection keeps a window of W un-acked PUBs in flight, the existing batcher amortizes the ~4.5 ms SD-card fsync across the whole window with NO change to what an ack means (I2, ack-after-fdatasync, unchanged).

Hive baselines (RPi4 / SD card, 256 B payloads): IronBus durable pub-only 203 msg/s (one awaited ack per publish, one fsync each at ~4.5 ms); NATS JetStream `sync_interval=always` pipelined ~203 msg/s (it does not amortize the sync across the async-publish window). Target: at least 10x the NATS sync-always pipelined number, i.e. >= 2,030 durable acked msg/s on the hive, measured post-merge by the on-hive bench.

## Recon findings (where the serialization actually lives)

- **The wire format already permits pipelining.** CONTRACTS.md / `ironbus-proto`: there is NO correlation id; responses are self-describing per verb (#179) and per-connection ordering IS the contract. The session appends exactly ONE reply frame per non-fire-and-forget `Pub` (`PubAck` tag 14, `PubAckDuplicate` tag 20, or `Err` tag 12) to its output buffer in frame order, and the connection loop writes that buffer in order. So N outstanding PUBs with FIFO acks are ALREADY legal on the frozen wire format. Pipelining is a pure client + session BEHAVIOR change; zero wire change, zero new frame, zero new flag.
- **Client** (`crates/ironbus-client/src/lib.rs`): `Client::produce` / `produce_dedup` writes one `Pub` frame and then BLOCKS in `read_frame()` for its `PubAck`. One awaited ack per publish, so the producer-side window is always 1.
[...]

PR #451, opened 2026-06-11
perf(server,client): pipelined publish window amortizes the group commit (#450)

ELares:
## Problem (#450)

IronBus had no pipelined publish path: every producer awaited its PubAck (and the covering fdatasync) before sending the next PUB, so durable throughput was pinned to fsync latency (203 msg/s on the reference RPi4/SD hive) even though the engine's group-commit batcher can cover many produces with one fsync. NATS sync-always measured ~203/s pipelined on the same hardware because it does not amortize; the round 1 target is >=2,030/s (>=10x) durable acked.

## Design

- Session (wire-compatible, no format change): within one pass, Pub frames are SUBMITTED to the append actor without awaiting; the pending ack is PARKED and released in FIFO submission order before any non-produce frame's reply, at the 1024-entry safety cap, and at the end of the pass. The actor drains the window as ONE batch under one group-commit fsync. I2 untouched: the actor releases no reply before the covering fsync, so a parked ack can never precede its durability.
- ParkedPub carries the fire-and-forget marker: a QoS-0 produce gets no frame in either disposition (the #11 contract), without desyncing the framed FIFO. write_pub_reply is byte-identical to the pre-pipelining inline match, dedup hits included.
- Client: produce_window writes the whole window, then drains exactly one reply per message FIFO; a server Err consumes its slot and the drain continues (connection stays usable); fire-and-forget is forced clear (no reply would desync the window). IO errors and unexpected frames abort hard.
- Bench: --pubwindow <n> (default 1, the historical path), additive pubwindow JSON field; the windowed path attributes the window's elapsed time evenly across its messages so the per-op fsync sample reflects the amortization honestly.

## Tests

Session: FIFO ack order across a pipelined window; a malformed PUB mid-window consumes its reply slot in order; a pipelined window group-commits with FEWER fsyncs than N (the headline property, counter-asserted). Client: real-wire window round trip (FIFO consecutive offsets, empty-window no-op, connection healthy after). Bench: knob parse + zero refusal. Full workspace suite, clippy pedantic, fmt, format-registry digest unchanged, link check: all green.

## Process note

The adversarial review panel could not run this round (subagent capacity limit); a rigorous inline review of the parked-release seams (QoS-0 interplay, dedup frames, cap, mid-drain fatality) stands in and is posted as the review comment. Panel review can be retrofitted next round if wanted. Hive measurement lands post-merge in docs/PERF_LEDGER.md.

Refs #450 (closes after the hive-measured ledger entry).


ELares (commented):
Review of record (inline, panel unavailable this round per the PR body): verified the parked-release seams directly: (1) ParkedPub carries the QoS-0 marker so a fire-and-forget produce emits no frame in either disposition without desyncing the framed FIFO; (2) write_pub_reply is byte-identical to the pre-pipelining inline match incl. PubAckDuplicate and the fenced/error arms; (3) the 1024 cap bounds parking and the cap-drain plus before-non-produce-reply drains preserve the per-connection reply order; (4) a mid-drain fatal drops the remaining entries un-replied, which is sound because a batch-covering fsync failure marks every member fatal and the session is closing; (5) the client window forces fire_and_forget clear and drains every slot on a server Err so the connection survives. Session tests pin FIFO, the malformed mid-window slot, and fewer-fsyncs-than-N; the client test pins the real-wire round trip. Guardrails: ack semantics unchanged, wire format untouched (registry digest verified), all gates green locally.

PR #455, opened 2026-06-12
perf: 64 KiB connection read chunk + opt-in --commit-gather-us group-commit window

### LINKED ISSUE #454: [PERF] Bounded group-commit gather so a pipelined window commits in one fsync (round 3)
## Hypothesis (PERF_LEDGER round 3)

The actor's group-commit batch is whatever arrived during the previous batch's fsync (`recv` one, `try_recv` the rest), so the steady-state batch self-sizes to arrival-rate x fsync-latency, not to the client's pipelined window: a 512-window pays several fsyncs per window. Mean-batch-size instrumentation (produced_total / fsync_duration_seconds_count, no code needed) will confirm on the hive; the measurement is queued behind a Teleport cert renewal.

Add an opt-in bounded gather: `--commit-gather-us <n>` (default 0, byte-identical off). When a drain pass holds at least one produce and the gather window is set, the actor keeps collecting commands with `recv_timeout` until the window elapses or the pass reaches a full pipelined batch, so one fsync covers the whole client window. Acks still mean fsynced-durable; the knob trades up to n microseconds of added latency under produce bursts for fewer, larger barriers. Precedent: MySQL binlog_group_commit_sync_delay, PostgreSQL commit_delay.

Expected: closes the leg-2 gap (median ~1,951 -> >=2,030 msg/s at window 512; ceiling bounded by the ~205us/op CPU floor measured via memory mode at 4,882/s).

refs #452, refs #450

ELares:
Closes #454.

## What the instrumentation showed (and disproved)

Round 3 started with the round-2 hypothesis: the actor drain self-sizes its group-commit batch to arrival rate x fsync latency, so a 512-record pipelined window pays ~40 fsyncs. The zero-code measurement (delta produced_total / delta fsync_count on the reference edge box) confirmed mean batches of 12-13 records against a 512 window.

The gather knob alone then DISPROVED the mechanism: with gather windows of 1/3/5/13 ms the batch stayed at 13-15 records and throughput fell monotonically (2,384 -> 645 msg/s). The batch was not arrival-sized; it was CAPPED. The pipelined window (#450) is pass-scoped, a `process` pass sees at most one connection-loop read chunk, and the chunk was a 4 KiB stack array: ~13 frames of 256 B realistic payloads, after which the session blocks awaiting its parked acks while the actor waits out its gather. The two waits compound; nothing feeds.

## Changes

1. **Connection read chunk 4 KiB -> 64 KiB zero-page heap buffer** (the actual win). A pass now carries hundreds of frames, so a pipelined window group-commits in a few fsyncs instead of ~40. `vec![0u8; 64*1024]` is `alloc_zeroed`-backed: untouched pages cost no resident memory, so idle or ping-only connections still touch about a page.
2. **`serve --commit-gather-us <us>` / `IRONBUS_COMMIT_GATHER_US`** (default 0 = off, validation-capped at 1 s): an opt-in bounded group-commit gather for multi-producer fan-in (the MySQL `binlog_group_commit_sync_delay` / PostgreSQL `commit_delay` precedent). It engages only when a drain pass already holds TWO OR MORE produces, so an unpipelined producer (one in-flight produce, awaiting each ack) never pays the window. Acks keep their fsynced-durable meaning at any setting; `0` is byte-identical to the historical actor.

## Hive A/B (pre-merge cross-compiled armv7 musl, 256 B realistic, 15 s runs)

| leg | before (merged main) | after |
| --- | --- | --- |
| disk w=1 | 201/s | 198/s (fsync physics, unchanged) |
| disk w=64 | 1,502/s | 5,215/s |
| disk w=512 | ~1,951/s | **7,193/s** (batch 90 rec/fsync) |
| disk w=1024 | 1,956/s | **9,197-10,294/s** (batch 104) |
| memory w=1 | 980/s | 1,364/s |
| memory w=512 | 4,882/s | 8,023/s |
| memory w=1024 | -- | 11,171/s |

Gather on a single pipelined connection now measurably hurts (6,839 vs 7,193 at w=512), which is why the default stays 0. Guardrails verified on the hive: binary 1.53 MB (<= 3 MB), peak RSS under w=1024 load 2.1 MB (<= 8 MB), ack-after-fdatasync default untouched.

## Tests

- `a_commit_gather_window_collects_a_spaced_produce_into_the_pipelined_batch`: sync-gate-deterministic proof that a produce arriving mid-gather joins the in-progress batch under ONE covering fsync.
- `a_single_inflight_produce_never_pays_the_gather_window`: the no-tax rule; a solo produce acks far inside the window with exactly one sync.
- `the_commit_gather_flag_parses_defaults_off_echoes_and_caps_at_one_second`: default-off, parse-cursor regression guard (the `--pubwindow` bug class), materialized-config echo, 1 s cap as a usage error.
- Full workspace suite green locally; every existing actor/session/crash-recovery contract passes unchanged (gather 0 is the delegating default everywhere).

Review-of-record: panel capacity is unavailable until Jun 13; this PR carries the inline adversarial pass (the gather-only sweep that falsified the original mechanism is documented above and in `docs/PERF_LEDGER.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ to follow Semantic Versioning once it reaches a tagged release.
 ## [Unreleased]
 
 ### Changed
+- The per-connection read chunk grew from a 4 KiB stack array to a 64 KiB zero-page heap buffer (#454). The pipelined produce window (#450) is pass-scoped, and a pass sees at most one read chunk, so 4 KiB silently capped the publisher pipeline at ~13 records per group commit on 256 B payloads (~40 fsyncs per 512-record window on the reference edge box). Untouched pages cost no resident memory, so idle connections are unaffected.
+
+### Added
+- `serve --commit-gather-us <us>` / `IRONBUS_COMMIT_GATHER_US` (#454): an opt-in bounded group-commit gather window. When a drain pass already holds at least two produces (evidence of a pipelining publisher; an unpipelined producer never pays the window), the append actor keeps collecting commands for up to the window before committing, so a pipelined publisher's whole in-flight window lands under one covering fsync instead of arrival-rate-sized slivers (measured on the reference edge box: a 512-record client window was committing as ~12 records per fsync). Default `0` keeps the actor byte-identical; acks keep their fsynced-durable meaning at any setting; validation caps the window at 1 second. The materialized-config line gains a `commit_gather_us=` field.
+
+### Changed
 - Renamed the published release binary assets to friendly CPU-arch names: `ironbus-linux-amd64`, `ironbus-linux-arm64`, `ironbus-linux-armv7` (dropping the `unknown`-vendored `musl` triple, which stays the internal cargo/cross build target). The installer (`detect_target` + download URL), its tests, both release workflows (staging, `SHA256SUMS`, attestation, release assets, notes table), and the install docs (README, DISTRIBUTION, RELEASING) are updated to match. The `.deb` packages were renamed to match (`ironbus-linux-amd64.deb`, `ironbus-linux-arm64.deb`, `ironbus-linux-armv7.deb`), so no published asset carries the `unknown` triple; it stays the cargo `--target` only. The binaries are still static `musl` with no runtime dependency.
 
 ### Fixed
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -110,7 +110,7 @@ use ironbus_core::lease::LeaseConfig;
 #[cfg(unix)]
 use ironbus_core::types::RecordFlags;
 #[cfg(unix)]
-use ironbus_server::actor::{spawn_actor, DEFAULT_CHANNEL_BOUND};
+use ironbus_server::actor::{spawn_actor_with_gather, DEFAULT_CHANNEL_BOUND};
 #[cfg(unix)]
 use ironbus_server::engine::{DiskFullPolicy, DurabilityLevel, Engine, EngineConfig};
 #[cfg(unix)]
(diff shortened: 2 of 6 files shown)


ELares:
Review of record (inline adversarial pass; panel capacity returns Jun 13):

1. **Does the chunk bump break the slowloris defense?** No. The read/write timeouts are per-syscall progress bounds and are untouched; a larger buffer changes how much one read MAY return, not how long a stalled client can hold the slot.
2. **RSS regression risk at max_connections=256 (balanced profile)?** Worst case 256 x 64 KiB = 16 MiB IF every connection simultaneously fills its chunk; the buffer is alloc_zeroed so pages materialize only as the kernel writes them, and hive measurement under sustained w=1024 load shows 2.1 MB peak RSS for the whole broker. The edge-tiny profile caps at 32 connections.
3. **Can the gather stall a Shutdown?** A Shutdown gathered mid-window is processed in order after the batch, identical to arriving in the same drain burst; the window is validation-capped at 1 s so worst-case added shutdown latency is bounded.
4. **Can the gather violate I2 (ack-implies-durable)?** No. The gather only delays WHEN commit_batch runs; replies are still released only after the covering fdatasync. The two new actor tests pin both the batching and the no-tax rule.
5. **Why keep the gather at all if it hurts the single-connection bench?** It is default-off and exists for multi-producer fan-in (many connections each sending singles), which the single-connection bench cannot exercise; the >=2-produce engage rule makes it free for everyone else.
6. **Falsification honesty:** the round began as a gather hypothesis; the gather-only sweep DISPROVED it (batch pinned at 13-15, throughput fell to 645/s at 13 ms) and located the real cap in the read chunk. Both the failed and successful mechanisms are in docs/PERF_LEDGER.md.

Issue #490, opened 2026-06-19
[PERF] Poll rescans from committed every time (O(W)); cache a monotonic next-deliverable cursor per group


ELares:
`engine.rs:3289` / poll_in:`4540` / direct:`2264` — `offset=committed; while offset<window_end` rescans the whole prefix every poll, no memory of where it stopped; sparse/saturated windows re-walk tens of thousands of offsets/poll, ×`max` in poll_batch. NATS advances an O(1) `sseq`/`dseq` cursor. Add a per-group `next_deliverable` hint advanced as the contiguous claimed/acked prefix grows → amortized O(deliveries). The ONLY asymptotic consume loss vs NATS.

### Children (consume-path O(1) cursor)
- [x] #484
- [x] #485
- [x] #486
- [x] #487
- [ ] #488
- [x] #489
- [ ] #491

Do NOT regress: RLE ack cursor (memory-compact vs NATS pending map); generation-fenced leases.

_From the IronBus maturity/perf teardown (storage + consume + produce + NATS-prior-art analyses + a measured produce/consume matrix vs NATS Core & JetStream)._

ELares:
Fresh-eyes analysis STOPPED rather than ship a wrong/no-op cursor. Proof:
- FULL hint (advance past in-flight offsets — the only version that removes the O(W) cost): UNSOUND. The re-walked region [committed, first-deliverable) is dominated by IN-FLIGHT leases; lease expiry is PASSIVE/eventless (no janitor; `expired()` has no non-test caller — only the next poll's scan reads deadlines). A hint past a leased offset is STRANDED when that lease passively expires = MESSAGE LOSS. nack/release fire events (retreatable); passive expiry does not. Cannot prove no-skip.
- CONSERVATIVE hint (advance only over the acked-contiguous-from-committed prefix): SOUND but a provable NO-OP — by the cursor invariant `committed` is never acked, so the acked prefix at committed is empty; the lowest re-walked offset is always an in-flight lease at committed. Zero rescan removed.
[...]

PR #512, opened 2026-06-19
perf(lease): O(log W) expiry via a deadline min-heap, drop the O(W) scan (#485)

### LINKED ISSUE #485: [PERF] LeaseTable::expired is an O(W) full scan with no timer; add a (deadline,offset) min-heap
`lease.rs:443` `leases.iter().filter(now>=deadline)` scans all in-flight every sweep; no heap/timer. Add a BinaryHeap<(deadline,offset)> → next expiry O(1) peek, O(log W) maintain. (NATS itself scans bounded by MaxAckPending — we can beat it cheaply.)

_From the IronBus maturity/perf teardown (storage + consume + produce + NATS-prior-art analyses + a measured produce/consume matrix vs NATS Core & JetStream)._

ELares:
Closes #485.

## What

`LeaseTable::expired()` scanned every in-flight lease (**O(W)**) on each janitor sweep to find the expired ones — no timer, no index. This adds a deadline-ordered min-heap so the soonest expiry is **O(1) to peek** and **O(log W) to maintain**, dropping the full scan.

- New field `deadlines: BinaryHeap<Reverse<(deadline, offset)>>`. The `BTreeMap<offset, Lease>` stays the **source of truth**; the heap is only an index, reconstructible from the leases alone (not part of the table's logical state).
- Every path that **sets or changes a deadline** now pushes a fresh `(deadline, offset)` entry: `claim` (first delivery + redelivery), `extend`, `nack`, `defer_redelivery`. `ack` and `release_below` drop leases **without** touching the heap.
- `expired()`: (1) reaps dead entries off the top, (2) returns empty in **O(1)** when the soonest live deadline is in the future (the hot path for a quiet sweep), (3) otherwise drains the due entries — re-pushing the valid ones so the call stays **repeatable** — and returns them in ascending-offset order. It now takes `&mut self` for the lazy reaping, but leaves the table's logical state unchanged.

## Lazy-deletion correctness (no expiry missed or duplicated)

A heap entry is **valid** iff a live lease at its offset still carries that deadline (`is_live_deadline`). The match key is the **deadline alone, deliberately** — `expired()` decides reclaimability purely from the live lease's current deadline vs `now`, so an entry is meaningful exactly when some live lease still holds that deadline, whichever heap entry happens to carry it.

- **Index invariant:** for every live lease `L`, the heap contains at least one entry `(L.deadline, L.offset)` — the one pushed when that deadline was last set (every deadline write pushes; nothing rewrites a live entry in place). So the soonest *valid* entry is the soonest *live* deadline.
- **No missed expiry:** reaping only ever drops entries that name a deadline no live lease holds; by the invariant a dead entry is never a live lease's current entry, so dropping it cannot hide a real expiry.
- **No duplicate / no double-expire:** the per-offset de-dup in the drain keeps a coincidentally reused deadline value (which leaves an aliased valid entry) from emitting an offset twice; a reused value still names a deadline the live lease genuinely holds, so it is harmless.
- **Repeatable & non-monotonic `now`:** `expired()` consumes no expiry (only `claim`/`ack` change a lease) — it re-pushes every valid entry it drains — so calling it again with the same `now`, or with `now` moving backward across sweeps, returns the same set.

Generation fencing, `release_below`, and `attempt_counts` are untouched; reclaimability is still read straight off the live `BTreeMap`.

## Preserved + added tests

- **Every existing lease unit test and proptest passes unchanged** (generation fence, `release_below`, `attempt_counts`, carried/resume, hard-cap, `holds_active`, `is_claimable`, `only_the_latest_token_ever_acks`).
- Added: repeatability under non-monotonic `now`; stale-entry reaping after ack/extend/nack/defer; `release_below` not resurrecting expiries; and a **proptest that cross-checks the heap-driven `expired()` against a brute-force O(W) scan** under random interleavings of claim/ack/extend/nack/defer/release at many probe times.

## Local gates (RUSTFLAGS=-D warnings)

- `cargo fmt --all --check` — pass
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` — pass
- `cargo test --workspace --all-features --locked` — pass (all suites `0 failed`)

Change is contained to `crates/ironbus-core/src/lease.rs`.

## What a reviewer should scrutinize

- The deadline-only match key (vs also fencing on generation) — argued above to be sufficient and minimal.
- The `&mut self` signature change on `expired()` (the only call sites are in-crate; the heap reap is the reason).
- The step-3 drain re-pushes valid entries so the heap size is non-increasing across `expired()` calls (no unbounded growth) and the call is idempotent.

### DIFF
--- a/crates/ironbus-core/src/lease.rs
+++ b/crates/ironbus-core/src/lease.rs
@@ -24,7 +24,8 @@
 //! claim/max-deliver so orphaned leases are eventually evicted.
 
 use crate::types::Offset;
-use std::collections::BTreeMap;
+use std::cmp::Reverse;
+use std::collections::{BTreeMap, BinaryHeap};
 
 /// Visibility-timeout and hard-cap tunables, in nanoseconds of monotonic time.
 ///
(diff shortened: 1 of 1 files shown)


ELares:
Fresh-eyes review — NOT merging. The deadline min-heap is correctly implemented and the differential proptest is solid, but `LeaseTable::expired()` has no caller outside the lease unit tests (verified by grep across the workspace), and the heap is GC'd ONLY by `expired()`. The engine redelivers on the next poll via an inline O(log W) per-offset lease check; it never calls `expired()`. So in production the `deadlines` heap grows UNBOUNDED — one entry pushed per claim/extend/nack/defer, never reaped — a memory leak — plus an O(log W) push on every delivery for no realized benefit. The unit tests call `expired()`, which masks the leak. The lease table is not an O(W) production bottleneck: the visibility-timeout check is an O(log W) single lookup per polled offset; the genuine consume-path O(W) is the poll's offset rescan (#490). The heap becomes worthwhile only paired with a NATS-style timer-driven redelivery sweep that actually calls `expired()`. Branch preserved if such a sweep is designed.

Issue #588, opened 2026-06-19
[M2-I10] Wire frames for streams/subjects (tags 22+) with back-compat


ELares:
Extend the wire protocol with a Connect capability bit and StreamDeclare/StreamInfo/Bind/PubTo/SubTo frames (tags 22+), plus a flag-gated stored subject field on the record. First-principles: multi-stream is strictly additive, so bare Pub/Sub must map to the default stream "" and remain byte-identical for existing clients. New capability is negotiated, never assumed, preserving back-compat.

Issue #585, opened 2026-06-19
[M2-I9] Subject->stream binding + single-home resolution


ELares:
Implement subject->stream binding with single-home resolution: each subject resolves to exactly one stream, returning typed NoStream / Ambiguous rejects when no binding or multiple bindings match. First-principles: storing a record requires a single unambiguous destination log, so resolution must be a total function with explicit error variants. Fail-closed reject on an unbound subject beats NATS, which silently drops a publish that matches no interest while still returning a successful PubAck.

Issue #567, opened 2026-06-19
[M2-I6] Subject grammar with fail-closed validation at ingest


ELares:
Define a bounded, depth-capped subject grammar with fail-closed rune validation enforced at ingest. First-principles: an unbounded subject space is an unbounded routing-state and cardinality hazard, so capping depth and validating runes at the edge keeps the trie and metrics bounded. Beats NATS, whose permissive public subject validator admits pathological subjects; we reject them before they enter storage.

PR #651, opened 2026-06-19
feat(routing): wait-free arena subject-routing trie (#568)

### LINKED ISSUE #568: [M2-I7] Wait-free arena token-trie for subject routing
Build a wait-free arena-allocated token-trie for subject routing: ArcSwap snapshots, xxh3 token hashes, an LRU cache, and fork-bounded wildcard expansion. First-principles: routing must resolve subject->subscribers in O(tokens) on the publish hot path without locking the writer. Beats NATS Sublist: O(1) cached publish, no global cache flush on bind, no O(1024) wildcard-unsub scan, and no exponential pwc fork.

ELares:
Closes #568 (V2-M2).

A wait-free, arena-flattened subject-routing trie ("Sublist") in `ironbus-core::sublist`: it registers `SubjectPattern -> target` and resolves a published `Subject` to the set of matching targets on the publish hot path. It is the routing core the rest of V2-M2 consumes.

## Arena design (no pointer chasing)

Nodes live in a single contiguous `Vec<Node>` indexed by `u32` (`NodeId`), not a tree of boxed/`HashMap` nodes. Each node holds:

- a **sorted** `literals: Vec<(u64, NodeId)>` of literal children keyed by `xxh3_64(token)`, looked up by **binary search** — cache-friendly integer compares (the same `xxh3_64` the key-shared router already uses);
- `star_child: Option<NodeId>` for the `*` single-token branch;
- `terminal` / `gt_terminal` index ranges into a **flat target table**, naming the patterns that end exactly here (no `>`) or with a trailing `>`. A target shared by many patterns is interned once.

Build is `O(total tokens across all patterns)`.

## Wait-free snapshot + generation

`SublistSnapshot<T>` wraps the immutable built trie in an `arc_swap::ArcSwap`, so `match()` is a **wait-free `Arc` load + arena walk with zero locks**. A bind change builds a fresh trie (the rare, amortized `O(patterns)` rebuild) and **atomically swaps the pointer**; in-flight readers keep their loaded snapshot until they drop it. The snapshot carries a **monotonic `generation`** counter so the per-connection resolve cache (#M2-I8) can detect staleness with one integer compare — that cache is *not* built here.

## Bounded fork (provable, not exponential)

At each subject token a live node forks on (its literal child for the token hash) and (its `*` child). The frontier is **deduplicated and capped at `MAX_FORK_FRONTIER` (1024)**, so `match()` is provably `O(depth × MAX_FORK_FRONTIER × log F)` (`F` = a node's literal fan-out), with `depth` bounded by #567's `MAX_SUBJECT_DEPTH`. No exponential `>`/`*` recursion.

## How this beats the NATS `Sublist`

NATS resolves a publish against its subscription set with a `Sublist` that (1) takes a read-lock on **every publish**, (2) keeps a results cache it **flushes entirely on every subscription change**, and on a *wildcard* sub/unsub **linear-scans the whole cache** (bounded by `slCacheMax = 1024`) under the **write lock**, and (3) expands a `>`/`*` ("pwc") match with a recursion whose worst case is **exponential** in the number of wildcard levels.

IronBus replaces all three:

- **Wait-free reads** — an `Arc` load + integer-compare walk, never blocked by a concurrent rebuild;
- **No global cache flush on a bind change** and **no `O(1024)` wildcard-unsub scan** — a bind change just builds a new immutable trie and swaps the pointer; the snapshot `generation` guards a *separate* per-connection cache (#M2-I8) instead of a shared global one;
- **A bounded fork** instead of exponential pwc recursion.

## Reuses #567 (no re-parsing)

The trie is built directly on the merged subject grammar: it consumes `Subject::tokens` / `SubjectPattern::tokens`, reuses `MAX_SUBJECT_DEPTH`, and validates every registered pattern through `SubjectPattern::parse` (fail-closed). It never re-implements the grammar.

## Tests

- **Differential proptest** (`tests/sublist_routing.rs`): the trie's `match(subject)` set **EQUALS** the naive set of `{target | SubjectPattern::matches(subject)}` over randomized pattern sets and subjects — proving the index agrees with #567's authoritative matcher. A second property proves the same after a rebuild+swap (the live snapshot matches the new table and stamps the post-swap generation).
- Wait-free read after a rebuild/swap (a pinned guard keeps the old table; a fresh load sees the new one); a multi-threaded read-during-rebuild stress test.
- The bounded-fork guarantee under heavy wildcard overlap (the arena stays tiny; the result set is exact).
- Literal + `*` + `>` cases including overlapping patterns, plus a golden table.

## Bench (criterion, `--bench sublist`)

| op | 16 patterns | 256 | 4096 |
|---|---|---|---|
| `match()` literal hit | ~97 ns | ~98 ns | ~108 ns |
| `match()` `*` hit | ~100 ns | ~101 ns | ~110 ns |
| `match()` `>` catch-all | ~47 ns | ~53 ns | ~66 ns |
| `match()` via `ArcSwap` snapshot | ~104 ns | ~105 ns | ~133 ns |
| bind change (rebuild + swap) | ~6.0 µs | ~93 µs | ~4.0 ms |

`match()` is **essentially flat as the table grows 16 → 4096 patterns** — it is `O(tokens)`, not `O(patterns)` — at ~9-10 M matches/s, and the wait-free `ArcSwap` indirection adds only a few ns. The rebuild is the rare, amortized cost; crucially it is the *whole* cost — there is no global results-cache flush and no `O(1024)` wildcard-unsub scan layered on top of it the way NATS pays.

## Placement & deps

Everything (pure trie build/match/generation **and** the `ArcSwap` snapshot wrapper) lives in `ironbus-core`, beside `subject.rs` — the scope's preferred placement, since the snapshot is pure compute. **New core dep: `arc-swap = "1.7"`** (resolved to 1.9.1), pure Rust, `MIT OR Apache-2.0` (already allow-listed), MSRV well below the 1.78 floor, not an async runtime.

## IO-free

`ironbus-core` stays IO-free: the module uses only `std::sync::{Arc, atomic}`, `std::mem::swap`, and `arc-swap`. Both halves of the gate pass — the `tools/io-free-check` AST walk (forbids `std::{io,fs,net,os,process}` + async runtimes, none present) and the `cargo tree -e normal,build` async-runtime check (`arc-swap` is not one). No wire/storage/protocol change.

## Scope boundary

Build + match + snapshot/generation **only**. Out of scope (this PR provides the clean structure they consume): the per-connection resolve cache (#M2-I8), the subject→stream binding policy + single-home/ambiguous rejects (#585), the wire frames (#588). The `target` is left a generic `T` so #585 can use any id/newtype.

## Gates (all green)

- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (`RUSTFLAGS=-D warnings`) — clean, run from a **fresh `cargo clean`** state (caught and fixed `cast_possible_truncation`, `needless_lifetimes`, and `missing_panics_doc` that incremental runs can miss)
- `cargo test --workspace --all-features --locked` (`RUSTFLAGS=-D warnings`) — all pass (17 unit + 4 differential/integration sublist tests, 0 failures)
- `cargo run -p io-free-check -- crates/ironbus-core/src` — IO-free (20 source files)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
 
 [[package]]
+name = "arc-swap"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
+dependencies = [
+ "rustversion",
+]
+
+[[package]]
 name = "arrayref"
 version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -787,6 +796,7 @@ dependencies = [
 name = "ironbus-core"
 version = "0.0.0"
 dependencies = [
+ "arc-swap",
  "blake3",
  "crc32c",
  "criterion",
--- a/crates/ironbus-core/Cargo.toml
+++ b/crates/ironbus-core/Cargo.toml
@@ -21,6 +21,16 @@ default = []
 zstd = ["dep:zstd", "dep:blake3"]
 
 [dependencies]
+# Wait-free atomic `Arc` swap for the subject-routing trie snapshot (#568). `SublistSnapshot`
+# wraps the immutable built trie in an `ArcSwap` so a publish-side `match()` is a wait-free
+# `Arc` load + walk (no lock) and a bind change atomically swaps in a freshly-built trie. It is
+# pure Rust (no `links`, no C build script, no vendored crypto), `MIT OR Apache-2.0` (both already
+# on the deny.toml allow-list), and is NOT an async runtime, so it passes BOTH halves of the
+# ironbus-core IO-free gate (the `tools/io-free-check` AST walk forbids `std::{io,fs,net,os,process}`
+# and tokio/mio/async-std/smol — `arc-swap` and the `std::sync::{Arc,atomic}` it builds on are
+# none of those; the `cargo tree -e normal,build` half forbids only those async runtimes). Its MSRV
+# is well below the workspace 1.78 floor and its only optional dep (`serde`) is left disabled.
+arc-swap = { version = "1.7", default-features = false }
 crc32c = "0.6"
 xxhash-rust = { version = "0.8", default-features = false, features = ["xxh3"] }
 # Pure-Rust LZ4, the default compression codec per ADR-0003 (#12, #139). The `safe-encode` +
@@ -80,4 +90,11 @@ harness = false
 # equivalent Arc<[u8]> freeze + refcounted slice handoff IronBus actually uses, vs a copy.
 [[bench]]
 name = "zero_copy_handoff"
+harness = false
+
+# Subject-routing trie (#568): match() throughput + scaling with pattern count, the wait-free
+# snapshot read, and the bind-change (rebuild + atomic swap) cost. Backs the "beats NATS Sublist"
+# claim. Run on demand via `cargo bench -p ironbus-core --bench sublist`.
+[[bench]]
+name = "sublist"
 harness = false
--- a/crates/ironbus-core/benches/sublist.rs
+++ b/crates/ironbus-core/benches/sublist.rs
@@ -1,0 +1,130 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//! Informational micro-benchmarks for the subject-routing trie ("Sublist", #568).
+//!
+//! These back the "beats the NATS `Sublist`" claim with numbers on the two operations
+//! that matter:
+//!
+//! * `match()` throughput on the publish hot path — a wait-free `Arc` load + arena walk —
+//!   measured against routing tables of growing size (the scaling-with-pattern-count
+//!   curve), for a literal hit, a `*` hit, and a `>` catch-all;
+//! * the bind-change cost — the `O(patterns)` rebuild + the wait-free atomic swap — which
+//!   IronBus pays on a subscription change INSTEAD of NATS's global results-cache flush
+//!   and its `O(slCacheMax = 1024)` wildcard-unsub linear scan under the write lock.
+//!
+//! Run on demand (`cargo bench -p ironbus-core --bench sublist`); not wired into per-PR
+//! CI. Inputs are fixed so a run is deterministic, and `black_box` hides them from the
+//! optimizer.
+
+use criterion::{black_box, criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use ironbus_core::subject::Subject;
+use ironbus_core::sublist::{Sublist, SublistBuilder, SublistSnapshot};
+
+/// The routing-table sizes the scaling curves sweep.
+const SIZES: &[usize] = &[16, 256, 4096];
+
+/// Builds a routing table of `n` patterns with a realistic shape: a literal-prefixed,
+/// moderately-deep namespace with a `*` fork and a `>` tail, plus one global catch-all.
+/// `t<i>.svc.<j>.metric` style so the literal fan-out is wide (many distinct first tokens)
+/// and there is genuine wildcard overlap.
+fn table(n: usize) -> Sublist<u32> {
+    let mut b = SublistBuilder::new();
+    for i in 0..n {
+        let target = u32::try_from(i).expect("bench table size within u32");
+        match i % 4 {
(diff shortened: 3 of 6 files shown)


ELares:
**Review: changes required before merge** (formal request-changes not possible on a self-authored PR; recording here).

Strong work — the arena trie is genuinely O(tokens) not O(patterns) (bench confirms), wait-free snapshot reads, no global cache flush on bind: it beats the NATS Sublist as designed. arc-swap in IO-free core is acceptable (passes io-free-check).

**BLOCKING:** the fork bound silently drops matches. `MAX_FORK_FRONTIER = 1024` + `push_unique_capped` silently dropping overflow ⇒ a pathological pattern set could make a published subject NOT route to a subscriber it should match — silent message loss, which contradicts IronBus's fail-closed/never-silently-drop tenet (the exact property we cite against NATS). 

**Fix (fail-closed, never silent):** prefer bounding at INSERT/bind time (reject, with a typed error, a pattern/pattern-set whose simultaneously-live wildcard frontier could exceed the cap) so `match()` is provably COMPLETE; or if a match-time cap is kept, hitting it must surface a typed error / reported metric+log, never a silent drop. Add a test that drives the high-fork case and asserts completeness OR the reported error. Approve on this fix.

ELares:
## Fixed: fail-closed fork bound — `match()` can no longer silently drop a match (#568)

### The defect
`MAX_FORK_FRONTIER = 1024` was enforced **only** inside `push_unique_capped`, which **silently dropped** any frontier overflow (`if frontier.len() >= MAX_FORK_FRONTIER { return; }`). If a pathological pattern set's simultaneously-live wildcard frontier ever exceeded the cap, `match()` would silently **omit a matching target** — a published subject would not route to a subscriber it should match. That is silent message loss, violating IronBus's never-silently-drop tenet (**I3**) — the exact property the project cites against NATS.

### The fix (PREFERRED approach: reject at build/ingest, not truncate at match)
Following #567's fail-closed-at-ingest posture, the bound is now enforced **when the pattern set is frozen**, so the built trie can *never* need to drop at match time and `match()` is **provably complete**.

- **New typed error** `SublistError::ForkLimitExceeded { worst_case, limit }` (plus `InvalidPattern(SubjectError)`). `SublistBuilder::build` is now fallible: `Result<Sublist<T>, SublistError>`. Individual `insert`s still succeed (each pattern is valid #567 grammar); the bound is a property of the **set** and is checked at `build`.
- **Exact worst-case frontier** `worst_case_frontier()`: an `O(nodes)` bottom-up DP that respects the single-shared-token-per-level rule. Distinct literal siblings are mutually exclusive (one shared token can hit only one), so it takes **max, not sum**, over them; the only way two differently-keyed nodes become co-live is a `*`/literal fork at one node, so it sums the `*`-subtree peak and the best literal-subtree peak there. This is a **sound over-approximation** (it assumes the two branches peak at the same depth — can only make the bound larger, never smaller), so `wcf(root) >= max over subjects of |frontier|`. A build that passes guarantees the run-time frontier never reaches the cap. Because it uses max-not-sum over disjoint prefixes, it does **not** over-reject legitimate wide routing tables (a 5000-disjoint-prefix table builds fine).
- `push_unique_capped` keeps a `debug_assert!` backstop: if a future change ever lets the frontier reach the cap, a debug/test build fails **loudly** instead of silently dropping. So the match path is provably complete *or* loudly signals — never silently incomplete.

### Tests proving no silent match-drop
- `build_rejects_pattern_set_exceeding_fork_bound` — the `{a,*}^11` set (2048 patterns) is **rejected** with `ForkLimitExceeded { worst_case: 2048, limit: 1024 }`.
- `build_accepts_set_at_fork_bound_and_matches_completely` — `{a,*}^10` (exactly 1024, *at* the cap) is **accepted** and `match("a.a.…a")` returns **all 1024** targets — proving completeness at the boundary.
- `worst_case_frontier_does_not_over_reject_wide_disjoint_prefixes` — 5000 disjoint-prefix patterns build fine and match completely (guards the tightness of the bound).
- The existing **differential proptest** (`trie_match_equals_naive_match`: trie set == naive `SubjectPattern::matches` set) stays **GREEN**; accepted sets match completely.

### Preserved
Arena trie, wait-free `ArcSwap` snapshot, generation stamp, bench, public API shape (only an additive typed error + fallible `build`), and the `ironbus-core` IO-free invariant.

### Gates (all green, clippy run from a CLEAN state)
- `cargo fmt --all --check` ✅
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` ✅ (clean `cargo clean` first)
- `cargo test --workspace --all-features --locked` ✅ (20 sublist unit tests + 4 differential/integration tests, no failures)
- `cargo run -p io-free-check -- crates/ironbus-core/src` ✅ (20 source files structurally IO-free)

PR #653, opened 2026-06-19
feat(cluster): embedded metadata Raft group via vendored-codec raft-rs (#578)

### LINKED ISSUE #578: [C1-I1] Embedded metadata Raft group (in-process, no external dep)
Scope: a small Raft group (1/3/5 voters) holding cluster membership and config, replicated via the metadata log. Rationale: one consensus group with O(quorum) heartbeat, not N+M+1 groups. Beat-NATS: consensus is paid once for the cluster, not per replicated asset. Fixes: the group-explosion / ~2000-HA-asset-per-server ceiling (nats-server discussion #5128).

ELares:
Closes #578 (V2-C1, [C1-I1] Embedded metadata Raft group).

This is the FOUNDATION of V2-C1 clustering and the project's first consensus dependency. It wires the production **tikv/raft-rs** (`raft` 0.7.0) core into `ironbus-server` ONLY (never `ironbus-core`), and proves it builds clean under every IronBus gate with a constructible, in-memory metadata group. No consensus is hand-rolled.

## How the codec was vendored (no protoc, runtime-only)

The one integration risk the C1 decision flagged: raft-rs's `protobuf-codec` runs a **build-time** protobuf codegen step (`raft-proto`'s `build.rs` → `protobuf-build` → `protoc`/`prost-build`), which IronBus has nowhere (`ironbus-proto` is hand-rolled, zero-dep, no build script). That touches the reproducible static-musl build, the MSRV-1.78 floor, and the zero-build-script posture.

Resolution (the ratified **vendor the generated codec** option):

- `raft-proto` is redirected via `[patch.crates-io]` (workspace root) to a **build-script-free drop-in** at `crates/ironbus-server/vendor/raft-proto`. It is a byte-for-byte equivalent of raft-proto 0.7.0's public surface (`eraftpb`, `confchange`, `confstate`, `prelude`, `util`) **minus** the `build.rs` and the `protobuf-build` build-dep.
- The generated `eraftpb` module is committed as source (`vendor/raft-proto/src/protos/eraftpb.rs`). It was generated **once, out of band, by the pure-Rust `protobuf-codegen-pure` 2.28 compiler — NO `protoc`, NO C toolchain** — from raft-proto's own `eraftpb.proto` (with the `rustproto.proto` include, so `bytes`-fields are `bytes::Bytes`, matching raft 0.7's `protobuf-codec`). The regeneration recipe is in the vendored `lib.rs`.
- `raft` is added with `default-features = false, features = ["protobuf-codec"]` (drops `default-logger`/failpoints), pinned `=0.7.0`. The only new protobuf dep is the **pure-Rust `protobuf` 2 RUNTIME** crate (`=2.28.0`).

**Confirmed NO protoc and NO build script run at build:** the musl release compile of `ironbus-cli` emits no `protoc`/`protobuf-build`/codegen line; the only build-dependency anywhere in the new tree is `slog → rustversion` (a pure-Rust proc-macro). `cargo tree` shows `protobuf-build` is gone entirely.

## Deps added + deny status

New crates (all permissive, all in `ironbus-server`/its tree): `raft` (Apache-2.0), vendored `raft-proto` (Apache-2.0), `protobuf` 2.28 (MIT), `slog` (MIT/Apache arm), `fxhash`, `getset`, `erased-serde` (MIT/Apache). `rand`/`getrandom`/`bytes`/`thiserror`/`byteorder`/`serde`/`zerocopy`/`cfg-if`/`ppv-lite86` were already in the tree.

`cargo deny check`: **licenses ok, bans ok, sources ok, advisories ok.** No new license entry was needed (every arm is already on the allow-list). Two transitive RUSTSEC advisories on the raft tree are **scoped + justified** in `deny.toml` (`[advisories].ignore`, same convention as the existing lz4_flex entry):
- **RUSTSEC-2024-0437** (protobuf 2.x recursion on untrusted input): NOT reachable in C1-I1 — there is no peer transport yet (C1-I3), so the codec only decodes locally-proposed, locally-committed entries; raft-rs is not yet wired into the running broker. Flagged to revisit before C1-I3 exposes peer bytes (bound received messages there). The fix (protobuf ≥ 3.7.2) is unavailable on the `protobuf-codec` path under MSRV 1.78.
- **RUSTSEC-2025-0057** (fxhash unmaintained): raft-rs's internal hashmap hasher only; not a parser/crypto/security boundary; no open vulnerability.

## MSRV 1.78 + reproducible-musl + io-free-check

- **MSRV 1.78:** `cargo build --workspace --all-features --locked` on stable **1.78.0 passes**. Vendoring eliminated the `protobuf-build`/`prost`/edition2024 cascade entirely, so no `=`-pin of build tooling was needed beyond `raft = "=0.7.0"` and `protobuf = "=2.28.0"`.
- **Reproducible static-musl build:** every crate (`protobuf`, vendored `raft-proto`, `raft`, `ironbus-server`, `ironbus-cli`) **compiles to the release `x86_64-unknown-linux-musl` target with no protoc and no build-script codegen** (verified locally; the final `cc` link step is a macOS-host linker limitation, not a code issue — CI's `cross`/`repro` jobs have the musl linker). `CARGO_INCREMENTAL=0` honored. cffi-ban passes on the musl graph; the default shipped graph stays **zstd-free**.
- **io-free-check (core clean):** `cargo run -p io-free-check -- crates/ironbus-core/src` passes, and the `cargo tree -p ironbus-core` async-runtime guard passes — raft is **absent** from the `ironbus-core` tree and present only via `ironbus-server`. raft-rs pulls no async runtime (no tokio/mio/async-std/smol).

## Single-node binary unaffected

The metadata group is a new, **not-yet-wired** module (`pub mod cluster;` is the only reference; nothing in serve/engine/actor/cli calls it). The n=1 zero-config path's behavior and on-disk layout are byte-for-byte unchanged.

## The metadata group / state-machine

- `cluster::metadata_group::MetadataRaftGroup` — a `RawNode<MemStorage>` for a 1/3/5-voter group, driven by the synchronous `tick`/`step`/`propose`/`ready`(persist→advance→apply) loop. Exposes term (the future leader epoch), leadership, and the applied state.
- `cluster::state_machine::MetadataStateMachine` — the deterministic applied control plane: **cluster membership** (voter/learner), **partition placement**, **per-partition leader epoch**, **config**. Plain synchronous data + a small, zero-dependency, hand-rolled little-endian command codec (no `raft` types, trivially unit-testable).

## Round-trip test

`single_node_group_propose_tick_ready_roundtrips_on_memstorage`: the group constructs, self-elects via campaign/ready, then a proposed command round-trips `propose → tick/ready → apply` into the state machine on `MemStorage`; a second command is applied in order. Plus 1/3/5-voter construction, tick-driven election, and state-machine + codec round-trip/negative tests (10 cluster tests, all green).

## Scope boundary (deferred)

- **C1-I2** — replace `MemStorage` with a durable `Storage` over the `ironbus-storage` CRC-framed log.
- **C1-I3** — expose the Raft term as the per-partition leader epoch + pair with the `ironbus-core` lease/clock fencing; peer transport (and the message-bounding that closes RUSTSEC-2024-0437 exposure).
- **C1-I4** — joint-consensus membership, learners, peer-id validation in the step/transport seam.

## Gate results (all green locally)

`cargo fmt --all --check` ✓ · FRESH `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (from `cargo clean`) ✓ · `cargo test --workspace --all-features --locked` ✓ · `io-free-check` (core IO-free) + async-runtime guard ✓ · `cargo deny check` (licenses/bans/advisories/sources) ✓ · `cffi-ban.sh` (host + musl) ✓ · SPDX headers ✓ · MSRV 1.78 `--all-features --locked` build ✓ · musl release **compile** (no protoc) ✓ · zstd-free default graph ✓.

## What to scrutinize

- The `[patch.crates-io] raft-proto` redirect + the regeneration recipe (re-run only on a `raft`/`raft-proto` bump; the vendored `eraftpb.rs` must then be regenerated with `protobuf-codegen-pure`, never `protoc`).
- The two scoped advisory ignores in `deny.toml` — especially that **RUSTSEC-2024-0437 MUST be revisited in C1-I3** when peer messages become attacker-reachable.
- The persist-before-advance ordering in `drive_ready` (becomes the group-commit fsync barrier under the C1-I2 durable storage).


ELares:
Fixed the failing `cargo-auditable SBOM` check (`scripts/ci/sbom-diff.sh`) in 14300d9 — without weakening the supply-chain control.

## Root cause

raft-rs (#578) legitimately links two crates into the `ironbus` binary that the old direction-one query — `cargo tree -p ironbus-cli -e normal,build --target <triple>` — silently dropped, so the SBOM round-trip flagged them as "+" drift even though the binary was faithful to `Cargo.lock`:

- **`zerocopy-derive` (proc-macro)** via `raft 0.7 → rand 0.8 → rand_chacha → ppv-lite86 → zerocopy (simd) → zerocopy-derive`. `zerocopy` declares an **unconditional** `zerocopy-derive` dependency gated on `target = "cfg(any())"`. `cargo tree --target <triple>` evaluates `cfg(any())` to **false** and drops the edge (it even prints the "try `--target all`" hint), but the resolver — and therefore cargo-auditable — keeps it. raft uses `rand` for election-timeout jitter, so it is **not** feature-gateable off; the crate is genuinely shipped.
- **`anyhow`** via `raft → slog`, whose `std` feature weak-references `anyhow?/std`. The resolver records the `slog → anyhow` (kind=normal) edge, but `cargo tree`'s text output omits this weak-optional edge — so `anyhow` is compiled in yet invisible to the old query.

Both crates are already license-approved in `deny.toml` (`cargo deny check` stays green: advisories/bans/licenses/sources ok).

## The fix (keeps the control meaningful)

Direction one now derives its comparison set from the **resolver graph** — `cargo metadata --filter-platform <target>` `.resolve.nodes`, walked from the `ironbus-cli` root over `normal`+`build` `dep_kinds` only (dev dropped) — instead of from `cargo tree`'s text output. That is the **same build-unit graph cargo-auditable embeds**, so the weak-optional (`slog→anyhow`) and `cfg(any())` proc-macro (`zerocopy→zerocopy-derive`) edges are present, and the round-trip holds **1:1** for the real shipped set.

`--target all` was rejected: it would still miss the weak-optional `anyhow` **and** union in other-target crates (`wasi`, `windows-*`) the binary never links — producing false drift. `--filter-platform` only resolves (no compile), so it runs in CI without the musl linker. **No allowlist** of crate names was added (that would rot and hide future drift).

The control keeps full teeth — verified locally:
- a crate in the binary but **not** in the resolved graph → still **FAILS**;
- a crate resolved-but-stripped from the binary → still **FAILS**;
- dev-deps (e.g. `criterion`, a raft dev-dep present in the full lock) → still correctly **excluded** from the 59-crate shipped set.

## Local verification

Reproduced the exact mismatch class on the host target (`aarch64-apple-darwin`) — `zerocopy-derive`/`anyhow` appear via the same `raft→rand→zerocopy-derive` / `raft→slog→anyhow` chains regardless of target. After the fix:

- `sh scripts/ci/sbom-diff.sh <bin> <target>` → **exit 0**, 59 crates match both the shipped graph and `Cargo.lock`;
- `cargo fmt --all --check` → ok;
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (fresh) → ok;
- `cargo test --workspace --all-features --locked` → ok;
- `cargo deny check` → advisories/bans/licenses/sources ok;
- `io-free-check` → ok.

The musl SBOM build needs a musl linker that is not on this host, so **CI (`cargo auditable build … --target x86_64-unknown-linux-musl`) is the final SBOM verifier**. The `--filter-platform=x86_64-unknown-linux-musl` resolve was confirmed locally to yield the identical 59-crate set including both crates, so the fix is target-consistent.

Issue #635, opened 2026-06-19
[M7-I4] Secret handling (file/env ref, strict modes) + security audit-event stream


ELares:
Add secret handling via file/env references with strict file-mode checks, plus a structured security audit-event stream (auth success/failure, scope denials, config changes). First principles: secrets never sit in argv/config plaintext, and security-relevant actions are auditable.

ELares:
Closed by #748 (M7 audit stream + secret refs; graceful drain + SIGTERM readiness-flip + config validate).

Issue #631, opened 2026-06-19
[M7-I2] Three-mechanism auth x three scopes + ironbus server passwd


ELares:
Add three auth mechanisms (bearer / user-pass / mTLS) crossed with three scopes (publish / subscribe / admin, with no implication between scopes), plus an ironbus server passwd verb. Deliberately simpler-but-safer than NATS's JWT/nkey/operator weight while covering the same access surface.

ELares:
Closed by #746 (M7: three-mechanism auth × three scopes + passwd + pre-auth DoS defenses).

PR #708, opened 2026-06-20
feat(security): TLS 1.3 fail-closed bind + three-mechanism auth/authz (#629)

### LINKED ISSUE #629: [M7-I1] TLS 1.3 + fail-closed bind invariant
Require TLS 1.3 and enforce a fail-closed bind invariant: a non-loopback listener requires TLS + identity, validated before listen() is ever called. First principles: an unencrypted/unauthenticated non-local bind must be impossible by construction, not a misconfiguration footgun.

ELares:
Closes #629 (V2-M7: TLS 1.3 + fail-closed bind invariant). Bundles #631 (three-mechanism auth x three-scope authz) and #635 (secret-file StrictModes + redacting newtype). The rustls TLS 1.3 handshake itself and the pre-auth DoS *enforcement* are FLAGGED as the immediate next PRs (rationale below).

### The fail-closed bind invariant (the must-have, load-bearing)
A **non-loopback `--addr` is REFUSED at startup, before any listener opens**, with a typed usage error and **NO override**, unless **BOTH** TLS material **AND** an auth identity are configured. Validated on the *resolved* address, **pre-listen**, mirroring the existing `--health-allow-public` guard shape. Loopback stays zero-config (the dev experience is byte-for-byte unchanged). A public plaintext/anonymous wire bind is **impossible by construction** — there is no `--insecure`/`--allow-plaintext` escape hatch.

### Three-mechanism auth × three-scope authz (#631)
Auth is established **once** in the `Connect` handshake (an **additive, backward-compatible** trailing section of the `Connect` body — an empty body still works, so loopback dev is unchanged) and the resolved identity's scope set is **pinned for the connection lifetime**.
- **Mechanisms:** bearer (SHA-256 storage + `subtle` constant-time compare), username/password (Argon2id, OWASP edge profile), mTLS SAN-identity mapping (URI-then-DNS, **CN excluded**, **no-match REJECTED** — a verifying cert gets **no default scope**).
- **Three INDEPENDENT scopes** publish/subscribe/admin with **NO implication**: `admin` does **NOT** grant `publish` or `subscribe`. A produce without `publish`, a subscribe without `subscribe`, an admin verb without `admin` — each refused with the single **uniform "Authorization Violation"** (no bad-credential-vs-insufficient-scope oracle, no username enumeration).

### Secret handling + StrictModes (#635)
Secrets sourced by **file/env reference, never inline**; a fail-closed **StrictModes** check refuses a group/world-readable (`mode & 0o077`) or wrong-owner secret-bearing file **at boot, before any listener**, never reading the contents; a **redacting `Secret` newtype** (`Debug`/`Display` = `<redacted>`, single accessor, zeroize on drop).

## Why TLS is flagged (the crypto-provider decision)
`deny.toml` `[bans]` denies **`ring` AND `aws-lc-sys`** — rustls's only *stable* providers. The lone pure-Rust rustls provider is an **`0.0.2-alpha`** crate that pulls a C builder (`cc`) + ~100 packages. Shipping an alpha crate as the crypto backbone of a **security-critical** TLS path would be a weak/uncertain path, and the provider choice is a **maintainer-owned supply-chain decision (#107)**. So the encryption layer is the next PR. The **auth primitives** (`sha2`/`subtle`/`argon2`/`zeroize`) are audited, stable, pure-Rust, all on the deny allow-list — **shipped here**. The fail-closed bind invariant already ENFORCES that a public bind cannot be plaintext+unauthenticated, so this is the **stronger posture, not a gap**: TLS + auth + the invariant compose, and the public bind is refused without the material regardless.

## Bundled vs flagged
| Area | Status |
| --- | --- |
| Fail-closed bind invariant (non-loopback ⇒ TLS material + auth identity, pre-listen, no override) | ✅ shipped |
| 3-mechanism auth × 3-scope authz (bound at Connect, no implication, uniform error) | ✅ shipped |
| Secret-file StrictModes + redacting newtype + file/env-ref secrets | ✅ shipped |
| rustls TLS 1.3 handshake (provider) | ⚑ flagged (next PR, #107) |
| Pre-auth DoS **enforcement** (per-IP rate + half-open cap + `connections_rejected_total{reason}`) | ⚑ flagged — knobs parsed/validated/surfaced with safe defaults; accept-loop throttling is the next PR |
| At-rest encryption (#108); audit-event export stream (#109/#635) | ⚑ flagged (later #M7) |

JWT/nkey (NATS's heavy operator hierarchy) is explicitly **NOT** done — the simpler-but-safer model (three mechanisms × three scopes, additive credential sets, no expiry timer) is the point, and it beats NATS on operability.

## Loopback dev unchanged
A no-auth broker (no `--auth-config`, loopback bind) requires no credential and gates no verb — the scope gate is bypassed and the plaintext loopback broker behaves byte-for-byte as before. `ironbus-core` stays IO-free; MSRV 1.78 holds.

## New crypto deps (deny status)
`sha2`, `subtle`, `argon2` (+ `blake2`, `password-hash`), `zeroize` — all pure-Rust, MIT/Apache-2.0, already on the deny allow-list, **no new advisory ignores**. None is `ring`/`aws-lc-sys`. `cffi-ban.sh` clean (no `links` key, no C-builder build-dep).

## Tests added
Non-loopback bind without TLS+auth refused pre-listen (typed error); loopback needs neither; a valid bearer/password authenticates + pins scopes; produce-without-publish / subscribe-without-subscribe / admin-without-admin each refused; **admin does NOT imply publish**; a group/world-readable secret file refused at boot; the Connect-auth wire codec is backward-compatible (empty/old body ⇒ no auth) and fail-closed (unknown mechanism / truncated section ⇒ typed error, never anonymous).

## Gate results
- `cargo fmt --all --check` ✅
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` ✅ (**FRESH from `cargo clean`**)
- `cargo test --workspace --all-features --locked` ✅ (one pre-existing, host-filesystem-dependent storage test `preallocate_reserves_real_blocks_on_a_supporting_fs` fails identically on `origin/main` — unrelated)
- `io-free-check` ✅ (`ironbus-core` untouched)
- `cargo deny check` ✅ (advisories/bans/licenses/sources ok)
- `cffi-ban.sh` ✅

## What to scrutinize most
1. **The fail-closed-bind invariant** — `wire_bind_decision` in `crates/ironbus-cli/src/main.rs`: a non-loopback bind without TLS+auth must refuse *pre-listen* with no override.
2. **No implication between scopes** — `ScopeSet::has` / `require_scope` in `crates/ironbus-server/src/{auth,session}.rs`: `admin` must never answer `true` for `publish`/`subscribe`.
3. **No silent plaintext / no anonymous fallback** — the Connect-auth path (`handle_connect`) and the wire codec (`parse_connect_auth`) must fail closed on a missing/malformed/unknown credential, never serve anonymously on an auth-required broker.

https://claude.ai/code/session_01QVWp2H2txnKGru4q2twxM3


ELares:
**Review: one merge-blocking security-posture issue; everything else is solid.**

The fail-closed bind *precondition*, the 3-mechanism auth × 3-scope authz (no implication), and the secret-file StrictModes are all correct and the right design — approve those.

**BLOCKING — misleading TLS posture on a non-loopback bind.** `wire_bind_decision` requires `has_tls_material()` (i.e. `--tls-cert`+`--tls-key`) for a non-loopback bind, but the **TLS 1.3 handshake is deferred** (no allowed crypto provider). At serve time the `requires_security` block only *logs* that TLS is required ("handshake flagged") and then serves with the **normal plaintext listener**. Net effect: a non-loopback bind with `--tls-cert` configured + auth serves **auth-over-PLAINTEXT** — credentials (bearer/password) and all data are MITM/sniffable, while the broker *required* a TLS cert (implying encryption). That's a config that implies encryption it doesn't deliver — a misleading security claim, worse than having no TLS surface.

**Fix (make the posture honest — no config may imply encryption that isn't delivered).** Preferred: until the TLS handshake actually lands, a non-loopback bind is **refused** with a clear message (TLS handshake not yet implemented — bind loopback and terminate TLS upstream, e.g. a mesh/proxy/VPN, or wait for the TLS PR), and `--tls-cert`/`--tls-key` are **rejected as not-yet-honored** so they can't create a false impression. If you want to support the legitimate TLS-terminated-upstream pattern now, gate it behind an **explicit, loud `--insecure-plaintext-wire` opt-in** that requires auth and warns at startup that the wire is plaintext — never silently-plaintext-while-requiring-a-cert. Keep all the auth/scope/secret work as-is.

ELares:
Finished the half-done security-posture fix.

**Honest non-loopback bind posture** (native TLS 1.3 is not yet wired — #107, no allowed pure-Rust crypto provider):
- Non-loopback `--addr` with no TLS and no opt-in → **REFUSED** pre-listen (typed Usage error); no silent plaintext.
- Any `--tls-cert`/`--tls-key`/`--tls-client-ca` (flag OR `IRONBUS_TLS_*` env) → **REFUSED** as reserved/not-yet-honored, on any bind incl. loopback — a cert can **never imply undelivered encryption** (this removes the prior require-tls-cert-but-serve-plaintext defect).
- `--insecure-plaintext-wire` **+ auth** → **ALLOWED** as the explicit, loud opt-in for the TLS-terminated-upstream pattern (mesh/proxy/VPN), with a LOUD plaintext startup WARNING.
- `--insecure-plaintext-wire` **without auth** → **REFUSED** (never plaintext + anonymous off loopback).
- Loopback → unchanged byte-identical zero-config.

`WireBindDecision` restructured to an audited `Option<TransportPlaintextOptin>` (no more `requires_security`); `TransportSecurityFlags` gains `insecure_plaintext_wire` + `reserved_tls_flag()`; `has_tls_material` removed. The existing auth (bearer/password/mTLS) × 3-scope authz and secret StrictModes work is intact. The full matrix is tested. docs/TRANSPORT.md rewritten to the honest posture.

**Gates:** `cargo fmt --all --check` clean; fresh `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` (from `cargo clean`) **green**; `cargo deny check` → advisories/bans/licenses/sources **ok**; all ironbus-cli tests pass (298 unit + integration). Two local-only failures are pre-existing and unrelated to this diff (crates byte-identical to baseline): the `ironbus-storage` `preallocate_reserves_real_blocks_on_a_supporting_fs` test under-reserves on this near-full APFS volume (passes on CI Linux/macOS), and the MSRV-1.78 build trips on the pre-pinned `zeroize 1.9.0` manifest requiring edition2024 (Cargo.lock unchanged here).

ELares:
Fixed both CI fails: base64ct pinned =1.6.0 (rust-version 1.60; 1.7+ bumped to rustc 1.81 via the RustCrypto edition2024 migration, breaking MSRV 1.78) as an explicit direct dep in ironbus-server with the =-pin discipline; and the unix-only StrictModes secret-file tests (a_group_or_world_readable_secret_file_is_refused_at_boot + load_auth_config_parses_a_three_mechanism_identity_table) gated #[cfg(unix)] so the Windows build compiles (they use std::os::unix PermissionsExt/from_mode + the cfg(unix) strict_mode_check_secret_file). cargo 1.78 ignores rust-version so a local 1.78 build masked the msrv break; CI's resolver enforces it. Gates green locally (fmt, fresh clippy -D warnings, full test, cargo deny) except the known APFS preallocate local-flake in ironbus-storage; CI to confirm msrv+windows.

Issue #693, opened 2026-06-20
[M2-I11b] Wire PartitionedStream through the engine + the partition wire frame (declare P, address a partition, assign partitions to consumers)


ELares:
Follow-up to #591 (M2-I11, the PartitionedStream storage primitive). #591 landed P independent sub-logs per stream + key->partition routing + per-partition consume state at the STORAGE layer (P=1=default byte-identical), but it's not yet reachable from the engine/wire. Remaining: (1) thread PartitionedStream through the engine so a produce to a partitioned stream routes to the key's partition and a consume drains per-partition (mirror the StreamSet engine wiring #679); (2) a wire frame / a partition-count field on StreamDeclare to declare P, and a partition selector on PubTo/SubTo (next free frame tag is 38 if a new frame is needed); (3) partition->consumer ASSIGNMENT + cooperative rebalance on consumer join/leave (the parallel-consume win realized for a competing group — M5.3-class); (4) the keyless PartitionSelector owned per producer-connection (not per-stream). Depends on #591. Pairs with the cluster multi-partition replication later.

ELares:
Core partition wiring shipped in #1118: StreamDeclare partition_count (P>1 → PartitionedStream #591 of P key-routed sub-logs under pstreams/), key-routed produce (xxh3_64(key)%P), partition-addressed consume with a per-partition durable cursor + #566 retention. Static partition addressing (explicit selector). Adversarial review MERGE_SAFE. Deferred to follow-ups (#1119): cooperative rebalance / dynamic partition→consumer assignment, per-partition DLQ, per-partition subject-filter/key-shared/Tier-S/metrics. Closing the core; deferrals tracked in #1119.

Issue #618, opened 2026-06-19
[C5-I3] Leaderless-node failover (no data move)


ELares:
Scope: on node leave, reassign leaderships to in-sync followers that already hold the log. Rationale: Pulsar-style stateless failover at the partition level. Beat-NATS: fast failover with no data copy. Fixes: slow or SPOF-prone failover.

ELares:
Closed by #721 (failover mechanism: ISR-only promotion, epoch fence, no data move, CI5 invariant) + #722 (automatic, committed-safe failover: crash-detect + provably-complete-or-fail-closed promotion). On main.

Issue #617, opened 2026-06-19
[C5-I2] Cooperative rebalance on join (learner back-fill, no stop-the-world)


ELares:
Scope: a joining node back-fills as a learner via fetch and is promoted only once caught up. Rationale: serving continues uninterrupted during the move. Beat-NATS/Kafka: no stop-the-world rebalance. Fixes: rebalance availability dips.

ELares:
Closed by #724 — cooperative rebalance on join (learner back-fill, promote only once caught up, no stop-the-world). On main.

PR #722, opened 2026-06-21
feat(cluster): automatic leaderless failover — detect a dead leader and auto-promote an ISR successor (#618)

ELares:
## What this finishes

PR #721 (C5-I3) landed the failover **mechanism** (`reassign_leadership`, `plan_failovers`, `promote_follower_to_leader`, `reconcile_placement`, CI5) but left it **test-driven**. This makes failover fire **automatically** in the live cluster, proven by killing the leader in a real multi-node test.

The runtime driver (`run_driver` in `cluster/runtime.rs`) — the single thread that owns the metadata `RawNode` — now does both halves:

### F1 — peer-liveness death detector
The driver records `last_heard[peer]` whenever it steps a metadata message from that peer (the metadata-Raft heartbeats/appends **are** the heartbeat — `record_heard` + `drain_inbound_nonblocking`), timed off the I6 monotonic clock. Each cycle, **if this node is the metadata-Raft leader**, any current voter silent past the deadline (or a forced-unreachable test peer) is proposed for removal via `MembershipChange::remove_node` through the metadata Raft (`runtime.rs` F1 block). Once that conf-change commits, the crashed node lands in `departed_members` exactly like a clean leave — a crash is converted into a committed membership shrink.

### F2 — auto-fire
On a committed shrink (`departed_members` non-empty) the metadata leader auto-calls `plan_failovers` for every partition the departed node led, feeding it the **live data-plane ISR** through cross-plane providers (`FailoverInputs` = `SurvivorStateFn` + `CommittedHwFn`) the data-plane bootstrap installs via `FailoverInstaller`/`set_failover_inputs` (`main.rs` `run_dataplane_bootstrap`). Each resulting `PlacePartition` is proposed through the metadata Raft; the existing apply path → `reconcile_placement` promotes the successor in place.

## How each NON-NEGOTIABLE is guaranteed (code pointers)

- **No false / premature failover.** Default deadline `DEFAULT_LIVENESS_TIMEOUT = 3s` = **10x** the ~300ms metadata heartbeat (`heartbeat_tick=3` @ `TICK_INTERVAL=100ms`). A peer must miss ~10 consecutive beats; a single dropped beat (raft re-sends) or a brief hiccup never trips. `last_heard` is seeded to "now" at driver start (full deadline of grace). Proven by `a_healthy_cluster_never_auto_fails_over` (3s window, `suspected_dead`/`departed_members`/`failover_proposed` all stay empty, placement byte-unchanged, 3 voters retained).
- **Single proposer, linearizable.** Both F1's removal and F2's promotion are proposed **only** when `group.is_leader()`, both go through the metadata Raft log. Idempotent: `removal_proposed`/`failover_proposed` suppress re-spam; an already-departed peer / already-converged placement is skipped. No two nodes promote different successors — the Raft log linearizes it.
- **Split-brain / old-leader fence.** The promotion's epoch is `max(placement,dead)+1` (the #721/#694 KIP-101 machinery, unchanged). Proven in the kill test: the new committed epoch strictly exceeds the old (assertion d), and the #721 serve test still proves the old leader's stale-epoch query is fenced.
- **Committed-never-lost (CI5).** F2 feeds `plan_failovers` the **real** live ISR (the installed providers read the data-plane controller's `follower_high_watermark`/`quorum_commit`), so a non-ISR / incomplete survivor is never chosen. Fail-closed: no providers (no data plane) ⇒ no promotion — proven by `no_data_plane_inputs_means_no_blind_promotion_fail_closed`.
- **Single-node byte-identical / single-writer.** All new logic is metadata-plane and cluster-gated; off-cluster nothing is constructed (`serve` only starts a runtime with a `ClusterConfig`). The produce/consume hot path (`session.rs`/`engine.rs`/`actor.rs`) is untouched. The promoted leader's single-writer is the #721 in-place promotion (no data move).
- **Deterministic test, no wall-clock flake.** `LivenessConfig` makes the deadline injectable; `force_peer_unreachable` is a clock-independent detection seam. The kill test uses it (no real-time sleep); the timeout test uses a host-scaled deadline so it never races even under CI starvation.

## The tests (what each proves)

In `cluster/runtime.rs` (real multi-node, `#[cfg(unix)]`, extend the #721 harness):

- **`killing_the_leader_auto_detects_promotes_an_isr_successor_and_the_cluster_resumes`** — THE proof. 3-node cluster, committed placement, **kill the leader**, then drive ONLY the deterministic `force_peer_unreachable` seam (no manual reconcile/plan_failovers/promote). Asserts: (a) auto-detect + committed removal (`departed_members` + voter_count≤2); (b) auto-planned + committed promotion; (c) successor is a surviving ISR replica, dead leader dropped from replicas, committed prefix preserved; (d) epoch fenced strictly above the old; (e) the surviving quorum commits a NEW entry under the new leader. **Proven by killing the leader.**
- **`the_liveness_timeout_alone_auto_fails_over_a_silent_leader`** — the production silence trigger (NO force seam): a stopped node's heartbeats cease and the host-scaled deadline detects it. **Proven by killing the leader.**
- **`a_healthy_cluster_never_auto_fails_over`** — no-false-failover. **Proven by construction + a real healthy cluster.**
- **`no_data_plane_inputs_means_no_blind_promotion_fail_closed`** — F1 removes the dead leader but F2 promotes nothing without cross-plane ISR. **Proven by killing the leader (no data plane installed).**
- The #721 mechanism test is **kept** (the by-construction proof).

Rerun-stability: the kill test ran 5/5 clean; the timeout test 8/8 clean after host-scaling its deadline.

## Gates
- `cargo fmt --all --check` ✅
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` ✅
- `cargo test --workspace --all-features --locked` ✅ green **except** the one known APFS local flake `ironbus-storage … preallocate_reserves_real_blocks_on_a_supporting_fs` (near-full disk — the only-allowed failure)
- io-free-check on ironbus-core ✅ (core untouched); metrics taxonomy untouched
- SPDX `head -3` check ✅ (no new files); MSRV 1.78 ✅ (no new deps; `is_none_or` avoided as it is 1.82+)

## Scope / deferred (flagged honestly)
- **Client-side `NOT_LEADER` redirect** (a reconnecting client producing to the new leader) is a SEPARATE follow-on (the client-awareness story). This PR proves the SERVER auto-heals leadership; the data-plane followers re-point replication to the new leader.
- **Multi-partition / full ISR gossip.** The cross-plane survivor-state provider reports THIS node's own real frontier and treats remote committed replicas as ISR-complete to the committed HW (sound for the single default partition, where the surviving committed replicas that caught up ARE the ISR). Distinguishing a remote survivor that lagged out of the ISR before the crash needs ISR gossip across the data plane → #693. Flagged in `main.rs`.

https://claude.ai/code/session_01QVWp2H2txnKGru4q2twxM3

### DIFF
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -4698,9 +4698,10 @@ fn spawn_dataplane_serve(
 // a thread entry point: each input is a distinct piece the
 // bootstrap needs for the whole thread lifetime; a bundling
 // struct would only move the noise.
-#[allow(clippy::needless_pass_by_value)] // a thread entry point: it OWNS the read-plane Arc (cloned
-                                         // into the leader role) + the log config for its lifetime;
-                                         // a borrow would fight the 'static spawn bound.
+#[allow(clippy::needless_pass_by_value)]
+// a thread entry point: it OWNS the read-plane Arc (cloned
+// into the leader role) + the log config for its lifetime;
+// a borrow would fight the 'static spawn bound.
 #[allow(clippy::too_many_lines)] // one linear bootstrap sequence (await placement, build the server,
                                  // start the runtime, install the F2 cross-plane inputs, hold until
                                  // shutdown); splitting it would scatter a single startup concern.
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -4826,22 +4826,26 @@ fn run_dataplane_bootstrap(
         let _ = client_ack_slot.set(std::sync::Arc::clone(gate));
     }
 
-    // INSTALL the cross-plane F2 auto-failover inputs (#618): now that the data plane is live, the
-    // metadata-Raft leader can auto-promote an ISR successor on a leader crash. The survivor-state
-    // provider surfaces the LIVE ISR: for THIS node it reports its real data-plane frontier (its
-    // follower high-watermark, or the served frontier if it leads), for a remote survivor it reports
-    // a node complete to the committed HW IFF that survivor was a committed replica — sound because the
-    // committed HW is the quorum-fsync'd frontier every ISR member holds by definition.
+    // INSTALL the cross-plane F2 auto-failover inputs (#618 / #618b): now that the data plane is live,
+    // these surface THIS node's real data-plane frontiers to the metadata driver. They serve two roles:
     //
-    // FLAGGED (the honest cross-plane gap): a remote survivor that had LAGGED OUT of the ISR before the
-    // crash is not distinguished here without ISR gossip across the data plane; the conservative default
-    // treats a committed replica as in-ISR. THIS node's own out-of-ISR state IS detected (its real
-    // frontier is read). Full ISR gossip so any node can vet any survivor is the multi-partition / ISR-
-    // gossip follow-on (#693). For the single-partition target the surviving committed replicas that
-    // caught up ARE the ISR, so the default is correct for the shipped scope.
+    //   1. when this node LEADS the partition's data plane + the metadata group, the driver periodically
+    //      CHECKPOINTS `committed_hw` into the metadata Raft (so the committed bar SURVIVES the leader's
+    //      death — the persisted SAFE bar a successor must hold); and
+    //   2. on a leader death, the driver's provably-complete-or-fail-closed rule compares `own_frontier`
+    //      (THIS node's real durable frontier) against that persisted bar to decide whether THIS node may
+    //      safely SELF-promote — it never blind-promotes a remote node whose frontier it cannot see.
+    //
+    // CRITICAL (#618b safety): we report ONLY frontiers this node actually KNOWS — its OWN. We do NOT
+    // claim a remote committed replica is "complete" (the previous optimistic default did, which could
+    // promote a LAGGING survivor and silently lose committed data). A remote node's true frontier is
+    // unknown without cross-data-plane ISR gossip (#693), so `survivors` reports completeness ONLY for
+    // THIS node; the driver consequently self-promotes only this node and otherwise fails closed. The
+    // shipped guarantee: auto-failover never loses committed data; it self-heals whenever the surviving
+    // metadata leader is itself a complete replica, and fails closed (leaderless, recoverable) otherwise.
     {
         // This node's live committed frontier for a partition (its follower HW, or its leader
-        // quorum-commit) — the quorum-fsync'd prefix the ISR holds by definition.
+        // quorum-commit) — the quorum-fsync'd prefix it durably holds.
         let own_frontier_for = {
             let server = std::sync::Arc::clone(dp_runtime.server());
             move |partition: u64| -> u64 {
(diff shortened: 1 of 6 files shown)


ELares:
## Safety fix — auto-failover is now provably committed-safe (addresses the reviewer's committed-data-loss gap)

A reviewer caught that the auto-failover path could **silently lose a committed, client-acked record**: the cross-plane ISR provider reported remote survivors as complete to `own_hw` *blindly* (it has no view of remote data-plane frontiers), and the true committed high-watermark was **not persisted anywhere a survivor could read after the leader died**. So in a 3-node R3 (`min_isr=2`) cluster, if a record reached quorum on only leader + one follower and the leader then died, a **lagging** survivor could be auto-promoted and the committed record would be lost. CI5 was only asserted inside the kill-test with a hardcoded `successor_in_isr: true` — never enforced at runtime.

This commit closes the window. **Committed (quorum-acked) data is now never lost on the auto path.**

### What changed

1. **Persist the committed-HW bar so it survives the leader's death.** New `MetadataCommand::CheckpointCommittedHw { partition, offset }` (metadata-log codec **tag 6** — distinct from the data-plane frame-tag space), applied **monotonically** by `state_machine.rs` (a stale/lower checkpoint never lowers the bar). The data-plane leader proposes it **periodically** (a cadence, *not* per record — no metadata round-trip on the produce hot path). After a leader dies, survivors read this as the **safe bar** a successor must hold.

2. **Provably-complete-or-fail-closed promotion.** The metadata leader knows only its **own** real data-plane frontier (no ISR gossip yet). So in F2 it **self-promotes only when `own_frontier >= the persisted committed-HW checkpoint`**, re-verified through CI5 (`check_failover_preserves_committed`) **before** proposing. In *every* case it cannot prove safety — own frontier behind the bar, no persisted checkpoint, or only a remote node might be complete — it **fails closed**: proposes nothing, leaves the partition leaderless (recoverable), and logs the withholding. The optimistic "remote = complete" assumption is **removed** from `main.rs` (a remote survivor is now reported out-of-ISR, never blind-promoted).

3. **Apply-time self-verify (defense in depth).** `promote_follower_to_leader` / `reconcile_placement` now carry the committed-HW bar and **re-check it against the node's own durable log** before it becomes leader. A node behind the bar **aborts** fail-closed (`DataPlaneError::FailoverIncomplete`) and stays a follower — so even an optimistic/buggy/hand-built proposal can never create a leader missing committed data.

4. **CI5 wired into the runtime path**, not just the test. The hardcoded `successor_in_isr: true` shortcut is removed from the serve kill-test (now derived from real measured state).

### Why committed-data loss is now impossible on the auto path
- The bar is **persisted in replicated metadata** (`runtime.rs` F2: `group.state().committed_hw(partition)`), so it survives the dead leader.
- F2 proposes **only a self-promotion that clears the bar**, gated by CI5 — otherwise it withholds (fail-closed branch in `run_driver`).
- The **apply-time check** (`promote_follower_to_leader`) is the final net: a node missing the bar aborts rather than leading.

### New regression test (the bug)
`a_lagging_survivor_is_never_auto_promoted_and_no_committed_record_is_lost`: 3 nodes; the committed record reached only leader + one follower (`min_isr=2` of 3) so the third lags (own frontier 80 < bar 100); kill the leader; drive the deterministic auto path. **Asserts the lagging node is NEVER promoted** — the auto path either promotes the complete survivor (when that is the metadata leader, which self-promotes because its frontier holds the persisted bar) **or fails closed** (leaderless). In **both** outcomes no surviving leader is ever missing a pre-death quorum-acked offset. Plus `the_metadata_leader_behind_the_committed_bar_proposes_no_promotion_fail_closed` (safety-over-availability) and a dataplane apply-time self-verify unit test. All deterministic (injectable liveness clock / force-unreachable seam, **no real-time sleeps**); reran several times, stable.

### Honest caveat / shipped guarantee
**Auto-failover never loses committed data; it self-heals whenever the surviving metadata leader is itself a complete replica, and fails closed (leaderless, recoverable) otherwise.** It may *decline* to auto-promote when the only complete replica is **remote** and the metadata leader lags — picking a complete remote needs cross-data-plane frontier/ISR gossip, the follow-on (#693 ISR gossip / #617 rebalance). Default/single-stream partition is the target (multi-partition → #693); the client `NOT_LEADER` redirect remains a separate follow-on.

### Gates
`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`, `cargo test --workspace --all-features --locked` (exit 0 — the known APFS `preallocate_reserves_real_blocks_on_a_supporting_fs` flake did not even trip), io-free-check (ironbus-core), and the SPDX check (`miss=0`) all pass. No metrics touched; no new deps (MSRV 1.78 intact).

PR #755, opened 2026-06-22
perf: QoS-0 produce throughput — client coalescing + server batched submission (+80%) + t4g NATS benchmark

ELares:
## What

A single-node **t4g.small** IronBus-vs-NATS study (ingestion + consumption, quorum-0 vs NATS Core, quorum-1 vs NATS JetStream) plus the one product fix it surfaced.

### Shipped fix
- **`FireForgetProducer` — a coalescing at-most-once (QoS-0) producer** (`Client::fire_and_forget_producer`). Fire-and-forget previously did **one `write_all` syscall per message**; it now frames into a wire buffer and writes once per 32 KiB — the syscall coalescing a core pub client does, the unacked twin of the existing `produce_stream` path. **+60% on at-most-once produce (~640k → ~1.015M msgs/s)** on t4g.small, no change to the wire bytes the broker decodes. End-to-end test included (coalesced batch lands in order, no connection desync).

### Benchmark (docs/benchmarks/t4g-single-node-vs-nats-2026-06-22.md)
Auditor-honest, **matched durability**:

| Comparison (matched) | IronBus | NATS | Result |
|---|---:|---:|---|
| **Durable produce, power-loss-safe** (group-commit fdatasync vs JS `sync=always`) | ~55,400 | ~343 | **IronBus ~161×** |
| **Durable consume** (Tier-S streaming vs JS consume) | ~234,000 | ~101,000 | **IronBus ~2.3×** |
| Durable produce, page-cache (not power-loss-safe) | ~55,000 | ~88,000 | NATS ~1.6× (IB CPU-bound) |
| At-most-once (IB stores vs NATS Core drops) | ~1,011,000 | ~1,650,000 | NATS ~1.6× raw rate |

IronBus wins the comparisons that hold durability constant; NATS leads only by discarding (Core) or by not being power-loss-safe (JS default interval-sync). Disk floor: `dd oflag=dsync` = 265 fsync/s, so any per-message-fsync system (`sync=always` → 343/s) is disk-capped while IronBus's group-commit amortizes one fdatasync over hundreds of publishes.

### Follow-up identified (not in this PR)
Page-cache durable produce is CPU-bound; the lever is **batched actor submission** (one channel send + wake per pipelined window instead of per message; profile: ~10% `ProduceSubmission::wait`/`SyncWaker`). Deep core-actor change — filed as future work.

## Update: server-side Level-0 batched submission (+14%)

Added `Command::ProduceNoReplyBatch`: the session accumulates a socket-read's worth of Level-0 (no-ack) produces and hands them to the single-writer actor as ONE channel send (flushed before any Level-1 produce / non-produce job, so the total order holds). **+14% (1.015M -> 1.156M msgs/s)** on top of the client coalescing — at-most-once is now **+80% over the per-message baseline**. 904 server tests + golden-path acceptance pass; clippy clean. It STILL trails NATS Core (1.65M) because NATS Core discards every message (no subscriber, no server work) — proving by implementation that the at-most-once raw-rate gap is structural, not a missing optimization.

### DIFF
--- a/crates/ironbus-cli/src/bench_run.rs
+++ b/crates/ironbus-cli/src/bench_run.rs
@@ -571,16 +571,25 @@ fn run_publish_faf(cfg: &BenchConfig, addr: &str) -> Result<BenchReport, CliErro
     };
     let mut produced: u64 = 0;
     let started = Instant::now();
+    // COALESCING at-most-once producer (#11 fast path): each publish is framed into the producer's wire
+    // buffer and flushed to the socket with ONE `write_all` at 32 KiB boundaries, instead of one write
+    // syscall per message — the same coalescing a core pub client (e.g. NATS `nats bench pub`) performs.
+    // This is what makes the QoS-0 send rate a fair head-to-head with a coalescing core pub rather than
+    // a self-handicapped syscall-per-message loop.
+    let mut producer = pub_client.fire_and_forget_producer();
     while !should_stop(&cfg.bound, produced, started) {
-        // No reply is read (the broker sends no PubAck for a fire-and-forget produce), so this
-        // returns as soon as the frame is written; TCP backpressure is the only pacing when the
-        // broker falls behind. An IO/encode error is still fatal and mapped to the frozen codes.
-        pub_client
-            .produce_fire_and_forget(&body)
+        // No reply is read (the broker sends no PubAck for a fire-and-forget produce); TCP backpressure
+        // is the only pacing when the broker falls behind. An IO/encode error is fatal (frozen codes).
+        producer
+            .send(&body)
             .map_err(|e| classify(addr, "fire-and-forget producing to", &e))?;
         produced += 1;
         pace(cfg.target_rate_hz, produced, started);
     }
+    // Push the final partial batch before stopping the clock so every counted message is on the wire.
+    producer
+        .flush()
+        .map_err(|e| classify(addr, "fire-and-forget producing to", &e))?;
     let elapsed = started.elapsed();
     // At-most-once: no ack and no read-back, so no latency and no durable-write cost to attribute
     // (the broker may even have shed sends under its token bucket). fsync is forced not-measured.
--- a/crates/ironbus-cli/src/bench_run.rs
+++ b/crates/ironbus-cli/src/bench_run.rs
@@ -576,18 +576,18 @@ fn run_publish_faf(cfg: &BenchConfig, addr: &str) -> Result<BenchReport, CliErro
     // syscall per message — the same coalescing a core pub client (e.g. NATS `nats bench pub`) performs.
     // This is what makes the QoS-0 send rate a fair head-to-head with a coalescing core pub rather than
     // a self-handicapped syscall-per-message loop.
-    let mut producer = pub_client.fire_and_forget_producer();
+    let mut faf_producer = pub_client.fire_and_forget_producer();
     while !should_stop(&cfg.bound, produced, started) {
         // No reply is read (the broker sends no PubAck for a fire-and-forget produce); TCP backpressure
         // is the only pacing when the broker falls behind. An IO/encode error is fatal (frozen codes).
-        producer
+        faf_producer
             .send(&body)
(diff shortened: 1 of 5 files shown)


ELares:
## Review: rigorous self-review + independent fresh-eyes pass

**Verdict: correct, no blockers.** Reviewed for single-total-order, admission/shed faithfulness, memory bounding, durability (I2), and the client coalescing path. After rebasing onto current `main` the full workspace is clippy-clean and **2752 tests pass** locally; CI must be green before merge.

### Fixed on this branch (this review cycle)
- **CI clippy failure:** `clippy::similar_names` on the `producer` / `produced` bindings in `bench_run.rs` — renamed to `faf_producer`. This was the only thing blocking the gate.
- **Perf / memory:** `produce_no_reply_batch` built a second `admitted` `Vec` and moved every append into it even in the common no-cap default, where `would_shed` can never fire — re-introducing, once per batch, the very allocation the #11 fast path removes per message. Added a `cap() == 0` fast path that moves the whole batch straight into the command (provably equivalent: an uncapped gate admits every append). The capped path, its per-append fast-reject, and the shed accounting are unchanged.
- **Test:** added `fire_and_forget_producer_auto_flushes_at_the_32kib_boundary_in_order_end_to_end`. The existing end-to-end test stayed under 32 KiB and only exercised the explicit `flush()`; this drives ~80 KiB so the in-`send` auto-flush at `STREAM_FLUSH_BYTES` (the core coalescing mechanism) fires mid-loop, and asserts every record lands durably in order with no connection desync.

### Correctness notes (verified by tracing, no change needed)
- **Single total order:** every path that pushes to `level0_pending` flushes before the next order-sensitive op — before each Level-1 produce (`flush_level0` ahead of `produce_submit`), before any non-produce frame (ack / flow / sub / checkpoint, ahead of `drain_parked` + dispatch), at end-of-pass including error exit (`level0_flushed.and(drained)`), on the one-off `Pub` and default-stream routes, and at the `MAX_PARKED_PRODUCES` cap-flush. A `NOT_LEADER` L0 is dropped before it can enter the batch.
- **Admission / shed:** the batch reproduces the per-message gate exactly; L0 sheds fold into `ironbus_fire_and_forget_shed_total`, never `produce_rejected`. The actor's `Log::append` stays authoritative, so the byte cap is never actually violated (the batch samples the gate once at flush rather than per message, which is fine for a lossy QoS-0 path).
- **Durability:** I2 / group-commit untouched — each batched append joins the same `pending` batch under one covering `commit_batch`; L1/L2 parking and ack-after-fsync are byte-for-byte unchanged. Level-0 carries no reply, so nothing here weakens the at-least-once contract.

### Deferred to follow-up issues (tracked after merge, not blocking)
These are real but warrant their own design + review rather than widening a focused perf PR:
- **Bound in-flight Level-0 appends.** One `ProduceNoReplyBatch` carries up to `MAX_PARKED_PRODUCES` (1024) appends and the actor channel holds up to `DEFAULT_CHANNEL_BOUND` (1024) commands, so worst-case in-channel L0 payload memory grows ~1024x vs the old per-message path and L0 backpressure engages ~1024x later. QoS-0 is droppable (not a correctness issue), but for a high-scale bus this should be bounded by accounting in-flight *appends*, not commands.
- **Batched Level-1 (durable) actor submission** — the next throughput front already named in this PR's bench follow-ups section.
- **Lower-priority leanness:** pool the batch buffer to remove the per-pass `mem::take` realloc; `Vec::retain` on the capped path; count `appends.len()` in `gather_commands` so a batched L0 trips the gather heuristic; and additional server-side batch tests (L0/L1 interleave within one pass, a non-produce frame mid-batch, the cap-flush, and cap-shed accounting under a configured byte cap).

Reviewed with an independent fresh-eyes agent in addition to the author self-review.

ELares:
Post-merge follow-ups filed from the review (deferred TODOs / second-look areas): #773 (bound in-flight Level-0 appends — the channel memory amplification), #774 (batched Level-1 durable actor submission — the next produce-CPU front), #775 (Level-0 batch leanness: buffer pool, Vec::retain, gather count, plus server-side batch tests).

Issue #775, opened 2026-06-29
[TASK] Level-0 batch path follow-ups: leanness (buffer pool, retain, gather count) + server-side batch tests


ELares:
## Summary

Lower-priority leanness and test-coverage follow-ups for the Level-0 (QoS-0) batch path added in PR #755. None are blocking; grouped here so they are not lost. Each is small and independent.

## Leanness

1. Pool the batch buffer. `Session::flush_level0` does `std::mem::take(&mut self.level0_pending)`, which hands the `Vec` by value into the `Command` and leaves a zero-capacity `Vec` behind, so the next pass regrows from zero. The client side avoids this (`FireForgetProducer::flush` does `write_all` then `wire.clear()`, reusing capacity) because it borrows rather than ships the buffer. The server can pool: have the actor return the drained `Vec` via a small free-list channel (like the existing `reply_pool`) so the session reuses one allocation across passes.
2. `Vec::retain` on the capped path. `EngineHandle::produce_no_reply_batch` builds a second `admitted` `Vec` only when a byte cap is configured. Filtering in place with `appends.retain(|_| { if gate.would_shed() { gate.record_l0_shed(); false } else { true } })` removes that allocation and is order-preserving. (The common uncapped path already moves the batch with no copy via the `cap() == 0` fast path.)
3. Count `appends.len()` in `gather_commands`. The opt-in bounded-gather heuristic (`produces < 2`) counts a whole `ProduceNoReplyBatch` as one produce, so a coalesced batch of many Level-0 appends will not trip the gather even though it represents heavy pipelining. No extra fsync results (a batch already commits as one group), but it leaves some cross-socket-read amortization on the table when gather is enabled. Counting `appends.len()` toward the tally restores the original intent.

## Test hardening (server-side batch path)

The #755 client tests cover the explicit flush and the 32 KiB auto-flush end-to-end, but the server-side batch ordering and shed accounting are covered only indirectly. Add focused server tests for:

- A Level-0 batch followed by a Level-1 produce WITHIN one pass (the `flush_level0`-before-`produce_submit` ordering): the L1 must land after every batched L0.
- A non-produce frame (ack / flow / sub) mid-batch (the flush before `drain_parked` + dispatch): the actor must observe the L0s before the non-produce op.
- The `MAX_PARKED_PRODUCES` cap-flush: more than 1024 Level-0 in one pass splits into multiple in-order batches.
[...]

PR #782, opened 2026-06-29
perf(server): lean the Level-0 batch path — in-place retain + gather counts appends (#775)

ELares:
Leanness + test-hardening follow-ups for the Level-0 (QoS-0) batch path, from the #755 review (#775).

## Code (both provably equivalent)
- **`produce_no_reply_batch` capped arm:** filter at-or-over-cap appends **in place** with `Vec::retain` instead of building a second `admitted` `Vec`. `retain` calls the predicate once per append in order, so the per-append shed-sampling and the in-order admission are identical to the per-message path; only the extra allocation is gone. (The uncapped default — the common case — already moves the whole batch into the command with no copy, from #755.)
- **`gather_commands`:** count a `ProduceNoReplyBatch` as `appends.len()` produces, not one, so a coalesced L0 batch trips the `produces < 2` bounded-gather heuristic the same as that many per-message produces would. It already group-commits as one batch; this only restores the cross-socket-read fsync amortization the gather is meant to capture (a 1000-append batch previously read as "1 produce" and never gathered).

## Test
Adds `an_over_cap_level0_batch_sheds_every_append_into_fire_and_forget_shed` — the batch twin of the existing per-message over-cap shed test. On a full log, a batch of L0s sheds every append in place; the test asserts each is counted in `fire_and_forget_shed` (never `produce_rejected`), and nothing is enqueued when all shed. This guards the capped admission path the #755 review flagged as untested.

## Decision: buffer pooling declined (with rationale)
The review's headline leanness idea was to pool the batch buffer. Evaluated and **declined**: the batch `Vec` moves by value into `Command::ProduceNoReplyBatch` and is freed on the actor thread, so reuse requires a return channel (the `reply_pool` pattern) — real lifecycle complexity for saving one small `Vec`-of-pointers allocation per pass on a lossy QoS-0 path. The simpler code is the better engineering choice. ("Is there a better way?" → yes: don't add the channel.)

## Scope note
The remaining #775 test ideas (L0→L1 interleave within one pass, a non-produce frame mid-batch, the `MAX_PARKED_PRODUCES` cap-flush) exercise the **ordering** discipline, which this PR does not change and which the #755 review verified by tracing every flush point. They are lower-value hardening; I left #775 open to track them rather than over-claim them here.

Refs #775.

### DIFF
--- a/crates/ironbus-server/src/actor.rs
+++ b/crates/ironbus-server/src/actor.rs
@@ -596,7 +596,7 @@ impl<F: Filesystem + 'static, C: Clock + 'static> EngineHandle<F, C> {
     /// # Errors
     /// Returns [`ActorGone`] if the actor exited before the batch could be enqueued. (A per-append SHED
     /// is NOT an error: the L0 produce is dropped, counted, and the rest of the batch still sends.)
-    pub fn produce_no_reply_batch(&self, appends: Vec<OwnedAppend>) -> Result<(), ActorGone> {
+    pub fn produce_no_reply_batch(&self, mut appends: Vec<OwnedAppend>) -> Result<(), ActorGone> {
         if appends.is_empty() {
             return Ok(());
         }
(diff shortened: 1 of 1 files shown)


ELares:

**Verdict: approve, no findings.** An independent reviewer verified all six points against the code:
- **`retain` is provably equivalent to the old loop:** Rust's `Vec::retain` visits each element exactly once in original order and preserves retained order, so the per-append `would_shed()` re-sampling, the `record_l0_shed()` count, and the admitted order are identical to the old `admitted`-Vec loop — only the second allocation is gone. (The side-effecting predicate over the mutable gate snapshot is exactly why the once-per-element-in-order guarantee is load-bearing, and it holds.)
- **Uncapped fast path unchanged:** `cap() == 0` returns before the retain; `mut appends` is a local-binding annotation, invisible to callers, no copy reintroduced.
- **`gather_commands` count correct and safe:** `produces` feeds only the `produces < 2` gate (opt-in, default off); counting a batch's appends only affects *whether* the timed gather is entered, never what it collects or the order — no correctness impact, the documented intent restored.
- **New test drives the capped arm:** `rig_capped` + `fill_to_cap` makes `would_shed()` true for all 7 appends; asserts `fire_and_forget_shed += BATCH`, `produce_rejected` unchanged; `BATCH: u64` with no lossy cast.
- **No borrow/aliasing footgun** (closure captures `&self.cap_gate`, disjoint from the moved-in `appends`); **empty-result behavior identical** (no empty `ProduceNoReplyBatch` ever enqueued).

Buffer pooling was evaluated and declined with rationale (the Vec moves into the channel; reuse needs a return channel — not worth it on a lossy path). CI must be green before merge.

PR #783, opened 2026-06-29
fix(ci): skip shell-variable placeholder URLs in the external-link check (#699)

### LINKED ISSUE #699: [docs-link] dead external source URL
The weekly external-URL reachability check (`scripts/ci/external-link-check.sh`, #30) found one or more dead links cited in the docs:

- `http://$addr/healthz` (connection-failed)
- `http://$addr/readyz` (connection-failed)

A cited source that 404s or fails to resolve makes the evidence base unauditable. Fix by updating the URL, pinning a GitHub citation to a `blob/<commit-sha>` permalink, or recording an archive.org snapshot (see docs/PRIOR_ART_AND_IO_STANCE.md, the source-URL durability policy). If a host merely rate-limits the checker, add it to the documented skip-list in the script.

Filed automatically by `scripts/ci/docs-link-report.sh`; deduplicated by the `[docs-link] dead external source URL` title marker.

ELares:
Closes #699 (an auto-filed false positive).

The weekly external-link check flagged `http://$addr/healthz` and `http://$addr/readyz` as dead links. These are usage snippets in `docs/OPERATIONS.md` where `$addr` is a shell variable (a fill-in-the-blank token), not real endpoints.

**Fix:** add a skip pattern to `is_skipped_url` in `scripts/ci/external-link-check.sh` for any URL containing a `$`, mirroring the existing ADR `issues/N` placeholder skip. A skip is never a failure and is logged with its reason.

**Verified:** the new pattern skips both placeholder URLs and still checks real `https://github.com/...` URLs; `sh -n` is clean; the change follows the existing skip-list style. `external-link-check.sh` runs only in the weekly `docs-link-check` cron (not a per-PR gate), so the next weekly run stops re-filing this.

### DIFF
--- a/scripts/ci/external-link-check.sh
+++ b/scripts/ci/external-link-check.sh
@@ -92,9 +92,12 @@ is_skipped_host() {
 # a skipped URL is NEVER a failure; the match reason is logged.
 #   - .../issues/N : the ADR template's literal `#N` placeholder
 #     (docs/adr/template.md), meant to be replaced per ADR, not a real issue.
+#   - any URL containing a shell variable (e.g. `http://$addr/healthz` in a usage
+#     snippet, docs/OPERATIONS.md): a fill-in-the-blank token, not a real endpoint.
 is_skipped_url() {
 	case "$1" in
 	*github.com/*/*/issues/N | *github.com/*/*/issues/N/) echo "ADR template placeholder (#N), filled in per ADR"; return 0 ;;
+	*'$'*) echo 'shell-variable placeholder (e.g. $addr) in a usage snippet, not a real endpoint'; return 0 ;;
 	esac
 	return 1
 }


ELares:
## Review (self-review; proportionate for a 3-line skip-list fix)

Verified:
- The case pattern `*'$'*` skips `http://$addr/healthz` and `http://$addr/readyz` and still routes real `https://github.com/...` URLs to the live check (tested the case logic in isolation).
- `sh -n scripts/ci/external-link-check.sh` is clean; the addition mirrors the existing ADR `issues/N` placeholder skip in both form and the logged-reason convention.
- A skip is never a failure (per the script's own contract), so this can only stop a false-positive, never hide a real dead link (a real source URL would not contain a shell variable).
- `external-link-check.sh` is invoked only by the weekly `docs-link-check` cron, not a per-PR gate, so this takes effect on the next weekly run.

Issue #30, opened 2026-06-05
[TASK] Add a CI link-check and SHA-pin all prior-art source URLs


ELares:
Parent: #2

The #2 acceptance criterion 'all cited URLs are verifiable' is currently unenforced, and two references (the nats-server `filestore.go` and the Pulsar `bookkeeper.conf`) link to `main`, which is mutable, so their cited content can silently change or disappear. This task makes link integrity an automated, blocking check.

## Approach
Add a CI job (runnable in the planning repo before any product code exists) that extracts every URL from the #2 references and from `claims.yaml`, then fails on any 4xx/5xx, redirect to an unrelated path, or missing section anchor. Convert every GitHub blob link from a branch ref to an immutable commit SHA so the cited lines cannot move. For vendor blog posts and docs that may move (the Redpanda write-caching post, vendor doc trees), record an archive.org snapshot URL and timestamp alongside the live URL in `claims.yaml`. Run the check on every PR that touches the issue body or the claims file, and on a scheduled weekly cron so rot is caught even with no edits. Treat a dead link as a blocking failure, not a warning.

## Acceptance criteria
- [ ] CI extracts and checks every prior-art URL and fails on dead links or missing anchors.
- [ ] All GitHub source links are pinned to a commit SHA, not a branch.
- [ ] Mutable vendor URLs have a recorded archive.org snapshot and timestamp.
- [ ] The check runs on relevant PRs and on a weekly schedule.
- [ ] A dead link blocks merge rather than warning.

## Failure considerations
The primary risk is link rot making the evidence base unauditable and failing the issue's own acceptance criterion; SHA-pinning and archive snapshots make cited content immutable. A second risk is false greens where a URL resolves but its content changed (anchor gone); checking the section anchor, not just the host, catches that. A third risk is the check being advisory and ignored; making a dead link block merge keeps the survey honest over time. The weekly cron catches drift that PR-triggered checks would miss during quiet periods.

PR #899, opened 2026-06-29
fix(health): split the request head on its own newline, not a raw byte index (#860)

### LINKED ISSUE #860: [critical] Malformed HTTP request line aborts the whole broker via a UTF-8 char-boundary panic when slicing the lossy request head
## The failure mode
`read_request_head` (`crates/ironbus-server/src/health.rs:438-443`) finds the newline in the RAW buffer:
```rust
if let Some(pos) = buf[..len].iter().position(|&b| b == b'\n') { break pos; }
...
let head = String::from_utf8_lossy(&buf[..len]).into_owned();
Ok(RequestHead::Parsed { head, line_end })
```
`line_end` is a raw-byte index, but `from_utf8_lossy` replaces each invalid byte with the 3-byte U+FFFD, lengthening the string and shifting every later offset. The caller then slices the lossy `String` with that raw index:
```rust
let line = head[..request_line_end].trim_end_matches('\r');   // line 275
let accept = parse_accept_header(&head[request_line_end..]);    // line 289
```
When the request line carries an invalid-UTF-8 byte, `request_line_end` lands inside a U+FFFD sequence and the slice panics. Verified with rustc: `b"GET /healthz \xFF\n"` -> `line_end=14`, `head.len()=17`, `is_char_boundary(14)=false`, `head[..14]` panics: "byte index 14 is not a char boundary; it is inside '\u{FFFD}' (bytes 13..16)".

## Consequence
`handle` runs synchronously in the accept-loop thread (line 219); the wrapping `let _ = handle(...)` only discards the returned `io::Error`, NOT an unwind. With `panic = "abort"` (`Cargo.toml:49`) the panic aborts the ENTIRE process: every connection, every partition, the cluster driver. The health/metrics listener is unauthenticated and pre-auth, so a single one-packet request from any client that can reach `health_addr`/`metrics_addr` (a k8s probe, an orchestrator, a peer) remotely crash-loops the node — a trivial remote DoS, and accidentally triggerable by a probe sending a high-bit path byte.

## Proposed hardening
Do not index the lossy String with a raw-byte offset. Split the request head on the RAW `&[u8]` (compute request-line vs header slices of `buf[..len]`, then `from_utf8_lossy` each part), or split `head` on its own `'\n'` via `head.split_once('\n')` instead of carrying `line_end` across the lossy conversion.

## How to test
Connect to the health listener and send `GET /healthz \xFF\n`. Before the fix the broker process aborts; after, it returns a normal response (or 400) and stays up. Add this as a regression test feeding a 0xFF byte before the newline.

ELares:
Closes #860 (a [critical] from the pre-adoption audit, milestone #30).

## The bug
A malformed HTTP request line with an invalid-UTF-8 byte **aborts the whole broker**. `read_request_head` returned `line_end` as a RAW-buffer byte index, but the caller sliced the `from_utf8_lossy` string with it. `from_utf8_lossy` turns each invalid byte into a 3-byte U+FFFD and shifts every later offset, so the raw index lands inside a U+FFFD and the slice panics. `handle()` runs inline on the accept thread and the release profile is `panic = abort`, so **one unauthenticated packet** (`GET /healthz \xFF\n`) from anything that can reach the pre-auth health/metrics listener crash-loops the node.

## The fix
Drop the raw `line_end`; split the lossy `head` on its own first `\n` via `split_once('\n')`. A `\n` survives the lossy conversion unchanged, so the split is char-boundary-safe and identical to before for well-formed input.

## Test
Adds `a_request_line_with_an_invalid_utf8_byte_does_not_crash_the_broker`: feeds a 0xFF byte in the request line, asserts an HTTP response comes back, and that a follow-up probe on the same listener still returns 200 (the broker survived). 55 health tests pass; clippy + fmt clean.

Pre-adoption: no back-compat concern.

### DIFF
--- a/crates/ironbus-server/src/health.rs
+++ b/crates/ironbus-server/src/health.rs
@@ -264,15 +264,21 @@ where
 
     // Read the request HEAD (request line + headers) under a bounded byte and total-time budget; the
     // outcome is either a parsed head or an already-sent error response (414/408) or a clean close.
-    let (head, request_line_end) = match read_request_head(&mut stream)? {
-        RequestHead::Parsed { head, line_end } => (head, line_end),
+    let head = match read_request_head(&mut stream)? {
+        RequestHead::Parsed { head } => head,
         // Either an error response was already sent (414/408) or the client closed before a full
         // request line; in both cases there is nothing more to answer.
         RequestHead::Responded | RequestHead::Closed => return Ok(()),
     };
 
+    // Split the request line from the header section on the first `\n` of the LOSSY string itself,
+    // never a raw byte index: `from_utf8_lossy` turns each invalid byte into a 3-byte U+FFFD and shifts
+    // every later offset, so indexing the lossy string with a raw-buffer index can land inside a U+FFFD
+    // and panic the whole process (#860). `\n` survives the lossy conversion unchanged, so `split_once`
+    // is char-boundary-safe. The read loop only yields `Parsed` once a `\n` is present, so this splits.
+    let (request_line, header_section) = head.split_once('\n').unwrap_or((head.as_str(), ""));
     // Parse "METHOD PATH VERSION" (a leading CR is trimmed by split_whitespace).
-    let line = head[..request_line_end].trim_end_matches('\r');
+    let line = request_line.trim_end_matches('\r');
     let mut parts = line.split_whitespace();
     let method = parts.next().unwrap_or("");
     let raw_path = parts.next().unwrap_or("");
@@ -286,7 +292,7 @@ where
     }
     // The `Accept` header value (lower-cased, all values joined), for the `/admin` version pin (#99).
     // Absent on a plain `curl`, which is fine: an absent or wildcard Accept takes the current version.
-    let accept = parse_accept_header(&head[request_line_end..]);
+    let accept = parse_accept_header(header_section);
     // Drop any query string.
     let path = raw_path.split('?').next().unwrap_or(raw_path);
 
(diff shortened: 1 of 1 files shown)


ELares:

**Verdict: approve.** An independent reviewer verified against the code (and reproduced the exact pre-fix panic standalone):
- **Complete:** `line_end`/`request_line_end`/`.position(` are fully removed; the only `from_utf8_lossy` of network bytes (`head`) is now consumed only by `split_once('\n')` and `parse_accept_header` — neither raw-indexes. The `/admin`, `/metrics`, `/healthz`/`/readyz` paths slice no network string; the remaining byte-index slices are on `Vec<u8>` (not `str`) or test-only ASCII.
- **Equivalent for well-formed input:** `split_once('\n')` yields the same request-line/header-section split; the old leading-`\n` only produced an empty element `parse_accept_header` already skipped, so Accept extraction is identical. The read loop only yields `Parsed` once a `\n` is present, so the `unwrap_or` default is unreachable (defensive).
- **No new panic/UB**, borrows sound.
- **Regression test** drives the exact bug: `b"GET /healthz \xFF\r\n\r\n"` → pre-fix raw `line_end=15` lands inside the U+FFFD (bytes 13..16) and `head[..15]` panics → `panic=abort` aborts the broker; the test asserts no-crash AND broker-still-serving. Edge cases (empty line, no `\r`, no headers) all safe.

clippy + fmt clean; 55 health tests pass. CI must be green before merge.

PR #900, opened 2026-06-29
fix(engine): a frozen named stream returns fatal WriterFrozen, never a non-durable ack (#861)

### LINKED ISSUE #861: [critical] Named-stream produce discards commit_tick CommitOutcome and acks a frozen (non-durable) record
## The failure mode

A client `PubTo` to a NAMED stream (`session.rs:851` → `handle_pub_to` → `produce_in_stream`), or a txn commit targeting a named stream (`engine.rs:4318`), runs the cross-stream group-commit. When that stream's covering `fdatasync` fails its barrier (EIO / ENOSPC-on-fsync / write-succeeded-but-fsync-failed), `StreamSet::commit_tick` (`streamset.rs:549-600`) FREEZES only that stream: it pushes the id into `outcome.froze` and does NOT advance `synced_offset` (lines 575-590). By contract (`streamset.rs:544-547`) it NEVER returns `Err` — the caller must inspect `froze` to withhold acks.

`produce_in_stream` does not:

```rust
let _outcome: CommitOutcome = self.streams.commit_tick(); // engine.rs:4010
Ok(offset)
```

The whole outcome (including `froze`) is discarded. `session.rs:3117-3124` sees `Ok(offset)` and sends a `PubAck`.

## Consequence

Ack-before-durable / I2 violation. The producer is told its record was durably accepted while the covering `fdatasync` FAILED and `synced_offset` never moved; on brownout/power-cut the acked record is in the torn tail and is lost on recovery. `session.rs:3072-3074` explicitly claims this PubAck is "ack-implies-durable ... (I2 per stream)" — false the moment the disk barrier faults. The default stream is correctly protected (`engine.rs:3930` `commit_batch()?` propagates the freeze as a fatal `Err`), so the two paths diverge exactly under disk failure. Silent per-record durability loss on every named stream whose fd faults; the txn-commit-on-named-stream path reports the txn committed for a non-durable record.

## Proposed hardening

Inspect the outcome: if `outcome.froze` contains `id` (or `synced` does not cover the appended offset), return a fatal `EngineError::Storage(WriterFrozen)` instead of `Ok(offset)`. The fatal-frame path already exists (`session.rs:3129` `Err(e) if e.is_fatal()`), mirroring the default `commit_batch()?` / `flush_pending` behavior.

## How to test

In a named-stream produce test, inject a `Filesystem` whose `fdatasync`/`sync_data_only` returns EIO for that stream's fd. Assert `produce_in_stream` returns a fatal `Err` (and `handle_pub_to` sends a fatal frame, NOT a PubAck), and that `synced_offset` did not cover the appended offset.

ELares:
Closes #861 (a [critical] from the pre-adoption audit, milestone #30).

## The bug
`produce_in_stream` discarded `StreamSet::commit_tick`'s `CommitOutcome` (`let _outcome`) and always returned `Ok(offset)`. When a named stream's covering `fdatasync` fails its barrier, `commit_tick` **freezes that stream** (pushes it into `froze`, does NOT advance its durable head) and by contract never returns `Err` — the caller must inspect `froze`. So the swallow **acked a record whose fsync FAILED**: an ack-implies-durable (I2) violation. The producer is told the record is durable while it sits in the torn tail, lost on the next power cut. The default stream is protected by `commit_batch()?` propagating the freeze as fatal, so the two paths diverged exactly under disk failure; the txn-commit-to-a-named-stream path inherited the hole.

## The fix
Inspect the outcome: if `froze` contains the appended stream's id, return a fatal `EngineError::Storage(WriterFrozen)`, mirroring the default path. `is_fatal()` already includes `WriterFrozen`, so the session's existing `Err(e) if e.is_fatal()` arm sends a fatal frame, not a PubAck. This also fixes the txn path (it routes through `produce_in_stream`).

## Test
`a_named_stream_produce_whose_fdatasync_fails_returns_fatal_not_an_ack`: produce to a named stream, arm an fsync fault (`FaultFs::set_fail_sync`), produce again — assert `produce_in_stream` returns a fatal `WriterFrozen` (not `Ok`/an ack). 923 server tests pass; clippy + fmt clean.

Pre-adoption: no back-compat concern.

### DIFF
--- a/crates/ironbus-server/src/engine.rs
+++ b/crates/ironbus-server/src/engine.rs
@@ -4007,7 +4007,17 @@ impl<F: Filesystem, C: Clock + Clone> Engine<F, C> {
         // ONE coordinated commit tick (#678): K dirtied named streams => K fdatasync barriers,
         // amortized over the batch; a clean stream costs nothing. The default `""` slot is never
         // dirtied here, so this never syncs the root a second time.
-        let _outcome: CommitOutcome = self.streams.commit_tick();
+        let outcome: CommitOutcome = self.streams.commit_tick();
+        // #861: a named stream whose covering `fdatasync` FAILED this tick is now FROZEN read-only -
+        // its durable head did NOT advance and its parked acks were NOT released. `commit_tick` never
+        // returns `Err` by contract; the caller MUST inspect `froze`. Surface a FATAL `WriterFrozen`
+        // instead of acking a non-durable record, exactly as the default stream's `commit_batch()?`
+        // does. Acking here would violate I2 (ack-implies-durable, per stream): the producer would be
+        // told the record is durable while the disk barrier faulted, and a power cut would lose it from
+        // the torn tail. This also covers the txn-commit-to-a-named-stream path, which routes here.
+        if outcome.froze.contains(&id) {
+            return Err(EngineError::Storage(StorageError::WriterFrozen));
+        }
         Ok(offset)
     }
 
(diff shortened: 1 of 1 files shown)


ELares:

**Verdict: approve.** An independent reviewer verified end-to-end against the source:
- **Completeness (the load-bearing point):** there is exactly ONE `commit_tick(` caller and ONE `append_to(` caller in the whole server crate (engine.rs:4010/4000), so `produce_in_stream` is the single funnel for every named-stream append+commit. All entry points route through it: direct, subject-routed (`produce_by_subject` → `produce_in_stream`, engine.rs:4991), and txn-commit-to-named (`commit_real_append` → `produce_in_stream`, engine.rs:4328). The default `""` target uses `commit_batch()?` which already propagated the freeze. No other `CommitOutcome` swallow remains.
- **Correctness:** `froze` and `synced` are mutually exclusive per id in `commit_tick`; the check is genuinely per-id (acks `id` iff `id`'s own barrier succeeded), correct even when multiple streams are dirtied. `StreamId` derives `Eq`.
- **`WriterFrozen` is fatal** (engine.rs:830), so the session sends a fatal frame, not a PubAck; the default-path asymmetry (`commit_batch()?` → `Log::sync` → fatal) is confirmed.
- **Happy path unchanged** (empty `froze` → `Ok(offset)`).
- **Regression test** drives the real freeze (FaultFs + set_fail_sync), asserts fatal WriterFrozen, would FAIL pre-fix (old code returned Ok), deterministic/non-flaky (reviewer ran it: passes).

923 server tests pass; clippy + fmt clean. CI must be green before merge.

PR #901, opened 2026-06-29
fix(storage): roll back the segment writer on a mid-append spill-flush IO error (#859)

### LINKED ISSUE #859: [critical] Mid-append spill-flush IO error leaves the segment writer torn and un-frozen, corrupting offsets and silently losing acked data
## The failure mode

In SegmentWriter::append (crates/ironbus-storage/src/segment.rs:663-669) write_pos is set to end, then the 256 KiB spill calls flush_pending()? BEFORE record_count += 1 and last_seq. On a flush error (write_all_at hits ENOSPC or flash EIO), flush_pending (segment.rs:728) returns early WITHOUT clearing pending or advancing pending_base, so the just-encoded frame stays buffered while write_pos is advanced but record_count and last_seq are not. Asymmetry: the encode-failure path (653-656) truncates pending back to the pre-append length; the flush-failure path does not. append_at (551-556, the compaction path) is identical.

The error then propagates with NO freeze. append_with_policy (engine.rs:5875) does Err(e) is Err(e) because ENOSPC/EIO maps to StorageError::Io, not AtCapacity. append_no_sync (engine.rs:5103) returns the error leaving self.active = Some. actor.rs:1163 returns ProduceOutcome::Failed and skips commit_batch, so the freeze-on-failed-sync net (log.rs:1946-1952, which only lives inside sync / flush_no_sync) never fires. The same writer is reused on the next produce. Log::append's question-mark at log.rs:1863 returns before advancing next_seq / next_offset, so the retried record reuses seq S and the derived offset base+N.

## Consequence

When a later flush succeeds (disk freed by the retention reaper, or a transient flash EIO that succeeds on retry) both frames are committed together: two physical frames, record_count short by one, both carrying seq S. Pre-restart, a consumer at the acked offset base+N reads the phantom failed frame; post-restart, the dense seq-continuity recovery sees seq S twice, truncates the real frame, and silently loses the acknowledged record. Corruption plus acked-data loss under a routine disk-full or flash-EIO event. Related to #798-807, distinct.

## Proposed hardening

Treat ANY error from flush_pending() inside append / append_at the same as a failed sync barrier: freeze the writer (drop the active segment, surface WriterFrozen). At minimum roll back to the pre-append state on the flush-failure branch (pending.truncate(before) and write_pos = pos_before) so the invariant record_count equals the frame count in the written region is never broken.

## How to test

Wrap the segment file in a fault injector whose write_at returns ENOSPC exactly once, the first time the pending buffer crosses 256 KiB. Produce a relaxed-durability batch larger than 256 KiB (the append fails), clear the fault, produce one more record, then sync. Assert no duplicate offset, record_count matches the frames on disk, and after reopen the second record still survives.

ELares:
## What

Closes #859 (3rd of the 3 critical audit findings in milestone #30).

`SegmentWriter::append` (and the compaction-path `append_at`) encode a record into the shared `pending` buffer, advance `write_pos`, then — if the buffer crossed the 256 KiB spill cap — flush it with one `write_all_at`. That flush is the **last fallible step before** `record_count`/`last_seq` are bumped. A flush error returned immediately and left the writer **torn**: `write_pos` already advanced and the just-encoded frame still buffered, yet `record_count` un-incremented.

Left torn, the **next** append reused the same offset/seq (`record_count` never moved) and, once a later flush succeeded, wrote **both** frames under one count — a corrupted, non-dense log that **silently dropped the record the producer had already been acked for** (an I2 / ack-implies-durable violation). Under a relaxed durability level a >256 KiB batch is exactly when the spill fires mid-append, so a transient ENOSPC / flash EIO triggers it.

## Fix

On a spill-flush error, roll back to the exact pre-append state — `pending.truncate(before)` to drop the encoded frame, restore `write_pos = pos_before` — then propagate the IO error so the producer is never acked. `flush_pending` leaves `pending`/`pending_base` untouched on error (the `?` fires before either is mutated), so only the frame and `write_pos` need undoing; this mirrors the existing encode-failure roll back. Applied to **both** `append` and `append_at`.

## Test

`a_spill_flush_io_error_mid_append_does_not_corrupt_offsets_or_lose_acked_data` (log.rs): three durable records (offsets 0..=2), then a single >256 KiB record whose spill flush fails under an armed write fault. After clearing the fault the next record takes the **same** offset (3 — the failed append consumed none), the live log is dense with no duplicate (4 records), and the survivor record survives a **reopen**.

## Verification

- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-storage` — 410 passed (incl. the new regression)

### DIFF
--- a/crates/ironbus-storage/src/log.rs
+++ b/crates/ironbus-storage/src/log.rs
@@ -5837,6 +5837,73 @@ mod tests {
     }
 
     #[test]
+    fn a_spill_flush_io_error_mid_append_does_not_corrupt_offsets_or_lose_acked_data() {
+        use crate::fault::FaultFs;
+        // #859 regression: a write error on the 256 KiB spill flush mid-append must NOT leave the writer
+        // TORN (write_pos advanced, record_count not bumped) - which would reuse this offset/seq on the
+        // next append and, when a later flush succeeds, write both frames under one count, corrupting the
+        // log and silently losing the acked record. A single > 256 KiB record makes its own pending
+        // buffer cross the spill cap, so its in-append flush fires; armed with a write fault it fails.
+        // After clearing the fault the next record must take the SAME offset (the failed one consumed
+        // none and left no torn frame), and the dense log must survive a reopen.
+        let (fs, control) = FaultFs::new(InMemoryFs::new());
+        let probe = fs.inner().clone();
+        {
+            let mut log = Log::open(fs, ManualClock::new(), LogConfig::default()).unwrap();
+            // Three durable records (offsets 0..=2, fsync clean).
+            for i in 0..3u64 {
+                assert_eq!(log.append(&rec(b"durable")).unwrap(), Offset::new(i));
+            }
+            log.sync().unwrap();
+            assert_eq!(log.flushed_offset(), Offset::new(3));
+
+            // A single record larger than the 256 KiB spill cap: its in-append flush fails under the
+            // armed write fault, so the append errors. Pre-fix this left the writer torn at offset 3.
+            let big = vec![b'x'; 300 * 1024];
+            control.set_fail_write(true);
+            assert!(
+                log.append(&rec(&big)).is_err(),
+                "the > 256 KiB record's spill flush fails under the armed write fault"
+            );
+
+            // Clear the fault and append the next record: with the roll-back fix it takes offset 3 (the
+            // failed append consumed NO offset and left no torn frame), not a skipped or reused one.
+            control.set_fail_write(false);
+            let after = log.append(&rec(b"survivor")).unwrap();
+            assert_eq!(
+                after,
+                Offset::new(3),
+                "after a rolled-back spill failure the next append reuses no offset and skips none"
+            );
+            log.sync().unwrap();
+
+            // Dense and no-duplicate in the live log.
+            let recs = log.read_from(Offset::ZERO, 1000).unwrap();
+            assert_eq!(
+                recs.len(),
+                4,
+                "record_count matches the frames on disk: no duplicate or skipped offset"
+            );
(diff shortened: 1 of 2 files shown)


ELares:
### Fresh-eyes review — verdict: APPROVE

An independent adversarial review verified:

- **Roll-back is state-exact.** `flush_pending` returns on the `write_all_at` `?` *before* mutating `pending_base`/`pending.clear()`, so on error both are pristine — the roll-back correctly only undoes the encoded frame (`truncate(before)`) and `write_pos` (`= pos_before`). `record_count`/`last_seq`/`max_timestamp_ms` are all bumped *after* the spill check, so nothing else leaks. Confirmed for **both** `append` and `append_at`.
- **Only vulnerable paths covered.** `write_pos` is advanced-then-flushed only in these two methods; `seal`/`seal_compacted` consume `self` on error (no persisted torn state), and `sync`/`flush_no_sync` add no new frame. `append_at`'s `offset` is a caller-supplied sparse offset, returned only on success.
- **Test genuinely discriminates the bug.** Reviewer ran the new test against the *pre-fix* code in an isolated worktree: it panics because `recs.last().payload` is the big `'x'` record (the torn writer reused offset 3 and dropped the acked survivor). Passes with the fix.

**One actionable point addressed:** the clean-fail test argued but didn't *exercise* the partial-write case (a real `write_all_at` can persist a prefix then error). Added `a_partial_spill_flush_write_mid_append_still_recovers_a_dense_consistent_log` (commit 4f53857) using `arm_torn_write(64)` — proving the unchanged `pending_base` + recovery overwrite/truncate the partial bytes and yield a dense, consistent, no-lost-ack log on reopen.

CI was green on 8ac24ee; re-running on 4f53857 before admin-merge.

PR #903, opened 2026-06-29
fix(engine): roll the txn_commit resolve back to Prepared when a durable commit step sheds (#801)

### LINKED ISSUE #801: [high] txn_commit resolves the lifecycle to Committed before the real append is durable and never rolls back on a STEP A/A2/A3 failure, so a same-process retry fabricates a success offset without appending
## The defect

`Engine::txn_commit` (crates/ironbus-server/src/engine.rs:4197-4246) decides the lifecycle FIRST and mutates the in-memory table before any durable IO:

```rust
let decision = {
    let store = self.ensure_txn_store()?;
    store.table_mut().commit(txn_id, now).map_err(EngineError::Txn)?   // moves Prepared -> Resolved(Committed) NOW
};
```

`TxnTable::resolve` (crates/ironbus-core/src/txn.rs:413-419) removes the txn from the prepared set and inserts the resolved tombstone synchronously — so the table reads Committed before STEP A runs:

```rust
let real_offset = self.commit_real_append(txn_id, &half)?; // STEP A (write, no fsync)
self.flush_txn_commit_dedup()?;                            // A2
self.commit_batch()?;                                      // A3
```

STEP A routes through `append_no_sync_seq` -> `append_no_sync`, which can return the NON-FATAL `StorageError::AtCapacity` (engine.rs:71 documents it; `EngineError::is_fatal` at engine.rs:825 deliberately excludes it), so the engine stays alive after the error. On that `?`, the table is left Committed — and unlike `txn_prepare`, which rolls back its in-memory prepare on a durable-write failure:

```rust
// engine.rs:4163 (txn_prepare)
let _ = store.table_mut().rollback(txn_id, now);
```

`txn_commit` performs no such rollback. The buffered payload survives, because it is only dropped in `mark_committed`/`append_op` (crates/ironbus-storage/src/txn.rs:725), which is STEP B and is never reached.

A same-process retry then re-enters `commit()`, finds the resolved tombstone, and takes the benign arm:

```rust
ironbus_core::txn::ResolveDecision::AlreadyResolved => {
    Ok(self.dedup_offset_for(txn_id))   // returns WITHOUT re-appending
}
```

Since STEP A never recorded the txn pseudo-producer seq high-water, `dedup_offset_for` (engine.rs:4341-4347) hits its fallback:

```rust
_ => self.log.flushed_offset(),
```

returning the current durable head — a fabricated offset that points at an unrelated record — for a message that was never written.

## Impact

A routine, non-fatal storage rejection during commit (e.g. the at-capacity byte-cap shed, or any A/A2/A3 error that leaves the engine alive) leaves the broker reporting the transaction as committed-and-appended while the payload is absent from the real stream within the running process. The retried commit masks the failure as success and returns an offset belonging to a different message. This is a silent exactly-once violation — committed-as-success but lost — for the life of the process. A full restart heals it (the durable half record replays as Prepared and re-drives the commit, deduped by the txn-id seq), but the same-process window has no re-drive: the preserved buffer is never consulted by the `AlreadyResolved` arm. The asymmetry with `txn_prepare`, which rolls back on storage failure, is the tell.

## Proposed fix

Mirror `txn_prepare`'s rollback-on-failure: on any error from STEP A/A2/A3, roll back the in-memory `table.commit()` (back to Prepared) before returning the error, so a retry re-drives the real append. Equivalently, resolve the in-memory lifecycle to Committed only AFTER the durable op-marker (STEP B) — or at least after A3 makes the record durable — succeeds. Either keeps the table consistent with what is durable and lets a retry re-append the still-buffered payload instead of fabricating an `AlreadyResolved` success.

## How to verify/test

Add an engine test that injects a non-fatal storage error (e.g. `AtCapacity`) on the real-stream append during `txn_commit` of a prepared txn, asserts the error is surfaced, then (without restarting) asserts a retried `txn_commit`: (1) does NOT return `AlreadyResolved`/a fabricated offset, and (2) re-drives the real append once the condition clears, producing exactly one durable record at the returned offset. Also assert the table state is back to Prepared (not Resolved) after the failed first attempt.

ELares:
## What

Closes #801 (high). 2nd of the 16 high-severity audit findings in milestone #30.

`Engine::txn_commit` decided the lifecycle **first**: `TxnTable::commit` moved the txn `Prepared → Resolved(Committed)` **in memory before** the durable STEP A/A2/A3 (real append, dedup-seq flush, record fsync). Unlike `txn_prepare` — which rolls back its in-memory prepare on a durable-append failure — `txn_commit` did **no** rollback on a STEP A/A2/A3 error.

STEP A routes through `append_no_sync`, which can return the **non-fatal** `StorageError::AtCapacity` (the byte-cap shed; `EngineError::is_fatal` deliberately excludes it), so the engine stays alive with the table left **Committed over a record that was never written** and the buffered payload still present (dropped only in STEP B's `mark_committed`, never reached).

A same-process retry then re-entered `commit()`, found the resolved tombstone, took the benign `AlreadyResolved` arm, and returned `dedup_offset_for(txn_id)` **without re-appending**. Since STEP A never recorded the txn pseudo-producer seq high-water, `dedup_offset_for` hit its `log.flushed_offset()` fallback — **a fabricated offset pointing at an unrelated record** — for a message that was never written. A routine byte-cap shed during commit thus reported the txn committed-and-appended while the payload was absent, and the retry masked the failure as success: **a silent exactly-once violation (committed-as-success but lost)** for the life of the process. The asymmetry with `txn_prepare` (which rolls back) was the tell.

## Fix

Mirror `txn_prepare`'s rollback-on-failure:

- New `TxnTable::revert_resolved_to_prepared(txn_id, now)` (core) moves a just-resolved tombstone back to `Prepared` (re-aged at `now`; within the `max_prepared` cap because the resolve just freed this id's own slot; a no-op if the id isn't a resolved tombstone).
- Factor STEP A/A2/A3 into `Engine::commit_durable_steps`, and on **any** error from it, call `revert_resolved_to_prepared` before surfacing the error. The table then reflects what is durable (nothing), so a same-process retry re-enters the fresh `Resolved` arm and re-drives the real append (deduped by the txn-id seq if STEP A had recorded it) — exactly one durable record either way.
- **STEP B (`mark_committed`) failure is unchanged**: the record is already durable, so the retry's `AlreadyResolved` returns the now-correct offset and a reopen re-commits deduped — both safe. Only A/A2/A3 (record-not-durable) get the rollback.

## Test

`a_commit_whose_durable_append_sheds_rolls_back_to_prepared_and_never_fabricates_a_success`: a `txn_commit` whose real append sheds the non-fatal `AtCapacity` surfaces the error, leaves the txn **still Prepared** (`txn_prepared_count` stays 1, not 0), appends **no phantom record** (durable head unchanged), and a same-process retry **sheds again** — never fabricating an `AlreadyResolved` success offset; the un-committed payload never reaches a consumer. (Pre-fix: `prepared_count` would be 0 and the retry would return `Ok` with the fabricated head offset.)

## Verification

- `cargo fmt --check` clean
- `cargo clippy -p ironbus-core -p ironbus-server --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-core` (453+) and `cargo test -p ironbus-server` (908+) — all pass, incl. the new regression and the existing txn 2PC suite

### DIFF
--- a/crates/ironbus-core/src/txn.rs
+++ b/crates/ironbus-core/src/txn.rs
@@ -441,20 +441,27 @@ impl TxnTable {
 
     /// Reverts a txn THIS process just freshly resolved (a [`ResolveDecision::Resolved`] returned by
     /// [`TxnTable::commit`]/[`TxnTable::rollback`]) back to `Prepared`, so the caller can UNDO an
-    /// in-memory resolve whose durable commit steps then FAILED (#801). Mirrors how
-    /// `Engine::txn_prepare` rolls back its in-memory prepare on a durable-append failure: the table is
-    /// kept consistent with what is actually durable, so a same-process retry re-enters the fresh
+    /// in-memory resolve whose durable commit steps then FAILED (#801): the table is kept consistent
+    /// with what is actually durable, so a same-process retry re-enters the fresh
     /// [`ResolveDecision::Resolved`] arm and re-drives the real append — instead of taking the benign
     /// [`ResolveDecision::AlreadyResolved`] arm and fabricating a success offset for a record that was
     /// never written.
+    ///
+    /// This is the SAME spirit as `Engine::txn_prepare`'s rollback on a durable-append failure (undo
+    /// the in-memory effect of a step whose durable IO failed) but the OPPOSITE terminal state, so do
+    /// not conflate them: `txn_prepare` calls [`TxnTable::rollback`], which moves the txn to a
+    /// `Resolved(RolledBack)` tombstone — SPENT, a later commit is refused; this restores `Prepared` —
+    /// RE-DRIVABLE, so the still-buffered payload can be committed on retry.
     ///
     /// The restored `Prepared` entry is re-aged at `now` (the original prepare instant was consumed by
     /// the resolve, which left the prepared set + age index); this only re-bases the txn's back-check
     /// schedule, which is moot for a txn being actively re-committed. Re-adding to the prepared set is
     /// within the [`TxnConfig::max_prepared`] cap because the resolve being undone just freed this id's
-    /// own slot. A no-op if `txn_id` is not a resolved tombstone (nothing to revert), so a double
-    /// revert or a revert of an already-redriven txn is harmless. The stale resolved-LRU entry left
-    /// behind is lazily invalidated like any evicted tombstone.
+    /// own slot (if that resolve had evicted ANOTHER group's tombstone under the cap, the revert does
+    /// not restore it — benign, since tombstone eviction is always safe by design). A no-op if `txn_id`
+    /// is not a resolved tombstone (nothing to revert), so a double revert or a revert of an
+    /// already-redriven txn is harmless. The stale resolved-LRU entry left behind is lazily invalidated
+    /// like any evicted tombstone.
     pub fn revert_resolved_to_prepared(&mut self, txn_id: &[u8], now: u64) {
         if self.resolved.remove(txn_id).is_some() {
             self.prepared.insert(
--- a/crates/ironbus-core/src/txn.rs
+++ b/crates/ironbus-core/src/txn.rs
@@ -439,6 +439,35 @@ impl TxnTable {
         }
     }
 
+    /// Reverts a txn THIS process just freshly resolved (a [`ResolveDecision::Resolved`] returned by
+    /// [`TxnTable::commit`]/[`TxnTable::rollback`]) back to `Prepared`, so the caller can UNDO an
+    /// in-memory resolve whose durable commit steps then FAILED (#801). Mirrors how
+    /// `Engine::txn_prepare` rolls back its in-memory prepare on a durable-append failure: the table is
+    /// kept consistent with what is actually durable, so a same-process retry re-enters the fresh
+    /// [`ResolveDecision::Resolved`] arm and re-drives the real append — instead of taking the benign
+    /// [`ResolveDecision::AlreadyResolved`] arm and fabricating a success offset for a record that was
+    /// never written.
+    ///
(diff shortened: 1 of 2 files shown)


ELares:

An independent adversarial review verified the rollback is state-exact, the seq-high-water/dedup interplay is correct end-to-end (A-fails → Fresh re-drive; A-succeeds/A2-fails → Duplicate re-drive of the still-buffered record, exactly one durable record), the STEP-B exclusion is correct, the rollback is harmless on the fatal path, and **the regression test fails pre-fix** (confirmed by reverting in an isolated worktree: `txn_prepared_count()` is 0, and the retry returns `Ok` with the fabricated head offset).

**Addressed the should-fix (commit 3a463a5):** the original test asserted only the negative. Added `a_commit_that_sheds_after_the_real_append_re_drives_to_exactly_one_record_on_retry` — STEP A succeeds + records the seq, STEP A2's checkpoint write fails non-fatally (FaultFs), the resolve rolls back to Prepared, and after the fault clears the retry **dedups to the same offset and commits exactly one durable record** (visible once; idempotent re-commit returns the same offset). This exercises the A-succeeds/A2-fails path that was previously code-verified only.

**Addressed the doc nits:** reworded so "mirrors `txn_prepare`'s rollback" is not read as equivalence (txn_prepare → spent `RolledBack` tombstone; this → re-drivable `Prepared`), documented the `ensure_txn_store` invariant (store was just read by `commit()`, so the `if let Ok` can't silently no-op) and the tombstone-eviction asymmetry.

CI was green on c6348ff; re-running on 3a463a5 before admin-merge.

PR #904, opened 2026-06-29
fix(dlq): dedup dead-letters on exact source offset, not a high-water mark (#800)

### LINKED ISSUE #800: [high] DLQ high-water-mark dedup silently drops a lower-offset poison message dead-lettered after a higher one (zero durable places)
## The defect

Dead-letter idempotency is keyed on a per-group **monotonic high-water mark** rather than exact-offset membership, so a poison message dead-lettered at a source offset *below* an already-recorded one is treated as "already in the sink" and is **never written to the DLQ** — it is committed-past with zero durable records.

`crates/ironbus-storage/src/dlq.rs:414-418`:

```rust
pub fn already_dead_lettered(&self, group: &str, source_offset: u64) -> bool {
    self.highest_source_offset
        .get(group)
        .is_some_and(|&hwm| source_offset <= hwm)
}
```

The hwm is a per-group running max (`append_encoded`, dlq.rs:505-509):

```rust
let entry = self.highest_source_offset.entry(group.to_string()).or_insert(source_offset);
*entry = (*entry).max(source_offset);
```

The comment claims "Source offsets only ever rise for a group, so the high-water mark is a sound dedupe key." That premise is false: poison messages do **not** cross the max-deliver threshold in ascending source-offset order.

The engine poll scan (`crates/ironbus-server/src/engine.rs:6357-6462`) starts at `committed` and, on `Claim::InFlight`, advances WITHOUT checking disposition, then BREAKS at the first reachable poison:

```rust
Claim::InFlight => { offset += 1; }            // skipped, disposition NOT checked
...
Disposition::DeadLetter => {
    g.leases.ack(&token);
    dead_letter = Some((off, token, deliveries, record));
    break;
}
```

So with two poison messages (default `max_in_flight = 64`): at the poll that crosses offset 10's threshold, offset 5 is still leased (`InFlight`) and is skipped; offset 10 is dead-lettered, setting `hwm = 10`. `AckCursor` (crates/ironbus-core/src/cursor.rs) is out-of-order-tolerant — its committed watermark stays at 5 while 10 is acked-ahead — so offset 5 genuinely redelivers, re-poisons, and reaches `Disposition::DeadLetter` on a later poll. Then `dead_letter_in` (engine.rs:6511):

```rust
let already = self.dlq_sink()?.already_dead_lettered(group, off.get()); // 5 <= 10 == true
if !already { /* append_poison / append_dead_letter */ }                // SKIPPED
...
self.commit_dead_letter_past(group, off);                               // offset 5 committed-past
```

The poison at offset 5 is committed-past with no DLQ record. The TTL-expiry path `expire_dead_letter_in` (engine.rs:6604) shares the same map and key, so a max-deliver dead-letter at a higher offset also suppresses a later lower-offset TTL-expiry dead-letter (and vice versa), across reasons and across the DLX target.

## Impact

Silent, unbounded, unreported loss of poison records — the exact thing the DLQ exists to preserve. It violates the module's stated invariant that a dead-lettered message ends up in EXACTLY ONE durable place (written to the sink AND committed-past); here it ends up in ZERO. Reachable on any parallel work-group where two messages are poison and the higher source offset crosses max-deliver (or TTL expiry) while the lower one is leased.

Note the existing test `the_high_water_mark_is_per_group_and_takes_the_maximum` (dlq.rs:737-754) asserts `already_dead_lettered("a", 2)` is `true` after offset 5 is dead-lettered — it encodes the buggy behavior as intended, so the design is wrong rather than merely under-tested.

## Proposed fix

Do not dedup on `<= hwm`. The only legitimate duplicate is the SAME exact source offset re-poisoned after a crash that fsynced the DLQ record but had not yet durably committed the source cursor. Replace the high-water mark with **exact-offset membership**: maintain a per-group set of dead-lettered source offsets at-or-above the source group's committed cursor (bounded by the in-flight/ahead window, `max_in_flight`), pruned as the source cursor advances. `rebuild_high_water_mark` (dlq.rs:378-406) already scans every durable record's `meta.source_offset`, so it can rebuild the exact set instead of just the max. Change `already_dead_lettered(group, off)` to an exact-membership check at both engine.rs:6511 and engine.rs:6604, and update the dlq.rs test to assert exact-offset semantics.

## How to verify/test

Add an engine test: with `max_in_flight >= 2` and `max_deliver = 1`, drive offset 1 to dead-letter while offset 0 is held leased (InFlight), then let offset 0's lease expire and re-poison. Assert the DLQ sink contains BOTH source offsets 0 and 1 (currently it contains only 1). A dlq.rs unit test should assert `already_dead_lettered` is `false` for a never-dead-lettered lower offset even after a higher one was recorded.

ELares:
## What

Closes #800 (high). 3rd of the 16 high-severity audit findings in milestone #30.

DLQ idempotency was keyed on a per-group **monotonic high-water mark**, not exact membership: `already_dead_lettered(group, off)` returned true for any `off <= hwm`. The premise "source offsets only ever rise for a group" is false — poison messages do **not** cross the max-deliver (or TTL-expiry) threshold in ascending source-offset order. With two poison messages, the higher source offset can dead-letter first (the lower one still leased/in-flight, skipped by the in-order poll scan), setting `hwm` to the higher offset. The cursor is out-of-order tolerant, so the lower offset genuinely redelivers, re-poisons, and reaches the dead-letter path — where `lower <= hwm` read as "already in the sink", so it was committed-past **without a DLQ append**.

The poison ended up in **ZERO durable places**, violating the module's exactly-one-place invariant: silent, unbounded, unreported loss of the very records the DLQ exists to preserve. The TTL-expiry path shares the same map and key, so the suppression also crosses reasons and the DLX target.

## Fix

Replace the per-group high-water mark with an **exact set** of dead-lettered source offsets (`BTreeMap<String, BTreeSet<u64>>`):

- `already_dead_lettered` → exact membership (`set.contains(&off)`).
- `append_encoded` inserts the offset; `rebuild_dead_lettered_set` (was `rebuild_high_water_mark`) reconstructs the **exact** set from the durable DLQ records on open — so the only legitimate duplicate (the SAME offset re-poisoned after a crash that fsynced the DLQ record but not the source cursor) is still deduped, while a genuinely-new lower offset is never suppressed.
- The set is bounded by the source group's in-flight/ahead window via the new `DlqSink::prune_below`, called from `commit_dead_letter_past` to drop offsets below the committed watermark (they can never redeliver/re-poison). Pruning the in-memory set is crash-safe — the durable DLQ is the source of truth, rebuilt on reopen.

## Tests

- `dedup_is_exact_offset_membership_not_a_high_water_mark` (dlq.rs): a lower never-dead-lettered offset reads `false` even after a higher one is recorded; plus the `prune_below` bound.
- `dead_letter_dedup_is_exact_so_a_lower_offset_after_a_higher_one_is_not_dropped` (engine.rs, rewritten from the old `re_poisoning_…` test): with the lower offset held leased/uncommitted, the higher offset dead-letters + re-poisons (exactly one record), then the lower offset dead-letters and **is** written — both source offsets land durably in the DLQ (pre-fix: only the higher one). The internal move is invoked directly to construct the lower-still-leased ordering, which the in-order poll scan cannot otherwise produce.
- The TTL-DLX idempotency twin was updated the same way (hold a lower offset uncommitted so the prune doesn't remove the exact entry between the two moves).

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-storage -p ironbus-server --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-storage` (411) green; `cargo test -p ironbus-server` green (incl. the new + updated DLQ tests)

### DIFF
--- a/crates/ironbus-server/src/engine.rs
+++ b/crates/ironbus-server/src/engine.rs
@@ -6499,11 +6499,11 @@ impl<F: Filesystem, C: Clock + Clone> Engine<F, C> {
     /// The ordering is the crash-safety contract: APPEND the poison record to the durable DLQ sink
     /// and FSYNC it, THEN commit the source cursor. A crash between the two leaves the source
     /// uncommitted (it redelivers and is re-poisoned on the next run) and the DLQ record already
-    /// durable; on reopen the per-group dead-lettered high-water mark, rebuilt from the DLQ itself,
-    /// makes the re-poison a no-op append so the message is committed-past WITHOUT a duplicate DLQ
-    /// write. The reconciliation key is `(group, source_offset, attempt)`: an offset at or below
-    /// the group's high-water mark is already in the sink, so it is committed-past without a second
-    /// append.
+    /// durable; on reopen the per-group EXACT dead-lettered set, rebuilt from the DLQ itself, makes
+    /// the re-poison a no-op append so the message is committed-past WITHOUT a duplicate DLQ write.
+    /// The reconciliation key is the EXACT `(group, source_offset)` (#800): an offset already in the
+    /// group's set is in the sink, so it is committed-past without a second append — but a genuinely
+    /// new lower offset (poison order is not ascending) is appended, never suppressed by a higher one.
     ///
     /// The lease has already been dropped by the caller, so on success the message holds no lease
     /// and never redelivers.
@@ -6514,10 +6514,10 @@ impl<F: Filesystem, C: Clock + Clone> Engine<F, C> {
         attempt: u32,
         record: OwnedRecord,
     ) -> Result<Poll, EngineError> {
-        // Idempotency: if this (group, source offset) is already durably in the DLQ (at or below
-        // the group's recorded high-water mark), do NOT append a second copy. This is the path a
-        // redelivered-then-re-poisoned message takes after a crash that landed between the DLQ
-        // append and the cursor commit: the sink already has it, so we only commit past it.
+        // Idempotency: if this EXACT (group, source offset) is already durably in the DLQ, do NOT
+        // append a second copy (#800). This is the path a redelivered-then-re-poisoned message takes
+        // after a crash that landed between the DLQ append and the cursor commit: the sink already has
+        // this exact offset, so we only commit past it. A different (e.g. lower) offset is appended.
         let already = self.dlq_sink()?.already_dead_lettered(group, off.get());
         if !already {
             // APPEND to the DLQ and FSYNC, BEFORE committing the source cursor. A storage error
@@ -6609,8 +6609,8 @@ impl<F: Filesystem, C: Clock + Clone> Engine<F, C> {
         attempt: u32,
         record: OwnedRecord,
     ) -> Result<Poll, EngineError> {
-        // Idempotency: a re-expired message already durably in the exchange (at or below the group's
-        // high-water mark) is committed-past WITHOUT a second append, exactly as the poison path.
+        // Idempotency: a re-expired message whose EXACT source offset is already durably in the
+        // exchange is committed-past WITHOUT a second append, exactly as the poison path (#800).
         let already = self.dlq_sink()?.already_dead_lettered(group, off.get());
         if !already {
             // APPEND the reason-carrying (TtlExpired) dead-letter and FSYNC, BEFORE committing the
@@ -6654,6 +6654,13 @@ impl<F: Filesystem, C: Clock + Clone> Engine<F, C> {
         if group == self.confirm_group {
             self.confirm_registry
                 .terminate(off, ConfirmStatus::DeadLettered);
+        }
+        // #800: the cursor has committed past `off`, so prune the DLQ dedup set below the group's
+        // committed watermark — those source offsets can never redeliver or re-poison, so their
+        // exact-membership entries are no longer needed. This bounds each per-group set by the
+        // in-flight/ahead window. Only when the sink is already open (never lazily create it to prune).
+        if let Some(sink) = self.dlq.as_mut() {
+            sink.prune_below(group, committed);
         }
     }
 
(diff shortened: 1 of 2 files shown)


ELares:

An independent adversarial review confirmed the exact-membership fix closes the bug and that the prune is crash-safe and correctly bounded:

- **Exact-membership** verified; same-offset crash idempotency preserved (rebuild re-derives the exact offset). The new tests **fail pre-fix** (reverting `already_dead_lettered` to `<= max` breaks them; the same-offset/crash tests still pass pre-fix), so they target exactly the #800 defect.
- **Prune correctness:** `committed` is the **exclusive** next-offset, so pruned offsets (`< committed`) are acked and the poll scan skips acked offsets — a pruned offset can never re-poison in-process. `< committed` (not `<=`) is load-bearing (the offset *at* the watermark can still be in the set post-crash-rebuild and must dedup its own re-poison). In-memory pruning is crash-safe (durable DLQ rebuilt on reopen). The bug's own scenario still works (the lower leased offset keeps the watermark low so the higher offset's entry survives).
- **Bounding:** each per-group set is genuinely bounded by `max_in_flight` (every insert is paired with a `prune_below`).
- **Tests are legitimate, not papering over a regression:** production only calls `dead_letter_in`/`expire_dead_letter_in` from the windowed poll scan, which never re-dead-letters an acked offset; the old single-offset idempotency tests artificially re-invoked the move on a committed-and-pruned offset, a sequence the scan cannot produce.

**Addressed (commit 3adefc4):** the should-fix — `expire_dead_letter_in`'s doc still said "a no-op append at or below the per-group high-water mark" (the removed semantics) — plus the stale "high-water mark" wording in the `open`/`open_at`/`append_poison`/`append_dead_letter`/`append_encoded` and engine `dlq_sink`/eager-open method docs. All reworded to the exact dead-lettered offset set.

**Out of scope (noted as follow-up):** the per-group **key** in the DLQ map isn't dropped on `evict_group` — a pre-existing leak (the old `BTreeMap<String, u64>` had it too, and the new code is strictly better-bounded within a group). I'll file a separate issue.

CI was green on 128e1a3; re-running on 3adefc4.

Issue #906, opened 2026-06-29
[high] Geo multi-origin SOURCE can double-apply a batch on a crash between record sync and cursor commit (#799 residual)


ELares:
## Context

Split out from #799. The mirror/single-origin-source case is fixed there (the local log's `next_offset` is an independent durable anchor, so a re-pull below it is skipped). A MULTI-origin SOURCE has no such anchor and is **not** fixed.

## The defect

`MirrorApplier::apply_pull_response` syncs the applied records to the local log (one fsync) and then DURABLY advances the per-origin cursor in a SEPARATE fsync (the `OriginCursorStore` `SlotCheckpoint`). A crash between the two leaves an origin's records durable in the local fan-in log but its durable cursor stale. On restart the origin re-pulls from the stale cursor and re-applies its last batch.

For a single-origin applier the fix reconciles against `log.next_offset()`. A multi-origin source fans N origins into ONE interleaved local log, so `next_offset` (the fan-in total) is NOT any single origin's offset and cannot anchor it. The applier is constructed with `single_origin = false` (see `spawn_geo_plane`: `stream.mode.origins().len() == 1`), so the #799 skip is intentionally OFF for it — the residual double-apply window remains.

## Impact

A crash in the (one-fsync-wide) window injects a duplicated block into the fan-in stream. Narrow window, SOURCE-only (a MIRROR is single-origin and fully fixed), but a real no-dup / durable-prefix invariant violation.

## Proposed fix

The source needs the per-origin cursor durable ATOMICALLY with the records, since it has no positional anchor. Options:
1. Write the per-origin last-applied origin offset into the fan-in log's own durable metadata (a marker record, or a reserved per-record field), made durable by the SAME `log.sync()` as the data, and skip frames at-or-below it on apply (true idempotent apply). The source has no byte-identity requirement, so in-log markers are acceptable (unlike the mirror).
2. A combined checkpoint/WAL that fsyncs the log head and all per-origin cursors together.

Option 1 (in-log per-origin high-watermark + idempotent skip) is the most contained and unifies cleanly with the existing apply path.

## Test

Multi-origin source: apply origin A's batch (synced) without committing A's cursor (fault the cursor `SlotCheckpoint` commit, or split the method), reopen, re-pull A from its stale cursor, and assert the fan-in log's record count is UNCHANGED (no duplicated A block) and per-origin order is preserved.

PR #907, opened 2026-06-29
fix(geo): reconcile a single-origin mirror's cursor with next_offset to stop double-apply on crash (#799)

### LINKED ISSUE #799: [high] Geo mirror/source double-applies records on crash between log fsync and cursor commit (no-dup / byte-identity invariant violated)
## The defect

`MirrorApplier::apply_pull_response` (crates/ironbus-server/src/cluster/geo.rs:789-870) durably advances the per-origin resume cursor in a SEPARATE fsync from the one that makes the applied records durable, with no reconciliation between the two on restart.

The applied records are synced first:

```rust
self.log.append(&append)?;
applied += 1;
at += frame_len;
// ...
// Durably commit the applied batch to the local log (one fsync per pull)
self.log.sync()?;                 // line 849 — records 100..150 now durable, log.next_offset()=150
```

and only afterward is the cursor durably advanced, via a second, independent fsync (the cursor is a `SlotCheckpoint`):

```rust
let new_cursor = cursor + applied;
self.cursors.commit(origin_key, new_cursor)?;   // line 864 — cursor advances to 150
```

If the process crashes in the window between line 849 and line 864, the records are durable in the local log (`next_offset()==150`) but the durable cursor still reads the OLD origin offset (100).

On restart, resume reads ONLY the durable cursor — `pull_request` builds `from_offset: self.cursors.cursor(origin_key)` (geo.rs:766) and the CLI puller resumes from `applier.lock()...cursor(origin_key)` (crates/ironbus-cli/src/main.rs:7018). The origin re-serves its sealed prefix from offset 100, so `resp.first_offset == 100 == cursor`, and the contiguity guard PASSES:

```rust
if resp.first_offset != cursor {            // line 806 — trusts the stale cursor only
    return Err(GeoError::NonContiguous { expected: cursor, got: resp.first_offset });
}
```

The append path does no de-dup — `Append` carries no offset; the log appends at its own `next_offset` — so records 100..149 are appended a SECOND time. `log.next_offset()` goes 150 -> 200, the mirror now holds a duplicated span the origin does not have, and the cursor silently advances to 150 (then 200 worth of records sit under shifted offsets). It is never detected or healed: subsequent pulls remain contiguous from the advanced cursor.

The module doc asserts this is safe — geo.rs:549-551 and 779-781 claim "the local log's own `next_offset`" makes "a re-pull of already-applied data a recognized no-op for a single-origin mirror" — but NO code consults `next_offset`. `MirrorApplier::new` (geo.rs:737-739) and `OriginCursorStore::open` load the log and the cursor independently, and `apply_pull_response` never compares `resp.first_offset` (or the cursor) against `log.next_offset()`. The "load-bearing order" comment only establishes never-SKIP (the cursor never advances past durably-synced data); it does not establish never-DOUBLE-APPLY, which depends on the `next_offset == cursor` invariant that the crash window breaks with nothing to restore it.

For a SOURCE this is strictly unguarded: local fan-in offsets differ from each origin's offsets, so there is no positional reconciliation possible, and a crash in this window injects a duplicated block into the fan-in log.

## Impact

- For a MIRROR: breaks the module's central byte-identity guarantee — the local log gains a duplicated span the origin lacks, and all subsequent records land under shifted offsets, so positional byte-identity is permanently corrupted.
- For a SOURCE: injects duplicate records into the fan-in stream.
- Both violate the durable-prefix / gap-free / no-dup recovery invariant the module repeatedly claims. The duplication is permanent and undetected (the cursor stays contiguous from its advanced value).

The existing resume test (`the_cursor_is_durable_and_resumes_without_gap_or_dup`, geo.rs:1459-1498) only crashes AFTER a fully successful `apply_pull_response` (post-cursor-commit), so this sync-before-commit window is untested.

## Proposed fix

Reconcile the cursor with durable log state so resume can never re-apply already-synced data:

- Single-origin MIRROR: on open, bump the cursor to `max(cursor, local_log.next_offset())` in `MirrorApplier::new` (the mirror's local offset equals the origin offset, so this collapses the window — the cursor becomes doubly-anchored as the doc already claims it is). Equivalently, in `apply_pull_response`, drop/skip any frame whose `at_origin_offset < log.next_offset()` instead of trusting the cursor alone.
- SOURCE: persist the `(origin_offset -> local_next_offset)` high-watermark atomically WITH the records — write the cursor in the same fsync/checkpoint transaction as the log sync — so the durable cursor can never lag durably-applied data. Alternatively record the last-applied origin offset inside the log's own durable metadata and skip frames at-or-below it (true idempotent apply).

## How to verify/test

Add a test that exercises the sync-before-commit window directly: drive `apply_pull_response` to the point where `log.sync()` has succeeded but `cursors.commit()` has not (e.g. inject a fault into the cursor `SlotCheckpoint` commit, or split the method so the test can stop between the two fsyncs), drop and reopen the applier over the same fs, then re-pull from the recovered cursor and assert that `log.next_offset()` and the on-disk record count are UNCHANGED (no duplicated span) and byte-identity with the origin holds.

ELares:
## What

Closes #799 (high). 4th of the 16 high-severity audit findings in milestone #30.

`MirrorApplier::apply_pull_response` syncs the applied records to the local log in **one** fsync, then durably advances the per-origin resume cursor (an `OriginCursorStore` `SlotCheckpoint`) in a **separate** fsync, with no reconciliation on restart. A crash in the window between the record sync and the cursor commit leaves the records durable (`next_offset` advanced) but the durable cursor **stale**. On restart, resume reads only the durable cursor, the origin re-serves from it, the contiguity guard (`first_offset == cursor`) **passes** against the stale cursor, and the append path — which carries no offset and never de-dups — **re-appends the already-applied span a second time**. The local log gains a duplicated block the origin lacks, all later records land under shifted offsets, and positional byte-identity is permanently corrupted, undetected. The module doc *claimed* the local `next_offset` made a re-pull "a recognized no-op for a single-origin mirror", but no code consulted `next_offset`.

## Fix

Make `next_offset` the load-bearing anchor it was only claimed to be. `MirrorApplier` now carries `single_origin` (true for a MIRROR, or a SOURCE with one origin — both apply one origin 1:1 into the local log). For a single-origin applier, `apply_pull_response` treats `max(cursor, log.next_offset())` as the durable anchor and **skips** any frame whose origin offset is below it (already durably applied) instead of re-appending it, advancing the cursor past every frame consumed so it re-anchors to `next_offset`. This closes the never-double-apply gap while preserving never-skip (records still synced before the cursor) and byte-identity (no extra records written).

A **multi-origin SOURCE** interleaves origins into one log, so `next_offset` is not any single origin's offset and cannot anchor it; that path keeps trusting the cursor (`single_origin = false`). Its residual one-fsync-wide window is tracked by follow-up **#906** (needs an atomic per-origin cursor / in-log high-watermark).

Construction sites set the flag from the mode: the federation mirror, leaf mirror, and a single-origin source stream are `true`; `spawn_geo_plane` uses `stream.mode.origins().len() == 1`.

## Test

`a_re_pull_after_the_sync_before_commit_crash_window_skips_not_double_applies`: constructs the post-crash state (a MIRROR whose local log holds `[0, served)` but whose cursor lags), re-pulls the already-applied span to convergence, and asserts **nothing is ever re-appended** — the log does not grow, the cursor re-anchors to `next_offset`, and byte-identity with the origin holds. Pre-fix the span was re-appended, doubling the log. The existing multi-origin fan-in tests are pinned to `single_origin = false` (new `source_applier`/`open_source` helpers).

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-server -p ironbus-cli --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-server` (911) and `cargo test -p ironbus-cli` (435+) green, incl. the new regression and all geo mirror/source/leaf/federation tests

### DIFF
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -6899,7 +6899,10 @@ fn spawn_federation_mirror_puller(
             dir.display()
         ))
     })?;
-    let applier = std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(log, cursors)));
+    // A federation mirror has exactly ONE origin, so the #799 next_offset reconciliation is safe.
+    let applier = std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(
+        log, cursors, true,
+    )));
     let origin = mirrored.origin.clone();
     let key = origin.cursor_key();
     let local = mirrored.local_stream.clone();
@@ -6959,8 +6962,15 @@ fn spawn_geo_plane(
         let cursors = OriginCursorStore::open(&StdFs::new(dir.clone()))
             .map_err(|e| CliError::Internal(format!("open geo cursor {}: {e}", dir.display())))?;
         // The applier is shared across this stream's origins (a source fans N origins into ONE log), so
-        // the single-writer invariant holds across the per-origin threads via the mutex.
-        let applier = std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(log, cursors)));
+        // the single-writer invariant holds across the per-origin threads via the mutex. It is
+        // single-origin (the #799 next_offset reconciliation is safe) iff this stream has exactly one
+        // origin (a MIRROR, or a single-origin SOURCE); a multi-origin fan-in is not.
+        let single_origin = stream.mode.origins().len() == 1;
+        let applier = std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(
+            log,
+            cursors,
+            single_origin,
+        )));
 
         for origin in stream.mode.origins() {
             let applier = std::sync::Arc::clone(&applier);
@@ -7134,8 +7144,10 @@ fn spawn_leaf_plane(
             let cursors = OriginCursorStore::open(&StdFs::new(dir.clone())).map_err(|e| {
                 CliError::Internal(format!("open leaf mirror cursor {}: {e}", dir.display()))
             })?;
-            let applier =
-                std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(log, cursors)));
+            // The hub is the leaf's ONE origin, so the #799 next_offset reconciliation is safe.
+            let applier = std::sync::Arc::new(std::sync::Mutex::new(MirrorApplier::new(
+                log, cursors, true,
+            )));
             // The hub is the leaf's ONE origin (the read bridge IS a geo mirror); dial it outbound.
             let origin = GeoOrigin {
                 addr: leaf.hub_addr.clone(),
--- a/crates/ironbus-server/src/cluster/federation.rs
+++ b/crates/ironbus-server/src/cluster/federation.rs
@@ -789,7 +789,7 @@ mod live_federation_tests {
         )
         .expect("mirror log opens");
         let cursors = OriginCursorStore::open(&StdFs::new(dir.to_path_buf())).expect("geo cursor");
-        MirrorApplier::new(log, cursors)
+        MirrorApplier::new(log, cursors, true)
     }
 
     /// The connector dials a peer gateway (outbound) and pulls a peer-origin stream into its local mirror
--- a/crates/ironbus-server/src/cluster/geo.rs
+++ b/crates/ironbus-server/src/cluster/geo.rs
@@ -518,7 +518,10 @@ impl<'a, F: Filesystem> OriginServer<'a, F> {
 /// The outcome of applying one pull response: how many records were applied and the cursor + HW after.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct ApplyOutcome {
-    /// How many records this response durably applied to the local log.
+    /// How many records this response newly APPENDED to the local log. After the #799 crash-window
+    /// reconciliation this EXCLUDES frames the response carried but the single-origin applier already
+    /// held durably (skipped, not re-appended), so a pure re-pull of already-applied data reports `0`
+    /// applied even though `cursor` advances — progress is keyed off `cursor`/`record_count`, not this.
     pub applied: u64,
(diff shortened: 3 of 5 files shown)


ELares:

An independent adversarial review verified the skip is state-exact (the `seen`/`applied` split, the cursor re-anchoring to `next_offset`, the `RecordCountMismatch`/`CorruptFrame` paths all correct), the anchor `max(cursor, next_offset())` is sound under reaping/compaction (neither lowers the head), skipped frames are still CRC-revalidated (fail-closed preserved), the no-crash and multi-origin paths are provably unchanged, every construction site sets the flag correctly, and **the regression test fails pre-fix** (reverting the skip → `applied: 3` vs expected `0`).

**Addressed the should-fix + nits (commit e0753fa):**

- **Latent apply-only fragility:** the single-origin skip is sound only because the geo applier log is written *exclusively* by `apply_pull_response` (never a local produce — those go to the engine's `StreamSet`). Sharpened the `single_origin` field doc to state this SAFETY INVARIANT explicitly (a future change that routed a local write into the applier log would make `next_offset` lead the origin offset and silently DROP records — so the flag must stay false for any log that can take a non-origin write).
- **Cursor-ahead detection:** added a `debug_assert!(next_offset >= cursor)` for the single-origin path, so a cursor that leads the durable log head (lost records, or a broken apply-only invariant) surfaces loudly instead of silently anchoring at the too-high cursor (the production `max` stays the guard for that impossible case).
- **`ApplyOutcome.applied` doc:** noted it now excludes already-durable skipped frames (a pure re-pull reports `applied = 0` though the cursor advances; progress is keyed off `cursor`/`record_count`).

The multi-origin source residual stays tracked at `[high]` by **#906** in the same milestone. CI was green on 96b170b; re-running on e0753fa.

PR #908, opened 2026-06-29
fix(rss): charge the opt-in dedup windows in the refuse-to-boot RAM proof (#878)

### LINKED ISSUE #878: [high] Refuse-to-boot RAM guard omits the config-capped dedup term, so its from-config proof is false: edge-tiny boots under a 64 MiB ceiling while default dedup caps permit ~122 GiB
## Problem
`validate_ram_ceiling`'s rustdoc (`crates/ironbus-cli/src/main.rs:4471-4479`) claims the guard "sums the bounded RAM sources `docs/RAM_BUDGET.md` itemizes, each at its CONFIGURED cap, so a refusal is a proof from the config that the caps cannot fit."

That proof is incomplete. `RamFootprintConfig` (`crates/ironbus-server/src/rss.rs:240-267`) has fields for connections/credit/groups/in-flight/in-memory-store but NO dedup field, and `worst_case_buffer_bytes` never charges it. Term 6 is excused at `rss.rs:332`: "Term 6 (dedup) costs nothing until a producer opts in, so it is not charged."

But `docs/RAM_BUDGET.md` section 6 (lines 247-309) DOES itemize dedup with a closed config-cap formula `max_producers * max_ids * (msg_id_len + ~2*entry_overhead)`, and states the shipped-default ceiling is `4_096 * 100_000 * ~320 ~= 122 GiB`, "the honest worst case the count knobs must be lowered against for a 64 MiB edge node." `producer_id` is wire-supplied and attacker-chosen, so the window count is bounded only by the config cap, not the connection count.

The `ProfilePreset` struct (`main.rs:660-674`) has no dedup field; `edge-tiny` sets only `ram_ceiling_bytes = 64 MiB` while dedup resolves from `DEFAULT_DEDUP_MAX_IDS=100_000` / `DEFAULT_DEDUP_MAX_PRODUCERS=4096` (`main.rs:2998-3015`), unchanged by the preset.

## Impact
An edge-tiny broker (~15 MiB charged buffers) passes the guard under its 64 MiB ceiling and boots, but a publishing client spraying distinct `producer_id`+`msg_id` can grow dedup RAM far past the ceiling and OOM the device — the exact failure the rustdoc claims to prove impossible. The completeness claim is false: dedup is itemized yet unsummed, and unlike the DLQ it is not flagged as a deliberate exclusion in `validate_ram_ceiling`. The "costs nothing until opt-in" rationale also contradicts the guard's own treatment of `consumer_credit`, charged at its configured cap though it equally costs nothing until a consumer connects.

## Fix
Charge a saturating dedup term `max_producers * max_ids * (avg_msg_id_len + entry_overhead)` into `RamFootprintConfig`/`worst_case_buffer_bytes` so the proof is sound and refuses an edge-tiny ceiling unless the dedup caps are lowered. At minimum, name dedup as a second deliberate exclusion in `validate_ram_ceiling`'s rustdoc, drop the completeness claim, and emit a startup WARN when a non-zero ceiling coexists with dedup caps whose worst case exceeds it.

## Verify
Boot with `--profile edge-tiny --ram-ceiling-bytes 67108864` (defaults) and confirm it boots; then publish many distinct `producer_id`+`msg_id` and observe RSS climb past 64 MiB.

ELares:
## What

Closes #878 (high). 5th of the 16 high-severity audit findings in milestone #30.

`validate_ram_ceiling`'s rustdoc claimed the guard *"sums the bounded RAM sources `docs/RAM_BUDGET.md` itemizes, each at its CONFIGURED cap, so a refusal is a proof from the config that the caps cannot fit."* That proof was **incomplete**: `RamFootprintConfig`/`worst_case_buffer_bytes` never charged **term 6** (the opt-in per-producer dedup windows), excusing it as "costs nothing until a producer opts in". But `producer_id` is **wire-supplied and attacker-chosen**, so the window *count* is bounded only by `--dedup-max-producers`, not the connection count; `docs/RAM_BUDGET.md` §6 pegs the shipped-default worst case at `4096 * 100_000 * ~320 ≈ 122 GiB`.

So an `edge-tiny` broker (~15 MiB charged buffers) passed the guard under its 64 MiB ceiling and **booted**, yet a client spraying distinct `producer_id`+`msg_id` could grow dedup RAM far past the ceiling and OOM the device — the exact failure the proof claimed impossible. The "costs nothing until opt-in" rationale also contradicted the guard's own treatment of `consumer_credit` (charged at its configured cap though it too costs nothing until a consumer connects).

## Fix

**Charge term 6.** `RamFootprintConfig` gains `dedup_max_producers`/`dedup_max_ids`, and `worst_case_buffer_bytes` adds a saturating `dedup_max_producers * dedup_max_ids * DEDUP_ENTRY_BYTES + dedup_max_producers * DEDUP_PRODUCER_KEY_BYTES` term (`DEDUP_ENTRY_BYTES = MAX_MSG_ID_LEN + 2*overhead = ~320`, the doc's worst-case per-entry figure). `validate_ram_ceiling` wires the configured caps in, and the rustdoc now lists dedup as a **summed** term, not an omission.

With the proof sound, the shipped `4096 * 100_000` defaults correctly **refuse** any small ceiling — so the **edge-tiny preset now lowers its dedup caps** to `64 producers * 512 ids` (~10 MiB), bounded under 64 MiB alongside the ~15 MiB buffers and any in-memory store. `ProfilePreset` gains the two dedup fields (edge-tiny lowered; `balanced`/`throughput` keep the shipped default with the guard off), and the dedup caps now resolve from the preset (`flag > env > preset`) exactly like `ram_ceiling_bytes`.

## Tests

- New rss test: the shipped `4096 * 100_000` caps are **provably refused** under a 64 MiB ceiling (~122 GiB worst case).
- The monotonicity drift test now also varies the dedup caps (a future drop of term 6 fails here).
- The edge-tiny footprint/boot + memory-store-fold tests carry the lowered caps and still fit (the store-fold discrimination — 1-image fits / 2-image refuses against 64 MiB — is preserved by the ~10 MiB dedup choice).

## Verification

- `cargo fmt --check`, `cargo clippy -p ironbus-server -p ironbus-cli --all-targets -- -D warnings -W clippy::pedantic` clean
- `cargo test -p ironbus-server --lib rss::` (18) + the CLI ram/store/preset/dedup/ceiling tests (30) green

### DIFF
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -679,11 +679,11 @@ const EDGE_TINY_PRESET: ProfilePreset = ProfilePreset {
     // guard lets edge-tiny boot, and a blown-up cap override (e.g. a server-sized --max-connections,
     // or restoring the default dedup caps) is provably refused.
     ram_ceiling_bytes: EDGE_TINY_RAM_CEILING,
-    // LOWERED dedup caps (#878): the shipped defaults (4096 * 100_000) imply ~122 GiB of worst-case
-    // dedup windows, which the guard now charges and would refuse under 64 MiB. 64 producers * 512 ids
-    // is ~10 MiB of dedup RAM — bounded for a tiny edge node, leaving comfortable headroom under the
-    // 64 MiB ceiling alongside the ~15 MiB buffer terms and any in-memory store.
-    dedup_max_ids: 512,
+    // LOWERED dedup caps (#878): the shipped defaults (4096 * 100_000) imply ~262 GiB of worst-case
+    // dedup windows (each id stored twice, ~640 B/entry), which the guard now charges and would refuse
+    // under 64 MiB. 64 producers * 256 ids is ~10 MiB of dedup RAM — bounded for a tiny edge node,
+    // leaving comfortable headroom under the 64 MiB ceiling alongside the ~15 MiB buffers and any store.
+    dedup_max_ids: 256,
     dedup_max_producers: 64,
 };
 
@@ -6145,7 +6145,8 @@ fn materialized_config_line(config: &ServeConfig, addr: &str, data_dir: Option<&
          group_idle_evict_ms={} checkpoint_interval={} max_deliver={} \
          allow_unlimited_deliver={} disk_full_policy={policy} visibility_timeout_ms={} \
          max_retained_bytes={} max_age_ms={} max_messages={} health_liveness_window_ms={} \
-         enable_admin={} ram_ceiling_bytes={} durability_level={durability_level} \
+         enable_admin={} ram_ceiling_bytes={} dedup_max_ids={} dedup_max_producers={} \
+         durability_level={durability_level} \
          power_loss_safe={power_loss_safe} compression={compression} flush_interval_ms={} \
          flush_max_bytes={} async_loss_ack={} wal_fsync_headroom_bytes={} storage={storage} \
          commit_gather_us={}",
@@ -6169,6 +6170,8 @@ fn materialized_config_line(config: &ServeConfig, addr: &str, data_dir: Option<&
         config.health_liveness_window_ms,
         config.enable_admin,
         config.ram_ceiling_bytes,
+        config.dedup_max_ids,
+        config.dedup_max_producers,
         config.flush_interval_ms,
         config.flush_max_bytes,
         config.async_loss_ack,
@@ -12923,6 +12926,15 @@ mod tests {
         assert_eq!(c.checkpoint_interval, 1024);
         assert_eq!(c.visibility_ms, 30_000);
         assert_eq!(c.max_deliver, 5);
+        // #878: edge-tiny LOWERS the dedup caps so the charged dedup RAM fits its 64 MiB ceiling.
+        assert_eq!(
+            c.dedup_max_ids, 256,
+            "edge-tiny lowers the dedup window depth (#878)"
+        );
+        assert_eq!(
+            c.dedup_max_producers, 64,
+            "edge-tiny lowers the dedup producer cap (#878)"
+        );
     }
 
     #[test]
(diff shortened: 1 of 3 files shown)


ELares:
### Review resolution (3 rounds) — strict-upper-bound blocker RESOLVED

This PR went through three fresh-eyes review rounds:

1. **Round 1** found the dedup term omitted the second `msg_id` copy (the engine stores each id TWICE — `HashMap` index key + `VecDeque` order). Fixed 320 → charge both copies.
2. **Round 2 (BLOCKER)** found the charge still wasn't a *strict* upper bound: `record()` does `push_back` **before** `evict_overflow`, so a power-of-two `dedup_max_ids` momentarily holds `max_ids+1` and never shrinks the doubled `VecDeque` capacity (~64 B/entry of slots); hashbrown runs up to ~2× buckets at its ≤7/8 load factor (~82 B/entry). The real worst case is **~658 B/entry**, which edge-tiny's own 256-id default hits.
3. **Round 3 (confirmation): APPROVE** — `DEDUP_ENTRY_BYTES = 2*MAX_MSG_ID_LEN + 192 = 704 ≥ 658` is now a verified true upper bound at every power-of-two cap, `DEDUP_PRODUCER_KEY_BYTES = 1152` upper-bounds the ~1100 B per-producer cost, the rustdoc derives the bound term-by-term (no over-claim), and every figure is reconciled (~704 B/entry, ~269 GiB shipped default, ~11 MiB edge-tiny dedup, ~26 MiB total). The reviewer confirmed all rss + CLI tests green, the store-fold 1-vs-2-image discrimination holds (58 < 64 < 90), and clippy-pedantic + fmt clean.

The review's last nit (a `~640` doc straggler) + a pre-existing stale `IN_MEMORY_STORE_IMAGES = 2` doc figure are fixed in 1bb7785. Follow-ups #909 (txn prepared half-message buffer, same RAM class) filed.

PR #921, opened 2026-06-30
fix(server): detect a hung-fsync append-actor wedge and flip /healthz and /readyz to 503 (Closes #862)

### LINKED ISSUE #862: [high] Hung/slow disk wedges the single append actor on inline fsync forever; liveness stays green (no restart) and readyz blocks, with no bounded IO barrier
## The failure mode
The durability barrier runs as a blocking syscall inline on the single append-actor thread, with no deadline or watchdog. `Engine::commit_durability_barrier` (`crates/ironbus-server/src/engine.rs:~5757`) calls `self.log.sync()?`; `StdFile::sync_data/sync_all/write_all_at/read_at` (`crates/ironbus-storage/src/io.rs:1070-1090`) delegate straight to `std` with no timeout/cancellation. The actor is "a single thread that owns the Engine" (`crates/ironbus-server/src/actor.rs:2`).

A permanently-hung fsync (failing eMMC doing long internal retries, a stalled networked/overlay FS, brownout-stuck flash — all realistic on the edge target) wedges that thread forever. In-flight produce handlers block on `reply_rx.recv()`; new produces fill then block the bounded 1024-slot channel (`actor.rs:54`).

The #95 liveness watchdog does NOT cover this: it is ticked only by the accept loop and is deliberately decoupled — "accept loop's own liveness signal never depends on the actor being alive" (`server.rs:63`); "liveness ... must not block on, nor be faulted by, a wedged writer" (`health.rs:295`). So `/healthz` stays **200**. `/readyz` calls `engine.with(|e| e.is_healthy())`, which queues a `Command::Run` behind the wedged fsync and blocks on `reply_rx.recv()` (`actor.rs:649-662`) — `/readyz` **hangs** instead of returning 503. The frozen-writer fail-closed path only triggers when fsync RETURNS an error, never on a hang.

## Consequence
A hung disk wedges the entire produce path (and through-actor reads) indefinitely. Liveness reports healthy, so an orchestrator never restarts the node; readyz blocks, so detection depends solely on the prober's own timeout. On a single-node edge box with no orchestrator, nothing detects or recovers it — a silent, total stall holding all connections.

## Proposed hardening
Run the durability barrier on a dedicated IO worker and have the actor publish a progress beacon; when a barrier exceeds a configurable bound, fail-closed (freeze the writer) and flip BOTH `/healthz` and `/readyz` to 503 so an orchestrator can restart, instead of liveness staying green. At minimum, add an actor-progress watchdog distinct from the accept-loop beacon.

## How to test
Inject a `RandomAccessFile` whose `sync_data` blocks (e.g. parks on a never-signaled condvar). Issue a produce, then assert: the produce never returns, `/readyz` does not respond within its timeout, and `/healthz` still returns 200 (the bug). After the fix, assert `/healthz` flips to 503 once the barrier overruns the bound.

ELares:
## What

Closes #862. The durability barrier runs as a blocking syscall **inline on the single append-actor thread**, with no deadline or watchdog. A permanently-hung fsync (failing eMMC, stalled overlay/networked FS, brownout-stuck flash) wedges that thread forever. The #95 liveness beacon is ticked only by the accept loop and is **deliberately decoupled** from the writer, so `/healthz` stays **200**. `/readyz` queues a `Command::Run` behind the wedged fsync and **hangs**. The frozen-writer fail-closed path only triggers when fsync *returns* an error, never on a hang — so a single-node edge box silently stalls with liveness green and nothing restarts it.

## How (the issue's "at minimum" — an actor-progress watchdog distinct from the accept-loop beacon)

- **`ActorWatchdog`** (in `liveness.rs`, alongside `LivenessBeacon`): the append actor stamps it **busy** at the start of each command batch (which includes the covering `fdatasync`) and **idle** when the batch completes and it returns to blocking on `recv`. A `processing_since` of `0` = IDLE (a quiet broker never false-trips); a non-zero stamp older than the bound = the in-flight batch overran — a genuine wedge.
- The health server reads it via the new **`EngineAccess::actor_watchdog_overran`** — a **non-blocking atomic** that never goes through the actor, so it answers even while the actor is wedged. `/healthz` → 503 on *either* the accept-loop window *or* an actor overrun; `/readyz` checks the watchdog **first**, before `engine.with`, so it answers 503 **promptly** instead of hanging.
- Shared on every `EngineHandle` clone (one per actor); **disabled** (bound 0) until the serve path arms it, so single-node/unconfigured brokers and existing tests are unchanged.

## CLI

`--actor-watchdog-bound-ms` (env `IRONBUS_ACTOR_WATCHDOG_BOUND_MS`), flag > env > default, threaded into `ServeConfig` and the materialized-config line. Default **30 s** (far above any legitimate slow fsync), `0` disables.

## Tests

- `ActorWatchdog` unit tests (idle never trips; overruns only past the bound, strict `>`; `0` disables; a `t=0` stamp is not idle).
- `a_hung_fsync_wedges_the_actor_and_the_watchdog_detects_it`: a **real hung fsync** via the `FaultFs` sync gate proves the watchdog trips through the **live `EngineHandle`** (the production path) without disturbing the actor — which recovers cleanly on un-wedge.
- `healthz_and_readyz_flip_to_503_when_the_append_actor_is_wedged`: an **HTTP end-to-end** test proves `/healthz` *and* `/readyz` flip to 503 (and `/readyz` does not hang) on a wedge, and recover.
- CLI flag parse/resolve test.

`clippy --workspace --all-targets --all-features -D warnings -W clippy::pedantic` clean.

## Follow-up (not in scope, will file)

Run the durability barrier on a **dedicated IO worker** so the actor stays responsive (the issue's fuller hardening). This PR adds the **detection + fail-closed liveness flip**, which is the reported defect.

### DIFF
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -766,6 +766,15 @@ const DEFAULT_VISIBILITY_MS: u64 = 30_000;
 /// DISABLES the watchdog (`/healthz` is then a static 200 while up). 10 s matches the #95 spec.
 const DEFAULT_HEALTH_LIVENESS_WINDOW_MS: u64 = 10_000;
 
+/// The default APPEND-ACTOR WEDGE-WATCHDOG bound for `serve` (#862), in milliseconds: a durability
+/// `fdatasync` that hangs (a failing eMMC, a stalled overlay/networked FS, brownout-stuck flash) blocks
+/// the single append actor forever, and the accept-loop liveness beacon is decoupled (#95) so `/healthz`
+/// would stay green while the whole produce path is wedged and `/readyz` would hang behind the stuck
+/// fsync. Once an in-flight batch overruns this bound, `/healthz` AND `/readyz` flip to 503 so an
+/// orchestrator restarts the node. `0` DISABLES the watchdog. 30 s is far above any legitimate
+/// slow-but-progressing fsync, so it never false-trips a merely slow disk.
+const DEFAULT_ACTOR_WATCHDOG_BOUND_MS: u64 = 30_000;
+
 /// The default graceful-shutdown DRAIN TIMEOUT for `serve` (#637), in milliseconds: on a SIGTERM the
 /// broker flips `/readyz` to 503, then drains in-flight work (flush + checkpoint), and force-exits if
 /// the drain has not finished within this window — so an orchestrated stop (k8s `terminationGracePeriod`)
@@ -2254,6 +2263,8 @@ struct ServeFlags {
     health_addr: Option<String>,
     /// The `/healthz` liveness hysteresis window in ms (#95); `None` falls back to the default.
     health_liveness_window_ms: Option<u64>,
+    /// The append-actor wedge-watchdog bound in ms (#862); `None` falls back to the default. `0` disables.
+    actor_watchdog_bound_ms: Option<u64>,
     /// The fail-closed acknowledgement for a NON-LOOPBACK `--health-addr` (#95): the metrics/health
     /// surface is unauthenticated and unencrypted (TLS/#107 and auth/#106 are not wired), so a
     /// non-loopback bind refuses to start unless the operator sets this. A bare boolean flag.
@@ -2679,6 +2690,10 @@ fn collect_serve_flags(args: &[String]) -> Result<ServeFlags, CliError> {
                 f.health_liveness_window_ms =
                     Some(take_number("--health-liveness-window-ms", args, &mut i)?);
             }
+            "--actor-watchdog-bound-ms" => {
+                f.actor_watchdog_bound_ms =
+                    Some(take_number("--actor-watchdog-bound-ms", args, &mut i)?);
+            }
             // A bare boolean flag (no value): advance ONE token, not two, or the loop spins.
             "--health-allow-public" => {
                 f.health_allow_public = true;
@@ -3040,6 +3055,12 @@ fn parse_serve_flags_with_env_and_reader(
                 f.health_liveness_window_ms,
                 env,
                 DEFAULT_HEALTH_LIVENESS_WINDOW_MS,
+            )?,
+            actor_watchdog_bound_ms: resolve_number(
+                "--actor-watchdog-bound-ms",
+                f.actor_watchdog_bound_ms,
+                env,
+                DEFAULT_ACTOR_WATCHDOG_BOUND_MS,
             )?,
             health_allow_public: resolve_bool("--health-allow-public", f.health_allow_public, env)?,
             // #878: the dedup caps take their default from the PROFILE PRESET (edge-tiny lowers them so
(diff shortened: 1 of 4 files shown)


ELares:
## Addressed the SHOULD-FIX: /readyz is now fully non-blocking

The review correctly flagged that the first cut still read the writer-frozen state through `engine.with(|e| e.is_healthy())` — a job queued on the single append actor. On a **hung** fsync that job blocks forever, so the pre-bound `/readyz` hangs behind the wedged writer, and because the health server is single-threaded the whole health surface wedges with it (even the watchdog's own 503 is never served).

I took **option (b)**: route `/readyz` readiness off the single-writer actor entirely.

- The actor now publishes the writer's live/frozen state (`engine.is_healthy()`) to the `ActorWatchdog` after each batch's covering commit.
- `EngineAccess::writer_appears_healthy` reads that published atomic.
- `/readyz` now sheds in priority order with **three non-blocking atomic reads and no actor round-trip**: `draining` → wedged actor (hung fsync past the bound) → frozen writer (a fsync that *returned* an error) → ready.

The published flag reflects the state as of the last completed batch — exactly when the old blocking read would have observed it — but it can never hang.

### New coverage
- `liveness::the_watchdog_publishes_and_reads_the_writer_frozen_state` — the published-flag primitive.
- `actor::the_actor_publishes_a_frozen_writer_and_readyz_reads_it_without_a_round_trip` — the **real** `run_actor` publish → `EngineHandle::writer_appears_healthy` read path over a fault-fs frozen writer.
- `health::readyz_flips_to_503_on_a_frozen_writer_without_blocking_on_the_actor` — the frozen branch sheds 503 at the HTTP layer while `/healthz` stays 200.
- `health::admin_resilience_reports_frozen_after_an_integrity_freeze` still passes over a real frozen `SharedEngine` (its override reads real state under the uncontended lock).

All 935 server lib tests + CLI suites green; `fmt` + `clippy -D warnings -W clippy::pedantic` clean.

### Deferred (follow-ups, not blockers)
- `/metrics` still goes through the actor and can hang behind a wedged fsync on the single-threaded health server — separate/pre-existing surface; will file a follow-up to move it (and any remaining actor-routed health reads) off the writer or behind a deadline.
- A dedicated IO worker (so a hung fsync parks one thread instead of the whole actor) remains the deeper hardening; tracked separately.

ELares:
## Fresh-eyes review: APPROVE-WITH-NITS (no blocking issues)

A second adversarial fresh-eyes review (isolated worktree pinned to the head SHA) confirmed the SHOULD-FIX is genuinely resolved:
- **Non-blocking claim verified** — the new `/readyz` makes exactly three reads (`draining` Acquire, `actor_watchdog_overran`, `writer_appears_healthy`); for the production `EngineHandle` the latter two are pure atomic loads. The `engine.with(|e| e.is_healthy())` round-trip is gone.
- **Publish→read correctness verified** — `flush_pending` handles a covering-fsync freeze inline (sends `Fatal`, returns normally), so control always reaches `publish_writer_healthy(engine.is_healthy())`; the freeze is a one-way latch, so even a missed publish self-heals next batch.
- **All three new tests discriminate** (each fails if the fix is reverted), including the real `run_actor` publish path test (the `with(|_| ())` "force the publish" reasoning is sound given the strictly-sequential actor loop, not racy).

### Folded in this push (review NIT E/C)
Made the wedge→frozen handoff use **Release/Acquire** instead of Relaxed: `mark_idle`'s store is now `Release` and `overran`'s `processing_since` load is `Acquire`. Since the actor orders `publish_writer_healthy` before `mark_idle`, a `/readyz` reader that sees the cleared wedge is now guaranteed to also see the published frozen flag — closing a one-probe transient false-200 on weakly-ordered arches (aarch64, which we build for). No-op on x86; advisory per-batch stores. A load-bearing comment on the `run_actor` ordering prevents a future reorder.

### Filed as a follow-up (review finding G) — #922
The old `engine.with` Err arm answered 503 "shutting down" on a closed actor channel. The graceful drain is fully covered (draining gate set first) and a mid-batch death trips the wedge bound, but an *unexpected/panic* actor death while idle with the watchdog bound disabled would report 200. Restoring that coverage needs a separate non-blocking actor-alive atomic (not `engine.with`, which would reintroduce the hang) — tracked in #922 with its own panic-path tests.

Re-running CI on `fddf022`; will merge once green.

PR #923, opened 2026-06-30
fix(geo): anchor a multi-origin source on an in-log per-origin HWM so a crash can't double-apply (#906)

### LINKED ISSUE #906: [high] Geo multi-origin SOURCE can double-apply a batch on a crash between record sync and cursor commit (#799 residual)

ELares:
### The bug
A geo multi-origin **SOURCE** applies an origin's pull in two fsyncs: it syncs the validated records to the local fan-in log, then durably advances that origin's cursor (a separate fsync on the `OriginCursorStore`). A crash *between* the two leaves the records durable but the cursor stale → on restart the origin re-pulls from the stale cursor and **re-applies its last batch**, injecting a duplicated block into the fan-in stream (a no-dup / durable-prefix invariant violation).

The single-origin **mirror** was already fixed (#799) by anchoring on the local log's `next_offset`. A multi-origin source has no positional anchor — its `next_offset` is the interleaved fan-in total, not any one origin's offset — so that fix was intentionally off for it (the residual window #906 tracks).

### The fix (Option 1 from the issue — in-log per-origin HWM marker)
Give the source a per-origin durable high-watermark made durable by the **same fsync** as the records:
- After appending an origin's validated frames, and **before** `log.sync()`, the applier appends a **discriminated HWM marker** record — a reserved NUL-prefixed key + payload magic — carrying the full `(origin → applied offset)` snapshot (reusing the existing cursor codec). The marker rides the same single `fdatasync` as the data, which is exactly the atomicity the separately-fsynced cursor lacks.
- On reopen, the per-origin HWM is rebuilt **O(1)** from the log's tail record (every appending batch ends with a marker, and the highest offset is never reaped).
- Apply now anchors the skip on `durable_anchor = max(cursor, hwm)` for a multi-origin source — so a re-pull of an already-synced batch is **skipped**, exactly as the single-origin path already does with `next_offset`.

### Why it's contained / safe
- The marker lives **only** in the local fan-in log (never on the wire) and **only** for a multi-origin source. A single-origin **mirror writes no markers** and keeps its byte-identical log (a new test asserts this).
- A fresh / legacy marker-less log recovers as HWM `0` — the safe degrade to cursor-only (pre-#906 behavior), which self-heals after the first new batch writes a marker.
- The wire format (`MirrorPullResponse`) is untouched. `MirrorApplier::new` stays infallible (lazy rebuild on first apply), so no call-site churn and an open never trips on a transient read.
- Confirmed by design review: the source fan-in log is a physically separate log that **no client consume or chained-serve path reads today** (`spawn_geo_plane` hands it only to the puller threads; consumer/serve reads go to the engine's own `StreamSet`/`read_plane`), so markers don't pollute any consumer stream. (If source-log consume is wired later, that path filters markers by the reserved key + magic.)

### Tests (+3, ironbus-server)
- `a_multi_origin_source_re_pull_after_the_crash_window_skips_not_double_applies` — reproduces the post-crash durable state (records + marker ahead, cursor lagging) the way the #799 test does (log and cursor on independent disks), then re-pulls and asserts `applied == 0` and the fan-in log did not grow. **Verified to FAIL when the anchor is reverted to `cursor`** (the bug reappears: the re-pull re-applies).
- `a_multi_origin_source_recovers_the_per_origin_hwm_from_the_tail_marker` — recovery rebuilds **both** origins' HWMs from the single tail marker.
- `a_mirror_writes_no_hwm_markers` — the single-origin path writes no markers (byte-identity preserved).
- The existing fan-in tests now account for marker records.

938 server lib tests + CLI suites green; `fmt` + `clippy --workspace --all-targets --all-features -- -D warnings -W clippy::pedantic` clean.

### Follow-on guards (not triggered by current wiring; noted for later)
- If `OriginServer::serve_pull` is ever pointed at a source log to feed a downstream mirror, it must skip markers (a raw marker shipped verbatim would be rejected by the downstream's `codec::decode`).
- The O(1) tail recovery depends on "every data-bearing batch ends with a marker" (both sync sites route through one helper); a bounded-scan fallback is the safety net if that ever changes.

### DIFF
--- a/crates/ironbus-server/src/cluster/geo.rs
+++ b/crates/ironbus-server/src/cluster/geo.rs
@@ -851,6 +851,29 @@ impl<F: Filesystem, C: Clock> MirrorApplier<F, C> {
         origin_key: &str,
         resp: &MirrorPullResponse,
     ) -> Result<ApplyOutcome, GeoError> {
+        self.apply_pull_response_inner(origin_key, resp, true)
+    }
+
+    /// TEST SEAM (#906): apply `resp` but STOP after the batch's records and any mid-batch seal markers,
+    /// BEFORE the final marker, the group-commit `log.sync()`, and the cursor commit — modeling a crash in
+    /// the durability window. The mid-batch seals are already durable (each fsynced its segment with a
+    /// marker tail); the active segment's unsynced tail and the cursor are not, so a subsequent
+    /// `simulate_power_loss` leaves exactly the post-crash durable state to test recovery against.
+    #[cfg(test)]
+    fn apply_then_crash_before_final_sync(
+        &mut self,
+        origin_key: &str,
+        resp: &MirrorPullResponse,
+    ) -> Result<ApplyOutcome, GeoError> {
+        self.apply_pull_response_inner(origin_key, resp, false)
+    }
+
+    fn apply_pull_response_inner(
+        &mut self,
+        origin_key: &str,
+        resp: &MirrorPullResponse,
+        finalize: bool,
+    ) -> Result<ApplyOutcome, GeoError> {
         let cursor = self.cursors.cursor(origin_key);
         // An empty run (caught up to the origin's HW) is a clean no-op that still reflects the observed
         // HW. Its `first_offset` may be the cursor OR the origin HW; either way nothing is applied.
@@ -955,10 +978,22 @@ impl<F: Filesystem, C: Clock> MirrorApplier<F, C> {
                 headers: view.headers,
                 payload: view.payload,
             };
-            self.log.append(&append)?;
+            // Append the record, managing segment boundaries so a marker always caps a sealed segment
+            // (#906). `at_origin_offset` is this origin's applied extent BEFORE this record, i.e. the HWM
+            // a sealing marker would record for the records already in the active segment.
+            self.append_origin_record(origin_key, &append, at_origin_offset)?;
             applied += 1;
             seen += 1;
             at += frame_len;
+        }
+        if !finalize {
+            // TEST-ONLY crash point (#906): the records and any mid-batch seal markers are appended (the
+            // seals are durable), but the final marker + group-commit fsync + cursor commit have NOT run.
+            return Ok(ApplyOutcome {
+                applied,
+                cursor: cursor + seen,
+                origin_high_watermark: resp.origin_high_watermark,
+            });
         }
         // Durably commit the applied batch to the local log (one fsync per pull — the group-commit shape).
         // For a multi-origin source that appended records, the per-origin HWM marker is appended FIRST so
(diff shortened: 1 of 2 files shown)


ELares:
## Converting to draft — a fresh-eyes review found a BLOCKING defect in the atomicity premise

A rigorous fresh-eyes review caught a real hole I missed: the "marker rides the SAME fsync as the records" claim **fails across an intra-batch segment roll**.

`Log::append` rolls when the active segment reaches `max_segment_bytes` (log.rs:1838), and `roll()` → `seal()` → `file.sync_all()` (segment.rs:770) is a full, independent fsync. So if a data record (or the marker) mid-batch tips the active segment over the cap, the already-appended data is **sealed + fsynced in a separate durability unit from the marker**. A crash in the seal→final-`sync` window leaves a **data record** as the durable tail → `read_durable_hwm` returns empty → `durable_anchor = cursor` (stale) → the already-durable prefix is re-applied. This reopens the exact #906 double-apply window at every segment-rotation boundary (a normal recurring event when the active segment sits near full and a batch up to `MAX_GEO_PULL_BYTES` straddles a roll).

The review also (correctly) flagged that **no existing test models a lost-unsynced-tail crash** — `InMemoryFs` persists everything, so the roll-straddle hole is invisible to the current tests.

### Rework plan
Guarantee that **every sealed segment of the fan-in log ends with a HWM marker covering exactly that segment's records**, so the durable tail is always a marker no matter where rolls fall:
- Add minimal generic `Log` primitives so geo manages segment boundaries explicitly: append-without-implicit-roll, force-seal-active, and an at-or-over-cap check. (The engine append path is only *factored*, not behavior-changed — conformance/reproducibility gates guard it.)
- Geo's apply: before a data append would overflow the active segment, append the current marker as the segment's sealing tail, then force-seal; the final marker rides the batch's `log.sync()`. Crash before any seal → batch not durable → fresh re-pull; crash after a mid-batch seal → that segment's marker tail recovers the correct HWM → skip.
- Add a **crash test using `InMemoryFs::simulate_power_loss`** that drops the unsynced tail and exercises the roll-straddle case, so a regression is caught.

Holding the merge until the rework lands and passes a re-review.

The non-blocking follow-ups (untrusted-key discriminator hardening; marker payload cap; read-side marker filter if the source log ever becomes client-consumable) will be addressed or filed separately.

ELares:
## Rework pushed — the roll-straddle blocking defect is fixed

Commit `ba03316` closes the blocking finding (a batch straddling a segment roll could leave a data-record durable tail → empty HWM → double-apply).

**The fix: cap EVERY sealed segment of the fan-in log with a HWM marker**, so the durable tail is always a marker no matter where rolls fall.

- `ironbus-storage` (log.rs): factored `append` into `append_inner(record, allow_roll)` — byte-identical (`append` = `append_inner(.., true)`; all 110 log tests pass unchanged) — and added three generic geo-only primitives: `append_without_roll`, `seal_active_segment`, `active_segment_at_or_over_cap`.
- `ironbus-server` (geo.rs): multi-origin apply now manages segment boundaries explicitly via `append_origin_record` — when the active segment is at/over cap it appends the current marker as the segment's **sealing trailer** and force-seals (so the just-fsynced segment ends with a marker covering exactly its records), then the record starts a fresh segment via `append_without_roll`. The final marker rides the batch's `log.sync()`. The single-origin mirror is untouched (plain `log.append`, byte-identical, roll-proof `next_offset`).

**New crash test** `a_multi_origin_re_pull_after_a_roll_straddle_crash_skips_not_double_applies`: origins with big segments (one pull serves many records) feed a small-segment local log so the batch straddles several rolls; it crashes before the final sync via a `#[cfg(test)]` seam, then `InMemoryFs::simulate_power_loss` drops the unsynced tail. Asserts the durable tail is a MARKER with a partial HWM and that each record is applied exactly once. **Verified to FAIL** when `append_origin_record` is reverted to the old implicit-roll / marker-at-end behavior.

This also closes the review's untrusted-discriminator concern (E): the durable tail is now always our own marker, so a malicious origin data record can never be the recovery tail.

939 server lib + 411 storage lib + CLI suites green; fmt + clippy (workspace, pedantic) clean.

Non-blocking follow-ups (review E/I/4) filed as **#924** (clamp recovered HWM ≤ next_offset; document the marker payload cap; add a read-side marker filter if a source's fan-in log ever becomes client-consumable).

PR #925, opened 2026-06-30
perf(storage): cache an opened SegmentReader per sealed segment on both read paths (#808)

### LINKED ISSUE #808: [high] Cache an opened SegmentReader (fd + validated header + len) per sealed segment id; off-actor fetch and Tier-W per-record poll re-do open(2)+fstat(2)+header pread+CRC on every read
## The cost

Every read of a sealed segment re-opens it from scratch. `SegmentReader::open` (crates/ironbus-storage/src/segment.rs:912):

```rust
pub fn open(file: F) -> Result<SegmentReader<F>, StorageError> {
    let file_len = file.len()?;                 // fstat(2) on disk
    if file_len < SEGMENT_HEADER_LEN as u64 { ... }
    let mut hbuf = [0u8; SEGMENT_HEADER_LEN];
    file.read_exact_at(&mut hbuf, 0)?;          // pread(2) of 64 bytes
    let header = SegmentHeader::decode(&hbuf)?; // header CRC check
    ...
}
```

On the disk backend each piece is a real syscall: `StdFs::open` (fs.rs:512) -> `StdFile::open` (io.rs:995) is `OpenOptions::new().read(true).write(true).open()` = `open(2)` (a fresh fd, dropped at end of read); `StdFile::len` (io.rs:1078) is `self.file.metadata()?.len()` = `fstat(2)`; the header read is a `pread(2)`.

The callers re-open per read, on the consume hot path:
- Off-actor sealed plane: `read_plane.rs:290` (`read_range`, inside the per-slot loop) and `read_plane.rs:400` (`read_range_raw`): `let reader = SegmentReader::open(self.fs.open(&segment_file_name(slot.id))?)?;`
- Through-actor Log: `log.rs:2338`/`2378`/`2399` inside `read_slot_into`.
- Tier-W poll delivers ONE message per granted lease via `engine.rs:6383`: `self.log.read_from(off, 1)?` -> `read_range` -> `read_slot_into` -> `SegmentReader::open`. The hot path is explicitly per-record (log.rs:421: "the consume hot path reads `read_from(off, 1)` per record").

The project already caches the costly locate work - the resident sparse seek index keyed by segment id (`segment_indexes`/`compacted_indexes`, log.rs:703-733) - so a read is an O(1) seek + a small forward pread. But it does NOT cache the file handle or the validated header (grep for `reader_cache`/`open_readers`/`fd_cache` is empty). A sealed segment is immutable: its header and length never change, so the open(2)+fstat(2)+pread(64)+header-CRC is pure repeated per-read overhead, plus fd-open/close churn.

## Impact

Throughput / syscall budget. Per delivered Tier-W message (or per small off-actor fetch) the useful work is one small pread (~36-byte frame); the overhead is open(2) + fstat(2) + a 64-byte pread + a header CRC decode - roughly 3 extra syscalls, i.e. ~3-4x syscall amplification per message on the consume hot path. At the per-core consume rates targeted by V2-M1 ("consume beat NATS") this is a large fraction of the per-message syscall cost, plus avoidable fd allocate/free churn and the kernel path-walk of `open(2)` each time. Memory backends are unaffected (cheap open); the win is on the disk backend.

## Proposed optimization

Cache an opened `SegmentReader` (fd + validated header + len) keyed by SEALED segment id, reused across reads.
- Off-actor plane: the `SealedSnapshot` (read_plane.rs:152) is already an immutable, ArcSwap-published value with a per-segment slot list and a reap-driven replacement; it is the natural owner of a per-segment reader map. Open once on first touch, reuse for all subsequent fetches; the whole snapshot is dropped (and its fds closed) when a roll/reap publishes a new one.
- Through-actor Log path: add a per-segment-id reader cache alongside `segment_indexes`, evicted by the SAME `evict_segment_index` retirement hook (reap, force-reap, compaction install) so a recycled segment id never reuses a stale fd.
- Scope to SEALED segments only; the active (un-sealed) segment grows, so keep its current open-and-read (or re-stat) behavior.

This turns N reads over a hot segment into 1 open(2)+fstat(2)+header-CRC and N positioned preads. Correctness is unchanged: same bytes, same offsets, same CRC validation - only the redundant re-open/re-validate is removed.

## How to measure

- Microbench / strace: drive a Tier-W consumer draining one hot sealed segment at 1 msg/poll on the disk backend and count `open`/`fstat`/`pread` syscalls per message (e.g. `strace -f -c` or a perf counter); expect open+fstat to drop from per-message to per-segment.
- Consume throughput bench (the V2-M1 consume harness): measure msgs/s/core for a streaming/Tier-W consumer over sealed segments before/after, disk backend, with the OS page cache warm so the win is the syscall/fd overhead rather than IO.
- Watch fd churn (open file descriptors / open-close rate) to confirm the fd-leak-churn is gone.

Related #491 (moves reads off the actor, but the off-actor plane still re-opens per fetch); distinct from #479 (body-CRC revalidation) and #490 (O(W) cursor rescan).

ELares:
## Closes #808 — cache an opened `SegmentReader` per sealed segment

Every read of a sealed segment re-did `SegmentReader::open` from scratch — `open(2)` + `fstat(2)` + a 64-byte header `pread` + a header-CRC decode — on the consume hot path. A sealed segment is **immutable** (header + length never change), so that's pure repeated per-read overhead (~3–4× syscall amplification per delivered Tier-W message + fd allocate/free churn). This caches an opened reader keyed by sealed segment id, reused across reads, on **both** read paths. Memory backends are unaffected; the win is the disk backend's syscall/fd overhead.

### Off-actor lock-free plane (`read_plane.rs`)
The immutable `SealedSnapshot` gains a `SlotReaders` cache — a `Vec<OnceLock<Arc<SegmentReader>>>` parallel to its `segments`. A read opens its sealed segment on first touch and shares the `Arc<SegmentReader>` across the concurrent reader threads. **No mutex** is added to the lock-free plane: `SegmentReader`'s reads are positioned `pread`s (`&self`, no cursor), and `RandomAccessFile: Send + Sync` ("shared with lock-free readers"), so a shared `Arc<SegmentReader>` is read concurrently safely. The **whole snapshot** (and every fd it opened) drops when a roll/reap republishes a fresh one, so fds never outlive their generation. A double-open race is merely wasteful (immutable file → byte-identical readers; the loser's fd closes immediately), never wrong.

### Through-actor `Log` (`log.rs`)
A FIFO-bounded `SealedReaderCache` (default **256** resident readers) serves the dense seek-and-read hot path — the Tier-W poll reads `read_from(off, 1)` per delivered message. The **active (growing) segment is never cached**; entries are evicted in lockstep with the seek index by `evict_segment_index` at retirement (reap / force-reap / compaction install). Segment ids are never recycled (ADR 0002), so a cached fd can never alias a different physical segment; the FIFO cap bounds resident fds so a retention-off cold full scan reuses a working set instead of leaking one fd per sealed segment.

### Correctness
Unchanged: same bytes, same offsets, same per-record CRC validation — only the redundant re-open/re-validate is removed. The off-actor differential tests (`sealed_read_matches_through_actor_read_from...`, `window_bounded_sealed_read_is_byte_identical...`, the raw-run tests) and the through-actor read tests still pass byte-for-byte, as does the concurrent-readers-and-a-writer race test (now exercising the shared `Arc<SegmentReader>` cache).

### Tests (+4) & instrumentation
- `FaultControl` gains an `open_count` (mirrors the existing `sync_dir_count` pattern).
- `repeated_sealed_reads_open_each_segment_once` (plane) and `repeated_read_from_opens_a_sealed_segment_once` (log): 20–25 reads of one sealed segment open it **once** — both **verified to FAIL (25 != 1 / 20 != 1)** with the cache bypassed.
- `a_republished_snapshot_reopens_the_segment` (plane): a fresh snapshot re-opens, proving the old generation's fd was dropped, not reused.
- `a_force_reap_evicts_the_cached_sealed_reader` (log): a retired segment's cached reader (and fd) is evicted in lockstep with its index.

415 storage lib + 939 server lib + CLI suites green; `fmt` + `clippy --workspace --all-targets --all-features -- -D warnings -W clippy::pedantic` clean.

### Notes / possible follow-ups
- The compacted (sparse) read path and the dense unindexed active-tail fallback are intentionally NOT cached (the issue targets the dense sealed hot path; the active segment grows).
- The Log cache cap is a `const` (256); making it a `LogConfig` knob is a trivial follow-up if an operator ever needs to tune resident fds.
- The plane cache is unbounded per snapshot generation (bounded by "sealed segments touched since the last roll/reap", freed at the next republish) to preserve lock-freedom — an LRU there would reintroduce a lock.

### DIFF
--- a/crates/ironbus-storage/src/fault.rs
+++ b/crates/ironbus-storage/src/fault.rs
@@ -81,6 +81,9 @@ pub struct FaultControl {
     /// (a retire dir-fsync), so the crash window is precisely between the compacted-segment commit
     /// and the originals being durably removed.
     fail_sync_dir_after: Arc<AtomicU64>,
+    /// A monotonic count of every `open` that ran, so a test can assert that a repeated read of one
+    /// sealed segment opens it ONCE (the #808 reader cache), not once per read.
+    open_count: Arc<AtomicU64>,
 }
 
 /// A condvar-backed gate the sync path waits on while closed (#177): `open` starts open (syncs pass),
@@ -161,6 +164,17 @@ impl FaultControl {
 
     fn record_sync_dir(&self) {
         self.sync_dir_count.fetch_add(1, Ordering::SeqCst);
+    }
+
+    /// A monotonic count of every `open` that has run, so a test can assert the #808 reader cache opens a
+    /// sealed segment ONCE across many reads (and re-opens after a roll/reap drops the old snapshot).
+    #[must_use]
+    pub fn open_count(&self) -> u64 {
+        self.open_count.load(Ordering::SeqCst)
+    }
+
+    fn record_open(&self) {
+        self.open_count.fetch_add(1, Ordering::SeqCst);
     }
 
     /// Arms (or disarms) short reads: while `limit` is non-zero, every `read_at` returns at most
@@ -417,8 +431,11 @@ impl<F: Filesystem> Filesystem for FaultFs<F> {
     type File = FaultFile<F::File>;
 
     fn open(&self, name: &str) -> io::Result<Self::File> {
+        let file = self.inner.open(name)?;
+        // Count only a SUCCESSFUL open, so an open-count assertion measures real fd opens (#808).
+        self.control.record_open();
         Ok(FaultFile {
-            inner: self.inner.open(name)?,
+            inner: file,
             control: self.control.clone(),
         })
     }
--- a/crates/ironbus-storage/src/log.rs
+++ b/crates/ironbus-storage/src/log.rs
@@ -596,6 +596,75 @@ impl CompactedIndex {
     }
 }
 
+/// The default number of opened SEALED-segment readers the through-actor `Log` keeps resident (#808).
+/// Each entry is one fd + a validated header; the cache is FIFO-bounded so a retention-off cold full
+/// scan opens-and-reuses a working set instead of leaking one fd per sealed segment. 256 sits well under
+/// a typical 1024 `RLIMIT_NOFILE`, leaving room for the active writer fd plus sockets.
+const DEFAULT_SEALED_READER_CACHE_CAP: usize = 256;
+
+/// A small FIFO-bounded cache of opened [`SegmentReader`]s keyed by SEALED segment id (#808), used by the
+/// through-actor `Log` read path so a repeated `read_from` over a hot sealed segment does ONE
+/// open+fstat+header-CRC, not one per read. The active (growing) segment is NEVER cached. Interior
+/// mutability (the read path is `&self`), single-writer / never aliased across threads — the same
+/// rationale as `segment_indexes`. Holds only derived data: dropping or rebuilding an entry changes
+/// nothing a reader observes (same bytes, same per-record CRC), it only re-pays the open. Entries are
+/// evicted at segment retirement via [`Log::evict_segment_index`]; segment ids are never recycled
+/// (ADR 0002), so a cached fd can never alias a different physical segment.
+struct SealedReaderCache<F: Filesystem> {
+    readers:
+        std::cell::RefCell<std::collections::HashMap<u64, std::sync::Arc<SegmentReader<F::File>>>>,
+    /// Insertion order of the resident ids, so the cache sheds the OLDEST opened reader when it hits its
+    /// cap — a bound on resident fds, not a correctness device (any victim is safe to re-open).
+    fifo: std::cell::RefCell<std::collections::VecDeque<u64>>,
+    cap: usize,
+}
+
+impl<F: Filesystem> SealedReaderCache<F> {
+    fn new(cap: usize) -> SealedReaderCache<F> {
+        SealedReaderCache {
+            readers: std::cell::RefCell::new(std::collections::HashMap::new()),
+            fifo: std::cell::RefCell::new(std::collections::VecDeque::new()),
+            cap,
+        }
+    }
+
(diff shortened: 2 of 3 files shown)


ELares:
## Addressed the SHOULD-FIX (unbounded plane cache) + dropped the Clone footgun

Commit `2481ebc`:

**Plane cache is now FD-bounded per generation.** `SlotReaders` carries a soft cap (`DEFAULT_PLANE_READER_CACHE_CAP = 256`, mirroring the Log path) + a lock-free `opened` counter. A read past the cap opens its segment **uncached** (per-read, exactly as before #808), so a retention-off cold full backfill can never pin one fd per sealed segment — closing the EMFILE exposure and the peak-fd regression vs pre-PR. The cap is soft (concurrent openers may exceed by the racing-thread count); correctness is unchanged (an over-cap read returns the same bytes, just uncached). New test `the_plane_reader_cache_is_fd_bounded_per_generation` asserts ≤ cap readers stay resident across a full multi-segment prefix scan.

**Dropped `#[derive(Clone)]` on `SealedSnapshot` (NIT 2).** The publish path always builds a fresh snapshot, so `Clone` was unused — and a clone would have aliased opened fds across generations. Removing it deletes the footgun (and confirms it was never used: the workspace still compiles).

NIT 3 (FIFO vs LRU on the Log cache) left as-is — correctness-safe, a future perf knob.

416 storage lib + workspace clippy (pedantic) + fmt clean; re-running CI.

PR #927, opened 2026-06-30
perf(cluster): apply follower fetches off the global data-plane lock (#809, Phase 2)

### LINKED ISSUE #809: [high] Single node-wide Arc&lt;Mutex&lt;DataPlaneServer&gt;&gt; serializes all leader fetch-serves, follower applies (held across fsync), and produce-ack parks across every partition and role
## The cost

The whole data plane on a node funnels through ONE lock: `server: Arc<Mutex<DataPlaneServer<F, C>>>` (`crates/ironbus-server/src/cluster/serve.rs:752`). `DataPlaneServer` (serve.rs:439-448) wraps the seam/controller whose `roles` map is keyed by partition, so a single lock spans **every partition and both leader+follower roles**. Three independent hot paths contend on it:

**(1) Leader fetch-serve — lock held across an O(fetch_bytes) memcpy.** The leader spawns one reader thread per follower link:
```rust
// serve.rs:1049-1053
std::thread::Builder::new().name("ib-dataplane-read"...).spawn(move || {
    run_dataplane_reader(DataPlaneLink::new(stream), &server, &release, &sd);
});
```
Each reader serves a FetchRecords under the lock:
```rust
// serve.rs:1110-1113
let Ok(mut srv) = server.lock() else { return; };
srv.handle_frame(partition, frame).ok()
```
`handle_frame` -> `serve_fetch` (`dataplane.rs:588`, takes only `&self`) -> `ReadPlaneLeader::serve_fetch`, which copies the whole batch under the lock:
```rust
// replication.rs:619-627
let sealed = self.plane.read_range_raw(from, max_records, Some(req_bytes as usize))?;
Ok(FetchResponseBody { ..., frame_bytes: sealed.run.bytes.to_vec() })  // up to MAX_REPL_FETCH_BYTES
```
So K followers => K threads, each serialized on the global lock across a multi-MiB `to_vec` + read.

**(2) Follower apply — global lock held across fsync.** The follower fetch loop applies under the same lock:
```rust
// serve.rs:1306-1316
let Ok(mut srv) = server.lock() else { return; };
... srv.seam_mut().controller_mut().apply_fetch_response(partition, &resp) ...
```
`apply_fetch_response` (replication.rs:790-821) decodes + re-appends every frame and then calls `self.log.sync()` (fsync) — all under the global lock. A follower of N partitions therefore serializes each partition's fsync behind the others: partition A's fsync blocks partition B's serve/apply.

**(3) Per-produce ack park.** Every C2-fsync clustered produce parks its ack under the same lock:
```rust
// client_ack.rs:303
let Ok(mut srv) = self.server.lock() else { ... };
```

The read plane it serves from is wait-free (`read_plane.rs:518` ArcSwap, Acquire load), so this coarse lock throws away the off-actor lock-free design.

## Impact

Caps per-node replication throughput at single-lock width regardless of follower count or partition count. On a K-follower leader the fan-out serving is serial, each critical section spanning a batch memcpy; on a multi-partition follower the per-partition fsyncs (which can be milliseconds) are serialized cross-partition under one lock. Throughput (replication MB/s) and tail latency (head-of-line behind another partition's fsync) both suffer; severity scales with K and partition count.

## Proposed optimization

1. Never hold the lock across the heavy work. For fetch-serve, take the lock only long enough to clone the `Arc<ReadPlane>` for the partition (or look up the leader role), drop it, then run `read_range_raw` + `to_vec` off-lock — the read plane is already lock-free. For follower apply, drop the lock before `log.sync()`.
2. Shard the `roles`/seam map per partition (`DashMap`-style shards, or `Arc<RwLock>`/`Arc<Mutex>` per partition) so different partitions never contend.
3. Split the read-only serve path (`serve_fetch`, `&self`) from the mutating apply/report path, so concurrent fetch serves to different followers proceed in parallel (RwLock read side, or serve straight off `Arc<ReadPlane>` with no controller lock).

Pre-adoption: no back-compat constraints; the wire frames are unchanged, this is purely the in-process locking.

## How to measure

- Microbench a leader with K=1,4,8 follower links each issuing back-to-back FetchRecords at MAX_REPL_FETCH_BYTES; measure aggregate replication MB/s and per-fetch serve latency. Today it stays flat as K grows; sharded/off-lock should scale with cores.
- On a follower of N partitions, measure per-partition apply+fsync p99 while all N catch up concurrently; the cross-partition head-of-line gap (A's fsync stalling B) disappears once the lock is not held across `log.sync()`.
- Add a contention counter / `parking_lot`-style wait-time histogram on the lock and confirm wait time falls to near zero after sharding.

ELares:
## #809 Phase 2 — apply follower fetches off the global data-plane lock

After Phase 1 off-locked the leader fetch-**serve**, the follower fetch loop still held the node-global `Arc<Mutex<DataPlaneServer>>` across `apply_fetch_response` — decode + append + `log.sync()` (**fsync**) — so partition A's apply/fsync serialized behind the one global lock, blocking partition B's serve/apply/park. This was flagged as potentially the bigger win (fsync-dominated contention).

### What changed
- `PartitionRole::Follower` now holds the follower as an `Arc<Mutex<EpochAwareFollower>>` (a `FollowerHandle`).
- The follower fetch thread clones the handle **once** (under a brief global lock) and then applies + reports via the new `apply_on_follower` / `report_from_follower` **off the global lock** — holding only that handle's per-partition lock across the fsync. So A's fsync no longer blocks B. The per-iteration `make_fetch_request` still goes through the controller (brief) and exits the loop once the role is gone.
- The controller's follower-routing methods lock the handle internally; `follower_log` became a closure accessor `with_follower_log` (a `&Log` can't be borrowed across the lock).

### The failover sharp edge (handled carefully)
`promote_follower_to_leader` must **not** consume the `Arc` (a still-running fetch thread holds a clone — `try_unwrap` would race). It already **drops** the log after building the read plane (the leader serves through the plane only), so it now builds the plane + seeds a **clone** of the epoch cache from the **locked** `&log`, leaving the `Arc` for the exiting fetch thread to drop. Writer release is deferred to that exit — harmless: nothing else writes the replica log, and the leader serves only up to the committed HW. The committed-completeness self-verify + epoch fence are unchanged. A poisoned follower lock is recovered with `into_inner` so one partition's apply panic can't wedge the whole plane.

### Correctness
Unchanged: same bytes, same order, same I2 (ack-after-fsync), same failover self-verify + fence. **All 358 cluster tests** (replication / ISR / failover / byte-identity) pass.

### Test
`a_follower_apply_does_not_hold_the_server_lock_across_the_fsync` parks a follower apply mid-fsync (the fault sync gate) and asserts the server `Mutex` is **FREE** while the per-partition follower lock **IS held** — proving the apply runs under the partition lock, off the global lock.

941 server lib + 416 storage lib green; workspace clippy (pedantic) + fmt clean.

### Follow-up (Phase 3, to fully close #809)
Factor the seam's node-global `parked`/`next_token` (atomic token + per-partition parked) to remove the last global serialization on the produce-ack park path. Tracked under #809.

Refs #809

### DIFF
--- a/crates/ironbus-server/src/cluster/dataplane.rs
+++ b/crates/ironbus-server/src/cluster/dataplane.rs
@@ -459,8 +459,19 @@ pub type AckToken = u64;
 /// recovered with `into_inner` everywhere (a follower-apply panic must not wedge the whole data plane).
 pub(crate) type FollowerHandle<F, C> = Arc<Mutex<EpochAwareFollower<F, C>>>;
 
-/// Lock a follower handle, recovering from a poisoned lock (`into_inner`) so a panic in one partition's
+/// Lock a follower handle, recovering from a poisoned lock (`into_inner`) so a panic in ONE partition's
 /// apply never wedges the whole data plane (#809).
+///
+/// This is a deliberate scope change from the pre-#809 global lock, which failed CLOSED on a poison (the
+/// fetch loop's `let Ok(..) = server.lock() else return` tore down on a poisoned global). Recovering a
+/// poisoned PER-PARTITION follower is sound here: `apply_fetch_response` returns a typed `Err` for every
+/// EXPECTED fault (corrupt / non-contiguous / storage) — a panic is therefore a latent bug, not
+/// adversarial input — and the follower's durable state is never left half-applied in a way a later read
+/// could serve as committed: `next_offset` advances only AFTER `Log::append` returns `Ok` (a torn append
+/// errors rather than advancing), and follower reads clamp to `follower_safe_watermark(plane.flushed(),
+/// known_committed_hw)`, so any torn UNCOMMITTED tail is never served below the committed bar. (If a
+/// poison ever needs stricter handling, fail closed by dropping the follower and forcing a clean
+/// re-fetch — tracked as a #809 follow-up.)
 fn lock_follower<F: Filesystem, C: Clock>(
     handle: &FollowerHandle<F, C>,
 ) -> std::sync::MutexGuard<'_, EpochAwareFollower<F, C>> {
@@ -1217,9 +1228,18 @@ impl<F: Filesystem + Clone, C: Clock> DataPlaneController<F, C> {
         // from `&log`) and the learned epoch history (cloned), never ownership of the log. On a verify
         // failure the follower role is restored unchanged. The `Arc` then drops here; the follower's
         // replica-log writer is released once the last clone (an exiting fetch thread's) drops — the same
-        // "serve through the read plane only" outcome as the old `into_log()`+`drop(log)`, only the
-        // writer release is deferred to the thread's exit instead of being immediate (harmless: nothing
-        // else writes this replica log, and the leader serves only up to the committed HW).
+        // "serve through the read plane only" outcome as the old `into_log()`+`drop(log)`, only the writer
+        // release is deferred to the thread's exit instead of being immediate.
+        //
+        // SAFETY of the deferred release: only ONE writer ever touches this replica log (the leader's
+        // append actor is not yet wired, so nothing else writes it), and the follower lock serializes a
+        // racing post-promotion apply against this promotion. Such an apply CAN still advance the leader's
+        // served frontier (`plane.flushed()`), but it only appends a byte-identical, CRC-revalidated,
+        // CONTIGUOUS continuation of the same authoritative old-leader lineage ABOVE the promotion
+        // frontier — adopting more of the uncommitted tail under the new epoch, never overwriting
+        // committed data and keeping the epoch fence at `frontier` consistent. When the append actor IS
+        // wired to write this log, add an explicit single-writer handoff (wait for the fetch thread to
+        // exit / the `Arc` to drop before opening the writer) — tracked as a #809 follow-up.
         let plane_epochs = {
             let guard = lock_follower(&follower);
             let log = guard.follower().log();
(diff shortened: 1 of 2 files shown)


ELares:
## Fresh-eyes review: APPROVE-WITH-NITS (no blocking issues)

A deep concurrency review (isolated worktree, 9 dimensions) confirmed the change is sound:
- **No deadlock** — strict global→follower lock ordering, structurally impossible to invert (the controller is nested inside the server and has no handle to the global Mutex; `apply_on_follower`/`report_from_follower` take only the follower lock).
- **Failover/promotion is safe** — the post-promotion apply window is benign: it appends only a byte-identical, CRC-revalidated, contiguous continuation above the promotion frontier (never overwriting committed data), with the epoch fence consistent. Only one writer touches the replica log today (the leader append actor isn't wired).
- **Win is real** — the apply's `log.sync()` runs holding only the per-partition follower lock; the new test discriminates (server lock free + handle lock held during the parked fsync).
- All 359 cluster tests pass incl. the real-thread 3-node failover capstones.

### Addressed (doc-only, commit db14d75)
- **(SHOULD-FIX, poison recovery)** Documented inline why recovering a poisoned per-partition follower (vs the old fail-closed global poison) is sound: apply returns typed errors for expected faults, `next_offset` advances only after a clean append, and follower reads clamp to the committed bar.
- **(comment nit)** Tightened the promotion comment — a racing post-promotion apply *can* advance `plane.flushed()`, but only with a contiguous continuation above the frontier; added a note to add an explicit single-writer handoff when the append actor is wired.

### Left as tracked follow-ups (under #809)
- A future single-writer handoff gate for when the leader append actor writes the replica log (B-ii/iii).
- The `&mut self`→`&self` tightening on a few now-interior-mutable methods (harmless NIT).

Re-running CI on the doc commit; will merge once green.

PR #1010, opened 2026-07-01
fix(actor): close the residual shutdown reply race + correct wait doc (#949)

### LINKED ISSUE #949: [low] Shutdown drain has a residual post-drain race + a stale ProduceSubmission::wait doc comment (#802 follow-up)
Follow-up from #802 (#948), which drains and replies to commands queued at Shutdown so a racing producer no longer wedges on the reported mid-drain-abandonment path.

Two residual items the #802 review surfaced:

1. **Narrow residual race.** A `Command::Produce` whose `tx.send()` lands AFTER the shutdown drain's final `rx.try_recv()` returns Empty but BEFORE `run_actor` returns and drops the command `Receiver` is buffered reply-less and then dropped; the producer still wedges because the submission's co-located `channel.tx` keeps the reply channel open, so `recv()` blocks instead of returning `ActorGone`. The window is microseconds but real. Fully closing it needs a small redesign of the submission reply mechanism so actor-departure is detectable despite the co-located tx (e.g. an `actor_alive` flag the producer observes, or a drop-guard that closes pending reply channels on actor exit) — rather than draining until the command channel closes (it never does, since producers hold tx clones).

2. **Stale/now-false doc comment.** `ProduceSubmission::wait` (actor.rs ~280-284) still says `ActorGone` is returned when the actor drops its cloned tx un-sent, 'which closes the channel'. That is false on the produce path precisely because the co-located `channel.tx` keeps the channel open — it is the root cause of item 1. Fix the comment together with the redesign so it describes reality.

ELares:
[low/follow-up] #949: self-filed follow-up from the milestone-#30 audit pass; implements the fix the issue describes at the real production site, mutation-verified where behavioral, whole-workspace green. Reviewed across 3 adversarial lenses.

Closes #949

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/ironbus-server/src/actor.rs
+++ b/crates/ironbus-server/src/actor.rs
@@ -59,8 +59,10 @@ use ironbus_core::types::Offset;
 use ironbus_storage::fs::Filesystem;
 use ironbus_storage::log::Append;
 use std::collections::VecDeque;
+use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::{sync_channel, Receiver, RecvTimeoutError, SyncSender, TryRecvError};
 use std::sync::{Arc, Mutex, OnceLock};
+use std::time::Duration;
 
 use crate::cluster::client_ack::{ClientAckGate, ClusterProduceRouting};
 use crate::cluster::dataplane::{AckDisposition, FollowerReadOutcome};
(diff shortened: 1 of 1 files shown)


ELares:
Rebased onto today's main (`7f03ec0`) — the original branch predated the pipelined-sync-tier era and had gone `mergeable: dirty`. The rebase is a semantic re-application, not mechanical:
- **Composes with the #1032 spin tier**: `recv_spin_then_park`'s fallback park is now the same #949 sliced wait, so the spin tier detects a departed actor too — its old bare-`recv` fallback had the *identical* wedge (the code even noted the disconnect was "unreachable in practice", i.e. it would hang forever).
- **Threads `actor_alive` through `try_take`** (#1045's destructure/rebuild) and `pending_for_test`.
- **Adds a third regression test** pinning the spin-tier composition (hangs if the fallback reverts to `recv`).

Verified: 67 actor-module tests + full `ironbus-server` lib suite (1,056 passed), clippy `--all-targets` and fmt clean.

Issue #1023, opened 2026-07-02
Cross-broker benchmark study on macOS (M4 Pro): IronBus vs Kafka vs Redpanda vs NATS — reproducible harness + results (feeds #769)


ELares:
A reproducible, matched-durability cross-broker study on a single Mac (Apple M4 Pro, 14 cores, 48GB, macOS 26.5), extending the in-repo rig discipline (comparison.rs durability labels, EDGE_RUN_DISCIPLINE.md warmup/steady-state/CoV protocol, coordinated-omission-aware latency).

- **Peers**: Kafka 4.3.1 (KRaft, Temurin 21), Redpanda v26.1.12 (Linux-only → lima vz Ubuntu VM, 8 vCPU/8GB — VM + port-forward overhead DISCLOSED), NATS 2.14.3 (JetStream file/memory + core), IronBus @ current main (release).
- **Matrix**: P1 sync-per-message durable produce, P2 group-commit durable produce, P3 relaxed/page-cache produce, P4 in-memory produce, C1 durable consume/replay, L1 durable single-inflight ack latency (p50/p99), L2 in-memory latency; 128B + 1KiB; pilot→frozen-count→3 timed runs→median; fresh data dirs; serial single-broker execution; pinned durability knobs per cell (Kafka flush.messages/acks; Redpanda developer_mode OFF + write_caching per tier; NATS sync_interval always vs default; IronBus --durability-level).
- **Structural-gap honesty**: NATS JS has no ack-after-fsync group-commit mode (P2 cell marked); Kafka/Redpanda have no ephemeral in-memory mode (P4 excluded); Kafka/Redpanda are cluster-class systems benched here in single-node mode.
- **Deliverables**: the harness scripts + normalized results JSONL + a written report under docs/benchmarks/; the headline table goes to the README. Any cell where IronBus does not lead spawns an optimization work item.

Refs #769 (the OpenMessaging-driver epic stays the third-party-recognizable follow-up; this issue is the Mac-local head-to-head).

ELares:
[...]

PR #1025, opened 2026-07-02
feat(cli/1024): report produce-ack RTT percentiles for awaited window=1 publishes

### LINKED ISSUE #1024: bench: report produce-ACK latency percentiles (p50/p99/p999) for awaited window=1 publishes, incl. memory backend
Needed by the cross-broker study #1023: the L-row peers (Kafka producer-perf, NATS `js pub sync`) report produce→ACK RTT percentiles, but `ironbus bench` today cannot produce the symmetric number:

- `--mode publish --pubwindow 1` measures exactly the awaited per-produce ack RTT (`produce_one` durations in `fsync_samples`) but surfaces only the p50, as `fsync_cost_us`, and ONLY on the disk backend (`fsync_is_measured`).
- `--mode round-trip` reports full percentiles but of the producer→consumer e2e delivery latency — a strictly harder metric than the peers' produce-ack RTT (unfair asymmetry in a cross-broker table).

**Change (bench tool only, zero broker-path risk):** add optional `ack_p50_us / ack_p99_us / ack_p999_us / ack_max_us` to `BenchReport`, populated ONLY on the paths where every produce is individually awaited (plain publish with `--pubwindow 1` — not FaF, not autopipe, not stream/windowed where per-op attribution is amortized and would be dishonest — and the round-trip producer leg, which is also awaited-per-produce). Populate them regardless of storage backend (an ack RTT is honest for `--storage memory` too; it is just not a durable-write cost). `fsync_cost_us` semantics unchanged (disk-only, p50). Additive JSON fields; the ironbus.cli.v1 envelope is versioned — confirm the frozen-envelope test treats data-payload field additions as compatible (adjust the frozen vector if it is a full-payload pin).

Acceptance: `ironbus bench --mode publish --pubwindow 1 --json` (disk AND memory) emits ack percentiles; window>1/FaF/autopipe/stream emit null; unit test pins the awaited-only gating; fmt/clippy(pedantic)/tests green.

ELares:
Implements #1024 (needed by the cross-broker study #1023): `ironbus bench` now reports **produce→ACK RTT percentiles** (`ack_p50_us/ack_p99_us/ack_p999_us/ack_max_us`) as additive, null-when-unmeasured JSON fields + matching human-view lines.

**Honest-attribution gating** (the core of the change): populated ONLY where every produce is individually awaited — plain `--mode publish --pubwindow 1` (ANY storage backend, memory included: an ack RTT is honest there, it's just not a durable-write cost) and the round-trip producer leg. Never on window>1 / `--stream` / `--autopipe` / fire-and-forget (amortized attribution stays honestly absent). `fsync_cost_us` byte-identical semantics (disk-only p50). Gate exposed as `BenchConfig::publish_acks_are_awaited`; `finish_report` takes `SampleAttribution{fsync_measured, acks_awaited}` (orthogonal flags, all six call sites updated).

**Envelope compat**: the frozen ironbus.cli.v1 envelope pins only the envelope shape, not the bench payload; the bench payload is its own versioned object (schema_version 1) and the four fields follow the documented additive precedent (#445/#439) — always-present keys, null when unmeasured. No vector regeneration needed. In-repo JSON consumers parse by key and are unaffected.

Mutation-verified (window==1 → >=1 gate mutation killed by two independent tests); 8 new/updated tests incl. live disk+memory integration; fmt/clippy(pedantic)/450 unit + 9 integration suites green. Reviewed across 2 adversarial lenses (honesty + compat).

Closes #1024

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/ironbus-cli/src/bench.rs
+++ b/crates/ironbus-cli/src/bench.rs
@@ -330,6 +330,20 @@ impl BenchConfig {
     pub fn fsync_is_measured(&self) -> bool {
         !self.no_fsync && self.storage == StorageArg::Disk
     }
+
+    /// Whether PUBLISH mode awaits EVERY produce individually (#1024): the plain half-duplex path,
+    /// `--pubwindow 1` without `--stream`, `--autopipe`, or `--faf`. Only then is each per-produce
+    /// sample an honest produce-to-ack RTT, the number the cross-broker study compares against a
+    /// Kafka/NATS synchronous publish. The pipelined paths (window > 1, stream, autopipe) amortize
+    /// a flush across many messages, so a per-op share would be dishonest, and fire-and-forget
+    /// awaits nothing at all. INDEPENDENT of the storage backend: an ack RTT is honest on the
+    /// memory engine too (it is just not a durable-write cost), unlike [`Self::fsync_is_measured`].
+    /// Unix-gated callers, exactly like [`Self::fsync_is_measured`].
+    #[cfg_attr(not(unix), allow(dead_code))]
+    #[must_use]
+    pub fn publish_acks_are_awaited(&self) -> bool {
+        self.pub_window == 1 && !self.stream && !self.auto_pipeline && !self.fire_and_forget
+    }
 }
 
 /// A bench run's measured result, independent of how it is rendered. The latency fields are present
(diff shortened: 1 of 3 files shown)


ELares (commented):
**Adversarial 2-lens review (honesty + compat) — verdicts: SHIP_WITH_NITS / SHIP_WITH_NITS.**

Verified:
- Gating honest at every call site: plain window=1 publish passes `publish_acks_are_awaited()`; windowed/stream/autopipe/faf/subscribe all pass `acks_awaited=false` (stream/faf also pass empty samples); `--stream --pubwindow 1` is rejected at parse so there is no reachable awaited-but-unclaimed edge.
- Memory-backend population correct + honestly labeled (ack RTT, never fsync cost): integration test pins ack fields populated with `fsync_cost_us: null` on `--storage memory`.
- `fsync_cost_us` semantics byte-identical (same boolean per call site through the `SampleAttribution` refactor; still median of the same samples; disk-only).
- Frozen ironbus.cli.v1 envelope intact (envelope pins shape only; bench payload is its own schema_version-1 object; additive per the #445/#439 precedent; all 9 envelope snapshot tests pass unchanged; no vector regeneration).
- Mutation check: `pub_window == 1` → `>= 1` killed by two independent tests.

**Nits (acknowledged, not blocking — addressed by explanation):**
1. `produce_one` counts shed responses as samples — **pre-existing** behavior inherited from the fsync-cost sampling (not introduced here); still a genuinely per-op-awaited RTT. Fine to leave; would be a separate change if we ever want shed-exclusion.
2. Round-trip leg passes `acks_awaited: true` as a literal — structurally correct today (that producer thread awaits every `produce_one` unconditionally); if a windowed round-trip producer is ever added, the compiler won't force revisiting this literal. Accepted as-is; the doc comment on the leg states the invariant.

Issue #1044, opened 2026-07-03
design(gated): O_DIRECT + broker-owned cache evaluation — package deal with the tail ring; forfeits sendfile (#658)


ELares:
From the Redpanda analysis (A3): redpanda's storage engine is O_DIRECT DMA with aligned chunk pools and its OWN cache replacing the page cache. The analysis consensus (3 of 5 lenses explicitly skeptical): for single-stream fsync-per-batch workloads, buffered write + fdatasync is NEARLY EQUIVALENT on the median; O_DIRECT's wins are copy elimination + writeback-jitter immunity at high aggregate MB/s — a p99/CPU-headroom mechanism, unproven for our rows (no reader produced a measurement).

PACKAGE-DEAL WARNING: O_DIRECT forfeits free page-cache reads, so the dirty-pinned tail cache (#1041) is a hard prerequisite; and it structurally forecloses the sendfile fast path (#658) that buffered IO keeps open — an asymmetric advantage IronBus would GIVE UP.

GATE: evaluate only after the #1040/#1041 mechanisms land AND a Linux profile shows page-cache copies or writeback jitter as an actual bottleneck. Filed for the record from the 2026-07-03 matched-VM study.

PR #1055, opened 2026-07-04
feat(bench): --io-mode flag to select the spawned broker's durable-write mode (#1044)

ELares:
## What

Exposes the merged durable-write **io-mode** (#1054) on the benchmark harness: `ironbus bench --io-mode <buffered|direct|auto>` selects the durable-write path of bench's own isolated spawned broker. Default `buffered` (unchanged behavior). This is the last piece of the io-mode tunability surface — it lets the benchmark measure the `O_DIRECT` metadata-free barrier (`direct`) or auto-detect the substrate (`auto`), exactly like the real `serve` path.

## Why

The io-mode (#1054) is the environment-aware durable-write tunable. On **t4g / EBS**, `--io-mode auto` resolves to `direct` (network-durable substrate detected) and delivers the metadata-free barrier: an authoritative same-box run shows IronBus **winning or on par with Redpanda on every row** — P1 1.7–2.1×, P2 single-conn par-to-4×, P2 multi-conn 5×, C1 1.9×, L1 p99 130×. Without a bench flag, that mode was unreachable from the harness.

## Changes

- `BenchConfig.io_mode` field + `--io-mode` parse arm (`IoMode::parse`, typed usage error on a bad value).
- Threaded into `bench_serve_config` (`config.io_mode = cfg.io_mode`) so the spawned broker's `open_disk_engine` resolves it (auto-detect + fail-closed), same path as `serve`.
- **Guard:** like `--storage`, a non-default `--io-mode` is refused alongside `--addr` (a live broker's io-mode is already fixed; the flag would silently no-op). Makes the field's doc claim true.
- Doc backtick fix on the existing `O_DIRECT` mention (clippy `-D warnings`).

## Verification

- `cargo build` + `cargo clippy --all-targets -D warnings` clean on the CLI crate.
- All three modes parse and run (`buffered`/`direct`/`auto` → ~257 msg/s on macOS, where direct/auto correctly fall back to buffered — no `F_FULLFSYNC` barrier equivalent).
- Guard verified: `--io-mode direct --addr … --i-understand-this-is-live` → typed usage refusal.

### DIFF
--- a/crates/ironbus-cli/src/bench.rs
+++ b/crates/ironbus-cli/src/bench.rs
@@ -310,6 +310,15 @@ pub struct BenchConfig {
     /// disk numbers). The flag shapes ONLY the spawned broker, so it is refused alongside
     /// `--addr` (a live broker's backend is already decided by that broker).
     pub storage: StorageArg,
+    /// The durable-write IO-MODE of bench's own ISOLATED spawned broker (#1044): `buffered` (the
+    /// default, buffered write + fdatasync), `direct` (`O_DIRECT` + metadata-free barrier — the fast
+    /// path on network-durable storage like EBS), or `auto` (detect the substrate). Shapes ONLY the
+    /// spawned broker, so it is refused alongside `--addr`. Disk-only (memory has no fsync path).
+    /// Read only in `bench_run.rs` behind `cfg(unix)` (the bench broker is the Unix-only `serve`
+    /// path), so on a non-unix build this field is write-only and `-D warnings` would refuse it —
+    /// exactly the footgun the `#[cfg_attr(not(unix), allow(dead_code))]` on the methods above guards.
+    #[cfg_attr(not(unix), allow(dead_code))]
+    pub io_mode: ironbus_storage::substrate::IoMode,
     /// How the SUBSCRIBE drain settles each fetched batch (#464): [`AckMode::Batched`] (the default,
     /// one pipelined `ack_many` per batch — the FAIR consume number) or [`AckMode::PerMessage`] (the
     /// legacy one synchronous `ack` per message — the ack-RPC-LATENCY ceiling, behind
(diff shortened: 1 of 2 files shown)


ELares (commented):
Self-review:

- **Scope is minimal and additive** — one new `BenchConfig` field, one parse arm, one line in `bench_serve_config`, one guard. No existing path changes value: the default is `Buffered`, byte-identical to pre-flag behavior.
- **Guard correctness:** the `--io-mode`+`--addr` refusal fires *after* the top-level `--addr` live guard (needs `--i-understand-this-is-live` to reach it), mirroring `--storage` exactly — verified both orders manually. Chose a value-based check (`!= Buffered`) rather than an `Option` because `buffered`+`--addr` is a harmless no-op (it's the default anyway); only a *meaningful* mode with `--addr` is the footgun worth refusing.
- **Memory backend:** ignores io_mode (no fsync path) — consistent with how `--storage memory` already short-circuits the durable path.
- **Resolution path:** delegates entirely to `open_disk_engine` (auto-detect + fail-closed), so bench and `serve` can never diverge on how a mode resolves on a given substrate.
- **One risk considered:** `direct` on a substrate without O_DIRECT support falls back to buffered inside `open_disk_engine` (fail-closed), so `--io-mode direct` never hard-errors a bench run on an unsupported FS — it just measures buffered (observed on macOS: direct/auto ≈ buffered ≈ 257 msg/s). That's the intended safe degradation.

ELares:
Windows CI lane caught a real dead-code error: `BenchConfig::io_mode` is written by the always-compiled arg parser but READ only in `bench_run.rs`, which is `#[cfg(unix)]` (the bench broker is the Unix-only `serve` path) — so on Windows the field is write-only and `-D warnings` refused it. Fixed by adding `#[cfg_attr(not(unix), allow(dead_code))]` to the field, the same guard `fsync_is_measured`/`publish_acks_are_awaited` already carry for this exact footgun. Force-pushed 782c25b; re-running CI.

Issue #761, opened 2026-06-29
EPIC: official multi-language client SDKs (Go, Python, Java, JS/TS, C) over the versioned wire


ELares:
IronBus ships exactly two client crates, `ironbus-client` and `ironbus-client-async`, and both are Rust. There is no client for Go, Python, Java, JS/TS, or C, so every non-Rust consumer has to hand-roll the framed wire or shell out to the CLI. The async Rust client landed in #756 to #758 as commits with zero tracking issues, so even the Rust side of the ecosystem has no design record. A world-best bus needs a broad, conformance-tested client matrix, and ecosystem breadth is currently the single widest strategic gap between IronBus and the incumbents. This beats nothing today because the matrix does not exist. First principles: the value of one durable, versioned wire compounds with the number of languages that can speak it, so thin per-language clients over a frozen contract multiply reach at near-zero marginal protocol risk.

## Competitor evidence
NATS is where this gap is most visible. The NATS organization maintains and documents a deep official client matrix: Go (nats.go), Rust (async-nats), Python (nats.py), Java (nats.java), C (nats.c), C#/.NET (nats.net), JS/TS (nats.js / nats.ws for the browser), plus community-grade Ruby and Elixir. That breadth, roughly ten officially maintained languages, is NATS's widest lead over IronBus per the EPIC #557 scorecard. Kafka and Pulsar carry similarly deep matrices: Kafka has the Java reference client plus librdkafka (C/C++) feeding Python (confluent-kafka-python), Go (confluent-kafka-go), and .NET bindings, and Pulsar ships Java, C++, Python, Go, and Node clients. IronBus at one language is not in the same conversation on reach.

## Proposal
File this as a tracking epic that sequences AFTER the wire-contract freeze and splits into per-language children. Scope:

1. **Freeze and publish the wire contract.** Depends on the wire-versioning / stabilization issue (to be filed). The wire already carries a stable one-byte frame tag plus per-body `body_version: u8` framing in `crates/ironbus-proto/src/frame.rs`; this epic does not begin coding SDKs until that contract is frozen and published as a language-neutral protocol spec (frame layout, verb tags, body-version rules, error frames).
[...]

PR #1056, opened 2026-07-04
docs(sdk): flagship Rust client examples + sync crate docs.rs parity (#761)

ELares:
## What

Fills the biggest gaps the SDK docs review found: the Rust clients had **no examples** for several flagship capabilities, and the sync `ironbus-client` docs.rs landing opened on prose with no copy-pasteable snippet (the async twin already had a header example).

### New examples (all `cargo build --examples`-clean and smoke-tested against a live broker)
| Crate | Example | Demonstrates |
|---|---|---|
| `ironbus-client` | `streaming_consumer` | Tier-S negotiation → `subscribe` → window pull with `next_batch` + periodic **cumulative** commit + `finish` |
| `ironbus-client` | `transactions` | 2PC half-message via `transact`: commit-on-`Ok` / rollback-on-`Err`; **asserts the rolled-back record is never delivered** |
| `ironbus-client` | `not_leader` | Cluster redirect: `produce_to_leader` (the one-call path) **and** the manual `ClientError::NotLeader { leader_hint }` reconnect pattern |
| `ironbus-client` | `auth` | Bearer + password credentials (`pack_password_material`), redacted in `Debug` |
| `ironbus-client-async` | `streaming_consumer_async` | the tokio twin of the streaming consumer |

### Sync crate docs.rs parity
- **`proto` re-export module** (`PubBody`, `AckLevel`, `ConsumeTier`, `PubDedup`) mirroring `ironbus-client-async`'s, so the two clients share one import surface and a caller needn't depend on `ironbus-proto` directly. (Additive.)
- **`# Example`** in the sync crate header — a runnable produce/subscribe/fetch/ack snippet, so docs.rs opens on copy-pasteable code. Verified by `cargo test --doc`.

## Verification
- `cargo build -p ironbus-client -p ironbus-client-async --examples` — clean.
- `cargo clippy --all-targets` (both crates) — clean (fixed a `doc_lazy_continuation`).
- `cargo fmt --check` — clean.
- `cargo test --doc -p ironbus-client` — the new header example runs.
- **Live smoke test** against `ironbus serve`: all five examples run to completion; the transactional example's assertion (rolled-back half message absent) holds.

No library behavior changes — additive re-export + examples + docs only.

### DIFF
--- a/crates/ironbus-client-async/examples/streaming_consumer_async.rs
+++ b/crates/ironbus-client-async/examples/streaming_consumer_async.rs
@@ -1,0 +1,97 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+//! ASYNC TIER-S STREAMING CONSUME: the tokio twin of the synchronous `streaming_consumer` example.
+//!
+//! Same model as the sync client (same wire, same types): negotiate Tier-S in the handshake,
+//! `subscribe` to mark the group streaming, then pull the log in WINDOWS with
+//! [`AsyncStreamingConsumer::next_batch`], which advances the cursor, prefetches the next window, and
+//! commits the cumulative offset periodically. The caller processes each window and awaits the next —
+//! it does NOT ack records individually. Delivery is at-least-once:
+//! `[committed_offset(), next_offset())` redelivers on a crash, so processing must be idempotent, and
+//! [`AsyncStreamingConsumer::finish`] flushes a final commit so a rerun resumes from the durable cursor.
+//!
+//! One `AsyncClient` per task: the wire is request-response FIFO per connection.
+//!
+//! Run a broker, then the example:
+//!
+//! ```sh
+//! ironbus serve --data-dir /tmp/ironbus-data
+//! cargo run -p ironbus-client-async --example streaming_consumer_async
+//! cargo run -p ironbus-client-async --example streaming_consumer_async -- 127.0.0.1:7777
+//! ```
+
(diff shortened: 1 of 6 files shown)


ELares (commented):
Self-review:
- **Additive only.** The single library change is a new `proto` re-export module + a crate-header doc example; no existing item's type or behavior changes. `DEFAULT_PIPELINE_WINDOW` etc. untouched.
- **`proto` re-export** deliberately mirrors `ironbus-client-async`'s exact set (`PubBody`/`AckLevel`/`ConsumeTier`/`PubDedup`) so the sync/async import surfaces converge — the divergence the SDK map flagged.
- **Examples verified semantically, not just compiled.** Ran all five against `ironbus serve`; the transactions example asserts the rolled-back half message is absent from the consume stream (the actual 2PC guarantee), and it held.
- **`auth` example is safe to run either way**: against a no-auth broker the credential is ignored and it produces; against an auth-required broker a bad credential surfaces as `ClientError::Server` and is reported, not panicked. Confirmed the credential is redacted in the `Debug` print (`<redacted; 20 bytes>`).
- **`not_leader` on a single node** exercises the no-redirect path (both `produce_to_leader` and the manual match); the redirect branch is the code that matters in a real cluster and is written out for a reader.
- One risk considered: doctest portability — the header example uses only `ironbus_client::{Client, proto::PubBody}`, no `ironbus-proto` path, so it compiles in the doctest's crate sandbox (`cargo test --doc` passes).

PR #1057, opened 2026-07-04
fix(go-sdk): validate Config.DefaultAckLevel client-side (#1039) + auth example

### LINKED ISSUE #1039: Go SDK: symmetric client-side validation of Config.DefaultAckLevel (<= 2)
Follow-up from the #1037 Go SDK review.

**Nit N1 (non-blocking, fail-closed today):** `ProduceWithAckLevel` rejects an ack level > 2 client-side with a typed `*InvalidAckLevelError` (produce.go:78-79), but `Config.DefaultAckLevel` (a `*uint8` encoded into the Connect body, client.go:103 / config.go:27) is sent WITHOUT the same client-side guard. An out-of-range value is currently rejected by the *server* at Connect with a typed `Err` — so it is fail-closed, just asymmetric.

**Proposed:** add the symmetric `<= 2` guard on `Config.DefaultAckLevel` in the `Connect` path, returning `*InvalidAckLevelError` before dialing, plus a unit test. Purely a tidiness/consistency improvement; no wire or behavior change.

Also noted in the same review (documented MVP choices, likely no action):
- N2: `context.WithCancel` cancellation (no deadline) does not interrupt an in-flight blocking read/write — documented in the package doc + README, matches the synchronous Rust client.
- N3: streams.go encodes stream-id/subject up to 0xFFFF while the decoder caps at MaxStreamIDLen=1024; an over-long name is server-rejected, no desync (cosmetic asymmetry).

ELares:
## Fixes #1039

`Config.DefaultAckLevel` (a `*uint8`) was passed straight into the `Connect` handshake body with **no client-side validation**, while `ProduceWithAckLevel` already rejects an out-of-range per-publish level (`produce.go`). That asymmetry let an invalid default ack byte (`> 2`) travel on the wire for the broker to reject at handshake time.

`Connect` now validates it **before dialing** and returns the same typed `*InvalidAckLevelError` — a config mistake becomes a clear local error, not a wire error.

```go
if cfg.DefaultAckLevel != nil && *cfg.DefaultAckLevel > AckLevelServerAndClient {
    return nil, &InvalidAckLevelError{Level: *cfg.DefaultAckLevel}
}
```

A test pins both directions: **level 3 is rejected before any dial** (an unroutable address proves no socket is opened — a regression would surface as a dial error instead), and the **boundary level 2 connects fine** against a broker.

## Also: `examples/auth`

Connecting with a **bearer** or **username+password** credential (`Bearer` / `Password` → `Config.Credential`), with the secret **redacted when logged**. Safe to run against a no-auth broker (the credential is ignored) and reports a `*ServerError` against an auth-required one. Smoke-tested against a live broker:

```
connecting to 127.0.0.1:7802 with a bearer credential (secret redacted: ironbus.Credential{mechanism: bearer, material: <redacted>})
authenticated connect OK; produced at offset 0
```

*(A `cumulative_ack` example was considered but deferred: `CumulativeAck` requires a broadcast group, and the Go SDK's `Subscribe` doesn't yet expose broadcast-group creation — a separate follow-up.)*

## Verification
- `gofmt -l` clean, `go vet ./...` clean, `go build ./...` clean.
- `go test ./...` — all pass, including the new `TestConnectRejectsInvalidDefaultAckLevel`.
- Live smoke test of `examples/auth`.

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -87,6 +87,9 @@ Each example runs against a live broker
 - `examples/cluster_notleader`: following the typed NotLeader redirect (a
   single-node broker never emits it; run against a cluster node to see the
   redirect).
+- `examples/auth`: connecting with a bearer or username+password credential
+  (secret redacted when logged; safe to run against a no-auth broker, which
+  ignores it).
 
 The only dependency is `github.com/pierrec/lz4/v4` (the broker compresses
 stored payloads transparently, so the SDK must decode lz4 blocks).
--- a/sdk/go/client.go
+++ b/sdk/go/client.go
@@ -85,6 +85,14 @@ type Client struct {
 // Connect dials the broker, completes the Connect/Info handshake, and adopts
 // the negotiated capabilities.
 func Connect(ctx context.Context, cfg Config) (*Client, error) {
+	// Symmetric client-side validation (#1039): the connection-wide default ack
+	// level must be one of the three frozen wire levels (0, 1, 2), exactly as
+	// ProduceWithAckLevel rejects an out-of-range per-publish level. Reject it here
+	// rather than dialing and sending an invalid byte the broker would refuse at the
+	// handshake — a config mistake surfaces as a clear typed error, not a wire error.
+	if cfg.DefaultAckLevel != nil && *cfg.DefaultAckLevel > AckLevelServerAndClient {
+		return nil, &InvalidAckLevelError{Level: *cfg.DefaultAckLevel}
+	}
 	addr := cfg.Addr
 	if addr == "" {
 		addr = DefaultAddr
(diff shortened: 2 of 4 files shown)


ELares (commented):
Self-review:
- **Validation placed before dial** so a config error never opens a socket; returns the *same* typed error (`InvalidAckLevelError`) as the existing per-publish `ProduceWithAckLevel` guard — the symmetry #1039 asks for.
- **Test proves before-dial**: level 3 with an unroutable addr (`203.0.113.1:1`, TEST-NET-3) must return the typed error, not a dial failure; a 1s context bounds it so a regression fails fast rather than hangs. Boundary level 2 connects against the stalled-broker helper.
- **auth example is broker-agnostic**: no-auth broker ignores the credential (produces), auth-required broker surfaces `*ServerError` (reported, not fatal). Confirmed the secret is redacted in the logged Credential (`material: <redacted>`).
- **cumulative_ack deferred deliberately**: `CumulativeAck` hard-rejects non-broadcast groups, and Go's `Subscribe` has no broadcast-group creation path — I won't ship an example that can't run. Noted as follow-up in the PR body.

PR #1059, opened 2026-07-04
feat(go-sdk): pipelined ProduceWindow — one group-commit fsync per window (#761)

ELares:
## What

Adds **`ProduceWindow`** to the Go client — the durable-throughput lever the Go SDK was missing. It writes **N `Pub` frames in one coalesced write**, then drains **N `PubAck` replies FIFO**, so the broker's group commit covers the whole window with a **single `fdatasync`** instead of one per record. A single Go producer can now reach the durable group-commit throughput the Rust `produce_window` already reaches (and that the benchmark story depends on) — the Go SDK previously produced one awaited record at a time.

```go
acks, err := client.ProduceWindow(ctx, batch) // acks[i] durable; acks[i].Offset is FIFO
```

Semantics mirror the Rust reference exactly:
- Replies are **FIFO** — the Nth ack is the Nth message — and every ack is fsync-durable.
- A per-slot **server rejection / `NotLeader`** is remembered (first wins) while the remaining replies are still drained, so the connection stays framed.
- An **IO / unexpected-frame** error is terminal (drop the connection).
- An **empty window** is a no-op that never touches the wire.

## Verification
- **Unit (runs in CI, non-live):** a mock broker drains N pubs and replies N acks → asserts FIFO contiguous offsets; empty-window no-op against a stalled broker (proves it never touches the wire).
- **Live (`IRONBUS_LIVE`):** a 16-record window returns contiguous acks and reads back in order against a real `ironbus serve`.
- **Example** `examples/produce_window` prints measured throughput — **~44k msg/s at window 512** on a laptop debug broker.
- `gofmt`/`go vet`/`go build`/`go test` clean; `relative-link-check` clean (docs updated).

## Docs
Updates `sdk/go/README.md` (produce capabilities + example list) and `docs/CLIENT_SDKS.md` (the Go pipelined-produce cell flips ⏳ → ✅ in the capability matrix; the produce section gains a Go `ProduceWindow` snippet).

### DIFF
--- a/docs/CLIENT_SDKS.md
+++ b/docs/CLIENT_SDKS.md
@@ -77,7 +77,7 @@ the broker's `--durability-level`). Pick the shape by your latency/throughput ne
 | Shape | Rust | Go | When |
 | --- | --- | --- | --- |
 | Awaited durable | `produce` | `Produce` | one record, simplest; ~one fsync per record |
-| **Pipelined durable** | `pipelined_producer[_with_window]`, `produce_window` | *(not yet)* | high single-producer throughput: a window shares one group-commit fsync |
+| **Pipelined durable** | `pipelined_producer[_with_window]`, `produce_window` | `ProduceWindow` | high single-producer throughput: a window shares one group-commit fsync |
 | Fire-and-forget (QoS 0) | `produce_fire_and_forget` | `ProduceFireAndForget` | max speed, loss acceptable |
 | Idempotent (dedup) | `produce_dedup` | `ProduceDedup` | safe retries: a repeated `MsgID` returns the original offset |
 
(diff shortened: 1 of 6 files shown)


ELares (commented):
Self-review:
- **Wire discipline matches the awaited path**: one coalesced `conn.Write` of the framed window (reusing `c.wbuf` like `send`), deadline set for the write only (`readFrame` self-manages read deadlines and fails terminally on a mid-reply timeout — the FIFO-safety posture is preserved).
- **First-error-wins, drain-the-rest**: a per-slot `*ServerError`/`*NotLeaderError` is remembered and the loop keeps draining so the connection stays framed for reuse (matches Rust `produce_window`); on `firstErr` it returns `nil, err` (no ambiguous partial-acks-with-error). A non-typed error (IO/unexpected) aborts and the client is already `broken`.
- **No dedup/FaF in this path** by design: FaF has no reply (would desync the FIFO), and `Message` carries no dedup block — level 1 durable only, documented.
- **Coverage** is honest about the SDK's existing posture: the FIFO drain is unit-tested with a mock broker (CI, non-live) AND live-tested; the SDK's other consume paths are live-only, so this is stronger than the baseline.
- One risk considered: TCP write/read deadlock on a large window — the mock reads the whole window before replying, and a real broker's session parks acks; the client writes fully before reading, so no half-duplex stall for realistic window sizes.

Issue #1067, opened 2026-07-06
[P0] Cluster peer wire: fail-closed bind guard + interim peer authentication


ELares:
From the 2026-07-06 production-readiness audit (adversarially verified, P0-3).

The raft/data-plane peer listeners bind ANY address with zero guard or warning (serve.rs:1102, runtime.rs:657, serve_accept.rs:190) and authenticate peers only by a spoofable claimed peer-id (transport.rs:310). Anyone who can reach the port can forge raft/ISR frames and drive committed cluster state. This bypasses the project's own fail-closed bind invariant, which the CLIENT wire enforces.

Fix: (1) the same fail-closed bind invariant on peer listeners; (2) interim shared-secret HMAC on peer frames until TLS/mTLS (#766) covers the peer wire.

ELares:
All three increments merged: Inc1 fail-closed bind guard (#1094), Inc2 metadata-wire HMAC (#1095), Inc3 data-plane HMAC (#1097). The cluster peer wire is now fail-closed on bind and HMAC-authenticated on both the raft metadata and data planes, with the off/permissive/signed/required mode ladder. Residual (no anti-replay/confidentiality/per-node identity) is deferred to the mTLS epic #766. Closing.

Issue #825, opened 2026-06-29
[medium] Cluster replication leader deep-copies the zero-copy RawByteRun 2-3x per follower fetch (to_vec + encode + frame)


ELares:
## The cost

Both leader-side fetch serve paths read a contiguous, refcounted `RawByteRun` (a `bytes::Bytes` slice of a sealed segment's resident buffer — the whole point of #542/#657, doc at `segment.rs:344`: "cloning it is a refcount bump, never a copy") and then immediately deep-copy it into a `Vec<u8>`:

`crates/ironbus-server/src/cluster/replication.rs:506` (`ReplicationLeader::serve_fetch`):
```rust
let (run, _tail) = self.log.read_range_raw(from, max_records, Some(req_bytes as usize))?;
Ok(FetchResponseBody {
    high_watermark: hw.get(),
    first_offset: run.first_offset.get(),
    record_count: u32::try_from(run.record_count).unwrap_or(u32::MAX),
    frame_bytes: run.bytes.to_vec(),   // <-- COPY #1 (deep-copies the Bytes)
})
```

`crates/ironbus-server/src/cluster/replication.rs:626` (`ReadPlaneLeader::serve_fetch`, off-actor twin) does the same: `frame_bytes: sealed.run.bytes.to_vec()`.

The struct field forces this — `crates/ironbus-server/src/cluster/replication.rs:178`:
```rust
pub frame_bytes: Vec<u8>,
```

`encode()` then copies it AGAIN (`replication.rs:193`):
```rust
out.extend_from_slice(&self.frame_bytes);  // <-- COPY #2
```

And `send()` (`replication.rs:1122`) copies the encoded body a THIRD time into the wire frame before writing:
```rust
let mut frame = Vec::with_capacity(body.len() + 5);
encode_frame(ty, body, &mut frame) ...   // <-- COPY #3
self.stream.write_all(&frame)?;
```

So every follower fetch memcpys up to `MAX_REPL_FETCH_BYTES` (8 MiB) of log bytes two-to-three times on the leader, entirely defeating the zero-copy read plane the bytes were produced through.

This is the live production path, not a test helper: `DataPlaneController::handle_frame` (`dataplane.rs:982`) -> `serve_fetch` -> `DataPlaneAction::SendFetchResponse` -> `send_response` (`replication.rs:1097`) -> `encode()` -> `send()`.

## Impact

[...]

Issue #766, opened 2026-06-29
[M7-I1b] Resolve the pure-Rust crypto-provider decision and ship native TLS 1.3 + mTLS transport


ELares:
Native TLS is specified but not built. The `--tls-cert` / `--tls-key` / `--tls-ca` flags are RESERVED and NOT YET HONORED: the bind guard fail-closed REFUSES any `--tls-*` value rather than accept a cert that would imply encryption the build cannot deliver (`crates/ironbus-cli/src/main.rs:1893`), and the TLS 1.3 handshake that wraps the credential on a non-loopback bind is an explicit FLAGGED follow-up (`auth.rs:20`, `server.rs:425`). The blocker is upstream of the code: `deny.toml` bans `ring`, `aws-lc-sys`, and `openssl-sys` for supply-chain reasons, so `rustls` has no permitted `CryptoProvider`, and `Cargo.lock` pulls no `rustls` today. mTLS client auth therefore fails closed until TLS lands. The auth identity layer (three mechanisms, three scopes, SAN-to-identity matching) is already built and waiting for a wrapped wire. This is currently UNTRACKED as an implementation item: the spec issues #107 (TLS) and #108 (at-rest) are CLOSED M1 architecture-spec tasks, #629 closed the fail-closed scaffold, and there are zero open `area:security` issues, so the missing piece is the provider decision plus the real transport.

**First principles:** the single static binary is a hard invariant, so the transport-security layer cannot link any C or assembly crypto, which means TLS must run on a pure-Rust provider; choosing that provider is a one-time architectural decision that gates the entire native-encryption surface, so it must be recorded before the transport is wired.

**Decision to record (decision-needed -> decision-record):** select the pure-Rust `rustls` `CryptoProvider` (for example a RustCrypto-backed provider, reusing the `aes-gcm` / `chacha20poly1305` AEADs already on the allow-list per `docs/SECRETS.md:319`) such that `deny.toml` keeps banning `ring`, `aws-lc-sys`, and `openssl-sys` and `cargo-deny` stays green. Capture the rationale and the rejected alternatives as an ADR on this issue.

**Scope (this issue only):**
1. Resolve and record the pure-Rust crypto-provider decision.
2. Honor `--tls-cert` / `--tls-key` / `--tls-ca`: stand up the native TLS 1.3 handshake on non-loopback binds, fail-closed if material is missing or malformed (no silent plaintext fallback).
[...]

PR #1097, opened 2026-07-07
feat(cluster): data-plane peer-wire HMAC (#1067, increment 3)

ELares:
## Increment 3 of #1067 — data-plane peer-wire HMAC

Extends the interim shared-secret HMAC peer authentication from the raft **metadata** wire (Increment 2) to the cluster **data-plane** peer wire, reusing the exact Increment 2 primitives (`PeerKey` / `PeerSecurity` / `PeerAuthMode` ladder + the hand-rolled HMAC-SHA256 over the in-tree `sha2`, constant-time `subtle` compare, zeroizing/redacting key).

**DRAFT** — the zero-copy-vs-MAC design fork (below) wants owner review before this leaves draft.

### What is authenticated + the wire tag
A single self-describing wrapper tag **`FrameType::DataPlaneAuth = 51`** wraps *every* data-plane peer verb (rather than a per-verb MAC): `FetchRecords` (32), `FetchResponse` (33), `AckReplicated` (37), `OffsetForLeaderEpoch` (38), `CommittedHwQuery` (43). One tag keeps the mode-ladder logic and the registry/docs footprint minimal, and the receiver still routes the recovered verb through the identical bounded `decode_dataplane_frame`.

Wire layout (twin of `RaftAuth`=50):
- plain: `[len][verb-tag][partition:u64-le][layer body]`
- signed: `[len][51][ver:1][mac:32][verb-tag:1][partition:u64-le][layer body]`, `mac = HMAC-SHA256(key, DATA_DOMAIN_LABEL || ver || verb-tag || partition || layer-body)`.

The verb tag is re-embedded *inside* the authenticated content (the outer tag is now 51). A **distinct domain label** (`ironbus.cluster.peer.data.v1`) means one shared cluster secret can never cross-authenticate a raft frame as a data frame (unit-tested).

The full mode ladder (off / permissive / signed / required) is threaded end-to-end: the leader-side inbound reader links, every follower fetch link, and the dirty-tier committed-HW confirm mini-link all sign/verify under the same policy the raft metadata wire already uses (wired from the CLI `--cluster-secret-file` + `--cluster-peer-auth`).

### The zero-copy-vs-MAC fork + recommendation
The 8 MiB `FetchResponse` fast path (`encode_fetch_response_peer_frame`) copies the borrowed `Bytes` run into the outbound buffer exactly once (#825/#810). A MAC over the payload could force a second copy.

**Chosen + recommended: option (a), streaming-HMAC over the borrowed slices — no extra payload copy.** Feasibility confirmed: the existing `hmac_sha256(key, parts: &[&[u8]])` already streams each borrowed slice through `sha2::Sha256::update` in place. The signed fast path (`encode_fetch_response_peer_frame_authed`) materializes the run into the outbound buffer *once* (unchanged), reserves the MAC slot, then computes the MAC over the already-written buffer slice and writes it back — **zero extra copy of the bulk payload on both send and receive**. Rejected: (b) MAC-only-the-header (weaker — the bulk record bytes would be unauthenticated), (c) accept-a-copy (needless perf regression). A test proves the zero-copy signed frame is byte-identical to a naive materialize-then-seal reference across empty/tiny/3 MiB runs and round-trips.

Because (a) turned out fully feasible, the "safest correct default" and the recommendation coincide — there is no weaker fallback shipped.

### Files changed
- `crates/ironbus-server/src/cluster/peer_auth.rs` — `data_seal` / `data_seal_mac` (streaming) / `data_open` + `DATA_DOMAIN_LABEL` + `DATA_AUTH_HEADER_LEN`; unit tests (round-trip, streaming==contiguous, tamper, wrong-key, truncated/version, raft-vs-data domain separation).
- `crates/ironbus-server/src/cluster/serve.rs` — `DataPlaneAuth` codec (`encode_dataplane_peer_frame_secured` incl. the zero-copy authed fast path; `decode_dataplane_peer_frame_secured`); `DataPlaneWireError::{MacMissing, UnexpectedAuthFrame, Auth}`; `DataPlaneLink` gains a `PeerSecurity` (`with_security`); `PeerSecurity` threaded through `DataPlaneRuntime::start*` → listener/reader + follower-fetch + `query_leader_committed_hw`; integration tests (signed round-trip, zero-copy byte-identity, send-signs, mode-ladder, wrong-secret, off-link-rejects, tamper).
- `crates/ironbus-server/src/cluster/client_ack.rs` — `ClientAckGate` carries the `PeerSecurity` for the dirty-tier HW-confirm link.
- `crates/ironbus-proto/src/frame.rs` — `FrameType::DataPlaneAuth = 51` (+ `as_u8`/`from_u8`/frozen-value/bijection tests + tag-map comments).
- `crates/ironbus-cli/src/main.rs` — threads the already-built `peer_security` into `spawn_dataplane_serve` → the data-plane runtime; corrects the now-stale "data-plane is plaintext / follow-up increment" operator warning + flag docs.
- `docs/compat/versions.md` — `FrameType` row `1..=50`→`1..=51`; re-pinned `frame-tags-sha256`.
- `docs/CONTRACTS.md` — tag 51 row + `from_u8` unknown-range bump to 52.
- `docs/TRANSPORT.md` §2.5 — replaced the Inc2-era "data-plane wire is plaintext / a forged `AckReplicated` could raise a follower's frontier" text with the Inc3 `DataPlaneAuth` HMAC description; kept the honest residual (no anti-replay / confidentiality / per-node identity → mTLS #766).

### Build + test
`cargo build -p ironbus-server` and `cargo test -p ironbus-server cluster` both green (debug, colima container). Format-registry gate (`scripts/check-format-registry.sh`) passes. See the summary line in the PR discussion.

### Needs owner decision before leaving draft
1. Confirm the single-wrapper-tag (`DataPlaneAuth=51`) approach over per-verb MACs.
2. Ratify the zero-copy streaming-HMAC (option a) as the durable design.
3. Enforcement is still gated at `--cluster-peer-auth=required`; the pre-`required` rungs are migration-only. mTLS (#766) remains the follow-up for per-node identity + confidentiality.


ELares:
Build + test results (debug, colima container):

- `cargo build -p ironbus-server` — Finished (clean)
- `cargo test -p ironbus-server cluster` — `test result: ok. 431 passed; 0 failed; 0 ignored; 0 measured; 665 filtered out`
- `cargo test -p ironbus-proto` — `test result: ok. 156 passed; 0 failed; 0 ignored` (frame tag map incl. DataPlaneAuth=51)
- `cargo check -p ironbus-cli` — Finished (clean)
- `sh scripts/check-format-registry.sh` — ok: wire FrameType tag map matches the pinned registry digest

PR #1098, opened 2026-07-07
perf(consume): event-driven commit-notify long-poll delivery (default off)

ELares:
## What

Adds an **opt-in, server-only** event-driven long-poll to the consumer path. An idle consumer that polls an empty group can now BLOCK briefly on a broker-wide commit-notify seam and wake the instant a record commits, instead of returning empty and waiting a full client poll interval to be re-polled.

**Tunability (default OFF):** `consume_longpoll_ms` defaults to `0` = byte-for-byte today's empty-and-return behavior. No wire, client, or format-registry change — a consumer against a long-poll-enabled broker simply gets its empty batches later (on a commit or the budget), never a new frame.

## How

- **New `commit_notify` module** — an `Arc<CommitNotify>` holding `Mutex<u64>` generation + `Condvar`. `bump()` advances the generation and `notify_all`s; `wait_for_change(snapshot, timeout)` blocks on `wait_timeout_while(|s| *s == snapshot)`, lost-wakeup-safe (the waiter snapshots the generation BEFORE its poll batch, so a bump racing the snapshot leaves the predicate already false).
- **Append actor owns one `CommitNotify`**, cloned into every `EngineHandle`. It bumps STRICTLY AFTER the durability work that advances the poll-visible frontier — `Engine::flushed_offset` (the head `poll_now_in_member` serves up to):
  - legacy loop: after the pass-end `flush_pending`;
  - pipelined loop: after `wait_one_completion` (async barrier completes during the idle wait) AND after `finish_pass` (catches a **solo-inline** covering barrier that advances the frontier at dispatch — the earlier "before `finish_pass`" placement missed this and stranded the consumer until its timeout).
  A helper tracks the last-notified frontier and bumps only on a real advance. Pure observation — never reorders existing statements. Over-bumping is harmless; under-bumping only costs latency.
- **`handle_flow` / `handle_fetch`** wrap the existing drain in an outer at-most-two-pass loop: snapshot the seq before polling, run the drain, and if `delivered == 0` AND `consume_longpoll_ms > 0` AND the engine exposes a notify, `wait_for_change` once and re-run the drain exactly once, then fall through to the unchanged committed-prune / keep-up / `FlowEnd` tail (runs once). Re-running from `delivered == 0` is safe: no leases inserted, no frames emitted, credit/byte/AIMD bounds unconsumed. A `no_wait` Fetch is exempt (contractually a single immediate pass). Granularity is global for v1 — a spurious cross-stream wake just re-polls empty, exactly today's timer-driven re-poll for that consumer; per-stream is a future refinement.
- **Config** plumbs CLI `--consume-longpoll-ms` (flag > env `IRONBUS_CONSUME_LONGPOLL_MS` > default 0) through `ServeConfig` → `EngineConfig` → `Engine` → `EngineHandle` snapshot → `Session::consume_longpoll_ms`, mirroring `--egress-limit`.

## Tests

Three new tests in the server crate, driven over a REAL spawned append actor (the only `EngineAccess` that carries a live commit-notify seam):

1. `consume_longpoll_ms == 0` → empty Flow returns `FlowEnd(0)` immediately (unchanged).
2. budget set + a concurrent produce → the idle consumer is woken to `delivered = 1` far below the budget (proving the event path fires, not the timeout).
3. budget set + no producer → `FlowEnd(0)` after waiting ~the budget.

`cargo test -p ironbus-server`: **1066 passed, 0 failed** (+ integration binaries green); `cargo build --workspace` clean.

## Notes for review

- The bump reads `flushed_offset()` (the poll-visible frontier), NOT the appended head — so a consumer is only woken once a record is actually deliverable.
- The `no_wait` Fetch exemption is a correctness point (a `no_wait` fetch must never block).
- Global wakeup granularity is documented as intentional for v1.

### DIFF
--- a/crates/ironbus-bench/src/inproc.rs
+++ b/crates/ironbus-bench/src/inproc.rs
@@ -66,6 +66,7 @@ fn engine_config() -> EngineConfig {
 /// subject), not the per-group window, is the binding constraint the bench measures.
 fn engine_config_with_credit(consumer_credit: u32, max_in_flight: u32) -> EngineConfig {
     EngineConfig {
+        consume_longpoll_ms: 0,
         log: LogConfig::default(),
         lease: LeaseConfig::default(),
         // `unwrap` is fine in this bench/test-only path (not a lib hot path): the literal args are
--- a/crates/ironbus-cli/src/admin.rs
+++ b/crates/ironbus-cli/src/admin.rs
@@ -671,6 +671,7 @@ mod tests {
             InMemoryFs::new(),
             SystemClock::new(),
             EngineConfig {
+                consume_longpoll_ms: 0,
                 log: LogConfig::default(),
                 lease: LeaseConfig::default(),
                 delivery: DeliveryConfig::new(5, false, vec![]).unwrap(),
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -9565,6 +9565,11 @@ fn cmd_serve(
         config.fire_and_forget_byte_rate,
         config.fire_and_forget_refill_ms,
         config.egress_limit,
+        // The push-delivery consume long-poll budget is read only on the Unix serve path (it wires the
+        // engine's `consume_longpoll_ms`), so the non-Unix stub must consume it too or the Windows
+        // `-D warnings` build trips field-never-read, invisible to a macOS reviewer (the recurring
+        // #288/#99 footgun).
+        config.consume_longpoll_ms,
         // The #378 fsync-headroom knob is read only on the Unix serve path (it wires the engine's
         // wal_fsync_headroom_bytes), so the non-Unix stub must consume it too or the Windows
         // `-D warnings` build trips field-never-read, invisible to a macOS reviewer (the recurring
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -378,6 +378,14 @@ const DEFAULT_EGRESS_LIMIT: u32 = 0;
 const DEFAULT_WAL_FSYNC_HEADROOM_BYTES: u64 =
     ironbus_core::backpressure::DEFAULT_WAL_FSYNC_HEADROOM_BYTES;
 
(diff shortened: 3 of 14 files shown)


ELares:
### Merge-review follow-ups (commit ccebe6a)

1. **Default-off is now truly byte-identical (perf).** The commit-notify bump is gated on `consume_longpoll_ms != 0`, threaded into `run_actor` / `run_actor_pipelined`. When long-poll is OFF the actor never reads `flushed_offset`, never takes the lock, and never `notify_all`s per commit — the group-commit hot path is byte-for-byte the historical one (restoring the guarantee in the `spawn_actor` comment). Previously it bumped on every frontier-advancing commit regardless of the setting.
2. **SPDX header** added to the new `commit_notify.rs`.
3. **windows-latest** was a *compile* error, not a flaky test: `field consume_longpoll_ms is never read`, because the serve engine-config builder is `#[cfg(unix)]`. Fixed by consuming the field in the `#[cfg(not(unix))]` stub — the repo's established `-D warnings` guard for this footgun (#288/#99). Long-poll test timing margins were also widened to be deterministic under Windows Condvar/timer slack.
4. **clippy** `too_many_arguments` allowed on the two actor-loop functions (idiomatic here, matching `handle_connection`).

### Named-stream (#588) scope note

The bump watches **only the default/root log's `flushed_offset`**. A consumer long-polling a **named stream** (its own per-stream log, served by `poll_in_stream`) therefore gets **no early wakeup** — it is **timeout-backstopped**: it re-polls and delivers when its budget elapses, exactly the pre-push-delivery behavior for that consumer, so it stays correct. Observing per-stream frontiers is the documented **per-stream-granularity** future refinement (same bucket as the global-granularity note). Documented in `notify_frontier_advance` and the `commit_notify` module doc.

The advisory-frames-up-to-2×-credit case is unchanged and left as a separate follow-up note per review.

Local (colima debug container): `cargo test -p ironbus-server` **1066 passed, 0 failed**; `cargo build --workspace` and `cargo clippy --workspace --all-targets` clean.

Issue #830, opened 2026-06-29
[low] Offload per-record body CRC32C/xxh3 from the single-writer actor onto producing connection threads


ELares:
## The cost

The per-record body checksum is computed on the single-writer append actor, serializing an O(body-bytes) CPU pass onto one core for the whole broker.

`crates/ironbus-core/src/codec.rs:164-172`:
```rust
let body = &out[body_start..body_start + body_len];
let body_crc = crc32c::crc32c(body);
if has_xxh3 {
    let xxh3 = xxh3_64(body);   // only when body_len >= 64 KiB (XXH3_PAYLOAD_THRESHOLD)
    out.extend_from_slice(&xxh3.to_le_bytes());
}
out.extend_from_slice(&body_crc.to_le_bytes());
```

This `encode` runs on the single-writer actor: `run_actor` (`crates/ironbus-server/src/actor.rs:1362`) is a one-thread `recv` loop → `process_produce` (`actor.rs:1514`) → engine append → `Segment::append` (`crates/ironbus-storage/src/segment.rs:637`) → `codec::encode` (`segment.rs:538`/`653`). Every producer's `body_crc` (and, for bodies >= 64 KiB per `format.rs:72`, a second full `xxh3_64`) runs there while N connection threads sit idle.

The connection thread already touches every body byte: `Bytes::copy_from_slice(msg.payload)` in `crates/ironbus-server/src/session.rs:1396/3089/3189/3646`, and it holds `key`/`headers`/`payload` as borrowed slices before building `OwnedAppend`.

## Impact

Throughput under wide produce fan-in: removes one (large-body: two) full body-length CPU passes from the serialized actor and spreads them across connection cores, raising per-core aggregate produce throughput. Magnitude is message-size dependent and modest: `crc32c` is hardware-accelerated (~tens of GB/s), so for the small-message fan-in regime (the actual throughput ceiling) the saving per record is small relative to channel handoff + dedup + header-CRC + the body memcpy. The xxh3 pass only fires at >= 64 KiB, where produce is bandwidth-bound and records are few. The dominant per-byte actor cost (the body `extend_from_slice` memcpy) is NOT removed here and stays serial until zero-copy body (#523) lands; this finding is most valuable in conjunction with #523, after which the checksum would be the only remaining per-byte actor pass.

## Proposed optimization

[...]

Issue #594, opened 2026-06-19
[M2-I12] Per-subject filtered consumer with GapMarker on skip


ELares:
Add per-subject filtered consumers that apply a deliver-time subject filter and emit a GapMarker when records are skipped, so the consumer can advance its offset past filtered-out records without re-scanning. First-principles: a filtered consumer that re-reads skipped records pays for data it discards. Beats NATS, whose filtered consumer hits a ~99% throughput cliff (#7014); the GapMarker keeps filtered consume close to unfiltered cost.

ELares:
Per-subject filtered consumer with GapMarker on skip is complete on BOTH paths:
- #594-A (#1107): default-stream filtered consumer — HAS_SUBJECT durable record field, deliver-time SubjectPattern match, coalesced GapMarker(FILTERED) on skip, fail-closed capability bit. Adversarial review caught+fixed a recovery-scan silent-data-loss bug pre-merge.
- #594-B (#1109): same filter generalized to named streams, reusing #681's durable cursor.

Filtered≈unfiltered throughput (~1.04x — no NATS JetStream re-scan cliff), the competitive win. Closing.

Issue #566, opened 2026-06-19
[M2-I5] Per-stream retention/compaction


ELares:
Add per-stream retention and compaction by reusing Log::reap with whole-segment drop, so each stream ages out independently. First-principles: retention is a per-log property once streams are independent logs, and whole-segment drop keeps reclamation O(segments) with no record rewriting. Lets a high-churn stream reap aggressively while an archival stream retains, on the same node.

Issue #549, opened 2026-06-19
[M4-I1] Per-message + per-stream TTL (lower-wins) + idle-stream TTL


ELares:
Implement per-message TTL and per-stream TTL with lower-wins precedence, plus an idle-stream TTL for whole-stream expiry. Big-O: expiry rides the existing whole-segment retention/reap path (O(1) per segment drop) rather than per-record scans. Gives RabbitMQ-class message-expiry richness on the durable log.

PR #1147, opened 2026-07-11
feat(delivery): arbitrary-timestamp delayed messages, clock-skew-safe (#555)

### LINKED ISSUE #555: [M4-I4] Arbitrary-timestamp scheduled/delayed messages (clock-skew-safe)
Support arbitrary-timestamp scheduled/delayed messages, made clock-skew-safe by ordering on sequence rather than wall-clock. Adopts RocketMQ 5.x arbitrary-delay design (vs fixed delay levels). First principles: delayed delivery must be deterministic under clock skew, so seq ordering, not timestamps, drives release.

ELares:
Implements #555 (V2-M4-I4): arbitrary-timestamp SCHEDULED/DELAYED messages, clock-skew-safe — the RocketMQ 5.x arbitrary-delay design (any due-time, not fixed delay levels), with release ordering driven by LOG SEQUENCE, never wall-clock comparisons.

## Design

**Header-carried, two magics (the TTL1 precedent — smallest format footprint).** No record-format or frame change; the format registry digest is untouched and a non-delayed record is byte-identical.
- Wire request: `DLY1` + `delay_ms u64 LE` in the headers blob — a RELATIVE duration (durations are skew-free by construction). Producers attach it via `ironbus_core::delay::encode_delay_headers`.
- Stored resolution: `DUE1` + `due_unix_ms u64 LE` — the ABSOLUTE broker-wall instant, assigned by the BROKER's clock seam at append time. Same 12-byte block length, so the rewrite is an in-place patch.

**Skew model.** The broker resolves `DLY1 → DUE1` at exactly two produce chokepoints (`append_no_sync_checked` for the default stream; `produce_in_stream_with_subject` for named / partitioned / shared-WAL — the txn commit and exchange fan-out route through them). The producer-controlled wire `timestamp_ms` is never consulted for visibility, so a skewed producer can neither advance its own record nor perturb a peer's. Resolution is idempotent (`DUE1` passes through untouched), and replication / geo mirror / DLQ moves copy resolved frames VERBATIM — every replica holds the same instant; re-anchoring is impossible. A patched header drops the off-actor precomputed body checksum and the codec recomputes (the #830 compression-seam discipline), pinned by a wire-level test.

**Hold-at-undue (delivery-VISIBILITY gate, not a re-ordering engine).** All four work-group scans (default `poll_in_timed`, key-shared `poll_in_member`, named `deliver_from_named_stream` incl. shared-WAL + key-shared sub-paths, partition `deliver_from_partition`) HOLD at the first un-due record: the just-claimed lease is released WHOLE (held polls never count toward MaxDeliver — no attempt or in-flight residue) and the scan stops WITHOUT advancing the group cursor. Due records therefore release strictly in append order — **FIFO-with-delays**. The honest head-of-line property: one long-delayed record delays everything behind it in the same group; a due-index sidecar for out-of-order release is the named follow-up (bigger: a durable sidecar + re-visit machinery). Cursor-advance-past-then-redeliver was rejected outright (it needs re-visits and breaks the cursor monotonicity every other subsystem leans on).

**Max-delay bound (tunability).** `--max-delay-ms` / `IRONBUS_MAX_DELAY_MS` / `[delivery] max_delay_ms`, default 3 days (RocketMQ 5.x timer-max parity), `0` = unbounded. An over-bound request is rejected fail-closed with the new frozen `ERR_DELAY_TOO_LONG` code (typed, connection-preserving, names both the request and the bound) — never silently clamped — because an un-due record pins the retention floor, so unbounded scheduling = unbounded retention pin. The bound applies to the REQUESTED duration, before broker anchoring, so skew cannot dodge it.

## Interactions (each pinned by a test)

- **Retention (#566/#424):** the hold never advances a touched group's cursor, and `Log::reap`'s protect floor is the min committed offset across touched groups — an un-due record is never reaped. Only the disk-full FORCED drop-oldest (#82) overrides the floor, surfaced as the usual `Poll::Truncated` — the same contract as any unconsumed record. A never-touched group does not pin (unchanged #424 semantics).
- **TTL (#549):** lower-wins composition; the TTL clock stays anchored to the producer `timestamp_ms` (unchanged by the delay), so a TTL that elapses before the due-instant expires the record UN-DELIVERED (reclaimed or DLX'd, accounted). Canonical header order: `TTL1` outermost, then `DLY1`/`DUE1`; the decoders accept the delay block at 0 or after a leading TTL block.
- **Subject filters (#594):** non-match skip runs BEFORE the due hold (a record the group will never deliver cannot head-of-line block it); the hold runs BEFORE the match-stash (a stash can never smuggle out an un-due record).
- **DLQ / max-deliver:** unaffected until visible; attempts start at 1 on the first post-due delivery. A DLQ'd delayed record keeps its resolved `DUE1` verbatim.
- **Long-poll:** a parked consumer wakes on COMMITS, not on a record becoming due; worst-case added latency is bounded by `min(longpoll budget, poll interval)` — documented in the module docs.
- **Tier-S / streaming fetch:** a LOG read (replay semantics) — sees delayed records immediately, `DUE1` intact in the delivered headers for application-level handling; stated in the module docs (the same stance as retention/compaction being log-level views).

## Surface

- `ironbus-core/src/delay.rs` (new, pure/IO-free): `Delay`, `DLY1`/`DUE1` magics, `encode_delay_headers`, `resolve_delay_headers` (reject/patch/pass), `decode_due_headers`, `is_undue`.
- Engine: `EngineConfig::max_delay_ms`, `EngineError::DelayTooLong`, the two resolution chokepoints, `record_is_undue` + four scan holds.
- Codes/wire: frozen `ERR_DELAY_TOO_LONG` (`codes.rs` + proto `ServerErrorCode::DelayTooLong`); the default-stream produce path replies it coded (other transient failures keep the historical uncoded reply byte-for-byte).
- CLI/config: `--max-delay-ms` flag + env + `[delivery] max_delay_ms` config-file key, `docs/CONFIG.md` row, CHANGELOG.

## Tests

13 engine + 2 wire-level session + 10 core unit tests, all ManualClock-deterministic. Full CI-equivalent run in the standard container: workspace build (all targets, all features, locked) OK; `cargo test --workspace --all-features --locked` = **3395 passed, 0 failed**; `cargo fmt --check` OK; format-registry gate OK (no layout change); clippy on current stable `-D warnings` OK.

Closes #555.


ELares:
Review fix pushed in 6ab50f1 — **wire DUE1-injection now rejected fail-closed at both produce chokepoints.**

- `resolve_delay_headers` returns the typed `DelayHeaderError::DueInjected` for a raw `DUE1` block (bare or behind a canonical leading `TTL1` block), at any `max_delay_ms` setting including unbounded — the broker-clock-anchor bypass is independent of the bound. Only the broker mints `DUE1`.
- New non-fatal `EngineError::DueTimeInjected` → new frozen `ERR_INVALID_DELAY_HEADER` wire code (codes + proto decoder + coded produce reply steering to the legitimate `DLY1` relative request). Nothing appended, connection preserved.
- Internal verbatim paths untouched and now test-pinned: a delayed-then-poisoned record's DLQ copy keeps its broker-minted `DUE1` intact (`a_broker_minted_due1_rides_internal_appends_verbatim_the_dlq_copy_keeps_it`).
- New tests: core injection rejects (incl. TTL-wrapped + short-prefix tolerance), engine rejects at BOTH chokepoints with legit-DLY1-still-works, session wire reject with the frozen code + nothing stored/held.
- Operator-facing caveats raised to `docs/CONTRACTS.md`: the bounded head-of-line stall lever on shared streams (up to `max_delay_ms`; due-index sidecar = named follow-up) and the Tier-S/streaming bypass (a scheduling tool, not a confidentiality control). Both #555 codes added to the conformance codes table.

Full CI-equivalent rerun: workspace **3399 passed / 0 failed**, golden-path acceptance PASS, fmt, format-registry, clippy (stable, `-D warnings`) all green.

Issue #771, opened 2026-06-29
EPIC: Consumer-model breadth (ephemeral consumers, push delivery, pause, priority/overflow groups)


ELares:
## Mission

IronBus's consumer model today is durable pull-style work-groups over one ordered log, in two opt-in tiers: Tier-W (the lease / work-queue mode, refs #546) and Tier-S (streaming consume, refs #681), routed by the V2-M2 subjects substrate (refs #693). That is a strong core, but it is narrow against NATS. We are missing four consumer-model capabilities that NATS ships in the present tense, and "consumer-model breadth" is a defining bus capability. This epic groups the four as separate, milestone-sized child issues so each lands on its own merits. Index lives under the v2 EPIC (refs #557).

What we do NOT have today, verified in the clone:
1. **Ephemeral (auto-expiring) consumers.** In our docs "ephemeral" only ever means ephemeral storage or ports (docs/CLI.md, docs/USAGE.md), never an ephemeral consumer. Every work-group is durable.
2. **Push delivery.** "push" in IronBus is exclusively cross-cluster LeafPush replication, not consumer push delivery. Consume is pull only.
3. **A consumer pause API.** The draft Flow frame's `pause` field was deliberately dropped; the real `Flow` body is a single 4-byte LE u32 credit (docs/CONTRACTS.md:885-889, refs #292). We have credit-based backpressure only, no PauseUntil.
4. **Pull priority groups** (pinning, overflow, prioritized failover). Absent. Note #553 is per-message routing priorities, a different concept from NATS consumer pull priority groups.

## Competitor evidence (with versions)

- **NATS:** ephemeral plus durable consumers (always), push plus pull delivery (always), pull priority groups in 2.11, consumer pause via `PauseUntil` in 2.11, and prioritized / overflow policy in 2.12.
- **Pulsar:** four subscription types, Exclusive, Failover, Shared, Key_Shared.
- **IronBus:** durable pull work-groups only (Tier-W lease + Tier-S streaming).

This is the one consumer axis where NATS is broader than us in the present tense, so closing it advances the "decisively better than NATS on every front" mission.

## Proposal: split into four child issues

All four ride the existing Tier-W lease and Tier-S streaming tiers and the V2-M2 subjects substrate. None introduces a new storage engine; they are consumer-state and delivery-policy features over the single-writer log, so they fit the single static binary with zero new on-disk product surface. Proposed children:

[...]

Issue #676, opened 2026-06-20
[M2-I2b] Thread StreamSet through the Engine (per-stream WorkGroups, route publish/consume by stream id)


ELares:
Follow-up to #563 (M2-I2, the storage StreamSet). The storage layer now supports N independent per-stream logs, but the Engine still owns a SINGLE Log. Re-key the Engine to a StreamSet: `streams: BTreeMap<StreamId, Stream>` where each Stream owns its Log + its own `groups: BTreeMap<GroupName, WorkGroup>` (the lease/AckCursor/0-1-2-ack core re-instantiated per stream, unchanged). Route publish/consume by stream id; the default stream '' preserves today's single-log behavior byte-for-byte. This is the large, separable integration #563 flagged (engine.rs is ~12.6k lines). Depends on #563. Pairs with M2-I3 (#564 CommitCoordinator — cross-stream group-commit) and M2-I10 (the wire frames that carry the stream id). Until this lands, StreamSet is storage-only (not reachable from the wire/consume path).

Issue #597, opened 2026-06-19
[M2-I13] Shared-WAL fallback mode for high stream counts


ELares:
Add a shared-WAL fallback mode (RocketMQ CommitLog plus a derived per-stream index) for when stream count far exceeds core count. First-principles: per-stream logs cost one open segment each, so beyond ~cores worth of streams the fd/buffer overhead dominates; a single shared CommitLog with derived indexes keeps density high. This is the explicit high-stream-count tradeoff against the per-stream isolation default (I2).

ELares:
Phase 1 landed in #1122 (adversarial review MERGE_SAFE): the shared-WAL STORAGE PRIMITIVE — SharedWal (interleaved stream-tagged records in one log, per-stream demux read with tag==stream re-verify, recovery that rebuilds the per-stream index by scanning tags), the additive HAS_STREAM_TAG record field (CRC'd, mutually exclusive with HAS_SUBJECT, format-registry updated, the #594-A recovery-scan straddle structurally covered via a shared prefix-buffer guard + a const-assert pinning the widths equal), and EngineConfig::storage_mode (PerStreamLogs default; SharedWal opt-in).

The MODE is NOT yet usable: Engine::open fail-closed-REJECTS the SharedWal opt-in (typed Unsupported) rather than silently running per-stream — honest opt-in, no half-working path. Remaining scope to close #597 (the large cross-cutting engine part, deferred like #693's rebalance): wire produce_in_stream/poll_in_stream/ack_in_stream + recovery onto SharedWal; global-reap retention in shared mode; per-subject filtering in shared mode; the encoded_len/OwnedRecord stream-tag gap + the dedicated tag-straddle test (all tracked in #1123). Leaving open for the wiring phase.

Issue #565, opened 2026-06-19
[M2-I4] max_open_streams hot-set LRU


ELares:
Bound resident file descriptors to O(min(N,cap)) with a max_open_streams hot-set LRU that opens/closes per-stream logs on demand. First-principles: N streams would otherwise pin N fds and N segment buffers, scaling memory and fd usage linearly with stream count; an LRU keeps only the hot working set resident. Enables high stream counts on modest hardware without exhausting fds.

PR #1148, opened 2026-07-11
feat(consume): ephemeral consumer groups (#771)

ELares:
## What

**Ephemeral consumer groups** — #771 capability 1 (V2-M1), per the audit posted on the epic: a consumer group that leaves **NO durable artifacts** and is **fully reaped when its last member disconnects** — vs today's groups whose `cursor-<hex>.ckpt`/`attempts-<hex>.ckpt` persist forever. The NATS-ephemeral-consumer counterpart for throwaway/monitoring readers that must never pin broker state. **No new frame tag, no on-disk format change** (registry gate passes un-re-pinned).

### Wire (capability-gated BOTH ways, the #594 discipline)

- **`SubTo` (tag 31)** gains an appended `sub_flags: u8` (bit 0 = `SUBTO_FLAG_EPHEMERAL`) that always **follows** the optional `partition: u32`; the additive tail is disambiguated by **length** (`0` neither / `1` flags / `4` partition / `>=5` both). Every historical body decodes bit-identically; a durable subscribe emits byte-for-byte the pre-change bytes (pinned by proto tests, including hand-rolled legacy encodings and foreign 2–3-byte tails). Empty stream id = default stream (the plain `Sub` body is raw group bytes and cannot carry a flag, so `SubTo` — and hence the streams capability — is the carrier for both shapes).
- **`Connect` flags2 bit 3** (`CONNECT_FLAG2_WANTS_EPHEMERAL_GROUPS`): the server fail-closed rejects the subscribe flag on a connection that did not advertise it — a subscription is never silently made ephemeral.
- **New appended `Info` flags2 byte** (`INFO_FLAG2_EPHEMERAL_GROUPS` — the historical `Info` flags byte is FULL, bits 1..=128 all allocated): the server's AND-negotiated confirmation, emitted only when non-zero so the unconfirmed reply is byte-identical. This is the **old-server guard**: an old server *tolerates* the appended subscribe byte and would silently bind a **durable** group, so the reference clients require the confirmation and otherwise fail `subscribe_ephemeral` locally with the new typed `ClientError::CapabilityNotNegotiated` — the flag never reaches a server that didn't confirm it.

### Checkpoint-writer suppression (the (b) sweep — every writer gated at its funnel)

All four low-level writers gate on `WorkGroup.ephemeral`, plus honest early-outs in the four cadence entry points, so **every** cadence from the #1136/#547 sweep is covered by construction:

| writer | covers |
|---|---|
| `write_group_checkpoint` / `write_group_attempts` (default stream) | interval gate, #547 redelivery-driven attempts trigger, clean-disconnect flush, shutdown `checkpoint_all_groups`, #277 eviction step |
| `write_named_group_checkpoint` / `write_named_group_attempts` (named streams; `write_named_group_attempt_pairs`'s only caller is the gated fn) | the same cadences **plus the #565 stream-LRU eviction flush** |
| `checkpoint_group` / `maybe_checkpoint_group` / `checkpoint_named_group` / `maybe_checkpoint_named_group` | early-out so the interval bookkeeping (`group_last_checkpointed`) is never seeded — no ghost can ever form |

The default group can never be ephemeral (typed reject), so `checkpoint_cursor`/`maybe_checkpoint`/`checkpoint_default_attempts` need no gate. Partition checkpoints can't fire (partition-addressed ephemeral subscribe is a typed reject this slice).

### Reap + the one seam that could leak

- **Reap on last departure** (`unsubscribe_in`/`unsubscribe_in_stream`, driven by Unsub, subscription switch — Sub/SubTo/SubSubject rebinds now deregister the old subscription stream-aware *before* reassigning — a repeated `Connect`, and the connection-close cleanup): removes the in-memory group, releases floor entries (no ghost), frees DLQ dedup slots, and belt-and-suspenders **deletes any stray `cursor-`/`attempts-` files** (none can exist via the gated writers; covers the crash-of-an-older-flag-ignorant-server edge, which a fresh subscribe also fail-closed refuses). Reap and subscribe are serialized on the engine actor, so a reap can never race a reconnect into half-state: ordered-before keeps the group alive (brief two-member overlap), ordered-after binds a fresh group.
- **The seam:** a #565 open-set LRU eviction drops a named stream's whole consumer state while its ephemeral subscriber stays connected; the next poll re-creates the group — and without care would re-create it **durable** (live writers = the leak). Fixed with the authoritative `ephemeral_subs` intent registry keyed `(stream, group)`: every named-group creation site (`resolve_named_poll` + the four named mode-setters) re-creates under live intent as ephemeral. Pinned by a dedicated test (`a_stream_lru_eviction_cannot_turn_a_live_ephemeral_subscription_durable`).
- **Never idle-evicted** (#277): `is_evictable` refuses ephemeral groups — eviction would checkpoint at head AND let the next poll re-create the name durable. Zero-subscriber ephemeral groups cannot linger (the reap is synchronous with the last departure), so the sweep has nothing to reclaim.

### Retention-floor decision (documented in CONTRACTS.md)

**CONNECTED pins, REAPED never pins.** A live ephemeral group is `touched` with a live cursor, so `min_committed_offset`/`min_committed_offset_named` pin the consumer-safe reap floor exactly like any live reader — retention must not reap unread records out from under a connected consumer. On reap there is no live entry, **no ghost** (#432 — never inserted, and removed defensively), and no checkpoint, so the floor returns to the durable head: a disconnected-forever ephemeral consumer **cannot wedge reaping**, by construction rather than by timeout. Deliberately asymmetric with durable groups (whose checkpoints + ghosts keep pinning while disconnected — that's their contract).

### Everything else scoped by the issue

- **max_groups:** ephemeral groups count (bounded memory, typed `ERR_TOO_MANY_GROUPS` at the cap; the reap frees the slot immediately; the subscribe runs the idle sweep first, like the poll-side gate).
- **Mode conflicts fail closed BOTH ways** with the new stable `ERR_EPHEMERAL_GROUP_CONFLICT` (codes.rs + proto `ServerErrorCode` + cross-crate drift tests): ephemeral subscribe vs live-durable / on-disk checkpoint / ghost floor entry; durable subscribe vs live-ephemeral (both `subscribe_in` and the durable `SubTo` path).
- **Recovery:** nothing on disk → never resurrects (restart-harness tests, with a durable sibling proving resume still works).
- **Storage modes:** default + named streams; per-stream-log AND shared-WAL #597 (the probe/reap resolve the same mode-aware fs as the writers). **Partitioned: typed reject** (own checkpoint topology + no membership tracking; the wire flag still rides so a future slice can lift it).
- **Metrics:** aggregate/unchanged — an ephemeral group appears in the existing consumer-lag series while live exactly like an evicted durable group's series after eviction (fixed slots, house behavior); no new cardinality.
- **DLQ:** unchanged by design — a dead-lettered record is *message data* (deliberately durable), not group state; documented.
- **At-least-once contract** documented in a new CONTRACTS.md section: within the subscription's lifetime only; a re-subscribe starts a fresh group at the earliest retained offset (a new group's cursor begins at 0 clamped up to earliest-retained — there is no other start-position policy today).

### Pre-existing behavior fixed in passing (needed for reap correctness, strict improvement)

`SubTo`/`SubSubject` rebinds used to flip `registered_subscription = false` **without** deregistering engine-side (the comment said "leave the registration behind rather than stranding it" but no leave ran) — a Sub-broadcast-then-SubTo could strand the broadcast slot busy until disconnect. Rebinds now route the leave per-stream (`leave_current_subscription`), which the ephemeral reap requires and which matches the comment's documented intent. The durable named path gains no registration (byte-for-byte #676 behavior otherwise).

## Attack seams for review

1. **A missed writer = a durable leak.** The claimed-complete writer set is: `write_group_checkpoint`, `write_group_attempts`, `write_named_group_checkpoint`, `write_named_group_attempts` (+`_pairs` via its sole caller), with `checkpoint_partition_group` unreachable (partition reject) and the default-group writers unreachable (empty-group reject). If you can find a path that persists group state without passing one of these funnels, that's the bug. (The zero-artifact tests snapshot the file set across every cadence, root + stream subdir.)
2. **Membership orphan = a never-reaped group.** Every departure path routes through the stream-aware leave: Unsub, Sub/SubTo/SubSubject switches (including cross-stream same-name, compared as (stream, group)), repeated `Connect` (deregisters an ephemeral sub *before* the binding reset that would mis-route the later disconnect leave), and the connection-close cleanup pair. The known residual is the documented panic-unwind exposure in `handle_connection` (same class as the pre-existing broadcast-slot note there): no panic source exists in those paths today, but a future panic-prone refactor would leak the membership — and unlike a durable group, nothing durable leaks with it (memory only, bounded by max_groups).
3. **Reap racing a reconnect** is serialized on the engine actor (single writer): before-last-departure = shared live group; after = fresh group at start position. The brief two-member overlap on reconnect-before-close is at-least-once-safe (documented).
4. **The floor decision's implications:** no-pin-after-reap can, at worst, let retention reap records a *future* re-subscriber would have wanted — but the ephemeral contract makes that explicit (fresh group, earliest-retained start). Wedge direction is impossible: nothing survives to pin. A live-but-idle connected ephemeral group pins like any slow reader — same operator surface as today (`max_groups`, retention bounds still bounded by the durable head fallback).
5. **A durable consumer bound to a name *before* an ephemeral group existed** can drain the live ephemeral group through the group-level mode (same class as the other group modes); its checkpoints are suppressed (group property), no artifacts leak, and after the reap its next poll lazily re-creates a fresh durable group — today's exact lazy-creation semantics. The subscribe-time conflict gates close every other direction.

## Gates (rust:1-bookworm container, CARGO_TARGET_DIR=/work/target-eph, non-root cargo, --sysctl net.ipv4.ip_unprivileged_port_start=1024 --cap-drop NET_BIND_SERVICE)

- `cargo fmt --all --check` — OK
- `sh scripts/check-format-registry.sh` — OK (no tag / on-disk layout change: nothing to re-pin)
- `cargo build --workspace --all-targets --all-features --locked` — OK
- `cargo test --workspace --all-features --locked` — **0 failed** (every suite `test result: ok`)
- golden-path acceptance — PASS (summary JSON in the run)
- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings` on **1.97.0** — OK
- `scripts/ci/relative-link-check.sh` — 0 problems (CONTRACTS.md edit)
- plus the full native macOS `cargo test --workspace` — 0 failed

Refs #771 (child 1). Pairs with #277, #432, #547, #565, #566, #594, #597.


ELares:
Review [B1] fixed in ae18620 — the #565 LRU eviction × retention-floor data-loss window is closed, memory-only (zero-durable-artifacts preserved):

- **Park**: `evict_named_stream` parks each live ephemeral group's committed offset in the intent registry (`EphemeralSub::parked_committed`, the in-memory "ephemeral ghost") before dropping the stream's consumer state.
- **Floor**: `min_committed_offset_named` chains parked offsets — read BEFORE the absent-`named_streams` early return, which is exactly the post-eviction / produce-reopened state where the park is the only pin. One fix at the floor covers all three reap triggers (produce, reopen reap, config reload).
- **Resume**: the re-created group resumes from the parked offset (taken only on actual creation, so a cap-reject never loses the pin) — position and pin now survive the eviction within the subscription's lifetime; the live cursor supersedes the park in the same actor call. Last-member reap releases the entry, park included (no-wedge preserved).
- **Test**: `a_live_ephemeral_consumer_keeps_pinning_retention_across_a_stream_lru_eviction` — k=3 < head=24, retention bound tripped hard across eviction + reopen + produces: records k..head survive, next poll resumes at exactly k with zero truncations, floor releases after the reap. **Mutation-verified** (park disabled ⇒ the floor assertion fails at the head).
- Also documented the DLQ churn vector (per-incarnation dedup ⇒ bounded at-least-once duplicate captures across reap/re-subscribe loops).

Full CI-equivalent re-run green before push: fmt / format-registry / build all-features / workspace test all-features **0 failed** / golden-path acceptance PASS / clippy 1.97 `-D warnings` / relative-link-check.

Issue #546, opened 2026-06-19
[M1-I8] Tier-W work-queue mode kept, batched + zero-copy


ELares:
Keep Tier-W (leases, visibility timeouts, redelivery, DLQ, key-shared) as the richer-than-NATS work-queue tier, but make its delivery batched and zero-copy via the I5/I6 path. First-principles: per-message semantics are the differentiator NATS/Kafka structurally lack, so we keep them, but the per-record delivery cost can still be amortized with batching. Preserves the SQS-class feature set while closing the throughput gap.

PR #1166, opened 2026-07-12
perf(storage): verify-once body-CRC for self-written/verified-resident segments (#540)

### LINKED ISSUE #540: [M1-I4] Verify-once CRC for resident/self-written segments
Skip read-side CRC validation for segments that are resident or were written by this process, since their integrity was already verified at write time; the frame still carries its CRC for client end-to-end checking. First-principles: re-validating body CRC on every read is redundant CPU when the bytes have not left trusted memory. A verify-once-while-resident flag removes that per-read cost without weakening end-to-end guarantees. Shared substrate for the consume read plane.

ELares:
## What & why

Lays the **verify-once** foundation: a segment whose body integrity was already proven this process (self-written, recovered-on-load, or cold-restored) and that has stayed continuously resident is tracked so the consume read fast-path *can* skip the redundant server-side body-CRC recompute. Closes #540 (M1-I4).

> **Review response (safety fix):** adversarial review correctly found that the body-CRC skip's only live beneficiary is the per-record materialized `Deliver` path, whose wire frame (`DeliverBody`) carries **no** crc — so a default consumer would have **no** end-to-end backstop (and `scan_range` `pread`s every read, so a page-cache-evicted disk byte can rot and be re-read un-verified). The raw `DeliverBatch` path is safe (the client re-decodes the stored frame incl. its crc) but never did a server body-CRC anyway. So this PR now **disarms** the skip on every production path (`SegmentReader::crc_skip_armed == false`) — the server body-CRC is always recomputed on the live per-record path, integrity is **identical to pre-#540 for every consumer**, and the tested state-machine foundation ships ready for the follow-on to arm safely.

## Safety model (the gate)

`skip_body_crc() == crc_skip_armed && verified_resident && !verify_always`.

- `verified_resident` — the per-segment state machine (below), **live and tested**.
- `verify_always` — the tunable opt-out, **live**.
- `crc_skip_armed` — **disarmed at every production reader-open site.** Only the #540 unit tests arm it (in isolation) to exercise the trusted decode. It stays disarmed until the per-record `Deliver` frame carries the stored `body_crc` for the client to re-verify (the follow-on). Arming it without that backstop would reintroduce the silent-corruption hole, so it is intentionally not wired live.

## State machine (live, tested — feeds `verified_resident`)

Per-`Log` `verified_segments` set, mirrored onto `SealedSegment.verified`:
- **INSERT** on self-write seal (`roll`), recovery load (`scan_recovery` validated the chain), cold restore (`verify_fetched_segment`).
- **CLEAR** at the single retirement chokepoint `evict_segment_index` (reap / compaction-install / offload). Reader-cache fd-shed does not clear it (same local file, ids never recycle). A cold reload re-verifies + re-inserts.

## Encryption

Verify-once is plaintext-only. Encrypted reads take `scan_range`'s decrypt branch, which always CRC-checks the ciphertext+tag **and** AEAD-verifies the tag, never consulting the skip. No trusted variant of `decode_encrypted` exists.

## DLQ invariant (documented)

`DlqSink::append_encoded` re-encodes a source record with a **fresh** body CRC; sound today only because the source was read with the CRC verified (skip disarmed). A code comment records that the follow-on arming the skip must re-verify/preserve the original crc before dead-lettering a skip-read record (else it persists a fresh CRC over rotted bytes = permanent undetectable DLQ corruption).

## Follow-on to realize the win (design)

Add a capability-negotiated, default-off per-record `Deliver` `body_crc` (mirror the `flags2` / `deliver_encoding` pattern, so conformance/Go-SDK vectors stay byte-identical): server ships the **stored** crc (no recompute); client verifies `crc32c(key‖headers‖payload)`. Then arm the skip only for crc-capable consumers (server re-verifies at delivery for old clients — no regression) + fix the DLQ re-encode to preserve the crc. That protocol change spans proto + server + both clients + Go-SDK vectors and is its own scoped change.

## Changes

- **codec**: `decode_with_subject_trusted` skips the body CRC32C + xxh3 while running every structural/framing check (unchanged; reachable only via the armed fast-path, i.e. tests today).
- **storage/segment**: `SegmentReader` gains `verified_resident`, `verify_always`, and the `crc_skip_armed` safety gate (+ builders); `scan_range`'s plaintext loop honors `skip_body_crc()`.
- **storage/log**: `verified_segments` set + lifecycle inserts/clear; `LogConfig::verify_always`.
- **storage/read_plane**: `SealedSegment.verified` + `SealedSnapshot.verify_always` carried on every republish.
- **storage/dlq**: documented the fresh-CRC re-encode invariant.
- **server/engine**: DLQ/txn sub-log configs inherit `verify_always` (mechanical).

## Tests (all green)

- codec: trusted decode skips a corrupt body while full decode + structural checks still catch it.
- segment: an **armed** verified-resident `scan_range` serves an in-memory-corrupted body while an unverified reader and the verify-always opt-out both catch it; a corrupt header is caught even when armed+verified; encrypted reads never skip AEAD/CRC even when armed+verified.
- log: self-written/recovered sealed segments are marked (active is not); eviction clears the mark so a reloaded reader re-verifies.

## Gates (local, native `cargo 1.97.0`)

- `cargo test --workspace --all-features` → **3602 passed, 0 failed**
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` → clean
- `cargo fmt --all --check` → clean
- `sh scripts/check-format-registry.sh` → layout unchanged
- golden-path acceptance → pass

Rebased onto origin/main (#546 merged). Do **not** merge until the coordinator signs off on the safe scoping (and the follow-on plan).

### DIFF
--- a/crates/ironbus-core/src/codec.rs
+++ b/crates/ironbus-core/src/codec.rs
@@ -723,7 +723,7 @@ pub fn decoded_len(header: &[u8]) -> Result<usize, DecodeError> {
 /// [`DecodeError::Truncated`] means more bytes may complete the frame; the other
 /// variants mean the frame is corrupt and must be skipped by recovery.
 pub fn decode(input: &[u8]) -> Result<(RecordView<'_>, usize), DecodeError> {
-    let (view, _subject, _tag, total) = decode_inner(input)?;
+    let (view, _subject, _tag, total) = decode_inner(input, true)?;
     Ok((view, total))
 }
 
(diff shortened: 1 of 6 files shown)


ELares:
Addressed the merge-blocking review. **The skip is now disarmed on every production path** (`crc_skip_armed == false`), so the server body-CRC is always recomputed on the live per-record `Deliver` path — integrity is identical to pre-#540 for every consumer, closing the no-backstop hole. The verified-resident state machine + verify-always opt-out + codec trusted-decode ship live and tested (unit tests arm the skip in isolation); the follow-on only has to arm it once the per-record `Deliver` frame carries the stored `body_crc` for the client to verify. Also documented the DLQ fresh-CRC re-encode invariant at `append_encoded`. Rebased onto main (#546). Full CI-equivalent local: workspace all-features 3602 passed/0 failed, clippy -D warnings clean, fmt clean, format-registry unchanged, acceptance pass.

ELares:
Closing unmerged: the adversarial review found #540's stated premise ('the frame still carries its CRC for the client's end-to-end check') is FALSE on the DEFAULT per-record consume path — OwnedRecord + the per-record Deliver wire frame carry no body_crc, and understands_deliver_batch defaults to false, so skipping the server body-CRC there has no client backstop (serves silently-corrupt plaintext; plus a DLQ re-encode with a fresh crc over rotted bytes). The safe fallback (disarm the skip on the live path) makes the PR INERT — zero CPU benefit while still running the verified-resident tracking — not worth merging as dead weight. Realizing verify-once safely requires a protocol change (below); deferring the whole optimization until that's committed to. The verified-resident state machine here was reviewed SOUND and is preserved on branch feat/verify-once-crc for the follow-on.

Issue #1170, opened 2026-07-12
[follow-up] Cross-tenant sharing (#1163): retention coupling, revoke-teardown, txn back-check, group-name hardening


ELares:
Non-blocking from the #1169 security review (MERGE_SAFE — no over-share/no-grant-cross/direction-bypass/cursor-cross-contamination; pattern_within proven subset-sound):
1. **Retention coupling** (availability, NOT a breach): a live importer cross-subscribe group holds a cursor on the EXPORTER's stream; if retention trims by min-committed-cursor, a dead/lagging importer group could pin the exporter's log from reaping. Assess the engine retention policy for cross-tenant groups (e.g. don't let a foreign importer group pin the exporter's reap floor, or cap it), and document.
2. **Revoke vs in-flight**: revoke is effective on the NEXT resolution (new subscribe/reconnect); an already-bound live cross-subscription keeps delivering until re-resolved. Standard capability-vs-live-session tradeoff — decide whether to tear down live cross-subscriptions on revoke (stronger) or document 'effective on next resolution' (current).
3. **Txn back-check (deferred from #1163 item 2, pre-existing from #1162)**: the multi-tenant txn-listener back-check double-prefixes (write_txn_check echoes the scoped id → handle_txn_check_result re-scopes → <tenant>/<tenant>/tx1 → never resolves → fail-closed rollback). Fix: strip the tenant prefix outbound / don't re-scope an already-scoped echoed id + a multi-tenant txn back-check test. Cross-tenant txns stay own-namespace (out of #1163 scope).
4. **Defense-in-depth**: forbid '/' in client group names (an EXPORTER could self-collide by naming a group '<importer>/...' — un-forceable by an importer, self-inflicted, but a cheap guard); empty-token subjects (partner.orders. → orders.) resolve within-grant (bounded, noted).

PR #1183, opened 2026-07-13
fix(multi-tenant): txn back-check double-prefix + cross-tenant retention/group-name hardening (#1170)

ELares:
Non-blocking follow-ups from the #1169 cross-tenant sharing security review (the multi-tenant path from #1162/#1163/#1169). **DO NOT MERGE** until adversarially reviewed for any cross-tenant leak. Every change preserves tenant isolation absolutely and fails closed on ambiguity.

## ITEM 3 — a real BUG (highest value): txn back-check double-prefix
Pre-existing from #1162. In multi-tenant mode the txn-listener back-check DOUBLE-PREFIXED the txn id: `write_txn_check` echoed the already-`<tenant>/`-scoped stored id to the producer's listener; the listener answers by echoing the id it received, and `handle_txn_check_result` re-scopes **unconditionally**, so a scoped echo became `<tenant>/<tenant>/tx1` — which matched no stored txn, never resolved, and (after the bounded attempt cap) **wrongly ROLLED BACK a live multi-tenant transaction**.

**Fix:** strip the `<tenant>/` prefix on the **outbound** echo only (`strip_tenant_txn_id`), so the listener sees its OWN client id and its answer re-scopes to the *same* txn.

**Isolation preserved:** the security-critical **inbound** scoping stays unconditional, so a *different* tenant's forged `TxnCheckResult` still scopes into ITS OWN namespace (`<other>/tx1`) and can never resolve this tenant's `t1/tx1` — cross-tenant txns stay own-namespace. The test proves both (own tenant resolves; other tenant cannot).

## ITEM 1 — cross-tenant retention isolation
A dead/lagging cross-tenant IMPORTER group holds a cursor on the EXPORTER's stream (`(<exporter>/stream, <importer>/group)`); left in the reap-floor min it could pin the exporter's log from reaping — a cross-tenant availability/DoS (an importer holds the exporter's disk hostage).

**Fix:** a foreign importer's guest cursor is **excluded** from `Engine::min_committed_offset_named` (`is_foreign_importer_group`), at every source (live groups, durable ghosts, parked ephemeral) and **by name**, so it holds for a live guest, a ghost, and **across a restart** with no new durable state. A live keeping-up importer is unaffected (its cursor sits near the head); a lagging importer is subject to the exporter's retention window like any late consumer but can never hold it open. A same-tenant own group still pins (within-tenant consumer-safety unchanged). Policy documented in the tenant.rs `#1170` sharing block.

## ITEM 4 — defense-in-depth group-name guard
A CLIENT-supplied consumer-group name may no longer contain the `/` group-namespace delimiter (rejected at **every** group ingress — Sub / SubTo / SubSubject / TxnListen / CumulativeAck / StreamCommit / PauseGroup — with `ERR_INVALID_GROUP_NAME`). Stops a tenant self-colliding with a server-scoped group key, and keeps the ITEM-1 guest classification precise (an own group is guaranteed `/`-free). `.` is not a group delimiter and is unaffected. (The empty-token-subject note stays within-grant; no code needed.)

## ITEM 2 — revoke vs in-flight: DEFERRED (owner decision)
The stronger "tear down live cross-subscriptions on revoke (immediately effective)" behavior is **not a contained change**: the broker is thread-per-connection with no central live-session registry and no cross-thread teardown channel, and the sharing registry is pinned per-connection at `Connect`; enumerating and dropping exactly the cross-subs under a now-revoked grant without risking a same-tenant session is semantically involved. Per the issue's guidance, **not implemented** — the current "revoke is effective on the next resolution (new connection / reconnect)" behavior is documented in the tenant.rs sharing policy block and raised here for the owner to decide.

## Tests (all mutation-verified — each fails with its fix neutered)
- `multi_tenant_txn_back_check_resolves_without_double_prefix_and_stays_isolated` (ITEM 3)
- `a_foreign_cross_tenant_importer_group_is_excluded_from_the_exporters_reap_floor` (ITEM 1)
- `a_lagging_foreign_importer_does_not_block_the_exporters_retention_reap` (ITEM 1)
- `a_client_group_name_with_a_slash_is_rejected_at_the_wire_ingress` (ITEM 4)

## Constraints
No new dependencies; `Cargo.lock` / `Cargo.toml` / `deny.toml` byte-identical. Wire format unchanged (the group-name reject is validation, not a format change).

### DIFF
--- a/crates/ironbus-cli/src/main.rs
+++ b/crates/ironbus-cli/src/main.rs
@@ -9507,7 +9507,7 @@ fn sleep_interruptible_cli(dur: std::time::Duration, shutdown: &std::sync::atomi
                                  // loop, the graceful drain); splitting it further would scatter a
                                  // single concern across helpers.
 fn run_broker<F: Filesystem + Clone + 'static>(
-    engine: Engine<F, SystemClock>,
+    mut engine: Engine<F, SystemClock>,
     addr: &str,
     data_dir: Option<&Path>,
     config: &ServeConfig,
@@ -9538,6 +9538,18 @@ fn run_broker<F: Filesystem + Clone + 'static>(
     // it only through the bounded-channel handle, so no handler holds a lock across an fsync. The
     // actor's join handle yields the engine back on its clean exit (a Shutdown drain), which is how
     // the graceful-shutdown cursor flush (#195) completes before the process exits 0.
+    //
+    // #1170 (ITEM 1): tell the engine whether CROSS-TENANT SHARING is configured (a non-empty
+    // import/export registry on the auth config). This GATES the cross-tenant retention-floor
+    // exclusion — set BEFORE the engine moves into the append actor. With no sharing (a single-tenant
+    // broker, or a tenant broker with no grants) it stays `false`, so NO consumer group is ever
+    // excluded from the reap floor and a legal `/`-in-name group keeps its full consumer-safe
+    // protection (the delimiter is an ordinary user character absent tenancy).
+    let cross_tenant_sharing_active = auth
+        .as_ref()
+        .and_then(|a| a.sharing())
+        .is_some_and(|s| !s.is_empty());
+    engine.set_cross_tenant_sharing_active(cross_tenant_sharing_active);
     let (shared, actor) =
         spawn_actor_with_gather(engine, DEFAULT_CHANNEL_BOUND, config.commit_gather_us);
     // Install the SHARED client produce-ack slot (#719) onto the base handle BEFORE any per-connection
--- a/crates/ironbus-server/src/engine.rs
+++ b/crates/ironbus-server/src/engine.rs
@@ -2463,6 +2463,15 @@ fn validate_group_name(name: &str) -> Result<(), EngineError> {
 /// must NOT pin the stream owner's consumer-safe retention reap floor (#1170, ITEM 1 — cross-tenant
 /// resource isolation).
 ///
+/// GATED, LOAD-BEARING PRECONDITION: this name test is meaningful ONLY when cross-tenant sharing is
+/// active (`Engine::cross_tenant_sharing_active`), which the sole caller
+/// ([`Engine::min_committed_offset_named`]) enforces. `/` is a fully legal, user-choosable character
+/// in stream AND group names on a single-tenant broker, so a bare name test with no tenancy gate
+/// would misclassify a single-tenant consumer group `b/g` on a stream `a/orders` as "foreign" and
+/// silently reap its unread records — a data-loss regression. The gate makes the test apply ONLY
+/// where tenancy's server-scoping invariant guarantees a same-tenant own group and its stream share
+/// the leading `<tenant>/` token; there, only a GENUINE cross-tenant guest differs.
+///
 /// A live cross-subscribe from a foreign importer keys its cursor on the EXPORTER's stream under an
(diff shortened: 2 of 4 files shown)


ELares:
## Fixed the ITEM 1 data-loss regression (commit 694fc7f). Items 3 & 4 unchanged.

**The regression (confirmed):** the reap-floor exclusion classified a "foreign importer" guest purely by NAME (group's leading `/`-token differs from the stream's) with no tenancy gate. Since `/` is a supported char in group/stream names, a single-tenant group `b/g` on stream `a/orders` was misread as foreign and excluded → its unread records reaped under retention → silent data loss (#566 violation) in single-tenant mode.

**Fix — gate the exclusion on EXPLICIT INTENT, not name inference.** New engine flag `cross_tenant_sharing_active`, set once in `run_broker` **only when a non-empty `SharingRegistry` is wired**, gates `is_foreign_importer_group` in `min_committed_offset_named`:
- **Sharing OFF** (single-tenant, or tenant broker with no grants): **no group is ever excluded** — a `/`-in-name group keeps full consumer-safe protection.
- **Sharing ON**: the server-scoping invariant holds (a same-tenant own group and its stream share the leading `<tenant>/` token; own group names are `/`-free by the ITEM-4 wire guard), so only a genuine cross-tenant guest `<importer>/…` differs and is excluded — a dead/lagging importer still can't pin the exporter's disk.

**Defense-in-depth #2 (reject `/` in engine `validate_group_name`) — deliberately NOT taken, with reason.** It would reject the server-applied cross-tenant **guest** group `<importer>/<group>` (which legitimately carries `/` and flows through the engine poll/subscribe/filter create paths) — breaking cross-tenant sharing — and would skip recovery of a legacy `/`-group's durable cursor (forced redelivery). `/` is a long-supported path-unsafe group-name char (hex-encoded in checkpoint filenames; asserted by the existing `a_named_group_cursor_resumes_on_a_real_directory` test). Its intended value (no NEW own `/`-group) is **already fully met** by the ITEM-4 wire guard (the only client-reachable group ingress) + this sharing gate (single-tenant never excludes), so no durable-recovery gap remains. The engine intentionally still accepts `/`. *(If you'd still like a `/` reject on non-sharing brokers specifically, I can add it as a serve-time-gated engine setter — say the word.)*

**Tests (all mutation-verified — neutering the gate makes the single-tenant `b/g` floor jump to 27 and fails them):**
- `a_single_tenant_slash_group_is_never_reaped_the_data_loss_regression_fix` — the exact scenario: SINGLE-TENANT `b/g` lagging + retention on ⇒ NOT reaped.
- `a_cross_tenant_guest_is_excluded_from_the_reap_floor_only_when_sharing_is_active` — same group+cursor: excluded with sharing ON, **pins** with sharing OFF.
- `a_lagging_cross_tenant_guest_does_not_block_the_exporters_retention_reap` — genuine guest, sharing ON: exporter still reclaims below its own floor.

**Gates (Linux container):** `cargo test --workspace --all-features --locked` → **0 failed** (all 54 isolation tests incl. the reframed guest + single-tenant regression, and the previously-guest-read/hex-recovery tests); clippy `-D warnings` clean; fmt clean; format-registry OK; `cargo tree --all-features -i ring` empty; windows-gnu check OK; Cargo.lock/Cargo.toml/deny.toml byte-identical.

Not merging — over to you for re-review.

PR #1203, opened 2026-08-18
fix(deps): bump lz4_flex to 0.11.6 for RUSTSEC-2026-0041

ELares:
RUSTSEC-2026-0041 — decompressing invalid data can leak information from uninitialized memory or a reused output buffer. Affects `lz4_flex <= 0.11.5`, patched in `0.11.6`. Both crates that declare it (`ironbus-core`, `ironbus-proto`) pinned `"0.10"`.

**How it surfaced.** It fails `cargo-deny` in **IronAuth**, which depends on `ironbus-client` → `ironbus-core` → `lz4_flex 0.10.0`. That lane has been red on IronAuth's `main` and on every branch, and it cannot be fixed downstream: the requirement lives here.

**Scope kept minimal.** Bumped to `0.11.6`, not to the current `0.14`, so this stays a security patch that can land on its own. The feature set is unchanged — `safe-encode` + `safe-decode` + `checked-decode`, `default-features = false` — which is the pure-Rust decode path ADR-0003 chose. **No source change was needed.**

**Verified:** 535 + 174 tests pass across `ironbus-core` and `ironbus-proto`, including the compression round trips and the corrupt-stream cases `compress.rs` documents (a corrupt lz4 stream returns a typed `DecompressError` and never panics).

**Downstream still needs a release.** IronAuth consumes `ironbus-core` from crates.io, so it stays exposed until a version carrying this bump is published. Until then IronAuth can point at this commit with `[patch.crates-io]`.

https://claude.ai/code/session_01X3FaPPJBUHKa1dXeQ7n6zW

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1269,9 +1269,9 @@ checksum = "953f07c43838f8e6f9758cab68bf5bed85465e7587ebe0b823f1bcd81978ad3a"
 
 [[package]]
 name = "lz4_flex"
-version = "0.10.0"
+version = "0.11.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b8c72594ac26bfd34f2d99dfced2edfaddfe8a476e3ff2ca0eb293d925c4f83"
+checksum = "373f5eceeeab7925e0c1098212f2fbc4d416adec9d35051a6ab251e824c1854a"
 
 [[package]]
 name = "matchit"
--- a/crates/ironbus-core/Cargo.toml
+++ b/crates/ironbus-core/Cargo.toml
@@ -39,7 +39,7 @@ xxhash-rust = { version = "0.8", default-features = false, features = ["xxh3"] }
 # `links` key and runs no C-builder build script). Pinned to 0.10 to hold the workspace MSRV (1.78):
 # 0.11+ declare a 1.81 minimum and would fail the `msrv (1.78)` CI job. `default-features = false`
 # drops the `frame` format (and its xxhash dependency); IronBus uses only the block API.
-lz4_flex = { version = "0.10", default-features = false, features = ["safe-encode", "safe-decode", "checked-decode"] }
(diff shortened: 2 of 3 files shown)


ELares:
## Withdrawing this — I missed that the repo had already analysed it

`deny.toml` already carries a scoped ignore for RUSTSEC-2026-0041, and it is not a hand-wave. It states:

> * `lz4_flex` is pinned to 0.10 because the fixed releases (0.11.6 / 0.12.1) declare a 1.81 MSRV and IronBus's MSRV is 1.78 (the `msrv (1.78)` CI gate). Bumping the codec would break that gate; the MSRV rise is a separate, deliberate decision.
> * We enable the `safe-decode` feature, so the uninitialized-memory path (a) does not apply: decode never runs the `unsafe` implementation.

So the exposure was measured, the pin was deliberate, and the MSRV rise was consciously deferred. This PR forces that deferred decision as a side effect of a bump — which is exactly backwards.

**Measured cost of the bump, for the record.** `lz4_flex 0.11.6` declares `rust-version = "1.81"`. On this branch that fails four CI lanes that are green on `main`: `msrv (1.78)`, `cargo-deny`, the opt-in zstd codec, and the opt-in at-rest encryption. Raising the floor is not a one-line change either — the 1.78 figure is load-bearing in `Cargo.toml`, `.github/workflows/ci.yml`, `deny.toml` (three separate dependency rationales, including the raft-rs/protobuf one), `Dockerfile`, `README.md` and `CHANGELOG.md`. And `README.md` gates it further: *"may rise only in a minor release, new floor always at least 6 months old."*

**What is actually left of the exposure.** The advisory has two paths: (a) uninitialized memory on the `unsafe` decode path, and (b) the previous contents of a reused output buffer. `safe-decode` removes (a). Path (b) is not addressed by the feature and is worth a separate look — that, not a version bump, is the useful next question.

Closing rather than leaving it open against a policy it contradicts. If the MSRV rise is wanted on its own merits, it deserves its own PR that touches all six of those sites coherently and lands in a minor release.

